PR #1, opened 2026-01-22
Minimal MLIR Integration Framework for MorphiZen

mingyueliuh:
## Motivation

This PR introduces **morphizen-mlir** - a minimal MLIR integration framework for the MorphiZen/VAIP ecosystem to enable MLIR-based graph transformations in ONNX Runtime execution providers.

**Goals:**
- Enable MLIR-based graph processing in VAIP framework
- Provide clean architecture without pattern matching or protobuf dependencies
- Establish foundation for future MLIR transformation passes

### Core Components

- **Level-1 MLIR Pass**: Parses ONNX graphs to MLIR ModuleOp and enables transformations
- **CMake Build System**: Modular configuration with FetchContent for LLVM/MLIR and MorphiZen
- **Test Infrastructure**: ORT integration tests with Conv and Conv+Gemm models
- **Documentation**: Testing guide with examples

### MLIR Integration

- Parses ONNX graphs to `mlir::ModuleOp`
- Loads func, arith, and unregistered dialects
- Walks and inspects all operations
- Prints complete MLIR IR with debug info

### Build Dependencies

1. ONNXRuntime (required - build first)
2. LLVM/MLIR (auto-fetched via FetchContent or pre-built)
3. MorphiZen (auto-fetched via FetchContent)

### Configuration

```json
{
  "target": "mlir-target",
  "name": "mlir-pass",
  "plugin": "morphizen-level1-pass-mlir"
}
```

Environment: `MORPHIZEN_ORT_BRIDGE_UNITTEST_BACKEND=mlir-backend`

### Test Cases

1. **LoadVitisAIProvider** - Verifies VitisAI EP registration
2. **CreateSessionWithVitisAIProvider** - Tests Conv model (1x3x8x8 → 1x16x8x8)
3. **CreateSessionWithConvGemmModel** - Tests Conv+Flatten+Gemm (1x3x8x8 → 1x32)

### Execution

```bash
set MORPHIZEN_ORT_BRIDGE_UNITTEST_BACKEND=mlir-backend
..\build\morphizen-mlir\bin\ort_integration_test.exe
```

### Build Status
✅ Successfully builds on Windows with Visual Studio 2022
- CMake configuration, LLVM/MLIR integration, compilation: All pass

### Test Results

```
[==========] Running 3 tests from 1 test suite.
[ RUN      ] OrtIntegrationTest.LoadVitisAIProvider
[       OK ] OrtIntegrationTest.LoadVitisAIProvider
[ RUN      ] OrtIntegrationTest.CreateSessionWithVitisAIProvider
Successfully parsed MLIR string to ModuleOp
Total operations in module: 6
[       OK ] OrtIntegrationTest.CreateSessionWithVitisAIProvider
[ RUN      ] OrtIntegrationTest.CreateSessionWithConvGemmModel
Total operations in module: 10
[       OK ] OrtIntegrationTest.CreateSessionWithConvGemmModel
[  PASSED  ] 3 tests.
```

### Verification
- ✅ MLIR parsing to ModuleOp
- ✅ Operation walking (6 ops for Conv, 10 for Conv+Gemm)
- ✅ Complete IR printing with debug info


README.md
```diff
@@ -0,0 +1,215 @@
...
+
+```
+onnx-hipdnn-ep/
+├── CMakeLists.txt              # Root CMake configuration
+├── README.md                   # This file
+├── LICENSE                     # Apache 2.0 License
+├── .gitignore                  # Git ignore rules
+├── cmake/                      # CMake modules
+│   ├── deps.cmake             # Dependency management
+│   └── llvm.cmake             # LLVM/MLIR configuration
```
MaheshRavishankar: I dont see the llvm.cmake file here.

cmake/deps.cmake
```diff
@@ -0,0 +1,245 @@
...
+file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/morphizen_deps.inc.h" "${MORPHIZEN_DEP_H_INC}")
+
+
+message(STATUS "Configuring LLVM/MLIR for morphizen-mlir")
+
+# LLVM configuration options
+set(LLVM_ENABLE_PROJECTS "mlir" CACHE STRING "LLVM projects to build")
+set(LLVM_TARGETS_TO_BUILD "host" CACHE STRING "LLVM targets to build")
+set(LLVM_ENABLE_ASSERTIONS ON CACHE BOOL "Enable LLVM assertions")
+set(LLVM_ENABLE_RTTI ON CACHE BOOL "Enable RTTI in LLVM")
```
MaheshRavishankar: Do we need to use RTTI? I'd start with disabling this for now.

cmake/deps.cmake
```diff
@@ -0,0 +1,245 @@
...
+# Enable ORT bridge and MLIR backend for morphizen
+set(morphizen_ENABLE_ORT_BRIDGE ON CACHE BOOL "Enable ORT bridge" FORCE)
+set(morphizen_ENABLE_MLIR_BACKEND ON CACHE BOOL "Enable MLIR backend" FORCE)
+set(morphizen_ENABLE_ONNX_BACKEND OFF CACHE BOOL "Enable ONNX backend" FORCE)
+set(morphizen_ENABLE_ONNX_SCHEMA_SUPPORT OFF CACHE BOOL "Enable ONNX schema support" FORCE)
+set(morphizen_ENABLE_RYZENAI_BIN_METADATA OFF CACHE BOOL "Disable ryzenai_bin_metadata submodule for version resource generation" FORCE)
+set(morphizen_ENABLE_BOOST OFF CACHE BOOL "Disable Boost dependency" FORCE)
+set(morphizen_OUTPUT_NAME "onnxruntime_morphizen_ep" CACHE STRING "Set morphizen output name of Morphizen" FORCE)
+
+FetchContent_MakeAvailable(morphizen)
```
MaheshRavishankar: Nit : Add new line at EoF.

cmake/deps.txt
```diff
@@ -0,0 +1,8 @@
+##
```
MaheshRavishankar: Nit: Remove blank line.

cmake/deps.cmake
```diff
@@ -0,0 +1,245 @@
+##
+# ** Copyright (C) 2023 - 2025 Advanced Micro Devices, Inc. All rights reserved.
```
MaheshRavishankar: If these files are new, they should have the year as 2026.

README.md
```diff
@@ -0,0 +1,215 @@
...
+- **Minimal Design**: No pattern matching, no protobuf dependencies
+- **Level-1 Pass**: Simple MorphiZen pass structure for MLIR integration
+- **Clean Architecture**: Based on morphizen-demo but simplified
+- **Ready for Extension**: Template for adding MLIR-based transformations
+
+## Building
+
+### Prerequisites
+
+- CMake 3.29 or later
```
MaheshRavishankar: Is there any specific reason for this?

cmake/deps.cmake
```diff
@@ -0,0 +1,245 @@
...
+# ** Licensed under the MIT License.
+##
+include(FetchContent)
+
+# Read and parse deps.txt for dependency version management
+file(STRINGS ${CMAKE_CURRENT_LIST_DIR}/deps.txt MORPHIZEN_DEPS_LIST)
```
MaheshRavishankar: This seems like a very complicated mechanism. Why dont we just use sub-modules to handle this instead of this complex logic to parse the hash etc.

level-1-pass-mlir/src/pass_main.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+using namespace morphizen;
+using namespace morphizen_cxx;
+
+namespace {
+
+struct Level1MlirPass {
+  Level1MlirPass(IPass& self) : self_{self} {}
+  
+  void process(IPass& self, Graph& graph) {
```
MaheshRavishankar: I dont understand this setup. This method is part of a `struct` that already has a handle to `IPass` and this method recieves an `IPass` object as well. Is this intended?

level-1-pass-mlir/src/pass_main.cpp
```diff
@@ -0,0 +1,76 @@
...
+    LOG(INFO) << "Level1MlirPass::process() called";
+   
+    // Save graph to file for MLIR processing
+    LOG(INFO) << "Saving graph to file...";
+    auto graph_ref = GraphConstRef(graph);
+    std::string graph_file = "graph_for_mlir.txt";
```
MaheshRavishankar: Please remove this by default. We dont want to have this on by default writing to disk. I think this was added for convenience or someone asked for something. It can be part of some debug or flag that allows you to dump intermediates, etc.

level-1-pass-mlir/src/pass_main.cpp
```diff
@@ -0,0 +1,76 @@
...
+    graph_ref.save(graph_file);
+    LOG(INFO) << "Graph saved to file: " << graph_file;
+    
+    // Parse MLIR file to mlir::ModuleOp
+    LOG(INFO) << "Parsing MLIR file to ModuleOp...";
+    mlir::MLIRContext context;
+    context.loadDialect<mlir::func::FuncDialect>();
+    context.loadDialect<mlir::arith::ArithDialect>();
+    context.allowUnregisteredDialects();
```
MaheshRavishankar, line 42: I am not sure we should allow unregistered dialects by default. This seems like a hack.

level-1-pass-mlir/src/pass_main.cpp
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+      LOG(INFO) << "Successfully parsed MLIR string to ModuleOp";
+      
+      // Get the module operation
+      mlir::ModuleOp module = *moduleRef;
+      LOG(INFO) << "ModuleOp created, ready for MLIR transformations";
+      
+      
+      // Print module with detailed flags
+      mlir::OpPrintingFlags flags;
```
MaheshRavishankar, line 66: Please remove these from the default execution mode. This should be done with a flag on request, and not by default. Also dont dump it to std::cout (and there is mix of llvm::outs and std::cout here which is not great).

level-1-pass-mlir/src/pass_main.cpp
```diff
@@ -0,0 +1,76 @@
...
+      flags.printGenericOpForm();
```
MaheshRavishankar: I dont think we want to use generic op form

level-1-pass-mlir/src/pass_main.cpp
```diff
@@ -0,0 +1,76 @@
...
+      flags.enableDebugInfo();
```
MaheshRavishankar: This and the following make it too verbose.

level-1-pass-mlir/src/pass_main.cpp
```diff
@@ -0,0 +1,76 @@
...
+  Level1MlirPass(IPass& self) : self_{self} {}
+  
+  void process(IPass& self, Graph& graph) {
+    LOG(INFO) << "Level1MlirPass::process() called";
+   
+    // Save graph to file for MLIR processing
+    LOG(INFO) << "Saving graph to file...";
+    auto graph_ref = GraphConstRef(graph);
+    std::string graph_file = "graph_for_mlir.txt";
+    graph_ref.save(graph_file);
```
MaheshRavishankar: As far as I can tell, it is this method that is converting the ONNX graph to MLIR? We need a way in which this is done without having to write to file. Looking through a little bit into Morphizen, the MLIRGraph structure in Morphizen is build a `func::FuncOp`. It should build a module and return the module. The structure here of having the Morphizen class write to file and then parsing it here is too round-about. We should keep it in memory and dump it only when needed.

MaheshRavishankar (changes requested):
I have left a bunch of comments so far. This is a big PR and I am reviewing it in phases. Also it pulls in a lot of infra from MorphiZen which is an external dependency for now, and really is not easy to review/adapt. I have a few blocking comments though. Ill update this list as I go along/things are addressed.

- [ ] Use Git submodules for dependencies instead of deps.txt + CMake FetchContent parsing
- [x]  Provide in-memory access to mlir::ModuleOp instead of file-based save/parse roundtrip (pass_main.cpp:28-42)

README.md
mingyueliuh: This version was changed to 3.28 to maintain consistency with the requirements of onnxruntime.
https://onnxruntime.ai/docs/build/inferencing.html#prerequisites

README.md
mingyueliuh: llvm.cmake is no longer needed; the README.md here is outdated.
Will update.

cmake/deps.cmake
mingyueliuh: Will apply this for all files.

cmake/deps.cmake
mingyueliuh: Yes, we don't need to use RTTI. Will disable it default.

level-1-pass-mlir/src/pass_main.cpp
mingyueliuh: it is a bad design. let's leave it as it is. it is a backward compatibility issue.  we will file an issue to clean it up.

level-1-pass-mlir/src/pass_main.cpp
mingyueliuh: Converting ONNX to MLIR is not done here; the graph is essentially already an MLIR graph.
At present, MorphiZen lacks  APIs for exporting the graph as a string or for directly accessing the FuncOp.
We plan to add APIs in MorphiZen’s MLIR implementation to support keeping the IR in memory.

level-1-pass-mlir/src/pass_main.cpp
mingyueliuh: Will remove it

level-1-pass-mlir/src/pass_main.cpp
mingyueliuh: Will remove it

level-1-pass-mlir/src/pass_main.cpp
mingyueliuh: Will change it.

cmake/deps.cmake
mingyueliuh: Will change to use sub-module.

level-1-pass-mlir/src/pass_main.cpp
mingyueliuh: Yes ，This is a dirty-hack.  We could solve this properly by loading the ONNX dialect, but that would require adding an onnx-mlir dependency, which would be challenging.

mingyueliuh:
> I have left a bunch of comments so far. This is a big PR and I am reviewing it in phases. Also it pulls in a lot of infra from MorphiZen which is an external dependency for now, and really is not easy to review/adapt. I have a few blocking comments though. Ill update this list as I go along/things are addressed.
> 
> * [ ]  Use Git submodules for dependencies instead of deps.txt + CMake FetchContent parsing
> * [x]   Provide in-memory access to mlir::ModuleOp instead of file-based save/parse roundtrip (pass_main.cpp:28-42)

@MaheshRavishankar  Thanks for your review.
 1: we will use git submodules for dependencies instead of deps.txt + CMake FetchContent parsing .
 2: It requires adding one or two new APIs to MorphiZen and is currently being planned.

level-1-pass-mlir/src/pass_main.cpp
mingyueliuh: Added an environment variable MLIR_PRINT_WITH_VERBOSE to allow users to control the behavior on demand. It is disabled by default.

level-1-pass-mlir/src/pass_main.cpp
MaheshRavishankar: If we are going to use onnx-mlir lets add that as a submodule. This is going to obfuscate any verifier etc. cause the onnx mlir ops will not be verified.

MaheshRavishankar (commented):
Ok, ill review again after the comments are addressed. I understand this is probably lower priority for now than PR#2, but if needed to land the other PR, I can review it son.

I am curious what the worfklow directorys are for?

mingyueliuh:
> Ok, ill review again after the comments are addressed. I understand this is probably lower priority for now than PR#2, but if needed to land the other PR, I can review it son.
> 
> I am curious what the worfklow directorys are for?

The workflow directory contains AI tool–specific rules and will be removed.

level-1-pass-mlir/src/pass_main.cpp
```diff
@@ -0,0 +1,73 @@
...
+namespace {
+
+struct Level1MlirPass {
+  Level1MlirPass(IPass& self) : self_{self} {}
+  
+  void process(IPass& self, Graph& graph) {
+    LOG(INFO) << "Level1MlirPass::process() called";
+   
+    auto graph_ref = GraphConstRef(graph);
+    auto graph_string = graph_ref.save_string();
```
MaheshRavishankar, line 34: Ok, this is slightly better, but why even serialize and deserialize. We already have a dependence on MLIR here. Just return the ModuleOp ?

level-1-pass-mlir/src/pass_main.cpp
```diff
@@ -0,0 +1,73 @@
...
+    context.loadDialect<mlir::arith::ArithDialect>();
+    context.allowUnregisteredDialects();
+    
+    auto moduleRef = mlir::parseSourceString<mlir::ModuleOp>(*graph_string, &context);
+    
+    if (!moduleRef) {
+      LOG(INFO) << "Failed to parse MLIR string to ModuleOp";
+    } else {
+      // Get the module operation
+      mlir::ModuleOp module = *moduleRef;
```
MaheshRavishankar, line 63: Could we guard the entire printing behind an env variable that is used for testing?

MaheshRavishankar (commented):
Apart from docs and stuff this is a fairly small PR. I am fine with landing this just a few additional comments left.

PR #3, opened 2026-02-04
morphizen rocm ep initial version

AMDmoore:
## Overview

MorphiZen-ROCm is an ONNX Runtime Execution Provider for AMD ROCm GPUs using MIOpen, hipBLASLt, rocBLAS, and custom HIP kernels. It integrates with the MorphiZen optimization framework to enable efficient ONNX model inference on AMD hardware.

**This project includes a GQA (Grouped Query Attention) layer test** demonstrating GPU-accelerated inference, serving as a reference implementation for LLM workloads on AMD ROCm hardware.

### 1. Level-1 Pass (`level-1-pass-rocm/`)
Graph optimization pass that orchestrates Level-2 sub-passes and performs subgraph merging.

- GPU availability checking
- Sub-pass orchestration for multiple operations
- Subgraph merging using Union-Find algorithm
- JSON metadata serialization

### 2. Level-2 Passes (`level-2-pass-rocm-*/`)
Operation-specific pattern matching passes for each supported operation.

- Conv pass (MIOpen)
- Gemm pass (hipBLASLt)
- MatMul pass (rocBLAS)
- Mul, Softmax, Tile, Reshape, Transpose passes (Custom HIP kernels)

### 3. Custom Operator Kernel (`custom-op-rocm/`)
HIP-based execution kernels for GPU inference.

- Shared HIP stream for implicit fusion
- MIOpen backend for convolutions
- hipBLASLt backend for GEMM operations
- Precompiled HIP kernels for element-wise operations
- GPU timeout protection

### 4. HIP Kernels (`kernels/`)
Precompiled GPU kernel implementations using hipcc (compiled at build time).

- Softmax kernel
- Transpose kernel with permutation support
- Tile kernel for tensor tiling
- Mul kernel for element-wise multiplication

### 5. Testing (`test/`)
- GQA layer end-to-end test
- ORT integration test
- Model verification with CPU baseline comparison

### Execution Flow

```
ONNX Graph → Level-1 Pass (GPU Check → Orchestrate Level-2 Passes → Subgraph Merge)
           → Level-2 Passes (Pattern Match → Serialize Params)
           → Custom Op (Load Params → Execute on GPU)
```

**Design Patterns:**
- Level-1/Level-2 pass system for modular pattern matching
- Operation dispatching for extensibility
- Clean separation between compile-time (passes) and runtime (custom op)

## Supported Operations

| Operation | Library | Status |
|-----------|---------|--------|
| Conv (2D) | MIOpen | Implemented |
| Conv + Bias | MIOpen | Implemented |
| Gemm | hipBLASLt | Implemented |
| Gemm + Bias | hipBLASLt | Implemented |
| MatMul | rocBLAS | Implemented |
| Mul | HIP kernel | Implemented |
| Softmax | HIP kernel | Implemented |
| Reshape | HIP Memory | Implemented |
| Transpose | HIP kernel | Implemented |
| Tile | HIP kernel | Implemented |

## Dependencies

**Required:**
- ONNX Runtime
- TheRock SDK (AMD ROCm: HIP, MIOpen, hipBLASLt, rocBLAS)
- MorphiZen Core framework (git submodule)
- LLVM/MLIR (for MorphiZen MLIR backend)
- Protobuf

**Build Tools:**
- CMake 3.29+
- Visual Studio 2022 with C++ workload
- Git with Git LFS

## Build Instructions

```bash
# Set environment
export THEROCK_DIST=$PWD/../therock
export HIP_ARCH=$($THEROCK_DIST/lib/llvm/bin/amdgpu-arch)

# Configure and build
cmake -B ../build/onnx-hipdnn-ep -S . \
  -DTHEROCK_DIST=$THEROCK_DIST \
  -DCMAKE_PREFIX_PATH=$PWD/../local \
  -DCMAKE_INSTALL_PREFIX=$PWD/../local \
  -DHIP_PLATFORM=amd \
  -DHIP_ARCHITECTURES=$HIP_ARCH

cmake --build ../build/onnx-hipdnn-ep --config Release --target install --parallel
```

## Testing

```bash
# Set environment
export PATH="$THEROCK_DIST/bin:$PATH"

# Run GQA test
cd ../local/bin
./test_gqa.exe ../../onnx-hipdnn-ep/test/models/gqa_layer_00.onnx
```

**Benchmark Results:**
```
=== Benchmark Results ===
  Iterations: 1
  Mean latency: 53.50 ms
  Throughput: 18.69 infer/sec
```

## Project Structure

```
onnx-hipdnn-ep/
├── level-1-pass-rocm/          # Level-1 orchestrator pass
├── level-2-pass-rocm-conv/     # Conv pattern matching (MIOpen)
├── level-2-pass-rocm-gemm/     # Gemm pattern matching (hipBLASLt)
├── level-2-pass-rocm-matmul/   # MatMul pattern matching (rocBLAS)
├── level-2-pass-rocm-mul/      # Mul pattern matching (HIP kernel)
├── level-2-pass-rocm-softmax/  # Softmax pattern matching (HIP kernel)
├── level-2-pass-rocm-reshape/  # Reshape pattern matching
├── level-2-pass-rocm-transpose/# Transpose pattern matching (HIP kernel)
├── level-2-pass-rocm-tile/     # Tile pattern matching (HIP kernel)
├── custom-op-rocm/             # GPU execution kernels
├── kernels/                    # Precompiled HIP kernel implementations
├── proto/                      # Protocol buffer definitions
├── test/                       # Test suite
│   └── models/                 # Test models (GQA layer)
├── 3rd-party/morphizen/        # MorphiZen framework (submodule)
└── etc/                        # Configuration files
```

## Key Features

- **Unified EP Architecture**: Single execution provider supporting multiple operation types
- **Level-1/Level-2 Pass System**: Modular pattern matching with orchestration
- **Shared HIP Context**: Operations share the same GPU stream for implicit fusion
- **GPU Timeout Protection**: Prevents indefinite hangs with configurable timeouts
- **Extensibility**: Easy to add new operations via Level-2 passes

## Known Limitations

- Windows only (Linux support planned)
- Static shapes only (no dynamic shapes)
- Single GPU execution

## Future Work

- Linux support
- Dynamic shape support
- Additional operations for full LLM support
- Multi-GPU execution
- INT8 quantization

## Validation

**Test Results:**
- ✅ GQA layer E2E inference
- ✅ ORT integration test
- ✅ Model verification (output matches CPU reference)
- ✅ Build verification (VS2022 on Windows)

---

**Summary:** Initial version of MorphiZen-ROCm Execution Provider for AMD ROCm GPUs with unified architecture supporting Conv, Gemm, MatMul, and custom HIP kernels for LLM workloads. Ready for testing and extension.


cmake/deps.cmake
```diff
@@ -0,0 +1,204 @@
...
+
+file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/version.txt" "${VERSION_INFO}")
+set(MORPHIZEN_VERSION_INFO_FILE "${CMAKE_CURRENT_BINARY_DIR}/version.txt")
+
+# LLVM configuration options (same as onnx-hipdnn-ep)
+message(STATUS "Configuring LLVM/MLIR for onnx-hipdnn-ep")
+set(LLVM_ENABLE_PROJECTS "mlir" CACHE STRING "LLVM projects to build")
+set(LLVM_TARGETS_TO_BUILD "host" CACHE STRING "LLVM targets to build")
+set(LLVM_ENABLE_ASSERTIONS ON CACHE BOOL "Enable LLVM assertions")
+set(LLVM_ENABLE_RTTI ON CACHE BOOL "Enable RTTI in LLVM")
```
MaheshRavishankar: Switch enable rtti OFF.

cmake/deps.cmake
```diff
@@ -0,0 +1,204 @@
...
+set(LLVM_ENABLE_LIBEDIT OFF CACHE BOOL "Enable libedit in LLVM")
+set(LLVM_BUILD_TOOLS ON CACHE BOOL "Build LLVM tools")
+set(LLVM_INSTALL_UTILS ON CACHE BOOL "Install LLVM utilities")
+set(LLVM_INCLUDE_TESTS ON CACHE BOOL "Build LLVM tests")
+set(LLVM_DISABLE_ASSEMBLY_FILES OFF CACHE BOOL "disable assembly")
+set(ZLIB_USE_STATIC_LIBS ON CACHE BOOL "Use static zlib")
+set(LLVM_ENABLE_ZSTD OFF CACHE BOOL "Enable zstd compression")
+
+# LLVM commit hash (same as onnx-hipdnn-ep)
+set(DEP_URL_llvm "https://github.com/llvm/llvm-project.git")
```
MaheshRavishankar: Please make this a submodule as well.

cmake/deps.cmake
```diff
@@ -0,0 +1,204 @@
...
+  string(APPEND VERSION_INFO "${COMP};${COMP_GIT_COMMIT};${COMP_VERSION}\n")
+  unset(COMP_GIT_COMMIT)
+  unset(COMP_VERSION)
+endforeach()
+
+file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/version.txt" "${VERSION_INFO}")
+set(MORPHIZEN_VERSION_INFO_FILE "${CMAKE_CURRENT_BINARY_DIR}/version.txt")
+
+# LLVM configuration options (same as onnx-hipdnn-ep)
+message(STATUS "Configuring LLVM/MLIR for onnx-hipdnn-ep")
```
MaheshRavishankar: Do we even need LLVM in this PR. If this isnt needed for this PR pleaes revmove and add as follow up.

custom-op-rocm/src/custom_op.hpp
```diff
@@ -0,0 +1,372 @@
...
+ * Stores device buffers, cached algorithms, and descriptors for a single
+ * operation
+ */
+struct NodeRuntimeData {
+  // Output buffers for this node (indexed by output_index)
+  std::vector<float *> output_buffers;
+  std::vector<size_t> output_sizes;
+
+  // Weight buffers (cached on GPU)
+  float *d_weight = nullptr;
```
MaheshRavishankar, line 240: The node runtime data doesnt have to be just weight and bias etc. That is very fixed function approach. Try to use a vector of pointers that represent the buffers so that you can reuse this irrespective of the specific computation being executed.

custom-op-rocm/src/custom_op.hpp
```diff
@@ -0,0 +1,372 @@
...
+  float *d_bias = nullptr;
+
+  // Host-side cached weight/bias data
+  std::vector<float> host_weight;
```
MaheshRavishankar, line 244: Same here. Dont make this specific to the node itself. Keep this general list of host-side cache data. That brings me to the next question. Does this need to be `std::vector<..>` Can we just make this is a `{pointer, size}`. At this point all you need is pointer + size in bytes right?

custom-op-rocm/src/custom_op.hpp
```diff
@@ -0,0 +1,372 @@
...
+  std::vector<float> host_bias;
+
+  // Workspace for this node
+  void *workspace = nullptr;
+  size_t workspace_size = 0;
+
+  // Cached algorithm for conv operations
+  // Avoids expensive miopenFindConvolutionForwardAlgorithm() on each inference
+  bool conv_algo_cached = false;
```
MaheshRavishankar, line 253: I would suggest making a separate struct called MIOpenConvRuntimeData to store this information.

custom-op-rocm/src/custom_op.hpp
```diff
@@ -0,0 +1,372 @@
...
+  miopenConvFwdAlgorithm_t cached_conv_algo = miopenConvolutionFwdAlgoGEMM;
+  size_t cached_conv_workspace_size = 0;
```
MaheshRavishankar, line 255: How is this different from workspace_size above?

custom-op-rocm/src/custom_op.hpp
```diff
@@ -0,0 +1,372 @@
...
+
+  // Cached algorithm for gemm operations
+  // Avoids expensive hipblasLtMatmulAlgoGetHeuristic() on each inference
+  bool gemm_algo_cached = false;
```
MaheshRavishankar, line 259: This too. Make this a HipBlasLtMatmulRuntimeData struct to store this information.

custom-op-rocm/src/custom_op.hpp
```diff
@@ -0,0 +1,372 @@
...
+
+  // Subgraph definition (either single-node or multi-node)
+  rocm::RocmSubgraphProto subgraph_;
+
+  // Fallback for single-node case (backward compatibility)
+  bool is_single_node_ = false;
+  rocm::RocmParamProto single_node_proto_;
+
+  // Per-node runtime data
+  mutable std::vector<std::unique_ptr<NodeRuntimeData>> node_data_;
```
MaheshRavishankar, line 358: I am not sure how you map the Node to the NodeRuntimeData.

custom-op-rocm/src/custom_op.cpp
```diff
@@ -0,0 +1,1616 @@
...
+  }
+
+  // Add bias if present
+  if (params.has_bias() && d_bias != nullptr) {
+    int64_t spatial_size = params.out_height() * params.out_width();
+    MY_LOG(2) << "[ROCm CustomOp] Node " << node_id
+              << ": Adding bias (batch=" << params.batch_size()
+              << ", channels=" << params.out_channels()
+              << ", spatial=" << spatial_size << ")";
+    rocm_kernels::add_bias_nchw(output, d_bias, params.batch_size(),
```
MaheshRavishankar, line 943: This does not need to be a separate kernel. This could be added directly to the Convolution call. MIOpen supports conv +  bias- add fusion directly in the conv kernel.

custom-op-rocm/src/custom_op.cpp
```diff
@@ -0,0 +1,1616 @@
...
+  // hipBLAS call: B'[N,K] @ A'[K,M] = D'[N,M] (column-major interpretation)
+  // Result: D'[N,M] col-major == D[M,N] row-major ✓
+
+  hipblasLtMatrixLayout_t layout_A, layout_B, layout_D;
+
+  // For hipBLAS column-major: we pass B as "A" and A as "B"
+  // B is [K, N] in row-major, viewed as [N, K] in column-major, ld = N
+  // A is [M, K] in row-major, viewed as [K, M] in column-major, ld = K
+  // D is [M, N] in row-major, viewed as [N, M] in column-major, ld = N
+  hipblasLtMatrixLayoutCreate(&layout_B, HIP_R_32F, N, K, N); // "A" for hipBLAS
```
MaheshRavishankar, line 997: Not necessary to do now, but something for later. All the handles and host-side data structures like these should be created at compile time (and not at runtime as it is done here). This approach is basically going to bottleneck you on the host-side rather than the device side. The question to answer is if we can store these handles in the final "compiled artifact" or not. Lets revisit this later.

custom-op-rocm/src/custom_op.cpp
```diff
@@ -0,0 +1,1616 @@
...
+    size_t max_workspace = 32 * 1024 * 1024; // 32 MB
+    hipblasLtMatmulPreferenceSetAttribute(
+        pref, HIPBLASLT_MATMUL_PREF_MAX_WORKSPACE_BYTES, &max_workspace,
+        sizeof(max_workspace));
+
+    hipblasLtMatmulHeuristicResult_t results[4];
+    int returned = 0;
+
+    // Swap layout_B and layout_A to match the swapped matmul call
+    hipblasStatus_t status = hipblasLtMatmulAlgoGetHeuristic(
```
MaheshRavishankar, line 1259: Such things definitely need to move out of the execute call. These are expensive and are the things that need to be done at compile time.

MaheshRavishankar (commented):
I left a few comments. I kind of understand what is happening here. Please respond to the comments here and we can land these. There is probably going to be a long-tail of follow ups on these.

cmake/deps.cmake
mingyueliuh: Yes, this PR no longer directly depends on LLVM and should be removed.

cmake/deps.cmake
mingyueliuh: Removed

cmake/deps.cmake
mingyueliuh: fetchcontent llvm was removed

custom-op-rocm/src/custom_op.cpp
AMDmoore: Thanks, noted. As this PR is an initial version provided for technology selection, we will create the other PRs to optimize it if this solution is finally selected.

custom-op-rocm/src/custom_op.hpp
AMDmoore: Thanks Mahesh, all the optimization comments in this file are noted. As this PR is an initial version provided for technology selection, we will create the other PRs to optimize it if this solution is finally selected.

custom-op-rocm/src/custom_op.hpp
AMDmoore: Thanks, noted.

custom-op-rocm/src/custom_op.cpp
AMDmoore: Ok, noted.

custom-op-rocm/src/custom_op.hpp
AMDmoore: Ok, noted.

custom-op-rocm/src/custom_op.hpp
AMDmoore: Ok, noted.

custom-op-rocm/src/custom_op.cpp
zpye: Thanks for your suggestion. Moreover, I think there is a situation that needs to be considered: when these inputs have dynamic shapes, the parameter values here may change with each inference, so the specific parameter values may not be available during compilation. Additionally, changes in shape may also affect the choice of the optimal algorithm, so perhaps we need to know the range of input shapes in advance.

MaheshRavishankar (approved):

PR #5, opened 2026-02-10
feat: MLIR AOT compiler with end-to-end ONNX Runtime integration

wcy123:
## Summary

Complete MLIR-based ahead-of-time compiler that transforms ONNX models into native AMD GPU code, with full ONNX Runtime integration and end-to-end validation.

## What Works Today

**End-to-end test passing:**
```bash
cd backend-mlir-compiler/test
../../build/test/bin/Debug/mlir-e2e-test.exe
# [PASSED] MlirE2ETest.TwoLayerConvSession (4158 ms)
```

- ✅ Full 8-stage compilation pipeline
- ✅ ONNX Runtime integration complete  
- ✅ Mock runtime validation (no GPU required)
- ✅ Single-file deployment (DLL in EPContext)

### Deployment Separation

**Compiler DLL** (`morphizen-mlir-compiler.dll`):
- Depends on LLVM+MLIR
- Used only during model compilation
- Not needed at inference time

**MorphiZen EP** (ONNX Runtime plugin):
- No LLVM/MLIR dependencies
- No ROCm dependencies
- Just orchestrates compilation and loads models

**Compiled Model DLL**:
- Only depends on ROCm libraries (HIP, MIOpen, hipBLASLt)
- Fully self-contained for deployment

**Enables:**
- Offline compilation (compile on server, deploy to edge)
- Cross-compilation (no GPU required during compilation)
- Small runtime footprint (no compiler infrastructure at inference)

### 8-Stage Compilation Pipeline

1. ONNX → HIP Dialect (high-level ops → GPU ops)
2. Buffer Deallocation (ownership-based)
3. Canonicalization (eliminate redundant copies)
4. Memory Pooling (60% memory savings via graph coloring)
5. HIP → LLVM Dialect (lower to runtime wrappers)
6. Generate C Interface (init/compute/cleanup exports)
7. LLVM Optimization (inline accessors, zero-cost abstraction)
8. Native Code Generation (DLL linking)

### Opaque State Pattern

Generated code uses opaque state pointers - never accesses RuntimeState fields directly. This enables:
- RuntimeState layout evolution without recompiling models
- Adding new GPU libraries without breaking compatibility
- Different RuntimeState versions coexisting

## Documentation

- **[DEMO.md](backend-mlir-compiler/doc/DEMO.md)** - Stakeholder presentation (meeting-ready format)
- **Design docs** - 25 detailed documents in morphizen submodule
- **Test README** - Comprehensive troubleshooting workflow

## Current Status

**What works:**
- Full compilation pipeline validated
- E2E test passing with mock runtime
- ONNX Runtime integration complete

**What's pending:**
- GPU validation (real ROCm hardware)
- Performance benchmarking
- Dynamic shape support


backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
```diff
@@ -0,0 +1,30 @@
...
+  size_t rank;    // Number of dimensions
+};
+
+struct span_t {
+  tensor_t *data; // Pointer to tensor array
+  size_t count;   // Number of tensors
+};
+
+// DLL function pointers
+typedef int (*init_fn)(void **);
```
MaheshRavishankar: Could you document this interface a little bit more?  What are the expected arguments etc.

backend-mlir-compiler/custom-op-mlir/src/plugin_registration.cpp
```diff
@@ -0,0 +1,21 @@
+/*
```
MaheshRavishankar, line 1: Why is this file called main.cpp. Typically main.cpp is reserved for executables that are needed?

backend-mlir-compiler/level-1-pass/src/MlirCompiler.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+// CRITICAL: morphizen.hpp must be included before any other morphizen headers
+#include "morphizen/morphizen.hpp"
+#include "morphizen/plugin.hpp"
+#include <chrono>
+#include <fstream>
+#include <glog/logging.h>
+#include <sstream>
+
+namespace hipdnn {
```
fhanuman: Nit: From C++ 17 onwards, we can collapse nested namespaces:

```
namespace hipdnn::level1pass {..}
```

Same comment in other files too.

backend-mlir-compiler/level-1-pass/src/MlirCompiler.cpp
```diff
@@ -0,0 +1,174 @@
...
+            << options_json;
+
+  // Get method with explicit types (avoids template forwarding ref issues)
+  // Signature: CompilerErrorCode (*)(const void*, size_t, const char*, const
+  // char*, CompilerError*)
+  auto func = plugin->get_method<CompilerErrorCode, const void *, size_t,
+                                 const char *, const char *, CompilerError *>(
+      "morphizen_mlir_compile");
+
+  std::cerr << "[MlirCompiler] get_method returned func = " << (void *)func
```
fhanuman: These are probably leftover debug statements? If we need these, we probably can wrap these in `MY_LOG` like you are doing in other places.

backend-mlir-compiler/custom-op-mlir/src/InferenceState.cpp
```diff
@@ -0,0 +1,117 @@
...
+    : state_(state), plugin_(std::move(plugin)), temp_dll_path_(temp_dll_path) {
+}
+
+std::unique_ptr<InferenceState>
+InferenceState::create(const std::vector<uint8_t> &dll_bytes) {
+  MY_LOG(1) << "Loading inference plugin from memory...";
+
+  // Write DLL to temp file (morphizen::Plugin loads from file path)
+  char temp_path[L_tmpnam];
+  if (!std::tmpnam(temp_path)) {
```
fhanuman: `std::tmpnam` is not recommend, please check this once.

backend-mlir-compiler/level-1-pass/src/MlirCompiler.cpp
```diff
@@ -0,0 +1,174 @@
...
+// Generate safe temporary filename (replaces deprecated std::tmpnam)
+std::string generateTempPath() {
+  auto now = std::chrono::system_clock::now();
+  auto timestamp = std::chrono::duration_cast<std::chrono::milliseconds>(
+                       now.time_since_epoch())
+                       .count();
+
+  static int counter = 0;
+  std::string filename = "morphizen_mlir_" + std::to_string(timestamp) + "_" +
+                         std::to_string(counter++) + ".dll";
```
fhanuman: Maybe an oversight, we hardcoding `.dll` extension. Is this intended?

fhanuman (commented):
Some initial comments, will continue looking into the changes.

etc/morphizen_config_mlir.json
```diff
@@ -0,0 +1,27 @@
...
+  "targets": [
+    {
+      "name": "mlir-target",
+      "pass": [
+        "init",
+        "mlir-pass"
+      ],
+      "provider_options": {
+        "pass.init.enable_dump": "1",
+        "pass.init.directory": "vaip_init",
```
MaheshRavishankar: Why is the pass init directory vait_init? There is no vait_init here

backend-mlir-compiler/level-1-pass/src/MlirCompiler.cpp
MaheshRavishankar: +1!

backend-mlir-compiler/level-1-pass/src/MlirCompiler.cpp
```diff
@@ -0,0 +1,174 @@
...
+    return std::nullopt;
+  }
+
+  LOG(INFO) << "Calling morphizen_mlir_compile with JSON options: "
+            << options_json;
+
+  // Get method with explicit types (avoids template forwarding ref issues)
+  // Signature: CompilerErrorCode (*)(const void*, size_t, const char*, const
+  // char*, CompilerError*)
+  auto func = plugin->get_method<CompilerErrorCode, const void *, size_t,
```
MaheshRavishankar: I dont understand where this function is defined. It does not seem to be in this repo/PR. I think this architecture is a problem. All changes that are relevant to the EP/PR dont seem to be in the EP/PR. The indirection through function names really makes it hard to review.

MaheshRavishankar: I see whats happening. This change is happening in the Morphizen repo. This kind of dependence is really hard to manage. The code changes that this PR needs has already been checked into Morphizen. We need clear separation of concerns. Morphizen AFAICS is just the infra for the EP. It shouldnt be containing logic that is relevant to the EP execution.

MaheshRavishankar (dismissed):

fhanuman (changes requested):
This comment is not about the code in this PR — the integration glue here is clean. It’s about the pattern this PR establishes: calling into a compiler that lives inside the MorphiZen submodule. Mahesh flagged a similar concern about this dependency in his review comment. Since all subsequent work will follow this pattern, I think it’s worth getting right before we land it.

I looked at the morphizen-mlir-compiler source in the MorphiZen submodule. The compiler doesn’t include anything from morphizen-core, morphizen-utils, or any other MorphiZen module — I checked  includes across the lib/ and include/ directories. Its dependencies are MLIR/LLVM, onnx-mlir for the ONNX dialect, protobuf, and GPU runtime libraries (HIP, MIOpen, hipBLASLt). It even declares its own project(morphizen-mlir-compiler) in CMake with its own find_package calls. In other words, this is a self-contained compiler that happens to live inside the MorphiZen repo.

The problem is that even though the compiler has no source-level dependency on MorphiZen, it’s not built or tested independently. It’s always built as part of MorphiZen’s tree. So anyone who wants to use these passes — from hip-opt, from onnx-mlir, or from a different framework — has to pull in all of MorphiZen. And every compiler change needs a MorphiZen PR first, then a submodule bump here.

There’s also a reuse angle worth considering. Some of this MLIR compiler infrastructure — the pipeline structure, the interface generation, the pass framework — could be reused for an NPU backend down the line. Decoupling the compiler from MorphiZen now would make that much easier than trying to untangle it later.

MorphiZen already loads the compiler as a plugin via dlopen at runtime. The source ownership should match that boundary — the compiler should be a standalone library that MorphiZen consumes, not a module it contains.

backend-mlir-compiler/custom-op-mlir/src/InferenceState.cpp
```diff
@@ -0,0 +1,117 @@
...
+  void *state = nullptr;
+  int ret = init_fn(&state);
+  if (ret != 0) {
+    LOG(FATAL) << "inference_init() failed with code: " << ret;
+  }
+
+  MY_LOG(1) << "Inference state initialized";
+
+  return std::unique_ptr<InferenceState>(
+      new InferenceState(state, std::move(plugin), dll_path));
```
fhanuman: Nit:

`std::make_unique<InferenceState>(state, std::move(plugin), dll_path)` 

is preferred over

 `std::unique_ptr<InferenceState>(new InferenceState(...))`

fhanuman (dismissed):

etc/morphizen_config_mlir.json
wcy123: Good catch on the naming. To clarify the intent:

`morphizen-pass-init` is a troubleshooting-only plugin — it is not part of the core compilation pipeline. All three `provider_options` (`enable_dump`, `directory`, `filename`) are optional.

The `directory` value only matters when `enable_dump` is `"1"`. When enabled, `PassContext::get_dump_directory()` uses it as the dump path and calls `std::filesystem::create_directories` to auto-create it if it does not exist. When `enable_dump` is `"0"`, the directory value is ignored entirely and the framework falls back to a default under the system temp directory (`C:\temp\morphizen_dumps\<cache_key>` on Windows, `/tmp/morphizen_dumps/<cache_key>` on Linux), also auto-created.

That said, I agree `vaip_init` is not a meaningful name for this context.

backend-mlir-compiler/level-1-pass/src/MlirCompiler.cpp
wcy123: Thank you both for raising this --  it is an important architectural point and you are right that the current hosting creates unnecessary friction for reviewers.

To clarify the technical boundary: `morphizen-mlir-compiler` has **zero source-level dependencies on MorphiZen**. It includes no morphizen-core, morphizen-graph, or any other MorphiZen headers. Its only dependencies are MLIR/LLVM, ONNX-MLIR, and Protobuf. It declares its own `project()` with its own `find_package` calls. The coupling is purely through two stable C ABI contracts - one at compile time, one at runtime.

**Compile-time contract** -- how the EP invokes the compiler (`morphizen-mlir-compiler.dll`):
```c
// Defined in morphizen-mlir-compiler: lib/CInterface/CompilerAPI.cpp
CompilerErrorCode morphizen_mlir_compile(
    const void*    input_mlir,    // MLIR bytecode (binary-safe, not null-terminated)
    size_t         input_size,    // size of input in bytes
    const char*    output_path,   // where to write the compiled artifact
    const char*    options_json,  // e.g. {"opt_level": 2, "output_mode": "OUTPUT_MODE_DLL"}
                                  // pass NULL for defaults
    CompilerError* error          // error details on failure, can be NULL
);
```

**Runtime contract** --  how the EP invokes the compiled model (`model.dll`):
```c
// Defined in this repo: backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
int inference_init(void** out_state);      // allocate GPU handles, upload weights
int inference_compute(void*    state,      // opaque context from inference_init
                      span_t*  inputs,     // host-memory input tensors
                      span_t*  outputs);   // host-memory output tensors
int inference_cleanup(void* state);        // release all GPU resources
```

The EP has no knowledge beyond these two contracts -- no MLIR, no LLVM, no compiler internals. Any compiler that produces a binary exporting the runtime contract works as a drop-in replacement.

The right long-term home for `morphizen-mlir-compiler` is its own standalone repository, with `compiler_api.h` visible there rather than buried in the MorphiZen submodule. The right long-term direction is to move it out, so that compiler changes are reviewable independently without requiring a MorphiZen PR and submodule bump. I will raise this with the team.

backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
wcy123: Done — each typedef now has a doc comment explaining the arguments, return value, and lifecycle role. See `backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp`.

backend-mlir-compiler/level-1-pass/src/MlirCompiler.cpp
MaheshRavishankar: ok, ill review it tomorrow, but can you point me to where the implementation of `morphizem_mlir_compile` is?

backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
wcy123: Done — \ now documents each typedef with parameters (name, direction, semantics), return value codes, and a lifecycle example at the top of the function-pointer block.

The full interface design (two-layer architecture, dynamic shape rationale, error handling policy) is in \ inside the MorphiZen repo — accessible here: https://github.com/ROCm/MorphiZen/blob/main/morphizen-mlir-compiler/doc/design/mlir/INTERFACE-DESIGN.md

backend-mlir-compiler/level-1-pass/src/MlirCompiler.cpp
MaheshRavishankar: Just spoke to Elliott. I think I missed the fact that this method is not implemented in this PR yet and it will be implemented later. Ill unblock it for now. I still have my review queue. Ill post some comments post landing which can be addressed later.

PR #19, opened 2026-03-07
Grow-on-demand memory pool via hip.get_pool

amd-mingw:
Replaces per-inference hip.alloc/hip.free with hip.get_pool(%ctx, %pool_size). The pool stays alive across calls and only reallocates when the requested size exceeds current capacity.

Pool size is now an SSA operand -- a compile-time constant for static shapes, runtime arithmetic for dynamic. The old module-level metadata attributes (hipdnn.pool_size, buffer_count, buffer_offsets) are removed since they are no longer consumed.

Compile-time liveness analysis and buffer offset assignment are unchanged.

Note: No runtime implementation yet. as #17 is designed for static mem pool.


tools/hip-mlir-opt/hip-mlir-opt.cpp
```diff
@@ -98,6 +98,8 @@ struct HipDstBufferizableModel
 
 void registerHipBufferizableOpInterfaceModels(mlir::DialectRegistry &registry) {
   registry.addExtension(+[](mlir::MLIRContext *ctx, mlir::hip::HipDialect *) {
+    mlir::hip::ConvOp::attachInterface<
```
MaheshRavishankar, line 101: This seems unrelated. Was this missed in previous PRs?

MaheshRavishankar (approved):

tools/hip-mlir-opt/hip-mlir-opt.cpp
BoarQing: Yes. This is missing from the Conv OP PR.

fhanuman (approved):
LGTM, thanks for making the changes. Just a heads-up: I made some changes to the `PoolAllocs` pass, so you may need to rebase to resolve merge conflicts.

PR #20, opened 2026-03-07
Add Conv operation and lit tests

mingyueliuh:
## Summary
This PR adds 2D convolution operation support to the HIP dialect with complete ONNX→HIP→LLVM lowering pipeline, comprehensive lit tests, and proper lit test infrastructure. The implementation follows PR-17 patterns and best practices.

## Architecture
- **onnx.Conv** → **hip.conv** → **llvm.call @wrap_miopenConvolutionForward**
- Uses Destination-Passing Style (DPS) interface via Hip_DpsOp base class
- Supports optional bias operand
- Forwards all Conv attributes (kernel_shape, strides, pads, dilations, group)
- Passes explicit shape parameters to runtime (23-parameter signature)

### 1. Hip_DpsOp Base Class
Added `Hip_DpsOp` base class in `HipOps.td` for all DPS operations, providing MemoryEffectsOpInterface and DestinationStyleOpInterface automatically.

### 2. Conv Operation
- `Hip_ConvOp` inherits from `Hip_DpsOp<"conv">` with declarative assembly format
- Implements `getDpsInitsMutable()` and `getEffects()` in `HipDialect.cpp`
- `ConvToHip` pattern converts ONNX to HIP with attribute extraction
- `ConvOpLowering` lowers to `wrap_miopenConvolutionForward` with 23-parameter signature (explicit shape passing, i32 return type)

### 3. Lit Test Infrastructure
- Set up `test/lit/` directory with CMake integration (following PR-17 pattern)
- Added ONNX→HIP conversion tests (5 test cases)
- Added HIP→LLVM lowering test (validates 23-parameter signature)
- Custom target: `check-hip-mlir-lit`

### Tensor Mode (pre-bufferization)
```mlir
func.func @conv_example(%ctx: !hip.context,
                        %input: tensor<1x3x224x224xf32>,
                        %weights: tensor<64x3x7x7xf32>,
                        %bias: tensor<64xf32>) -> tensor<1x64x112x112xf32> {
  %init = tensor.empty() : tensor<1x64x112x112xf32>
  %result = hip.conv(%ctx) ins(%input, %weights, %bias :
                                tensor<1x3x224x224xf32>,
                                tensor<64x3x7x7xf32>,
                                tensor<64xf32>)
                           outs(%init : tensor<1x64x112x112xf32>)
                           {kernel_shape = [7, 7],
                            strides = [2, 2],
                            pads = [3, 3, 3, 3],
                            dilations = [1, 1],
                            group = 1 : i64} : tensor<1x64x112x112xf32>
  return %result : tensor<1x64x112x112xf32>
}
```

### Memref Mode (post-bufferization)
```mlir
func.func @conv_memref(%ctx: !hip.context,
                       %input: memref<1x3x224x224xf32, 1>,
                       %weights: memref<64x3x7x7xf32, 1>,
                       %bias: memref<64xf32, 1>,
                       %output: memref<1x64x112x112xf32, 1>) {
  hip.conv(%ctx) ins(%input, %weights, %bias :
                     memref<1x3x224x224xf32, 1>,
                     memref<64x3x7x7xf32, 1>,
                     memref<64xf32, 1>)
                 outs(%output : memref<1x64x112x112xf32, 1>)
                 {kernel_shape = [7, 7], strides = [2, 2],
                  pads = [3, 3, 3, 3], dilations = [1, 1], group = 1 : i64}
  return
}
```

### Manual Testing (Verified ✅)
```bash
# ONNX to HIP conversion
hip-mlir-opt test/lit/Conversion/onnx-to-hip/test_conv.mlir \
  --hip-add-context-arg --convert-onnx-to-hip | FileCheck test_conv.mlir

# HIP to LLVM lowering
hip-mlir-opt test/lit/Conversion/hip-to-llvm/test_conv.mlir \
  --convert-hip-to-llvm | FileCheck test_conv_lowering.mlir
```

### Build and Run Lit Tests
```bash
cmake --build . --config Release --target check-hip-mlir-lit
```


include/hip/Dialect/IR/HipOps.td
```diff
@@ -15,6 +15,19 @@ include "mlir/IR/OpBase.td"
 class Hip_Op<string mnemonic, list<Trait> traits = []> :
     Op<Hip_Dialect, mnemonic, traits>;
 
+// Base class for all HIP destination-passing-style (DPS) compute ops.
+// Provides MemoryEffectsOpInterface, DestinationStyleOpInterface, and the
+// Variadic<AnyRankedTensor>:$result_tensors result (non-empty only at the
+// tensor level; empty after bufferization).
+class Hip_DpsOp<string mnemonic, list<Trait> traits = []> :
```
fhanuman, line 22: Could we also please update the already existing compute op definitions (e.g., `HipblasltMatmulOp`, `MiopenSoftmaxOp`, `MiopenRmsNormOp`) to inherit from `Hip_DpsOp`? This would reduce duplication and keep all ops consistent.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -60,6 +69,194 @@ static Value extractMemRefPtr(Value memrefDesc,
...
+      biasPtr = LLVM::ZeroOp::create(rewriter, loc, ptrType);
+    }
+
+    // Extract shapes from memref types (static shapes known at compile time)
+    auto inputType = cast<MemRefType>(op.getInput().getType());
+    auto weightsType = cast<MemRefType>(op.getWeights().getType());
+    auto outputType = cast<MemRefType>(op.getOutput().getType());
+
+    // Input shape: [N, C, H, W]
+    auto inputShape = inputType.getShape();
```
fhanuman: I know we have limited support for dynamic shapes, however, we should have proper guards. 

```
if (!inputType.hasStaticShape())
  return op.emitError("hip.conv lowering requires static shapes");
```

lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -488,6 +488,111 @@ SoftmaxToHip::matchAndRewrite(mlir::Operation *op,
...
+                           mlir::PatternRewriter &rewriter) const {
+  auto ctxOrFailure = getContextArg(op, rewriter);
+  if (mlir::failed(ctxOrFailure))
+    return mlir::failure();
+  mlir::Value context = *ctxOrFailure;
+
+  mlir::Location loc = op->getLoc();
+  mlir::Value input = op->getOperand(0);
+  mlir::Value weights = op->getOperand(1);
+  mlir::Value bias = (op->getNumOperands() > 2) ? op->getOperand(2) : nullptr;
```
fhanuman: Isn't this condition always `true`?

Maybe the below check helps?

```
bool hasBias = !mlir::isa<NoneType>(op->getOperand(2).getType());
```

test/lit/lit.cfg.py
```diff
@@ -0,0 +1,75 @@
...
+# Attributes guaranteed to exist when this file runs (set by lit.site.cfg.py):
+#   config.llvm_tools_dir  — path to FileCheck, not, etc.
+#   config.hip_build_dir   — CMAKE_BINARY_DIR
+#   config.hip_build_mode  — "Debug" or "Release" (resolved at runtime for VS)
+
+import os
+import lit.formats
+import lit.llvm
+
+# ruff: noqa: F821  (config/lit_config injected by lit framework)
```
fhanuman, line 28: There is an already existing (`lit.cfg.py`) file, could we have updated it with the new content instead of creating a new file?

fhanuman (changes requested):
This is pretty neat, thanks for creating the small PR so it’s easy to review.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -60,6 +69,194 @@ static Value extractMemRefPtr(Value memrefDesc,
...
+    if (inputShape.size() != 4) {
+      return op.emitError("Input must be rank-4 tensor [N, C, H, W]");
+    }
+    Value inputN = createI64Const(inputShape[0]);
```
MaheshRavishankar: What if the shape is dynamic?

lib/Dialect/IR/HipDialect.cpp
```diff
@@ -264,6 +264,24 @@ static void printSingleInitDpsOp(OpAsmPrinter &p, Operation *op, Value ctx,
...
+// Uses declarative assemblyFormat - parse/print auto-generated by TableGen
+//===----------------------------------------------------------------------===//
+
+MutableOperandRange ConvOp::getDpsInitsMutable() { return getOutputMutable(); }
+
+void ConvOp::getEffects(
+    SmallVectorImpl<SideEffects::EffectInstance<MemoryEffects::Effect>>
+        &effects) {
+  // Read inputs (conservative: assumes reads from memory)
+  effects.emplace_back(MemoryEffects::Read::get(),
```
MaheshRavishankar: The memory effect is only needed if the operand types are `memref`. For `tensor` types there is no memory effect.

fhanuman (approved):

include/hip/Dialect/IR/HipOps.td
mingyueliuh: 11 Ops changed to use Hip_DpsOp base class

lib/Conversion/OnnxToHip/OnnxToHip.cpp
mingyueliuh: changed

lib/Dialect/IR/HipDialect.cpp
mingyueliuh: Changed ConvOp::getEffects() to use the existing emitDpsMemoryEffects() helper which only adds memory effects when operands are memrefs (not tensors), matching the pattern used by all other DPS ops.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
mingyueliuh: Changed to use MemRefDescriptor for dimension extraction, which supports both static shapes (using compile-time constants for optimization) and dynamic shapes (extracting from runtime descriptor)

lib/Conversion/HipToLLVM/HipToLLVM.cpp
mingyueliuh: Changed to use MemRefDescriptor for dimension extraction, which supports both static shapes (using compile-time constants for optimization) and dynamic shapes (extracting from runtime descriptor)

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -60,6 +69,211 @@ static Value extractMemRefPtr(Value memrefDesc,
...
+    // Helper: Get dimension value (static constant or dynamic descriptor size)
+    auto getDim = [&](MemRefDescriptor &desc, MemRefType type,
+                      unsigned dimIdx) -> Value {
+      int64_t dimSize = type.getShape()[dimIdx];
+      if (!ShapedType::isDynamic(dimSize)) {
+        // Static dimension: use compile-time constant
+        return createI64Const(dimSize);
+      }
+      // Dynamic dimension: extract from runtime descriptor
+      return desc.size(rewriter, loc, dimIdx);
```
MaheshRavishankar, line 183: I guess this works too, but what I was suggesting was more like

```
auto getDim = [&](Value v, int64_t dim) {
  Value dim = memref::DimOp::create(rewriter, loc, v, dim);
}
```

test/lit/lit.cfg.py
mingyueliuh: @fhanuman  
`test/lit/lit.cfg.py` replaces the old `test/lit.cfg.py` (which was never integrated with CMake). I've also moved the orphaned `test/Dialect/` and `test/Pipeline/` tests into `test/lit/` so they're now discovered and run by the build system.

Do you think this approach works?

lib/Conversion/HipToLLVM/HipToLLVM.cpp
mingyueliuh: By the time ConvOpLowering runs, `memref.dim` ops have already been lowered into LLVM descriptor size extractions. At this point, we are already past the memref dialect.

The operands are already represented as LLVM structs. MemRefDescriptor::size() is the equivalent of memref.dim after lowering.

test/lit/lit.cfg.py
amd-mingw: Hi @fhanuman , for the lit.cfg.py, I think we just going with @mingyueliuh mentioned now, and I land this PR to unblock the next sets PRs. If you have more concerns, we can raise another PR to update it in the future. Thanks!

PR #22, opened 2026-03-07
Add hip-add-context-arg pass; replace per-function handle lifecycle with !hip.context argument

BoarQing:
**Why:** Each function was creating and destroying its own MIOpen/hipBLASLt handles internally. This is expensive and redundant for multi-function models and repeated inference. Moving to a caller-provided context lets the runtime create handles once and reuse them across calls.

Replace explicit hip.create_handle/hip.destroy_handle with !hip.context passed as func arg 0.

- Rename Hip_HandleType to Hip_ContextType, remove CreateHandleOp/DestroyHandleOp
- Add hip-add-context-arg pass (stage 0 in onnx-to-hip-pipeline)
- Update OnnxToHip, LowerAllocs, HipToLLVM to use context from arg 0
- Update all tests (8/8 pass), examples, and docs

**Suggested review order:** The core changes are in `lib/Dialect/Transforms/AddHipContextArg.cpp` (new pass), `include/hip/Dialect/IR/HipOps.td` + `HipTypes.td` (type/op changes), then `lib/Conversion/OnnxToHip/OnnxToHip.cpp` and `lib/Conversion/HipToLLVM/HipToLLVM.cpp` (pass updates). The rest is mechanical renaming in tests, examples, and docs.


lib/Dialect/Transforms/AddHipContextArg.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+struct HipAddContextArgPass
+    : public impl::HipAddContextArgPassBase<HipAddContextArgPass> {
+
+  void runOnOperation() override {
+    ModuleOp module = getOperation();
+    MLIRContext *ctx = &getContext();
+    Type hipCtxType = ContextType::get(ctx);
+
+    module.walk([&](func::FuncOp funcOp) {
```
fhanuman, line 30: Not a blocker — we currently have a single `func.func` per module for ONNX models, so no call sites are affected. Something to keep in mind for the future: once we support multi-function modules, hip-add-context-arg should also update `func.call` ops to pass the new `!hip.context` argument to callees.

lib/Dialect/Transforms/LowerAllocs.cpp
```diff
@@ -97,11 +97,14 @@ void LowerAllocsPass::runOnOperation() {
...
-    funcOp.emitRemark("no hip.create_handle found; skipping alloc lowering");
+  // Get !hip.context from function argument 0.
+  Value ctx;
+  if (!funcOp.getArguments().empty() &&
+      isa<ContextType>(funcOp.getArgument(0).getType()))
+    ctx = funcOp.getArgument(0);
+  if (!ctx) {
+    funcOp.emitRemark(
+        "no !hip.context argument found; skipping alloc lowering");
     return;
```
fhanuman: Consider `return signalPassFailure()` here instead of a silent `return`. If the pipeline is correctly configured, `hip-add-context-arg` should have already inserted `!hip.context`. A missing `context arg` likely indicates a pipeline misconfiguration, and failing fast would make it easier to diagnose than a silent no-op followed by downstream failures in `convert-hip-to-llvm`

fhanuman (approved):
Thanks for making the changes. Except for minor comments, changes look good.

docs/hip_tools_usage_guide.md
```diff
@@ -283,16 +282,15 @@ carved from a single memory pool via `memref.view`:
 ```mlir
 module {
   memref.global "private" constant @weight : memref<64x64xf32> = dense<"0x...">
-  func.func @main_graph(%X: memref<2x64x64xf32, strided<[?,?,?], offset: ?>>,
+  func.func @main_graph(%ctx: !hip.context,
```
MaheshRavishankar, line 285: Ok, so the change now is that the context is constructed before the "main_graph" is called and passed as an argument so that the context create/destroy is not done within the `main_graph`. Thats a good idea!

MaheshRavishankar (commented):
Nice! Thanks for making PR that was easy to review!

docs/hip_tools_usage_guide.md
fhanuman: Hi Mahesh, Yes, this was one of the things we discussed in the yesterday's meeting.

PR #23, opened 2026-03-09
Add module metadata generation to convert-onnx-to-hip pass

BoarQing:
## Summary

- Emit `hipdnn.{input,output}_{count,shapes,element_sizes}` module attributes from `@main_graph`'s original function signature before lowering.
- Metadata is captured before the function-processing loop since pattern rewrites change the signature.
- Required by two downstream consumers to be ported next:
  - `HipToLLVM::transformMainFunction()` -- rewrites `@main_graph` into an array-based C ABI wrapper using input/output counts and shapes.
  - `GenerateInterface` pass -- reads all six attributes to generate `inference_init/compute/cleanup/metadata` entry points.

## Test plan

- [x] Build succeeds with no new warnings
- [x] All existing lit tests pass (no regressions)
- Dedicated metadata lit tests deferred; coverage will come with e2e tests when downstream consumers (GenerateInterface, HipToLLVM metadata usage) are ported

### DIFF
--- a/lib/Conversion/OnnxToHip/OnnxToHip.cpp
+++ b/lib/Conversion/OnnxToHip/OnnxToHip.cpp
@@ -621,6 +621,76 @@ static mlir::LogicalResult convertComputeOps(mlir::func::FuncOp funcOp,
 }
 
 //===----------------------------------------------------------------------===//
+// Module metadata generation
+//===----------------------------------------------------------------------===//
+
+/// Generate module metadata attributes required by GenerateInterfacePass.
+/// Must be called BEFORE patterns transform function signatures.
+static mlir::LogicalResult generateModuleMetadata(mlir::ModuleOp module) {
+  auto mainFunc = module.lookupSymbol<mlir::func::FuncOp>("main_graph");
+  if (!mainFunc)
+    return mlir::success();
+
+  auto originalFuncType = mainFunc.getFunctionType();
+  mlir::OpBuilder builder(module.getContext());
+
+  int64_t inputCount = originalFuncType.getNumInputs();
+  llvm::SmallVector<mlir::Attribute> inputShapes;
+  llvm::SmallVector<int64_t> inputElementSizes;
+
+  for (mlir::Type inputType : originalFuncType.getInputs()) {
+    if (mlir::isa<mlir::hip::ContextType>(inputType)) {
+      --inputCount;
+      continue;
+    }
+    if (auto tensorType = mlir::dyn_cast<mlir::RankedTensorType>(inputType)) {
+      llvm::SmallVector<int64_t> shape(tensorType.getShape().begin(),
+                                       tensorType.getShape().end());
+      inputShapes.push_back(builder.getDenseI64ArrayAttr(shape));
+      inputElementSizes.push_back(
+          tensorType.getElementType().getIntOrFloatBitWidth() / 8);
+    } else {
+      llvm::errs() << "[ONNX→HIP] Warning: non-tensor input in @main_graph, "
+                      "skipping metadata\n";
+      return mlir::success();
+    }
+  }
+
+  int64_t outputCount = originalFuncType.getNumResults();
+  llvm::SmallVector<mlir::Attribute> outputShapes;
+  llvm::SmallVector<int64_t> outputElementSizes;
+
+  for (mlir::Type resultType : originalFuncType.getResults()) {
+    if (auto tensorType = mlir::dyn_cast<mlir::RankedTensorType>(resultType)) {
+      llvm::SmallVector<int64_t> shape(tensorType.getShape().begin(),
+                                       tensorType.getShape().end());
+      outputShapes.push_back(builder.getDenseI64ArrayAttr(shape));
+      outputElementSizes.push_back(
+          tensorType.getElementType().getIntOrFloatBitWidth() / 8);
+    } else {
+      llvm::errs() << "[ONNX→HIP] Warning: non-tensor output in @main_graph, "
+                      "skipping metadata\n";
+      return mlir::success();
+    }
+  }
+
+  module->setAttr("hipdnn.input_count", builder.getI64IntegerAttr(inputCount));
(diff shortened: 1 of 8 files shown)


lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -510,6 +510,82 @@ static mlir::LogicalResult convertComputeOps(mlir::func::FuncOp funcOp,
...
+//===----------------------------------------------------------------------===//
+// Module metadata generation
+//===----------------------------------------------------------------------===//
+
+/// Generate module metadata attributes required by GenerateInterfacePass.
+/// Must be called BEFORE patterns transform function signatures.
+static mlir::LogicalResult generateModuleMetadata(mlir::ModuleOp module) {
+  auto mainFunc = module.lookupSymbol<mlir::func::FuncOp>("main_graph");
+  if (!mainFunc)
+    return mlir::success();
```
fhanuman: If `main_graph` is not found, this should return `mlir::failure()` (or emit an error). In our pipeline, `main_graph` is always expected to exist at this point?

lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -510,6 +510,82 @@ static mlir::LogicalResult convertComputeOps(mlir::func::FuncOp funcOp,
...
+    if (mlir::isa<mlir::hip::ContextType>(inputType)) {
+      --inputCount;
+      continue;
+    }
+    if (auto tensorType = mlir::dyn_cast<mlir::RankedTensorType>(inputType)) {
+      llvm::SmallVector<int64_t> shape(tensorType.getShape().begin(),
+                                       tensorType.getShape().end());
+      inputShapes.push_back(builder.getDenseI64ArrayAttr(shape));
+      inputElementSizes.push_back(
+          tensorType.getElementType().getIntOrFloatBitWidth() / 8);
```
fhanuman: I think this is not correct right? We support quantized models as well, so this should query `getIntOrFloatBitWidth`?

Also, it would be nice to have below guard to make the intention clear here. Also, not sure if ORT allows `complex`  or `quantized` types as input today. We probably need to have proper validation to guard against these kind of issues.

```
if (!tensorType.getElementType().isIntOrFloat()) {
    mainFunc.emitError("unsupported element type in @main_graph: ")
        << tensorType;
    return mlir::failure();
}
```

lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -510,6 +510,82 @@ static mlir::LogicalResult convertComputeOps(mlir::func::FuncOp funcOp,
...
+  llvm::SmallVector<mlir::Attribute> outputShapes;
+  llvm::SmallVector<int64_t> outputElementSizes;
+
+  for (mlir::Type resultType : originalFuncType.getResults()) {
+    if (auto tensorType = mlir::dyn_cast<mlir::RankedTensorType>(resultType)) {
+      llvm::SmallVector<int64_t> shape(tensorType.getShape().begin(),
+                                       tensorType.getShape().end());
+      outputShapes.push_back(builder.getDenseI64ArrayAttr(shape));
+      outputElementSizes.push_back(
+          tensorType.getElementType().getIntOrFloatBitWidth() / 8);
```
fhanuman: Same comment as above

lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -510,6 +510,82 @@ static mlir::LogicalResult convertComputeOps(mlir::func::FuncOp funcOp,
...
+    } else {
+      mainFunc.emitError("non-tensor output type in @main_graph: ")
+          << resultType;
+      return mlir::failure();
+    }
+  }
+
+  module->setAttr("hipdnn.input_count", builder.getI64IntegerAttr(inputCount));
```
fhanuman, line 691: Could you please add example IR to the PR description? I am just curious why we need this information.

lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -510,6 +510,82 @@ static mlir::LogicalResult convertComputeOps(mlir::func::FuncOp funcOp,
...
+  module->setAttr("hipdnn.input_element_sizes",
+                  builder.getDenseI64ArrayAttr(inputElementSizes));
+  module->setAttr("hipdnn.output_count",
+                  builder.getI64IntegerAttr(outputCount));
+  module->setAttr("hipdnn.output_shapes", builder.getArrayAttr(outputShapes));
+  module->setAttr("hipdnn.output_element_sizes",
+                  builder.getDenseI64ArrayAttr(outputElementSizes));
+
+  auto remark = module.emitRemark("module metadata: ");
+  remark << "input_count=" << inputCount
```
fhanuman: Does it makes sense to make this `debug-only` instead?

lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -510,6 +510,82 @@ static mlir::LogicalResult convertComputeOps(mlir::func::FuncOp funcOp,
...
+          tensorType.getElementType().getIntOrFloatBitWidth() / 8);
+    } else {
+      mainFunc.emitError("non-tensor output type in @main_graph: ")
+          << resultType;
+      return mlir::failure();
+    }
+  }
+
+  module->setAttr("hipdnn.input_count", builder.getI64IntegerAttr(inputCount));
+  module->setAttr("hipdnn.input_shapes", builder.getArrayAttr(inputShapes));
```
fhanuman, line 692: Also, please see if we can add a test point to lock-down the change.

fhanuman (changes requested):

MaheshRavishankar (commented):
This needs a test.

Also not sure how we are going to handle the cases of the dynamic shapes.

fhanuman (approved):

BoarQing:
@MaheshRavishankar  @MaheshRavishankar This PR is updated to address your comments.

PR #31, opened 2026-03-11
Add hip.get_constant and index-based constant resolution

BoarQing:
## Summary

- Add `hip.get_constant(%ctx, %index)` op to retrieve pre-uploaded GPU constants by index
- Rewrite `hip-resolve-extern-constants` to emit `hip.get_constant` instead of `memref.view`
- Externalize constants via `morphizen::FileSystem`, emitting `hip.external_data{index, offset, size}` and `hipdnn.constant_sizes`/`hipdnn.constant_offsets` module attributes
- Lower `hip.get_constant` to `hipdnn_ep_constant_get(state, index)` runtime call in HipToLLVM

## Suggested review order

1. **Interfaces** — `file_io.hpp`, `DiskFileSystem.h`
   The abstract `FileSystem` interface (EP runtime provides its own impl in deployment). `DiskFileSystem` is a disk-backed impl for CLI tools/debugging only.

2. **New op** — `HipOps.td`, `HipDialect.cpp`
   Defines `hip.get_constant(ctx, index) → memref`. Understanding this makes the passes below straightforward.

3. **Externalization** — `OnnxToHip.cpp` + tests
   Writes constants to sidecar binary via `FileSystem`, tags globals with `hip.external_data`. Produces the IR that step 4 consumes.

4. **Resolution** — `ResolveExternConstants.cpp` + tests
   Replaces `memref.get_global` → `hip.get_constant` using the index from step 3.

5. **LLVM lowering** — `HipToLLVM.cpp` + test
   Lowers `hip.get_constant` to `hipdnn_ep_constant_get` call + memref descriptor.

## Test plan

- [x] All 29 LIT tests pass

### DIFF
--- a/include/hip/Dialect/IR/HipOps.td
+++ b/include/hip/Dialect/IR/HipOps.td
@@ -96,6 +96,39 @@ def Hip_FreeOp : Hip_Op<"free",
   }];
 }
 
+// ===== Constant Management ===================================================
+
+def Hip_GetConstantOp : Hip_Op<"get_constant", [
+  DeclareOpInterfaceMethods<MemoryEffectsOpInterface>
+]> {
+  let summary = "Retrieve constant from GPU memory";
+  let description = [{
+    Retrieves a pre-uploaded constant from state->gpu_constants[index].
+
+    Arguments:
+    - ctx: HIP execution context (contains gpu_constants array)
+    - index: Index in gpu_constants array (0, 1, 2, ...)
+
+    Returns:
+    - memref: GPU memory reference to the constant
+
+    Example:
+    ```mlir
+    %weight = hip.get_constant(%ctx, %index_0) : memref<64x3x3x3xf32, 1>
+    ```
+  }];
+
+  let arguments = (ins
+    Hip_ContextType:$ctx,
+    I64:$index
+  );
+  let results = (outs AnyMemRef:$result);
+
+  let assemblyFormat = [{
+    `(` $ctx `,` $index `)` attr-dict `:` type($result)
+  }];
+}
+
 // ===== Region ops (structural grouping, no runtime) =========================
 
 def Hip_MiopenGraphOp : Hip_Op<"miopen.graph", [SingleBlock, NoTerminator]> {
--- a/include/hip/Support/DiskFileSystem.h
+++ b/include/hip/Support/DiskFileSystem.h
@@ -1,0 +1,116 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+#pragma once
+
+#include "morphizen-foundation/file_io.hpp"
+#include <cstdio>
+#include <filesystem>
+#include <string>
+#ifdef _WIN32
+#include <io.h>
+#endif
+
+namespace mlir::hip {
+
+class DiskFileReader : public morphizen::FileReader {
+public:
+  explicit DiskFileReader(const char *path)
+      : file_(std::fopen(path, "rb")), size_(0) {
+    if (!file_)
+      return;
+#ifdef _WIN32
+    _fseeki64(file_, 0, SEEK_END);
+    size_ = static_cast<std::size_t>(_ftelli64(file_));
+    _fseeki64(file_, 0, SEEK_SET);
+#else
+    fseeko(file_, 0, SEEK_END);
+    size_ = static_cast<std::size_t>(ftello(file_));
+    fseeko(file_, 0, SEEK_SET);
+#endif
+  }
+
+  ~DiskFileReader() override {
+    if (file_)
+      std::fclose(file_);
+  }
+
+  bool ok() const { return file_ != nullptr; }
+  std::size_t size() const override { return size_; }
+
+  void rewind() const override {
+    if (file_) {
+#ifdef _WIN32
+      _fseeki64(file_, 0, SEEK_SET);
+#else
+      fseeko(file_, 0, SEEK_SET);
+#endif
+    }
+  }
+
+  std::size_t fread(void *buffer, std::size_t size) const override {
+    if (!file_)
+      return 0;
+    return std::fread(buffer, 1, size, file_);
+  }
+
+private:
+  std::FILE *file_;
+  std::size_t size_;
+};
+
+class DiskFileWriter : public morphizen::FileWriter {
+public:
+  explicit DiskFileWriter(const char *path) : file_(std::fopen(path, "wb")) {}
+
+  ~DiskFileWriter() override {
+    if (file_)
+      std::fclose(file_);
+  }
+
+  bool ok() const { return file_ != nullptr; }
+
+  std::size_t fwrite(const void *buffer, std::size_t size) const override {
+    if (!file_)
+      return 0;
+    return std::fwrite(buffer, 1, size, file_);
+  }
+
+private:
+  std::FILE *file_;
+};
+
+class DiskFileSystem : public morphizen::FileSystem {
+public:
+  explicit DiskFileSystem(const char *base_dir)
+      : base_dir_(std::filesystem::path(base_dir).lexically_normal()) {}
+
+  morphizen::FileReader *create_reader(const char *path) override {
+    std::string full_path = (base_dir_ / path).string();
+    auto *r = new DiskFileReader(full_path.c_str());
+    if (!r->ok()) {
+      delete r;
+      return nullptr;
+    }
+    return r;
+  }
+
+  morphizen::FileWriter *create_writer(const char *path) override {
+    std::string full_path = (base_dir_ / path).string();
+    auto *w = new DiskFileWriter(full_path.c_str());
+    if (!w->ok()) {
+      delete w;
+      return nullptr;
+    }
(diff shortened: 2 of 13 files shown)


lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -988,22 +1006,19 @@ void ConvertOnnxToHipPass::runOnOperation() {
     manifest["total_bytes"] = extState->currentOffset;
     manifest["constants"] = std::move(extState->manifestEntries);
 
-    std::error_code ec;
-    llvm::raw_fd_ostream jsonFile(jsonPath, ec);
-    if (ec) {
-      module.emitError("failed to open constants manifest: " + jsonPath + " (" +
-                       ec.message() + ")");
-      return signalPassFailure();
+    auto jsonWriter = diskFs.create_writer_template(jsonPath.c_str());
```
fhanuman, line 1448: What happens in this case of write failure?

lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -988,22 +1006,19 @@ void ConvertOnnxToHipPass::runOnOperation() {
...
+    if (jsonWriter) {
+      std::string jsonStr;
+      llvm::raw_string_ostream jsonOs(jsonStr);
+      jsonOs << llvm::formatv("{0:2}", llvm::json::Value(std::move(manifest)));
+      jsonWriter->fwrite(jsonStr.data(), jsonStr.size());
     }
-    jsonFile << llvm::formatv("{0:2}", llvm::json::Value(std::move(manifest)));
-    jsonFile.close();
 
     module.emitRemark("externalized ")
```
fhanuman: Could we fix this to emit log only for debug build?

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -1393,6 +1394,71 @@ struct MemRefDeallocOpLowering
...
+
+    // Cast to GPU address space only when the memref declares one (AS != 0).
+    unsigned targetAS = 0;
+    if (auto memSpace = memRefType.getMemorySpace())
+      if (auto intAttr = dyn_cast<IntegerAttr>(memSpace))
+        targetAS = intAttr.getInt();
+
+    Value dataPtr = gpuPtr;
+    if (targetAS != 0)
+      dataPtr = rewriter.create<LLVM::AddrSpaceCastOp>(
```
fhanuman: Could we use the following form as `rewrite.create<LLVM::AddrSpaceCastOp>(...)` is deprecated based on the feedback from earlier?

`LLVM::AddrSpaceCastOp::create(rewriter, ...)`

Same in other places too.

include/hip/Support/DiskFileSystem.h
```diff
@@ -0,0 +1,116 @@
...
+#include <cstdio>
+#include <filesystem>
+#include <string>
+#ifdef _WIN32
+#include <io.h>
+#endif
+
+namespace mlir::hip {
+
+class DiskFileReader : public morphizen::FileReader {
```
fhanuman, line 36: We could use `unique_ptr<FILE, ...>` for the file handle to make the class safe by construction — if the constructor body ever throws, the member destructor runs and closes the file. This also eliminates the manual destructor.

```
class DiskFileReader : public morphizen::FileReader {
  using FilePtr = std::unique_ptr<std::FILE, decltype(&std::fclose)>;

  FilePtr file_;
  std::size_t size_;

public:
  explicit DiskFileReader(const char *path)
      : file_(std::fopen(path, "rb"), &std::fclose), size_(0) {
    if (!file_)
      return;
    // ... fseeko to compute size_ ...
  }

  // No destructor needed
  bool ok() const { return file_ != nullptr; }
  std::size_t size() const override { return size_; }
  // fread/rewind use file_.get() instead of file_
};
```

include/hip/Support/DiskFileSystem.h
```diff
@@ -0,0 +1,116 @@
...
+      return 0;
+    return std::fread(buffer, 1, size, file_);
+  }
+
+private:
+  std::FILE *file_;
+  std::size_t size_;
+};
+
+class DiskFileWriter : public morphizen::FileWriter {
```
fhanuman, line 71: Above comment I made for `DiskFileReader` applies here as well.

include/hip/Support/DiskFileSystem.h
```diff
@@ -0,0 +1,116 @@
...
+    std::string full_path = (base_dir_ / path).string();
+    auto *r = new DiskFileReader(full_path.c_str());
+    if (!r->ok()) {
+      delete r;
+      return nullptr;
+    }
+    return r;
+  }
+
+  morphizen::FileWriter *create_writer(const char *path) override {
```
fhanuman, line 103: Above comment I made for `create_reader`  applies here as well.

include/hip/Support/DiskFileSystem.h
```diff
@@ -0,0 +1,116 @@
...
+      : file_(std::fopen(path, "rb")), size_(0) {
+    if (!file_)
+      return;
+#ifdef _WIN32
+    _fseeki64(file_, 0, SEEK_END);
+    size_ = static_cast<std::size_t>(_ftelli64(file_));
+    _fseeki64(file_, 0, SEEK_SET);
+#else
+    fseeko(file_, 0, SEEK_END);
+    size_ = static_cast<std::size_t>(ftello(file_));
```
fhanuman: Won't this return `-1` on failure? So, is this correct?

fhanuman (changes requested):

include/hip/Dialect/IR/HipOps.td
```diff
@@ -96,6 +96,39 @@ def Hip_FreeOp : Hip_Op<"free",
...
+def Hip_GetConstantOp : Hip_Op<"get_constant", [
+  DeclareOpInterfaceMethods<MemoryEffectsOpInterface>
+]> {
+  let summary = "Retrieve constant from GPU memory";
+  let description = [{
+    Retrieves a pre-uploaded constant from state->gpu_constants[index].
+
+    Arguments:
+    - ctx: HIP execution context (contains gpu_constants array)
+    - index: Index in gpu_constants array (0, 1, 2, ...)
```
MaheshRavishankar, line 110: I am pointing this out for now, but can be addressed later on. I think using some bespoke index this way is very very fragile. This is OK for now but we need to move to using global symbols to represent constants. That is the most accepted practice and with good merit.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -1393,6 +1394,71 @@ struct MemRefDeallocOpLowering
...
+    FailureOr<LLVM::LLVMFuncOp> funcOp = LLVM::lookupOrCreateFn(
+        rewriter, module, kHipGetConstant, paramTypes, ptrType);
+    if (failed(funcOp))
+      return failure();
+
+    SmallVector<Value, 2> args = {adaptor.getCtx(), adaptor.getIndex()};
+    auto callOp = LLVM::CallOp::create(rewriter, loc, *funcOp, args);
+    Value gpuPtr = callOp.getResult();
+
+    // Cast to GPU address space only when the memref declares one (AS != 0).
```
MaheshRavishankar: What is GPU memory space?

include/hip/Support/DiskFileSystem.h
BoarQing: create_writer_template construct a unique_ptr for writer. It looks strange because it is trying to resolve the cross-DLL destruction issue.
This API is mainly for C style. Personally, I don't use it.

include/hip/Dialect/IR/HipOps.td
BoarQing: I will keep that in mind. Thanks for the knowledge! I will think about a proper solution for that.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
BoarQing: Thanks for the knowledge. I i will keep that in mind

BoarQing:
@fhanuman Thanks for your programming tips. I have addressed nearly all comments.

include/hip/Support/DiskFileSystem.h
```diff
@@ -0,0 +1,112 @@
...
+      return;
+#ifdef _WIN32
+    _fseeki64(file_.get(), 0, SEEK_END);
+    auto pos = _ftelli64(file_.get());
+    size_ = (pos >= 0) ? static_cast<std::size_t>(pos) : 0;
+    _fseeki64(file_.get(), 0, SEEK_SET);
+#else
+    fseeko(file_.get(), 0, SEEK_END);
+    auto pos = ftello(file_.get());
+    size_ = (pos >= 0) ? static_cast<std::size_t>(pos) : 0;
```
fhanuman: This still doesn't look right to me. We are just masking the issue with the new change now.

`ftello` returning -1 on error is still silently mapped to size_ = 0. The ternary masks the failure rather than handling it.

fhanuman (approved):

PR #46, opened 2026-03-19
Port Target/LLVM backend, Compiler pipeline, and CInterface API

BoarQing:
## Summary

Port the Target/LLVM backend, Compiler driver, and C interface from the reference repository. Refactor `hip-compiler` CLI to use the new `CompilerDriver` instead of 200+ lines of inline logic.

### New
| File | Purpose |
|------|---------|
| `lib/Target/LLVM/LLVMBackend.h/.cpp` | MLIR → LLVM IR translation, optimization, object emission |
| `lib/Target/LLVM/DLLLinker.h/.cpp` | LLD-based linking (COFF/ELF) to produce DLL/SO |
| `lib/Compiler/CompilerDriver.h/.cpp` | End-to-end orchestration: parse → passes → LLVM IR → link |
| `lib/Compiler/Pipeline.h/.cpp` | MLIR pass pipeline construction |
| `lib/CInterface/CompilerAPI.cpp` | C-ABI wrapper delegating to `CompilerDriver` |
| `include/hip/compiler_api.h` | C entry points (`hip_compile_with_fs`, `hip_get_version`) |
| `include/hip/compiler_types.h` | C-compatible error codes and structs |
| `include/hip/InitAllPasses.h` | Centralized dialect loading and pass registration |
| `include/hip/Conversion/Passes.h` | Conversion pass declarations (HipToLLVM, OnnxToHip) |
| `lib/Conversion/Passes.cpp` | Pass registration |

### Modified
| File | Change |
|------|--------|
| `tools/hip-compiler/hip-compiler.cpp` | ~220 → ~47 lines, delegates to `CompilerDriver` |

### DIFF
--- a/lib/CInterface/CompilerAPI.cpp
+++ b/lib/CInterface/CompilerAPI.cpp
@@ -1,0 +1,104 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+#include "hip/Compiler/CompilerDriver.h"
+#include "hip/compiler_api.h"
+#include "hip/compiler_types.h"
+#include "hip/flatbuffers_json.h"
+
+#include "morphizen-foundation/file_io.hpp"
+
+#include "compilation_options_schema.h"
+
+#include "llvm/ADT/StringRef.h"
+
+#include <cstring>
+#include <string>
+
+using namespace hip::compiler;
+
+static const char *COMPILER_VERSION = "1.0.0";
+
+// Parse JSON into CompilationOptionsT (defined in schemas/compilation_options.fbs).
+// Key fields:
+//   opt_level      — LLVM optimization level 0-3 (default 2)
+//   output_mode    — DLL or LLVM_IR (default DLL)
+//   constants_file — externalized weights filename (default "constants.bin")
+static bool parseOptions(const char *options_json,
+                         mlir::hip::CompilationOptionsT &opts,
+                         std::string &error_message) {
+  if (!options_json || strlen(options_json) == 0)
+    return true;
+
+  return mlir::hip::fromJson<mlir::hip::CompilationOptionsT>(
+      options_json, mlir::hip::k_compilation_options_schema(), opts,
+      error_message);
+}
+
+static void setError(CompilerError *error, const std::string &message) {
+  if (error) {
+    size_t len = message.length();
+    if (len >= sizeof(error->message)) {
+      len = sizeof(error->message) - 1;
+    }
+    std::memcpy(error->message, message.c_str(), len);
+    error->message[len] = '\0';
+  }
+}
+
+extern "C" {
+
+COMPILER_API CompilerErrorCode hip_compile_with_fs(
+    const void *input_mlir, size_t input_size, const char *output_path,
+    const char *options_json, CompilerError *error, void *fs) {
+  if (!input_mlir || input_size == 0 || !output_path) {
+    setError(
+        error,
+        "Invalid input: input_mlir, input_size, and output_path must be valid");
+    return COMPILER_ERROR_INVALID_INPUT;
+  }
+  if (!fs) {
+    setError(error, "Invalid input: fs (FileSystem*) must be non-null");
+    return COMPILER_ERROR_INVALID_INPUT;
+  }
+
+  try {
+    mlir::hip::CompilationOptionsT options;
+    std::string parse_error;
+    if (!parseOptions(options_json, options, parse_error)) {
+      setError(error, parse_error);
+      return COMPILER_ERROR_INVALID_INPUT;
+    }
+
+    CompilerDriver driver;
+    driver.setFileSystem(static_cast<morphizen::FileSystem *>(fs));
+
+    std::string error_message;
+    llvm::StringRef input_ref(static_cast<const char *>(input_mlir),
+                              input_size);
+    std::string output_str(output_path);
+
+    bool success =
+        driver.compile(input_ref, output_str, options, error_message);
+
+    if (!success) {
+      setError(error, error_message);
+      return COMPILER_ERROR_COMPILATION_FAILED;
+    }
+
+    return COMPILER_SUCCESS;
+
+  } catch (const std::exception &ex) {
+    setError(error, std::string("Exception: ") + ex.what());
+    return COMPILER_ERROR_INTERNAL;
+  } catch (...) {
+    setError(error, "Unknown exception occurred");
+    return COMPILER_ERROR_INTERNAL;
+  }
+}
+
+COMPILER_API const char *hip_get_version(void) { return COMPILER_VERSION; }
+
+} // extern "C"
--- a/lib/CInterface/CompilerAPI.cpp
+++ b/lib/CInterface/CompilerAPI.cpp
@@ -21,8 +21,8 @@ using namespace hip::compiler;
 
 static const char *COMPILER_VERSION = "1.0.0";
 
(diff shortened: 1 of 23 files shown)


lib/Compiler/Pipeline.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include "mlir/Conversion/BufferizationToMemRef/BufferizationToMemRef.h"
+#include "mlir/Dialect/Bufferization/Pipelines/Passes.h"
+#include "mlir/Dialect/Bufferization/Transforms/Passes.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/Passes.h"
+
+namespace hip::compiler {
+
+void populateMorphizenPipeline(mlir::OpPassManager &pm,
```
fhanuman: My understanding was that we will use the existing pipelines. This new function duplicates both buildOnnxToHipPipeline (stages 0-10) and buildHipToLLVMPipeline (stages 10-11) from:

https://github.com/ROCm/onnx-hipdnn-ep/blob/910237a23416b66b0e2e9f75c246617ce18e6596/lib/Dialect/Transforms/Pipelines.cpp#L20

Can we delegate to those existing functions instead of reimplementing the same stages? Something like:

```
void populateMorphizenPipeline(mlir::OpPassManager &pm,
                               const mlir::hip::CompilationOptionsT &options) {
  OnnxToHipPipelineOptions onnxOpts;
  buildOnnxToHipPipeline(pm, onnxOpts);

  HipToLLVMPipelineOptions llvmOpts;
  llvmOpts.constantsFile = options.constants_file;
  buildHipToLLVMPipeline(pm, llvmOpts);
}
```

lib/CInterface/CompilerAPI.cpp
```diff
@@ -0,0 +1,99 @@
...
+static const char *COMPILER_VERSION = "1.0.0";
+
+static bool parseOptions(const char *options_json,
+                         mlir::hip::CompilationOptionsT &opts,
+                         std::string &error_message) {
+  if (!options_json || strlen(options_json) == 0)
+    return true;
+
+  return mlir::hip::fromJson<mlir::hip::CompilationOptionsT>(
+      options_json, mlir::hip::k_compilation_options_schema(), opts,
```
fhanuman, line 36: Could we add comments clarifying what `CompilationOptionsT` contains? I know we are parsing the JSON file to get the compilation options, but looking at the code, it’s not clear what options we need.

lib/Compiler/CompilerDriver.cpp
```diff
@@ -0,0 +1,297 @@
...
+  hip::compiler::registerAllPasses();
+
+  mlir::MLIRContext context;
+  hip::compiler::loadAllDialects(context);
+  mlir::registerLLVMDialectTranslation(context);
+
+  COMPILER_DEBUG_LOG("[CompilerDriver::compile] Input size: "
+                     << input_mlir.size() << " bytes\n");
+
+  auto memBuffer = llvm::MemoryBuffer::getMemBuffer(input_mlir, "", false);
```
fhanuman, line 56: Could we add some comments clarifying what this buffer is and why we need it? It's not clear from the code.

lib/Compiler/CompilerDriver.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+  llvm::SourceMgr sourceMgr;
+  sourceMgr.AddNewSourceBuffer(std::move(memBuffer), llvm::SMLoc());
```
fhanuman, line 60: This is not clear either -- what is this doing and why do we need this?. Some clarification comments in the code would help.

lib/Compiler/CompilerDriver.cpp
```diff
@@ -0,0 +1,297 @@
...
+  if (!llvmModule)
+    return false;
+
+  if (!linkRuntime(llvmModule.get(), error_message))
+    return false;
+
+  optimizeLLVMIR(llvmModule.get(), options.opt_level);
+
+  std::string base_path = output_path;
+  if (base_path.size() >= 4 &&
```
fhanuman: Not sure what this constant `4` means here, please add some comments. Same comments in other places, unless something is very obvious, please add some code comments. These comments would help when someone looks at this code later.

lib/Compiler/CompilerDriver.cpp
```diff
@@ -0,0 +1,297 @@
...
+      return false;
+    return true;
+  }
+
+  if (!compileToObject(llvmModule.get(), obj_path, error_message))
+    return false;
+
+  std::vector<std::string> export_symbols = {
+      "inference_init", "inference_compute", "inference_cleanup",
+      "inference_get_metadata_json", "test_hip_from_dll"};
```
fhanuman, line 140: Where is this symbol `test_hip_from_dll` getting introduced, and why do we need this? I am aware of the other symbols, but I'm not clear about this one. Please add some comments

lib/Compiler/CompilerDriver.cpp
```diff
@@ -0,0 +1,297 @@
...
+  std::vector<std::string> libraries;
+  std::vector<std::string> library_paths;
+
+  if (const char *therock = std::getenv("THEROCK_DIST")) {
```
fhanuman: If we expect more libraries to get added in the future, does it makes sense to move this logic into its own function for better readability?

lib/Compiler/CompilerDriver.cpp
```diff
@@ -0,0 +1,297 @@
...
+bool CompilerDriver::runMLIRPasses(
+    mlir::ModuleOp module, const mlir::hip::CompilationOptionsT &options,
+    std::string &error_message) {
+  mlir::PassManager pm(module.getContext());
+
+  if (options.verbose) {
+    COMPILER_DEBUG_LOG("Running ONNX->HIP->LLVM->Interface passes\n");
+  }
+
+  mlir::hip::OnnxToHipPipelineOptions onnxToHipOpts;
```
fhanuman, line 163: Nice!

lib/Target/LLVM/DLLLinker.cpp
```diff
@@ -0,0 +1,332 @@
...
+      argStrings.push_back(lib); // full path — pass as-is
+    } else if (lib.size() >= 4 && lib.substr(lib.size() - 4) == ".lib") {
+      argStrings.push_back(lib);
+    } else {
+      argStrings.push_back(lib + ".lib");
+    }
+  }
+
+  // Dynamic CRT (/MDd) so the DLL shares CRT state with the host EXE.
+  for (const char *sysLib : {"msvcrtd.lib", "ucrtd.lib", "vcruntimed.lib",
```
fhanuman, line 178: I don't understand why these libraries are needed, probably you understand it better.

fhanuman (approved):
How are we testing all this C++ code, considering we don't have the full pipeline yet? Is there any way we can unit test some of this logic?

lib/Compiler/Pipeline.cpp
fhanuman: I see you already addressed this review feedback while I was looking at the code.

BoarQing:
> How are we testing all this C++ code, considering we don't have the full pipeline yet? Is there any way we can unit test some of this logic?

I believe the subsequent testcases would cover these.

lib/Compiler/CompilerDriver.cpp
BoarQing: agree!

lib/Compiler/CompilerDriver.cpp
BoarQing: It would be a part of the mock.

PR #50, opened 2026-03-20
feat: add LIT e2e tests and E2E compile+execute test infrastructure

AMDmoore:
## Summary

- Add LIT e2e tests covering 15 HIP ops: cast, conv, gather, gqa, matmul, mlp, mul, reduce_sum, rotary_embedding, sigmoid, simplified_layer_norm, single_layer_decode, single_layer_small_size, skip_simplified_layer_norm, sub
- Port E2E compile+execute test framework (test/e2e/) with CMake auto-discovery of .mlir files
- Register missing bufferizable op interfaces (ConvOp, CastOp, SigmoidOp, SubOp, ReduceSumOp) and add OnnxToHip/HipToLLVM lowering patterns for new ops
- Enable constant externalization (externalizeMinNumElements = 1) so all tensor constants are written to constants.bin sidecar file
- Fix splat constant handling: expand splat data via staging buffer instead of inlining in host memory (previously caused GPU error 719)
- Thread EPContext FileSystem through CompilerDriver into ConvertOnnxToHipPass so constants.bin is written into EPContext archive instead of CWD
- Fix GenerateInterface metadata filename mismatch: prefer hip.constants_file module attribute over FlatBuffers schema default
- Remove deprecated Relu, MatMulNBits, QMoE ops from HIP dialect
- Use shared kDefaultExternalizeMinNumElements constant to avoid hardcoded values

## Test Plan

- [x] All 56 LIT tests pass
- [x] LIT E2E execute tests pass with GPU
- [x] ORT E2E: onnxruntime_perf_test.exe with single layer model


include/hip/Dialect/IR/HipBufferize.h
```diff
@@ -60,6 +60,7 @@ struct HipDstBufferizableModel
 inline void
 registerHipBufferizableOpInterfaceModels(DialectRegistry &registry) {
   registry.addExtension(+[](MLIRContext *ctx, HipDialect *) {
+    ConvOp::attachInterface<HipDstBufferizableModel<ConvOp>>(*ctx);
```
fhanuman, line 63: Was this missing before, or is it only needed now with the `e2e` tests? Curious if we have a gap in test coverage that let this slip through.

include/hip/Dialect/Transforms/Pipelines.h
```diff
@@ -8,9 +8,18 @@
...
+class FileSystem;
+} // namespace morphizen
+
 namespace mlir {
 namespace hip {
 
+/// Default minimum number of tensor elements for constant externalization.
+/// All tensor constants with at least this many elements are written to
+/// the sidecar constants.bin rather than inlined in the DLL.
+constexpr int64_t kDefaultExternalizeMinNumElements = 1;
```
fhanuman, line 22: The commit message explains this fixes GPU error 719 from inline splat constants. Could we add a comment here noting that? Without context, = 1 looks overly aggressive constant externalization.

include/hip/Dialect/Transforms/Pipelines.h
```diff
@@ -49,6 +58,13 @@ struct HipToLLVMPipelineOptions
 void buildOnnxToHipPipeline(OpPassManager &pm,
```
fhanuman, line 65: Both overloads make sense to me -- one for CLI/standalone, one for EP integration. A brief doc comment on each clarifying when to use which would be helpful.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -1720,6 +1727,172 @@ struct GetConstantOpLowering : public ConvertOpToLLVMPattern<GetConstantOp> {
...
+      return module.emitError("Metadata mismatch: shapes array size != count");
+
+    unsigned expectedParams = 1; // context
+    for (auto shapeAttr : inputShapesAttr) {
+      int64_t rank = cast<DenseI64ArrayAttr>(shapeAttr).size();
+      expectedParams += 2 + 1 + rank + rank;
+    }
+    for (auto shapeAttr : outputShapesAttr) {
+      int64_t rank = cast<DenseI64ArrayAttr>(shapeAttr).size();
+      expectedParams += 2 + 1 + rank + rank;
```
fhanuman: Could we give these constants meaningful names? Something like kMemRefPtrs = 2 (alloc + aligned), kMemRefOffset = 1, then expectedParams += kMemRefPtrs + kMemRefOffset + rank + rank -- makes the memref descriptor layout self-documenting.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -1720,6 +1727,172 @@ struct GetConstantOpLowering : public ConvertOpToLLVMPattern<GetConstantOp> {
...
+
+    auto castToAs0 = [&](Value ptr) -> Value {
+      auto ptrTy = cast<LLVM::LLVMPointerType>(ptr.getType());
+      if (ptrTy.getAddressSpace() == 0)
+        return ptr;
+      return builder.create<LLVM::AddrSpaceCastOp>(loc, as0PtrType, ptr);
+    };
+
+    Value allocPtr = builder.create<LLVM::ExtractValueOp>(loc, memrefStruct,
+                                                          ArrayRef<int64_t>{0});
```
fhanuman: Same as the earlier comment -- these struct field indices (0 = allocPtr, 1 = alignedPtr, 2 = offset, etc.) would be a lot easier to follow with named constants. Something like kAllocPtrIdx = 0, kAlignedPtrIdx = 1, kOffsetIdx = 2, kSizesIdx = 3, kStridesIdx = 4.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -1720,6 +1727,172 @@ struct GetConstantOpLowering : public ConvertOpToLLVMPattern<GetConstantOp> {
...
+    args.push_back(castToAs0(allocPtr));
+
+    Value alignedPtr = builder.create<LLVM::ExtractValueOp>(
+        loc, memrefStruct, ArrayRef<int64_t>{1});
+    args.push_back(castToAs0(alignedPtr));
+
+    args.push_back(builder.create<LLVM::ExtractValueOp>(loc, memrefStruct,
+                                                        ArrayRef<int64_t>{2}));
+
+    for (int64_t dim = 0; dim < rank; dim++)
```
fhanuman: Nit: `llvm::seq<int64_t>(0, rank)` would be a bit more idiomatic here -- same feedback as in previous PRs. Also, same comment in few other places.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -1720,6 +1727,172 @@ struct GetConstantOpLowering : public ConvertOpToLLVMPattern<GetConstantOp> {
...
+    if (actualParams != expectedParams) {
+      return module.emitError()
+             << "[HipToLLVM] Parameter count mismatch: expected "
+             << expectedParams << ", got " << actualParams;
+    }
+
+    OpBuilder builder(module.getContext());
+    Location loc = mainFunc.getLoc();
+
+    mainFunc.setName("main_graph_internal");
```
fhanuman, line 1836: Could we add a brief comment explaining the rename -- e.g., the original expanded-args function becomes private, and the new 3-pointer wrapper takes its name for C ABI compatibility?

lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -1398,21 +1453,37 @@ void ConvertOnnxToHipPass::runOnOperation() {
...
+    llvm::StringRef name = op->getName().getStringRef();
+    if (name == "onnx.NoValue" && op->use_empty())
+      toErase.push_back(op);
+    else if (name == "onnx.EntryPoint")
+      toErase.push_back(op);
+  });
+  for (auto *op : toErase)
+    op->erase();
+
+  // Strip ONNX result/arg attributes from func.func operations so that
```
fhanuman: For my own understanding -- is this a new issue surfaced by the e2e test models (which carry onnx.name attributes on results)? I don't recall hitting this when the onnx-to-hip-pipeline was first introduced. A short comment here about why the attributes block buffer-results-to-out-params would be helpful context.

lib/Dialect/Transforms/GenerateInterface.cpp
```diff
@@ -316,10 +316,13 @@ class GenerateInterfacePass
       return;
     }
 
-    const std::string constantsFile =
-        !compilationOptions_.constants_file.empty()
-            ? compilationOptions_.constants_file
-            : "constants.bin";
+    std::string constantsFile = "constants.bin";
+    if (auto attr =
```
fhanuman, line 322: Could we add some comments in what scenario we don't expect `hip.constants_file` attribute? Do you mean to handle the stand-alone case -- not going through EP?

fhanuman (approved):

lib/Dialect/Transforms/GenerateInterface.cpp
AMDmoore: No, I think this should be a design issue. Now, there are two ways to name the contants file, one is from compile option(doesn't work now), the other is from module attribute which is set by onnxtohip. The compile option may be used by EP or lit test case(unnecessary I think). The way by EP or module attribute has default value.
In order to run through the ORT E2E, it's easier to make module attribute prior, but I think we should have another feature to define the priority, maybe compile option if defined, then the default from onnxtohip pass.
I think the same issue for MinElements.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
AMDmoore: fixed, mark as resolved.

include/hip/Dialect/Transforms/Pipelines.h
AMDmoore: comment added, mark as resolved.

include/hip/Dialect/Transforms/Pipelines.h
AMDmoore: comment added, mark as resolved.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
AMDmoore: change to use llvm::seq<int64_t>(0, rank), resolved.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
AMDmoore: modified it as this comment, resolved.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
AMDmoore: comment added, resolved

lib/Conversion/OnnxToHip/OnnxToHip.cpp
AMDmoore: Yes, you're right. Comment about the reason added, mark this as resolved.

include/hip/Dialect/IR/HipBufferize.h
AMDmoore: most of the cases didn't touch one-shot-bufferize pass, and also Hip_ConvOp : Hip_DpsOp<"conv">

PR #52, opened 2026-03-23
ci: add GPU performance test job using MorphiZen EP

zz002:
## Summary

Add a GPU-based CI job that automatically benchmarks ONNX model performance using the MorphiZen EP on a self-hosted AMD GPU machine (StrixHalo), and publishes QPS results to the workflow Summary and PR comments.

- Add `gpu-test` job (`runs-on: StrixHalo`) that runs `onnxruntime_perf_test` via MorphiZen EP against all `*seq128.onnx` models after each build
- Publish QPS results (2 decimal places) to workflow run Summary tab and as a PR comment via `gh` CLI
- Build ORT v1.24.3 from source to obtain `onnxruntime_perf_test.exe` with Plugin EP support; install into `prebuilt-local` cache
- Stage self-contained GPU test package: MSVC release CRT import libs + `hip_custom_kernels.lib` + `amdhip64_7.dll` — GPU machine needs no VS Build Tools installed
- Disable OpenSSL at configure time (`-DCMAKE_DISABLE_FIND_PACKAGE_OpenSSL=ON`) to remove `libcrypto-3-x64.dll` dependency
- Fix JIT DLL linking: add `/NODEFAULTLIB` + explicit release CRT libs in `DLLLinker.cpp` to prevent debug CRT dependency
- Set `THEROCK_DIST` so `discoverLibraries()` resolves `amdhip64`/`MIOpen` import libs for JIT linking
- Clean `gpu-test-package` directory before artifact download to avoid stale results from previous runs

## Test plan

- [x] `build-windows` passes on `windows-latest`, uploads `gpu-test-package` artifact
- [x] `gpu-test` picks up on `StrixHalo` self-hosted runner
- [x] MorphiZen EP JIT-compiles and runs all `*seq128.onnx` models successfully
- [x] QPS table appears in workflow run Summary tab
- [x] No VS Build Tools required on GPU machine

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -482,7 +482,8 @@ jobs:
         run: |
           $resultsDir = "gpu-test-package/results"
           $runUrl = "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
-          $sha = "${{ github.sha }}".Substring(0, 7)
+          $rawSha = if ("${{ github.event_name }}" -eq "pull_request") { "${{ github.event.pull_request.head.sha }}" } else { "${{ github.sha }}" }
+          $sha = $rawSha.Substring(0, 7)
           $prNumber = "${{ github.event.pull_request.number }}"
 
           # Build table rows from result files
@@ -493,9 +494,9 @@ jobs:
             $qpsMatch = [regex]::Match($content, 'Number of inferences per second:\s+([\d.]+)')
             if ($qpsMatch.Success) {
               $qps = "{0:F2}" -f [double]$qpsMatch.Groups[1].Value
-              $tableRows += "| ``$model`` | $qps |`n"
+              $tableRows += "| $model | $qps |`n"
             } else {
-              $tableRows += "| ``$model`` | failed |`n"
+              $tableRows += "| $model | failed |`n"
             }
           }
           if (-not $tableRows) { $tableRows = "| (no results) | - |`n" }
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -162,8 +162,7 @@ jobs:
             --skip_tests ^
             --disable_memleak_checker ^
             --use_dml ^
-            --cmake_generator Ninja ^
-            --cmake_extra_defines onnxruntime_BUILD_UNIT_TESTS=ON
+            --cmake_generator Ninja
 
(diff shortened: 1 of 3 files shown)


.github/workflows/windows-build.yml
```diff
@@ -289,3 +305,243 @@ jobs:
...
+          New-Item -ItemType Directory -Force -Path $dist | Out-Null
+          curl.exe -fsSL `
+            "https://repo.amd.com/rocm/tarball/${{ env.THEROCK_VERSION }}.tar.gz" `
+            -o therock.tar.gz
+          tar -xzf therock.tar.gz -C $dist --strip-components=1
+          Remove-Item therock.tar.gz
+
+      # Run perf test against seq128 models via MorphiZen EP.
+      # Models are pre-placed on the runner at C:\actions-runner\_work\model\.
+      # -t 60        : run for 60 seconds per model
```
fhanuman: A couple of things I noticed in the gpu-test job:

- The comment says -t 60 but the actual command uses -t 30 -- stale from an earlier iteration?
- The "Diagnose gh CLI" step seems like a debugging step left in from development. Should we remove it, or is it meant to stay?

.github/workflows/windows-build.yml
```diff
@@ -289,3 +305,243 @@ jobs:
...
+      #                (MSVC release CRT + hip_custom_kernels); no VS required.
+      # Output is tee'd to results/ for QPS extraction. Non-zero exit code
+      # sets FAILED but continues so all models are tested.
+      - name: Run onnxruntime_perf_test (GPU)
+        shell: cmd
+        run: |
+          set THEROCK_DIST=${{ runner.workspace }}\therock-dist
+          set PATH=${{ runner.workspace }}\therock-dist\bin;%CD%\gpu-test-package\bin;%PATH%
+          set LIB=%CD%\gpu-test-package\lib;${{ runner.workspace }}\therock-dist\lib
+          set MODEL_DIR=${{ runner.workspace }}\model
```
fhanuman, line 447: What happens if models aren't present on the runner or MODEL_DIR doesn't exist? I think the for /r loop would do nothing and FAILED stays 0, so the step would pass even though nothing was tested. Would it make sense to add a quick existence check at the top?

fhanuman (approved):

fhanuman (commented):
Please remember to squash the commits before merge

PR #55, opened 2026-03-23
Address unresolved review comments from PR #44

mingyueliuh:
## Overview

This PR addresses the unresolved review comments from PR #44 (which has been merged to main). 

### ✅ RuntimeState struct synchronization (HIGH Priority)
**Problem:** Two different `RuntimeState` definitions in separate files caused undefined behavior (ODR violation).

**Fix:** 
- Created shared internal header `lib/Runtime/runtime_state_internal.h`
- Both files now include this header for single source of truth
- Eliminated struct duplication and UB

**Files modified:**
- `lib/Runtime/runtime_state_internal.h` (NEW)
- `lib/Runtime/hipdnn_ep_runtime_state.cpp`
- `lib/Runtime/hipdnn_ep_runtime_tensor.cpp`

---

### ✅ Conv group parameter not applied (HIGH Priority)
**Problem:** The `group` parameter was accepted but never applied via `miopenSetConvolutionGroupCount()`, causing incorrect results for grouped/depthwise convolutions.

**Fix:**
- Added group parameter application when `group > 1`
- Verified no similar issues exist in other runtime files

**Files modified:**
- `lib/Runtime/real/miopen.cpp`

---

### ✅ Resource leaks in runtime implementations (MEDIUM Priority)
**Problem:** Multiple runtime functions leaked MIOpen/hipBLAS descriptors when errors occurred, as `CHECK` macros returned directly without cleanup.

**Fix:** Applied goto cleanup pattern to 7 functions across 6 files:

1. **lib/Runtime/real/miopen.cpp** - `wrap_miopenConvolutionForward`
   - Fixed leaks: 4 descriptors + GPU workspace memory
   
2. **lib/Runtime/real/hipblas.cpp** - `wrap_hipblasLtGemm`
   - Fixed leaks: 4 hipBLAS descriptors

3. **lib/Runtime/real/matmul.cpp** - `wrap_hipblasLtMatmul`
   - Fixed leaks: 4 descriptors across 9 error points (most complex fix)

4. **lib/Runtime/real/activation.cpp** - `wrap_miopenActivationForward`
   - Fixed leaks: 3 descriptors across 7 error points

5. **lib/Runtime/real/elementwise.cpp** - `wrap_miopenOpTensor`
   - Fixed leaks: 3 descriptors across 7 error points

6. **lib/Runtime/real/miopen.cpp** - `wrap_miopenActivationForward_relu`
   - Fixed leaks: 3 descriptors across 7 error points

[...]

PR #57, opened 2026-03-23
feat: runtime algo caching, shared workspace, GQA refactor

zpye:
## Summary

Runtime-stage performance optimizations for hipBLASLt (MatMul, GQA) operators. Algorithm heuristics are queried once on the first inference call and cached; a lazily-grown shared GPU workspace eliminates per-call `hipMalloc`/`hipFree` overhead on subsequent runs.

### 1. Shared Workspace Infrastructure (`RuntimeState`)

- **`lib/Runtime/hipdnn_ep_runtime_state.cpp`** / **`lib/Runtime/hipdnn_ep_runtime.h`**
  - Added `workspace` (GPU pointer) and `workspace_size` fields to `RuntimeState`
  - New API: `hipdnn_ep_state_ensure_workspace(state, needed_size)` -- grows workspace if needed (never shrinks)
  - New API: `hipdnn_ep_state_get_workspace()` / `hipdnn_ep_state_get_workspace_size()` -- accessors
  - Workspace freed in `hipdnn_ep_state_cleanup()`
  - Shared across all operators (MatMul, GQA GEMMs)

### 2. MatMul Algorithm Caching (`real/matmul.cpp`)

- File-static `std::unordered_map` keyed by `(M, N, K, batch_count, elem_size)`
- On first call: `hipblasLtMatmulAlgoGetHeuristic` with 256 MB max workspace preference
- Caches optimal `algo` + `workspaceSize`; calls `ensure_workspace` to grow shared buffer
- Subsequent calls: cache hit, skip heuristic, reuse algo + shared workspace

### 3. GQA Refactor (device kernels + runtime orchestration)

- **`3rd-party/custom_kernels/hip/gqa_kernel.hip`** -- device kernels with full algorithm documentation and ONNX spec references:
  - `kv_cache_append_kernel` -- append new tokens into in-place/aliased KV cache (BSHD->BNSD transpose + scatter)
  - `kv_cache_concat_kernel` -- concatenate past data + new tokens into a fresh present buffer when past and present are separate allocations with different strides (replaces host-side B*G `hipMemcpyAsync` loop with a single kernel launch)
  - `rope_kernel` -- half-rotated RoPE
  - `transpose_mid_dims_kernel` -- 4D middle-dim transpose
  - `expand_kv_kernel` -- KV group expansion (G groups to H heads)
  - `causal_mask_kernel` -- causal attention mask
  - `softmax_inplace_kernel` -- column-wise softmax

- **`3rd-party/custom_kernels/include/hip_custom_kernels.h`** -- individual kernel launcher declarations with `present_seq` / `past_seq` parameters clarifying buffer strides

- **`lib/Runtime/real/gqa.cpp`** -- full 11-step hipBLASLt GQA pipeline:
  - hipBLASLt GEMM setup (score + value), algorithm caching per shape
  - All temp buffers packed into the shared workspace at fixed offsets, eliminating all `hipMalloc`/`hipFree` on the hot path
  - KV cache update: `kv_cache_append` for same-buffer (aliased), `kv_cache_concat` for separate-buffer with stride mismatch
  - `present_seq` used throughout (not `max_seq`) to clarify it is the present buffer's actual sequence dimension
  - `HIP_CHECK` on all kernel launcher return values with `goto cleanup` for safe resource deallocation
  - FP16-only restriction documented with future FP32/BF16 path

- **`lib/Runtime/real/error_check_macros.h`** -- `HIP_CHECK_GOTO` uses `static_cast<hipError_t>` so it works for both `hipError_t`-returning HIP APIs and `int`-returning kernel launchers

### 4. Test CMake Cleanup (`test/CMakeLists.txt`)

- Removed stale test target references

## Test Plan

- [x] Build succeeds (`_build.bat`)
- [x] Pre-commit hooks pass (trailing whitespace, clang-format, license headers)
- [x] GQA-only perf tests pass (all sequence lengths: seq1 through seq3072)
- [x] Full model perf tests pass (seq1, seq128, seq1024)

### DIFF
--- a/3rd-party/custom_kernels/hip/gqa_kernel.hip
+++ b/3rd-party/custom_kernels/hip/gqa_kernel.hip
@@ -7,489 +7,272 @@
 #include <hip/hip_fp16.h>
 #include <cstdint>
 #include <cstdio>
-#include <cfloat>
+#include <algorithm>
 
 #include "hip_custom_kernels.h"
+
+static constexpr int GQA_THREADS = 256;
 
 // =============================================================================
 // KV Cache Update Kernel
 // =============================================================================
-// Copies current key/value into the present_key/present_value buffer at the
-// correct offset (after past sequence).
-//
-// ONNX com.microsoft.GroupQueryAttention uses BNSD layout for KV cache:
-//   past_key:    [batch, kv_num_heads, past_seq, head_dim]       (BNSD)
-//   present_key: [batch, kv_num_heads, total_seq, head_dim]      (BNSD)
-//   key (current): [batch, seq_q, kv_num_heads * head_dim]       (BSD)
-//
-// The kernel copies past entries and new entries into the present buffer,
-// converting current key/value from BSD to BNSD in the process.
+// Writes new K/V from BSHD [B,sq,G,d] into BNSD KV cache [B,G,max_seq,d]
+// at positions [past_len .. past_len+sq). Fuses the layout transpose.
 
-template <typename T>
 __global__ void kv_cache_update_kernel(
-    const T* __restrict__ past,
-    const T* __restrict__ current,
-    T* __restrict__ present,
-    const int32_t* __restrict__ seqlens_k,
-    int64_t batch_size,
-    int64_t past_seq_len,
-    int64_t current_seq_len,
-    int64_t total_seq_len,
-    int64_t kv_num_heads,
-    int64_t head_dim) {
+    const __half* __restrict__ src,
+    __half* __restrict__ cache,
+    int sq, int G, int d, int max_seq, int past_len) {
+  int batch_idx = blockIdx.y;
+  int idx = blockIdx.x * blockDim.x + threadIdx.x;
+  int total_per_batch = sq * G * d;
+  if (idx >= total_per_batch) return;
 
-  int64_t idx = blockIdx.x * blockDim.x + threadIdx.x;
-  int64_t total = batch_size * kv_num_heads * total_seq_len * head_dim;
+  int di = idx % d;
+  int remaining = idx / d;
+  int g = remaining % G;
+  int s = remaining / G;
 
-  if (idx >= total)
-    return;
+  int src_idx = ((batch_idx * sq + s) * G + g) * d + di;
+  int dst_idx = ((batch_idx * G + g) * max_seq + (past_len + s)) * d + di;
+  cache[dst_idx] = src[src_idx];
+}
 
-  // Decompose flat index into BNSD coordinates
-  int64_t d = idx % head_dim;
-  int64_t remaining = idx / head_dim;
-  int64_t s = remaining % total_seq_len;
-  remaining = remaining / total_seq_len;
-  int64_t n = remaining % kv_num_heads;
-  int64_t b = remaining / kv_num_heads;
+// =============================================================================
+// RoPE Kernel
+// =============================================================================
+// Half-rotated RoPE: for each position s, pair_d in [0, half_rot):
+//   out[d] = in[d]*cos - in[d+half]*sin
+//   out[d+half] = in[d+half]*cos + in[d]*sin
 
-  int64_t actual_past_len = seqlens_k ? (int64_t)seqlens_k[b] : past_seq_len;
+__global__ void rope_kernel(
+    const __half* __restrict__ input,
+    __half* __restrict__ output,
+    const __half* __restrict__ cos_cache,
+    const __half* __restrict__ sin_cache,
+    int seq_len, int num_heads, int head_dim,
+    int half_rot, int past_len) {
+  int idx = blockIdx.x * blockDim.x + threadIdx.x;
+  int batch_idx = blockIdx.y;
+  int elems = seq_len * num_heads * half_rot;
+  if (idx >= elems) return;
 
-  // present is BNSD — idx already maps to the right position
-  if (s < actual_past_len && past) {
-    // past is BNSD: [batch, kv_num_heads, past_seq, head_dim]
-    int64_t past_idx = ((b * kv_num_heads + n) * past_seq_len + s) * head_dim + d;
-    present[idx] = past[past_idx];
-  } else if (s >= actual_past_len && s < actual_past_len + current_seq_len) {
-    // current is BSD: [batch, seq_q, kv_num_heads * head_dim]
-    int64_t cur_s = s - actual_past_len;
-    int64_t cur_idx = (b * current_seq_len + cur_s) * (kv_num_heads * head_dim)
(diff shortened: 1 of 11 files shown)


3rd-party/custom_kernels/hip/gqa_kernel.hip
```diff
@@ -7,489 +7,272 @@
...
+#include <algorithm>
 
 #include "hip_custom_kernels.h"
 
+static constexpr int GQA_THREADS = 256;
+
 // =============================================================================
 // KV Cache Update Kernel
 // =============================================================================
-// Copies current key/value into the present_key/present_value buffer at the
```
fhanuman, line 17: The GQA kernel I shared had detailed documentation -- ONNX spec references, BNSD layout diagrams, thread mapping, the 11-step algorithm overview. We probably missed all of it when porting it here. Could we restore all the code and algorithm documentation? The layout docs and operator spec references are especially useful for anyone debugging these later.

lib/Runtime/real/gqa.cpp
```diff
@@ -6,19 +6,368 @@
...
 #include "hip_custom_kernels.h"
+#include "runtime_types.h"
 
+#include <algorithm>
 #include <cstdio>
+#include <cstring>
+#include <functional>
+#include <unordered_map>
+
+#define GQA_HIP_CHECK(cmd)                                                     \
```
fhanuman: Same issue we fixed in PR #55 -- GQA_HIP_CHECK and GQA_HIPBLAS_CHECK use return -1, which skips cleanup. In gqa_forward_hipblaslt, we create 10 hipBLASLt descriptors and have a proper cleanup: block, but any macro failure in the matmul or memcpy calls will return -1 and leak all of them. Same thing in queryGemmAlgo -- pref leaks if the set-attribute or heuristic calls fail.

The goto cleanup after the algo queries handles this correctly, but the rest of the function doesn't. Could we make all error paths go through cleanup?

lib/Runtime/real/gqa.cpp
```diff
@@ -6,19 +6,368 @@
...
+      hip_gqa_kv_cache_update(stream, kSrc, present_key,
+                              (int)B, (int)sq, (int)G, (int)d,
+                              (int)max_seq, (int)past_len);
+      hip_gqa_kv_cache_update(stream, value, present_value,
+                              (int)B, (int)sq, (int)G, (int)d,
+                              (int)max_seq, (int)past_len);
+
+      if (past_key && past_len > 0) {
+        size_t past_bytes = (size_t)B * G * past_len * d * elem_sz;
+        GQA_HIP_CHECK(hipMemcpyAsync(present_key, past_key, past_bytes,
```
fhanuman: The kv_cache_update_kernel from the GQA kernel I shared assumes a single pre-allocated KV cache buffer. I might be wrong, but I think past_key and present_key can be separate buffers in ORT with different per-head strides (past_len * d vs max_seq * d). If that's the case, wouldn't the flat hipMemcpyAsync of B * G * past_len * d bytes corrupt data across head boundaries when the strides don't match?

Could we either restore the strided copy from the previous implementation or add a past_key == present_key guard here?

lib/Runtime/real/gqa.cpp
```diff
@@ -35,10 +384,30 @@ int wrap_group_query_attention(RuntimeState *state, void *query, void *key,
...
+    fprintf(stderr, "wrap_group_query_attention: null required argument\n");
+    return -1;
+  }
+  if (num_heads % kv_num_heads != 0) {
+    fprintf(stderr, "wrap_group_query_attention: num_heads (%lld) must be "
+            "divisible by kv_num_heads (%lld)\n",
+            (long long)num_heads, (long long)kv_num_heads);
+    return -1;
+  }
+  if (element_size_bytes != 2) {
```
fhanuman, line 428: The GQA kernel I shared was FP16-only since that's what the Llama benchmarks needed, but I think the previous implementation here was templated for both float and __half. Since this rewrite now rejects FP32 entirely, is that intentional? If so, could we add a comment here noting that? Without context it could look like an oversight.

fhanuman (changes requested):
The description mentions MIOpen Conv Algorithm Caching and MIOpen Tensor Descriptor Fix, but I don't see these in the final diff. It looks like the layernorm changes might have been reverted by the second commit, and the miopen caching may not have been added yet? Could we update the description to match what's actually here?

zpye:
> The description mentions MIOpen Conv Algorithm Caching and MIOpen Tensor Descriptor Fix, but I don't see these in the final diff. It looks like the layernorm changes might have been reverted by the second commit, and the miopen caching may not have been added yet? Could we update the description to match what's actually here?

PR message has updated to describe the actual changes.

3rd-party/custom_kernels/hip/gqa_kernel.hip
zpye: The missing documents have been restored.

lib/Runtime/real/gqa.cpp
zpye: Fixed. The GQA_HIP_CHECK and GQA_HIPBLAS_CHECK macros have been removed entirely. gqa.cpp now uses the shared error_check_macros.h definitions:
```
#define HIP_CHECK(cmd)     HIP_CHECK_GOTO(cmd, cleanup)
#define HIPBLAS_CHECK(cmd) HIPBLAS_CHECK_GOTO(cmd, cleanup)
```
All error paths -- including hipBLASLt calls, hipMemcpyAsync, and every kernel launcher -- now route through goto cleanup

lib/Runtime/real/gqa.cpp
zpye: I have used two GPU kernels to solve the issue:

- kv_cache_append_kernel (same-buffer / aliased case): Only appends new tokens at [past_len, past_len+sq) via BSHD->BNSD transpose. Past data is already in place.

- kv_cache_concat_kernel (separate-buffer case): A single kernel that fills the entire present buffer -- copying past data at [0, past_len) with correct stride handling AND transposing new tokens at [past_len, past_len+sq).

The dispatch uses past_key != present_key to choose between the two paths. The concat kernel accepts separate past_seq and present_seq stride parameters to handle the mismatch correctly.

lib/Runtime/real/gqa.cpp
zpye: I have added a comment here.
The main reason is all matrix layouts of hipBLASLt are created with hard-coded HIP_R_16F.
Previous implementation could be templated because the whole GQA used custom kernels. Current one should stay consistent with hipBLASLt.
FP32 support could be a future work.

fhanuman (approved):

PR #63, opened 2026-03-26
build: stop installing hip_custom_kernels.h to prevent version mismatch

zpye:
## Summary

- Remove the `install(FILES hip_custom_kernels.h DESTINATION include)` directive from `3rd-party/custom_kernels/CMakeLists.txt`
- The header is no longer copied to `<install_prefix>/include/` during `cmake --install`
- The static library `hip_custom_kernels.lib` continues to be installed normally

## Problem

When `hip_custom_kernels.h` is installed to the prefix, EP builds may pick up a stale installed copy of the header instead of the current in-tree version. This causes version mismatch issues — for example, when a new function is added to the header (like an extra parameter to `hip_gather`), the installed header still has the old signature, leading to linker errors or silent ABI breakage.

## Solution

Stop installing the header to the prefix. All EP build targets already include the in-tree header via `HIP_CUSTOM_KERNELS_INCLUDE_DIR` (exported by the same CMakeLists.txt at line 48), so the installed copy was redundant. This guarantees the header always matches the compiled static library.

## Test plan

- [x] Full configure + build + install succeeds
- [x] Install output no longer contains `hip_custom_kernels.h`
- [x] All 62 LIT tests pass

### DIFF
--- a/3rd-party/custom_kernels/CMakeLists.txt
+++ b/3rd-party/custom_kernels/CMakeLists.txt
@@ -12,7 +12,6 @@
 # Build flow:
 #   .hip files → hipcc -c → .obj → static lib (hip_custom_kernels.lib)
 #   cmake --install → <prefix>/lib/hip_custom_kernels.lib
-#                    → <prefix>/include/hip_custom_kernels.h
 
 cmake_minimum_required(VERSION 3.18)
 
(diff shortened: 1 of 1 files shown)


amd-mingw (approved):

amd-mingw:
this PR fixed a build error, so just approve and merge it.

PR #65, opened 2026-03-26
SkipSimplifiedLayerNorm: full MS spec alignment + GatherOp axis default value

amd-genmingz:
## Motivation
1. The current `hip.skip_rms_norm` does not match the full Microsoft `com.microsoft.SkipSimplifiedLayerNormalization` spec — the optional `bias` input is missing and `input_skip_bias_sum` output is always required. Models using bias cannot be loaded.
2. `hip.gather` requires `axis` to be explicitly specified, while the ONNX Gather spec defaults it to 0.
## Details
**SkipSimplifiedLayerNormalization alignment** (10 files):
- **HipOps.td**: Add `AttrSizedOperandSegments`, optional `bias` input, rename `skip_output` → `input_skip_bias_sum` (optional)
- **HipDialect.cpp**: Dynamic operand offset computation in `getDpsInitsMutable()`
- **OnnxToHip.cpp**: Handle 3-4 inputs (NoneType for bias), 1-4 outputs (mean/inv_std_var mapped to empty), build `operand_segment_sizes`
- **HipToLLVM.cpp**: Signature expanded from 10 to 11 params; optional inputs/outputs pass nullptr
- **Runtime**: `wrap_skip_simplified_layer_norm` gains `bias` param with full `input + skip + bias` via extra `miopenOpTensor(ADD)`; auto-allocates temp buffer when `input_skip_bias_sum` is nullptr
- **Tests**: New cases for with-bias, output-only, and 4-output ONNX pattern

**GatherOp default axis** (1 file):
- **HipOps.td**: `I64Attr:$axis` → `DefaultValuedAttr<I64Attr, "0">:$axis` to match ONNX spec
**Other** (1 file):
- **custom_kernels/CMakeLists.txt**: Install header to `<prefix>/include/`

## todo:
- [x] run ctest
- [x] accuracy test

### DIFF
--- a/lib/Dialect/IR/HipDialect.cpp
+++ b/lib/Dialect/IR/HipDialect.cpp
@@ -329,23 +329,11 @@ LogicalResult RmsNormOp::verify() {
 }
 
 //===----------------------------------------------------------------------===//
-// SkipRmsNormOp: ins(input, skip, gamma, [bias])
-//                outs(output, [input_skip_bias_sum])
+// SkipRmsNormOp: ins(input, skip, gamma, [bias])  outs(...variadic...)
 //===----------------------------------------------------------------------===//
 
 MutableOperandRange SkipRmsNormOp::getDpsInitsMutable() {
-  // Fixed inputs: ctx(1) + input(1) + skip(1) + gamma(1) = 4
-  // Optional input: bias(0|1)
-  unsigned numInputs = 4;
-  if (getBias())
-    ++numInputs;
-
-  // DPS inits: output(1) + input_skip_bias_sum(0|1)
-  unsigned numInits = 1;
-  if (getInputSkipBiasSum())
-    numInits = 2;
-
-  return MutableOperandRange(*this, /*start=*/numInputs, /*length=*/numInits);
+  return getOutputsMutable();
 }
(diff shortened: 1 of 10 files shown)


lib/Dialect/IR/HipDialect.cpp
```diff
@@ -329,13 +329,23 @@ LogicalResult RmsNormOp::verify() {
...
+  unsigned numInputs = 4;
+  if (getBias())
+    ++numInputs;
+
+  // DPS inits: output(1) + input_skip_bias_sum(0|1)
+  unsigned numInits = 1;
+  if (getInputSkipBiasSum())
+    numInits = 2;
+
+  return MutableOperandRange(*this, /*start=*/numInputs, /*length=*/numInits);
```
MaheshRavishankar: This sort of works, but one way to have made this more ergonomic is if there was an specific `Variadic<...>` operands list called `outputs` (or something like that) that was just part of each op definition ,and then we interpret the outputs/results based on whether the number of outputs is 1 or 2. But that is more an ergonomic change. The current version of the op is more readable

MaheshRavishankar (approved):

PR #66, opened 2026-03-26
Fix build warnings: hipError_t, getenv, and LLVM JSON type conversions

mingyueliuh:
## Summary
Eliminates 30+ compiler warnings related to unused `hipError_t` return values, deprecated `getenv()` usage, and MSVC type conversion warnings from external LLVM headers.

### 1. Fixed unused `hipError_t` return value warnings (21 instances)
Added explicit `(void)` casts to HIP API calls in cleanup and error handling paths to indicate intentional best-effort cleanup where errors can be safely ignored:

- **lib/Runtime/hipdnn_ep_runtime_state.cpp** (11 warnings)
  - `hipStreamDestroy()` in error paths
  - `hipFree()` / `hipHostFree()` in cleanup
  - `hipStreamSynchronize()` in cleanup
  
- **lib/Runtime/hipdnn_ep_runtime_tensor.cpp** (3 warnings)
  - `hipFree()` calls in error and cleanup paths
  
- **lib/Runtime/real/simplified_layer_norm.cpp** (1 warning)
  - `hipFree()` in cleanup path
  
- **lib/Runtime/real/skip_simplified_layer_norm.cpp** (1 warning)
  - `hipFree()` in cleanup path
  
- **lib/Runtime/real/test_hip_from_dll.cpp** (5 warnings)
  - `hipStreamDestroy()` in error paths

### 2. Fixed deprecated `getenv()` warning
- **3rd-party/custom_kernels/include/debug_log.h**
  - Replaced `getenv()` with Windows-safe `_dupenv_s()` on Windows platforms
  - Maintained `getenv()` for non-Windows platforms (cross-platform compatibility)
  - Added proper memory cleanup for allocated buffer

### 3. Suppressed MSVC C4244 warnings from LLVM JSON.h (9 warnings)
Added MSVC pragma warnings around LLVM JSON.h includes to suppress type conversion warnings that originate from LLVM's internal implementation:

- **lib/Conversion/OnnxToHip/OnnxToHip.cpp**
- **tools/hip-test-dll/hip-test-dll.cpp**
- **tools/hip-inspect-dll/hip-inspect-dll.cpp**

These warnings occur in LLVM's `json::Value::getAsNumber()` and `getAsInteger()` methods, which internally convert between `int64_t`/`uint64_t` and `double`. Since these conversions are in external library code, suppressing the warnings at the include site is the correct approach.

## Build Impact
- **Before**: 47+ warnings
- **After**: 17 warnings (only benign warnings from submodules/protobuf remain)
- **Total eliminated**: 30+ warnings

Remaining warnings breakdown:
- 7 warnings: `/src: warning: directory does not exist` (protobuf/LLVM build artifacts)
- 6 warnings: `cl : Command line warning D9025` (MSVC flag conflicts)
- 2 warnings: protobuf unused imports
- 2 warnings: From `3rd-party/morphizen` git submodule (external code)

## Testing
- Full clean rebuild completed successfully
- All targets built without errors
- Verified all targeted warnings are eliminated

### Why `(void)` cast for cleanup paths?
In cleanup and error handling code, we perform best-effort cleanup where failure to free resources is non-critical (process is likely terminating or already in an error state). The explicit `(void)` cast documents this intentional behavior and silences the `nodiscard` attribute warning.

### Why `_dupenv_s()` instead of `getenv()`?
The Windows CRT marks `getenv()` as deprecated for security reasons. `_dupenv_s()` is the recommended alternative that:
- Allocates memory for the result (preventing buffer overruns)
- Returns proper error codes
- Requires explicit cleanup (which we handle with `free()`)

### Why pragma suppress for LLVM JSON warnings?
The C4244 warnings originate from LLVM's JSON library implementation, not our code. The warnings occur when LLVM's internal methods convert between integer types and double in `std::optional` return values. Since we cannot modify external library code, suppressing warnings at the include site is the standard practice for handling third-party library warnings.

### DIFF
--- a/lib/Runtime/hipdnn_ep_runtime_state.cpp
+++ b/lib/Runtime/hipdnn_ep_runtime_state.cpp
@@ -14,12 +14,13 @@
 #include <cstring>
 
 // Macro for best-effort cleanup: logs errors but continues cleanup
-#define HIP_CLEANUP(expr) do { \
-    hipError_t _err = (expr); \
-    if (_err != hipSuccess) { \
-        fprintf(stderr, "Warning: " #expr " failed with error %d\n", (int)_err); \
-    } \
-} while(0)
+#define HIP_CLEANUP(expr)                                                      \
+  do {                                                                         \
+    hipError_t _err = (expr);                                                  \
+    if (_err != hipSuccess) {                                                  \
+      fprintf(stderr, "Warning: " #expr " failed with error %d\n", (int)_err); \
+    }                                                                          \
+  } while (0)
 
 // Runtime state management implementation
 
--- a/lib/Runtime/hipdnn_ep_runtime_state.cpp
+++ b/lib/Runtime/hipdnn_ep_runtime_state.cpp
@@ -123,7 +123,7 @@ int hipdnn_ep_state_init_with_fs(RuntimeState **out_state, void *fs,
   // Create MIOpen handle
   if (miopenCreate(&state->miopen_handle) != miopenStatusSuccess) {
     fprintf(stderr, "Failed to create MIOpen handle\n");
-    hipStreamDestroy(state->stream);
+    (void)hipStreamDestroy(state->stream);
     free(state);
     return 7;
   }
@@ -133,7 +133,7 @@ int hipdnn_ep_state_init_with_fs(RuntimeState **out_state, void *fs,
       miopenStatusSuccess) {
     fprintf(stderr, "Failed to set MIOpen stream\n");
     miopenDestroy(state->miopen_handle);
-    hipStreamDestroy(state->stream);
+    (void)hipStreamDestroy(state->stream);
     free(state);
     return 8;
   }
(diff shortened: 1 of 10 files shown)


lib/Runtime/real/skip_simplified_layer_norm.cpp
```diff
@@ -193,7 +193,7 @@ int wrap_skip_simplified_layer_norm(RuntimeState *state, void *input,
   if (rstdDesc)
     miopenDestroyTensorDescriptor(rstdDesc);
   if (rstd_buf)
-    hipFree(rstd_buf);
+    (void)hipFree(rstd_buf);
```
MaheshRavishankar: We should be catching and throwing errors here.

lib/Runtime/hipdnn_ep_runtime_state.cpp
```diff
@@ -123,7 +123,7 @@ int hipdnn_ep_state_init_with_fs(RuntimeState **out_state, void *fs,
   // Create MIOpen handle
   if (miopenCreate(&state->miopen_handle) != miopenStatusSuccess) {
     fprintf(stderr, "Failed to create MIOpen handle\n");
-    hipStreamDestroy(state->stream);
+    (void)hipStreamDestroy(state->stream);
```
fhanuman: Instead of silently ignoring the error, could we do something following? Same in other places too.

```
#define HIP_CLEANUP(expr) do { \
    hipError_t _err = (expr); \
    if (_err != hipSuccess) { \
        fprintf(stderr, "Warning: " #expr " failed with error %d\n", (int)_err); \
    } \
} while(0)
```

fhanuman (approved):

PR #67, opened 2026-03-27
fix: add BUILD_HIP_UNIT_TESTS option and fix decode layer f16 overflow

zpye:
## Summary

- Add **BUILD_HIP_UNIT_TESTS** cmake option (default ON) to control LIT and E2E test registration with CTest. Previously, `enable_testing()` was only called conditionally via `morphizen_ENABLE_UNIT_TEST` in `deps.cmake`, which meant tests were silently unregistered when that flag was OFF.
- Move `enable_testing()` out of `cmake/deps.cmake` into the top-level `CMakeLists.txt` under the new option, so test registration is independent of morphizen unit test settings.
- Reduce MLP weight constants (gate_proj, up_proj, down_proj) in `test_single_layer_decode.mlir` from 1.0 to 1e-3 to prevent f16 overflow in the large matmul chain (4096x14336x4096), matching the pattern from the small-size variant.

## Changed files

- **CMakeLists.txt** - Add `BUILD_HIP_UNIT_TESTS` option wrapping `enable_testing()`, LIT and E2E subdirectories
- **cmake/deps.cmake** - Remove `include(CTest)` / `enable_testing()` block (now controlled by top-level option)
- **test/lit/e2e/test_single_layer_decode.mlir** - Scale MLP weights to avoid f16 overflow

## Test plan

- [x] Fresh configure + build + install passes
- [x] 62/62 LIT tests pass
- [x] 32/32 E2E tests pass with freshly compiled DLLs
- [x] 33/33 all tests pass (clean run)
- [x] Verified 0 tests discovered when `enable_testing()` is disabled (confirming the fix is needed)
- [x] pre-commit run --all-files passes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,9 +131,27 @@ if(BUILD_HIP_TOOLS)
   # Executable targets
   add_subdirectory(tools)
 
-  # LIT tests for MLIR passes
-  add_subdirectory(test/lit)
+  # ==============================================================================
+  # HIP Unit Tests (LIT + E2E)
+  # ==============================================================================
+  # Controls registration of LIT (MLIR pass verification) and E2E (compile MLIR
+  # to DLL and execute on GPU) tests with CTest.
+  #
+  # Usage:
+  #   cmake -DBUILD_HIP_UNIT_TESTS=OFF ...   # disable tests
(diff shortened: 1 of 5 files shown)


amd-mingw (approved):
just fix a ctest issue, approved and merged.

PR #81, opened 2026-04-02
fix: Gather multi-index, GQA pre-allocated KV cache, shared constants, custom kernels discovery

zpye:
## Summary

- **Gather**: support multi-index embedding lookups (variable-length index tensors)
- **GQA**: use `seqlens_k` to determine actual sequence length for pre-allocated KV caches; add `B > 1` guard for non-uniform batch sequences; introduce `HIPBLAS_CLEANUP` macro for consistent hipBLASLt resource cleanup
- **Shared constants cache**: in OGA pipeline mode (prefill + decode), the second model reuses the first model's constants blob via process-wide named shared memory, avoiding a redundant ~15 GB allocation
- **Custom kernels discovery**: add `HIP_CUSTOM_KERNELS_DIR` environment variable for runtime library path override, with fallback to compile-time path and name-only linker search

### Gather multi-index (`gather_kernel.hip`, `gather.cpp`, `HipToLLVM.cpp`, `test_gather.mlir`)
- Extend `gather_axis0_kernel` to support multi-dimensional index tensors (not just scalar/1D)
- Add `indices_num_elements` parameter so `slice_size` and `data_axis_size` are computed correctly for multi-index gathers (e.g. embedding lookups with seq_len > 1)
- Update MLIR lowering to extract and pass the index count
- Update LIT test to match new function signature

### GQA pre-allocated KV cache (`gqa.cpp`, `HipToLLVM.cpp`, `test_gqa.mlir`, `error_check_macros.h`)
- Read `seqlens_k` from device memory (`hipMemcpyAsync` + `hipStreamSynchronize`) to determine actual sequence length, enabling correct attention over pre-allocated (larger-than-used) KV caches
- Use `total_seq` (from `seqlens_k`) for GEMM dimensions and workspace sizing; keep the buffer dimension (`present_seq` / `past_buf_seq`) for memory strides in concat/append/expand kernels
- Zero present KV buffers with `hipMemsetAsync` when `past != present` (separate buffers) so unused slots are deterministic
- Add `B > 1` guard: log error and `goto cleanup` when batch elements have non-uniform `seqlens_k` (batched GEMM pipeline assumes single `total_seq`)
- Add `HIPBLAS_CLEANUP` macro for `hipblasLtMatrixLayoutDestroy` / `hipblasLtMatmulDescDestroy` error checking in cleanup sections
- Extract `past_buf_seq` from MLIR and pass as a separate parameter
- Update LIT test to match new signature

### Shared constants cache (`hipdnn_ep_runtime_state.cpp`, `runtime_state_internal.h`)
- Use Win32 named shared memory (`CreateFileMappingA` / `OpenFileMappingA`) keyed by process ID and blob size to share the GPU constants blob between models in the same OGA pipeline
- Atomic reference counting (`__sync_add_and_fetch` / `__sync_sub_and_fetch`) ensures the last model to clean up frees the GPU memory
- Add `hipStreamSynchronize` before workspace reallocation to prevent use-after-free
- Add null checks on `state->stream` before `hipStreamDestroy` in error paths

### Custom kernels library discovery (`CompilerDriver.cpp`)
- Add 3-tier discovery: (1) `HIP_CUSTOM_KERNELS_DIR` env var, (2) compile-time `HIP_CUSTOM_KERNELS_LIB_PATH`, (3) name-only linker fallback
- When the env var is set, its directory is added to `library_paths` so the JIT linker can find the library at runtime
- Comments document the discovery priority for both end-users and developers

## Test Plan

- [x] Pre-commit checks passed
- [x] Build verified
- [x] LIT conversion tests passed
- [x] E2E runtime tests passed

### DIFF
--- a/lib/Runtime/hipdnn_ep_runtime_state.cpp
+++ b/lib/Runtime/hipdnn_ep_runtime_state.cpp
@@ -218,35 +218,120 @@ int hipdnn_ep_state_init_with_fs(RuntimeState **out_state, void *fs,
   bool is_igpu = (prop.integrated == 1);
 
   if (is_igpu) {
-    // iGPU: allocate pinned host memory. GPU reads the same physical DRAM
-    // directly -- no hipMemcpy needed. hipHostMalloc gives a reliable pinned
-    // allocation; hipHostRegister on mmap'd or VirtualAlloc'd memory is
-    // unreliable on some iGPU platforms (GPU reads zeros despite success).
-    if (hipHostMalloc(&state->gpu_constants_blob, total_size,
-                      hipHostMallocDefault) != hipSuccess) {
-      fprintf(stderr, "hipHostMalloc failed for constants blob (%zu bytes)\n",
-              total_size);
-      hipdnn_ep_state_cleanup(state);
-      *out_state = nullptr;
-      return 1;
-    }
+    // iGPU: try hipMalloc first (1x VRAM pool consumption), then fall
+    // back to hipHostMalloc (2x pool consumption but can spill to GTT).
+    //
+    // On iGPU the VRAM pool is carved from system RAM by the BIOS.
+    // hipHostMalloc consumes ~2x the blob size from this pool (host
+    // allocation + device mapping).  When the pool is large enough to
+    // absorb all constants it gets exhausted and kernel launches fail
+    // with hipErrorLaunchFailure (719).  hipMalloc only consumes 1x,
+    // leaving more pool for kernels.  If hipMalloc fails (pool too
+    // small), hipHostMalloc is used as a fallback — the driver will
+    // partially use the pool and spill the rest to GTT (system RAM),
+    // which the GPU can access directly on iGPU.
+    bool used_hip_malloc = false;
 
-    const void *src = reader->mmap();
-    if (src) {
-      memcpy(state->gpu_constants_blob, src, total_size);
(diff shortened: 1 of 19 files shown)


lib/Runtime/hipdnn_ep_runtime_state.cpp
```diff
@@ -215,86 +255,152 @@ int hipdnn_ep_state_init_with_fs(RuntimeState **out_state, void *fs,
...
     }
 
-    if (hipMemcpy(state->gpu_constants_blob, src, total_size,
-                  hipMemcpyHostToDevice) != hipSuccess) {
-      fprintf(stderr, "hipMemcpy failed for constants blob\n");
-      free(cpu_buf);
-      hipdnn_ep_state_cleanup(state);
-      *out_state = nullptr;
-      return 1;
+#ifdef _WIN32
```
MaheshRavishankar, line 377: Why do you need the `#ifdef _WIN32` ?

AMDmoore: This is the function to share the same weights between two models when OGA integration, because the current solution is to pipeline prefill and decode models in OGA. Here, these are windows specific APIs like CreateFileMappingA and MapViewOfFile, so I wrap them with _WIN32. If hipdnn can support dynamic shape in the future, we can have one model for OGA and may not need this function anymore.

amd-mingw:
https://github.com/ROCm/onnx-hipdnn-ep/pull/109 replace this PR and merged, so close this PR.

PR #87, opened 2026-04-03
ci: disable ORT graph optimization and remove CPU L2 cache

zz002:
## Summary
- Add `-o 0` flag to all perf and L2 test commands to disable ORT graph optimizations for more consistent results
- Remove CPU output caching logic from L2 accuracy tests — always run CPU inference fresh to avoid stale cache issues
- Temporary push trigger included for debugging (to be removed before merge)

## Test plan
- [ ] Workflow runs successfully with `-o 0` flag
- [ ] L2 accuracy tests complete without cache-related issues
- [ ] Remove temporary push trigger after verification

### DIFF
--- a/.github/workflows/gpu-perf-accuracy-test.yml
+++ b/.github/workflows/gpu-perf-accuracy-test.yml
@@ -5,6 +5,7 @@
 name: GPU Performance & Accuracy Test
 
 on:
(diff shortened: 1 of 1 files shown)


amd-mingw:
not used, close it.

PR #91, opened 2026-04-03
perf: persistent runtime state, fused GQA decode kernel, and diagnostic instrumentation

fhanuman:
## Summary

Reduce per-inference CPU dispatch overhead and GPU memory churn in the MorphiZen EP runtime. Persistent descriptor and buffer caching replaces per-call allocation across all runtime operators, and a fused GQA decode kernel replaces the multi-step hipBLASLt decomposition for single-token inference. Also includes opt-in hipBLASLt auto-tuning, HIP Graph capture/replay, and diagnostic instrumentation.

## Design

The central change is making GPU library state -- descriptors, buffers, algorithm selections -- persistent across inferences instead of created and destroyed per call.

1. **Descriptor caching** -- A `queryOrCreate*` pattern caches MIOpen and hipBLASLt descriptors for matmul, GQA, activation, element-wise, layer norm, and skip-layer-norm. Scratch buffers share a `RuntimeState` workspace instead of per-call `hipMalloc`/`hipFree`.

2. **Fused GQA decode** -- For `sq == 1`, a single HIP kernel replaces decomposed hipBLASLt pipeline with cooperative dot product via warp shuffle + online softmax in log2e space, one block per (batch, head\_q), templated on head dimension `d ∈ {64, 128, 256}`. Prefill (`sq > 1`) stays on the decomposed path with cached descriptors.

3. **Auto-tuning and graph replay** -- hipBLASLt requests up to 60 algorithm candidates and benchmarks each on first invocation per shape; winner is cached. HIP Graph (`HIPDNN_EP_GRAPH=1`) captures the `main_graph` GPU work on run 2, replays on run 3+, with codegen support in `GenerateInterface.cpp`.

## Key changes

- **`queryOrCreate*` caches** (`matmul.cpp`, `gqa.cpp`, `activation.cpp`, `elementwise.cpp`, `simplified_layer_norm.cpp`, `skip_simplified_layer_norm.cpp`) -- persistent MIOpen/hipBLASLt descriptor maps
- **Fused GQA decode kernel** (`gqa_kernel.hip`) -- templated `fused_gqa_decode_kernel<d>` with wave shuffle reduction and log2e online softmax
- **GPU KV cache kernels** (`gqa_kernel.hip`) -- `kv_cache_concat` / `kv_cache_append` replace host-side `hipMemcpyAsync` loops
- **WMMA GEMM kernel** (`gemm_wmma_kernel.hip`) -- fp16 wave matrix multiply (currently disabled, for future short-prefill use)
- **Buffer pool** (`hipdnn_ep_runtime_tensor.cpp`) -- size-keyed `pool_alloc` / `pool_release` for GPU I/O tensors
- **HIP Graph replay** (`hipdnn_ep_runtime_tensor.cpp`, `GenerateInterface.cpp`) -- stream capture on run 2, replay on run 3+, with conditional skip codegen
- **Diagnostic env vars** -- `HIPDNN_EP_TIMING` (compiler + session timing), `HIPDNN_EP_PERF` (per-inference H2D/compute/D2H), `HIPDNN_EP_GRAPH` (graph capture)

## Benchmark (AMD W7900, Llama 3.1 8B fp16)

GPU: gfx1150 (Strix Halo / Radeon 8060S)   ROCm 7.1   ISA: gfx1151

| Seq Len | HipDNN EP no autotune (inf/s) | HipDNN EP (inf/s) | DML EP (inf/s) | HipDNN EP (no autotune) vs DML EP | HipDNN EP vs DML EP |
|---------|-------------------------------|-------------------|----------------|-----------------------------------|---------------------|
| 1       | 10.2                          | 9.8               | 11.44          | 0.89x                             | 0.86x               |
| 128     | 7.03                          | 7.54              | 8.0            | 0.88x                             | 0.94x               |
| 256     | 5.05                          | 5.42              | 5.75           | 0.88x                             | 0.94x               |
| 512     | 3.19                          | 3.21              | 2.36           | 1.35x                             | 1.36x               |
| 1024    | 1.53                          | 1.54              | 0.8            | 1.91x                             | 1.93x               |
| 2048    | 0.67                          | 0.68              | 0.305          | 2.19x                             | 2.23x               |
| 3072    | 0.42                          | 0.43              | 0.125          | 3.36x                             | 3.44x               |

Benchmarked with TheRock 7.13.0. Short sequences (`sq ≤ 256`): DML EP is 6-14% faster; hipBLASLt underperforms on small matrix sizes. Long sequences (`sq ≥ 512`): HipDNN EP outperforms DML EP by 1.35-3.44x as hipBLASLt GEMMs dominate compute.


lib/Runtime/real/matmul.cpp
```diff
@@ -88,157 +304,63 @@ int wrap_hipblasLtMatmul(RuntimeState *state, const void *A, const void *B,
...
-  }
-  if (matC_layout) {
-    hipblasLtMatrixLayoutDestroy(matC_layout);
-  }
-  if (matmul_desc) {
-    hipblasLtMatmulDescDestroy(matmul_desc);
-  }
-
-  return result;
+  return 0;
```
tejus26, line 390: Can we add operator level unit tests? It will help benchmark performance and validate the implementation here.

fhanuman: Good point. Current e2e test coverage is not that great; we should add more tests. A lot of these implementations were done by @amd-mingw 's team, and they probably can help here. If they don't have bandwidth, I can look into it in a follow-up task as this PR is already quite big.

backend-mlir-compiler/custom-op-mlir/src/InferenceState.cpp
```diff
@@ -62,6 +69,11 @@ InferenceState::create(const std::vector<uint8_t> &dll_bytes,
     dll_out.close();
   }
 
+  if (timing) {
```
zpye: You can try using a macro to wrap the call here, similar to `RUNTIME_DEBUG_LOG`.

backend-mlir-compiler/level-1-pass/src/pass_main.cpp
```diff
@@ -213,16 +215,30 @@ struct Level1MlirPass {
   void process(IPass &self, Graph &graph) {
     MY_LOG(1) << "Level1MlirPass::process() called";
 
+    const bool timing = hipdnn_ep_timing_enabled();
+    auto t0 = std::chrono::steady_clock::now();
```
zpye: If timing is not enabled, `std::chrono::steady_clock::now()` does not need to be called.

lib/Compiler/CompilerDriver.cpp
```diff
@@ -102,18 +110,36 @@ bool CompilerDriver::compileImpl(mlir::ModuleOp module,
                                  const std::string &output_path,
                                  const mlir::hip::CompilationOptionsT &options,
                                  std::string &error_message) {
+  const bool timing = hipdnn_ep_timing_enabled();
+  auto phaseStart = std::chrono::steady_clock::now();
+  auto totalStart = phaseStart;
+
+  auto logPhase = [&](const char *name) {
```
zpye, line 117: As mentioned above, you can try to wrap using macros to unify the way logs are printed.

lib/Runtime/real/activation.cpp
```diff
@@ -24,13 +24,14 @@ static miopenDataType_t hipdnn_ep_to_miopen_type(int64_t data_type) {
...
     return miopenFloat;
   }
 }
 
-// Maps HIPDNN_EP_ACTIVATION_* to miopenActivationMode_t.
-// MIOpen calls sigmoid "logistic" (miopenActivationLOGISTIC).
-static miopenActivationMode_t hipdnn_ep_to_miopen_activation(int64_t mode) {
+static miopenActivationMode_t hipdnn_ep_to_miopen_activation(int64_t mode,
+                                                             bool *ok) {
+  *ok = true;
```
zpye: Please check if it is a null pointer, or use a reference instead.

lib/Runtime/real/activation.cpp
```diff
@@ -41,18 +42,109 @@ static miopenActivationMode_t hipdnn_ep_to_miopen_activation(int64_t mode) {
...
+  miopenActivationMode_t act =
+      hipdnn_ep_to_miopen_activation(key.activation_mode, &act_ok);
+  if (!type_ok || !act_ok)
+    return nullptr;
+  int n = static_cast<int>(key.num_elements);
+
+  ActivationCacheEntry e{};
+  miopenStatus_t st;
+
+#define ACT_CACHE_CHECK(call)                                                  \
```
zpye: Please try using MIOPEN_CHECK_GOTO.

lib/Runtime/real/elementwise.cpp
```diff
@@ -47,13 +49,122 @@ static miopenTensorOp_t hipdnn_ep_to_miopen_op(int64_t tensor_op) {
...
+    return &it->second;
+
+  bool type_ok;
+  miopenDataType_t dt = hipdnn_ep_to_miopen_type(key.data_type, &type_ok);
+  if (!type_ok)
+    return nullptr;
+  OpTensorCacheEntry e{};
+  miopenStatus_t st;
+
+#define OP_CACHE_CHECK(call)                                                   \
```
zpye: Please try using MIOPEN_CHECK_GOTO.

lib/Runtime/hipdnn_ep_runtime_tensor.cpp
```diff
@@ -319,15 +494,81 @@ int hipdnn_ep_tensor_prepare_output(RuntimeState *state, span_t *outputs,
...
-  if (hipMalloc(&gpu_ptr, size_bytes) != hipSuccess) {
+  // PERF: record H2D end on first output alloc (after all H2D copies queued)
+  if (hipdnn_ep_perf_enabled() && index == 0 && g_perf.initialized) {
+    (void)hipEventRecord(g_perf.h2d_end,
+                         static_cast<hipStream_t>(state->stream));
+  }
+
+  // GRAPH: begin stream capture on inference #2 (after warm-up)
+  // Placed AFTER the PERF event so h2d_end is NOT inside the captured graph.
+  if (hipdnn_ep_graph_enabled() && index == 0 &&
```
zpye: There is a question about whether HIP Graph can handle situations where the kernel call flow changes. For example, the prefill and decode processes in LLM inference might use different kernels.

zpye (commented):
Most updates look good to me. There is a question about whether HIP Graph can handle situations where the kernel call flow changes. For example, the prefill and decode processes in LLM inference might use different kernels.

amd-mingw:
Hi @fhanuman , this is a test CI about this PR91: https://github.com/ROCm/onnx-hipdnn-ep/actions/runs/24121537835/job/70376443119?pr=101
the perf result as following:
image
image
seems it has a little uplift, but not big. and not as expect to this PR description.
cc @tejus26

lib/Runtime/hipdnn_ep_runtime_tensor.cpp
tejus26: This should ideally be two different HIP Graph compilations. right, @fhanuman ?

tejus26:
> Hi @fhanuman , this is a test CI about this PR91: https://github.com/ROCm/onnx-hipdnn-ep/actions/runs/24121537835/job/70376443119?pr=101 the perf result as following: image image seems it has a little uplift, but not big. and not as expect to this PR description. cc @tejus26

lib/Runtime/hipdnn_ep_runtime_tensor.cpp
fhanuman: You both are probably right. I was experimenting, and this was one of the things I tried to see if it benefited. I did this to avoid kernel launch overheads; however, I didn’t see much benefit. I have reverted the changes for now. Let me think this through properly, and I will have a follow-up PR this.

fhanuman:
> Hi @fhanuman , this is a test CI about this PR91: https://github.com/ROCm/onnx-hipdnn-ep/actions/runs/24121537835/job/70376443119?pr=101 the perf result as following: image image seems it has a little uplift, but not big. and not as expect to this PR description. cc @tejus26



> > Hi @fhanuman , this is a test CI about this PR91: https://github.com/ROCm/onnx-hipdnn-ep/actions/runs/24121537835/job/70376443119?pr=101 the perf result as following: image image seems it has a little uplift, but not big. and not as expect to this PR description. cc @tejus26

Hi @tejus26, Hi @amd-mingw, I have clarified this here.

https://teams.microsoft.com/l/message/19:meeting_ZGQ2OTNiZTYtMTRhMi00NTQ0LWIyZDctYTFhOTU4ZTcwMDg4@thread.v2/1775665245939?context=%7B%22contextType%22%3A%22chat%22%7D

backend-mlir-compiler/custom-op-mlir/src/InferenceState.cpp
fhanuman: Thanks for this and other feedbacks. I have made the changes.

zpye (approved):

PR #94, opened 2026-04-07
ci: disable CPU EP fallback

zz002:
## Summary
- Upgrade TheRock dist version from 7.11.0 to 7.12.0
- Add `-C "session.disable_cpu_ep_fallback|1"` to all perf test invocations to ensure ops not supported by MorphiZen EP are not silently falling back to CPU

## Test plan
- [ ] gpu-perf-accuracy-test workflow runs with new TheRock version
- [ ] perf_test correctly fails on unsupported ops instead of falling back to CPU


amd-mingw:
https://github.com/ROCm/onnx-hipdnn-ep/pull/96 replace this PR, so close it.

PR #96, opened 2026-04-07
CI:disable cpu fallback

amd-mingw:
Add -C "session.disable_cpu_ep_fallback|1" to all perf test invocations to ensure ops not supported by hipdnn EP are not silently falling back to CPU


amd-mingw:
this PR just update one CI args, just merged it.

PR #101, opened 2026-04-08
ci: perf test with PR#91 build (GQA + full_model only)

zz002:
## Summary
- Temporarily use gpu-test-package from PR #91 build (run 24106718554) to validate performance improvements
- Only test GQA ops + full_model for GPT-OSS-20B, llama3.1-8B-fp16, phi4-int4-b32_space
- Trigger on push instead of schedule

**DO NOT MERGE** — this is a temporary hack PR for testing purposes.


amd-mingw:
not used, close it.

PR #102, opened 2026-04-08
Add Gemm op support

amd-genmingz:
## Motivation

Support the ONNX Gemm(https://github.com/onnx/onnx/blob/main/docs/Operators.md#gemm) operator (Y = alpha * A' * B' + beta * C) in the HIP dialect

## Technical Details

Define `hip.gemm` DPS op in HipOps.td with alpha, beta, transA, transB attributes and optional input C.
Define conversion from onnx.Gemm to llvm.gemm
Impl runtime

## Test Plan

- [x] Build succeeds 
- [x] Convert onnx.Gemm to hip.gemm success
- [x] convert hip.gemm to llvm.gemm success
- [x] run gemm in gpu success
- [x] accuracy good enough


lib/Conversion/OnnxToHip/GemmConversion.cpp
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+#include "OnnxToHipUtils.h"
+
```
fhanuman, line 7: Could we please add a `lit` test to lock down `onnx` to `hip` conversion?

lib/Conversion/HipToLLVM/GemmLowering.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+#include "HipToLLVMUtils.h"
+
+namespace mlir {
+namespace hip {
+namespace {
+
+// hip.gemm(handle, input_A, input_B, input_C, output, alpha, beta, transA,
+// transB, typeCode)
+struct GemmOpLowering : public ConvertOpToLLVMPattern<GemmOp> {
```
fhanuman, line 14: Could we please add a `lit` test to lock down `hip` to `llvm` conversion?

lib/Runtime/real/gemm.cpp
```diff
@@ -0,0 +1,284 @@
...
+    hblB_cols = K;
+    hblB_ld = M;
+  }
+
+  HIPBLAS_CHECK(hipblasLtMatrixLayoutCreate(&matA_layout, dataType, hblA_rows,
+                                            hblA_cols, hblA_ld));
+  HIPBLAS_CHECK(hipblasLtMatrixLayoutCreate(&matB_layout, dataType, hblB_rows,
+                                            hblB_cols, hblB_ld));
+  // C and output: [M,N] row-major → col-major [N,M] ld=N
+  HIPBLAS_CHECK(hipblasLtMatrixLayoutCreate(&matC_layout, dataType, N, M, N));
```
fhanuman, line 307: matC_layout assumes C is always [M, N], but the [spec](https://onnx.ai/onnx/operators/onnx__Gemm.html#gemm-13) allows C to be broadcastable to (M, N) -- e.g. [N] for a linear layer bias. If C is 1-D, wouldn't this read out of bounds? Could we broadcast C into the output buffer first and do the matmul with C_ptr = output, beta = 1.0?

fhanuman (changes requested):

lib/Conversion/OnnxToHip/GemmConversion.cpp
amd-genmingz: Added in a later commit

lib/Conversion/HipToLLVM/GemmLowering.cpp
amd-genmingz: Added in a later commit

lib/Runtime/real/gemm.cpp
amd-genmingz: I have added broadcast support in the latest commit.

fhanuman (approved):

PR #104, opened 2026-04-09
perf: defer constant writes to reduce I/O interleaving in OnnxToHip pass

BoarQing:
## Summary
- Defers all constant writes in `ConvertOnnxToHipPass` until after MLIR op traversal is complete, separating the read phase (walking ops) from the write phase (sequential I/O to `constants.bin`).
- Raw constant data uses zero-copy descriptors: only the pointer is stored, with `DenseElementsAttr` kept alive to prevent MLIR from freeing the underlying buffer before flush.
- Splat constants and zero-padding are expanded at flush time using dedicated scratch buffers.

## Motivation
The original code interleaves MLIR attribute reads with `fwrite` calls through a virtual `FileWriter` interface. Batching all writes to the end reduces per-constant virtual call overhead and avoids interleaved read/write patterns on the filesystem layer.

## Test plan
- [ ] CI E2E tests pass
- [ ] Run `onnxruntime_perf_test` with `full_model_seq1.onnx` and compare session creation time vs main
- [ ] Verify numerical correctness with `hip-onnx-runner --l2norm`

### DIFF
--- a/lib/Conversion/OnnxToHip/OnnxToHip.cpp
+++ b/lib/Conversion/OnnxToHip/OnnxToHip.cpp
@@ -154,6 +154,68 @@ struct ExternalizationState {
   std::string binFileName;
   llvm::SmallVector<int64_t> constantSizes;
   llvm::SmallVector<int64_t> constantOffsets;
+
+  struct WriteEntry {
+    enum Kind { kRaw, kSplat, kZeros };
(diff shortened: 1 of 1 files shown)


BoarQing:
perf: https://github.com/ROCm/onnx-hipdnn-ep/actions/runs/24175257640

PR #105, opened 2026-04-10
Fix/no random fp16 nan inf

lishixlnx:
## Motivation

previous version: when produce random data from input tensor,
there will be bad data ( NaN & Inf) for fp16 type input tensor;
this fix will prevent this by limit the random scope in valid  scope.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


amd-mingw (approved):
this PR fix a data range bug, just approved and merged it.

PR #109, opened 2026-04-12
fix: GQA pre-allocated KV cache, MIOpen 7.12 layout, shared constants, OGA CI

zpye:
## Summary

- Add GQA pre-allocated KV cache support with zero-D2H decode path for LLM models
- Fix MIOpen 7.12+ layout warnings for genuinely 4D NCHW tensors without breaking hipBLASLt/Tensile on gfx1151
- Add shared constants cache via Win32 named shared memory, saving ~15GB in OGA pipeline mode
- Integrate OGA (onnxruntime-genai) build and benchmark into CI pipeline, including DirectML EP support in the OGA build
- Optimize GQA prefill by removing redundant KV-cache tail memset
- Validate `seqlens_k` against KV buffer capacity in both kernels and host-side (defense-in-depth against malformed values)
- OGA integration refinements: EP name autodetection from genai_config.json / provider_options, shallow submodule init, and OGA_REF updates to pick up the latest feat/oga_hipdnn_experiment fixes

### Commits

1. **`feat: add HIP_CUSTOM_KERNELS_DIR env var for runtime library discovery`** — Allow user to override the custom kernels library path at runtime via `HIP_CUSTOM_KERNELS_DIR` environment variable, with a three-tier fallback chain (env var → CMake path → name-only linker search).

2. **`fix: GQA pre-allocated KV cache support with zero-D2H decode path`** — Restructure GQA to support pre-allocated KV caches. Decode path passes `seqlens_k` device pointers directly to kernels, eliminating host-device copies. Prefill path retains D2H for hipBLASLt's host-side dimension requirements. Adds `past_buf_seq` to the `wrap_group_query_attention` signature.

3. **`ci: remove -I flag from onnxruntime_perf_test`** — Fix perf test CI in both GPU perf and windows-build workflows by removing the unsupported `-I` flag.

4. **`fix: use miopenSetNdTensorDescriptorWithLayout for explicit NCHW layout`** + **`fix: restrict miopenSetNdTensorDescriptorWithLayout to genuinely 4D tensors`** — Replace `miopenSet4dTensorDescriptor` with `miopenSetNdTensorDescriptorWithLayout(..., miopenTensorNCHW, ...)` for genuinely 4D tensor sites (convolution, activation, elementwise). Layer norm files retain `miopenSetTensorDescriptor` with proper ND shapes and explicit strides — padding 1D/2D tensors to 4D NCHW caused hipBLASLt/Tensile kernel loading failures on gfx1151.

5. **`feat: shared constants cache via Win32 named shared memory`** — In OGA pipeline mode, prefill and decode models share the same `constants.bin`. Uses `CreateFileMappingA` / `OpenFileMappingA` with atomic ref counting so the second model reuses the first's GPU blob, saving ~15GB.

6. **`ci: integrate OGA build and benchmark into CI pipeline (#125)`** — Add onnxruntime-genai (OGA) build steps and `model_benchmark.exe` to the CI pipeline. OGA artifacts are packaged into gpu-test-package and benchmarked on the GPU runner with MorphiZen EP. Results (TTFT, TPS, peak memory) are published alongside existing perf results. Includes sccache, artifact caching, and expanded result table formatting.

7. **OGA CI refinements** (4 commits) — Follow-up fixes to the OGA CI integration:
   - `ci: read EP name from genai_config.json for OGA benchmark` — Autodetect EP name instead of hardcoding.
   - `ci: fix EP name extraction from provider_options object key` — Handle provider_options as an object keyed by EP name.
   - `ci: skip OGA checkout on cache hit and use shallow submodule init` — Speed up warm builds and reduce submodule fetch cost.

8. **OGA build with DirectML EP support** (2 commits) —
   - `ci: build OGA with DirectML EP support` — Enable `--use_dml` when building OGA so the benchmark can exercise the DML-backed decoder-only pipeline.
   - `ci: update OGA_REF to latest feat/oga_hipdnn_experiment commit` and `ci: update OGA_REF to pick up DML EP pipeline fix` — Bump OGA_REF to `5b547c7ec3be521d8590ae0c5f3e0ac450af0b7a` (AMDmoore/onnxruntime-genai) to include the DML EP pipeline mode support.

9. **`perf: remove redundant KV cache tail memset in GQA prefill`** — Remove `hipMemsetAsync` zeroing of the unused KV-cache tail `[total_seq, present_seq)`. The concat/append kernels fully write `[0, total_seq)`, and downstream `expand_kv` only reads that range, so zeroing is pure memory-bandwidth waste during prefill.

10. **`fix(gqa): validate seqlens_k against KV buffer capacity`** — Defense-in-depth against malformed `seqlens_k` values (negative, or exceeding the pre-allocated KV cache buffer):
    - `fused_gqa_decode_kernel`: clamp `eff_skv` to `[0, max_seq]` so a bad value cannot drive the attention loop past the current `(batch, head_kv)` cache slice.
    - `kv_cache_append_kernel`: early-return when `eff_past_len < 0` or the write range would overflow the cache, rather than scribble the next slice.
    - `gqa_forward_hipblaslt` (decomposed and fused-decode concat paths): after the `seqlens_k` D2H copy, reject `total_seq < 1`, `past_len < 0`, `total_seq > present_seq`, or `past_len > past_buf_seq` with a clear diagnostic.

#### Custom Kernels
- `3rd-party/custom_kernels/hip/gqa_kernel.hip` — Device-side `seqlens_k` support in `kv_cache_append` kernel; in-kernel bounds clamp/guard for malformed `seqlens_k` in `fused_gqa_decode_kernel` and `kv_cache_append_kernel`
- `3rd-party/custom_kernels/include/hip_custom_kernels.h` — Updated declaration for `kv_cache_append` signature

#### MLIR Lowerings
- `lib/Conversion/HipToLLVM/GqaLowering.cpp` — Pass `past_buf_seq` to GQA runtime wrapper
- `lib/Conversion/HipToLLVM/ElementwiseLowering.cpp` — Comment update (miopenSetNdTensorDescriptorWithLayout reference)
- `lib/Conversion/HipToLLVM/HipToLLVMUtils.h` — Comment update (miopenSetNdTensorDescriptorWithLayout reference)

#### Runtime
- `lib/Runtime/real/gqa.cpp` — Pre-allocated KV cache with `past_buf_seq`, zero-D2H decode via device `seqlens_k` pointer, removal of redundant KV-cache tail memset in prefill, and host-side validation of `seqlens_k` against `present_seq` / `past_buf_seq`
- `lib/Runtime/real/activation.cpp` — `miopenSetNdTensorDescriptorWithLayout` for 4D NCHW descriptors
- `lib/Runtime/real/elementwise.cpp` — `miopenSetNdTensorDescriptorWithLayout` for 4D NCHW descriptors
- `lib/Runtime/real/miopen.cpp` — `miopenSetNdTensorDescriptorWithLayout` for convolution 4D NCHW descriptors
- `lib/Runtime/real/error_check_macros.h` — Comment update (miopenSetNdTensorDescriptorWithLayout reference)
- `lib/Runtime/hipdnn_ep_runtime.h` — Updated `wrap_group_query_attention` signature (`past_buf_seq`)
- `lib/Runtime/hipdnn_ep_runtime_state.cpp` — Shared constants cache via Win32 named shared memory, add stream sync before workspace realloc, null guards on `hipStreamDestroy`
- `lib/Runtime/runtime_state_internal.h` — Shared constants fields (`constants_is_shared`, `shared_constants_mapping`, `shared_constants_view`)
- `lib/Runtime/mock/mock_gpu.cpp` — Add `miopenSetNdTensorDescriptorWithLayout` mock alongside existing `miopenSet4dTensorDescriptor`
- `lib/Runtime/mock/mock_types.h` — Add `miopenTensorLayout_t` and `miopenTensorNCHW`
- `lib/Compiler/CompilerDriver.cpp` — `HIP_CUSTOM_KERNELS_DIR` env var with three-tier discovery fallback

#### Tests
- `test/lit/Conversion/hip-to-llvm/test_gqa.mlir` — Updated CHECK lines for `past_buf_seq`

#### CI
- `.github/workflows/gpu-perf-accuracy-test.yml` — Remove `-I` flag
- `.github/workflows/windows-build.yml` — Remove `-I` flag; integrate OGA build (with DirectML EP support), benchmark, artifact caching, EP-name autodetection, shallow submodule init, and expanded result table

## Test plan

- [x] Build verified
- [x] LIT conversion tests passed (83/83)
- [x] E2E runtime tests passed (41/42, pre-existing QMoE numerical-stability failure unchanged)
- [x] Verified MIOpen layout fix does not cause hipBLASLt/Tensile crashes on gfx1151

### DIFF
--- a/3rd-party/custom_kernels/hip/gqa_kernel.hip
+++ b/3rd-party/custom_kernels/hip/gqa_kernel.hip
@@ -191,9 +191,14 @@ __global__ void rope_kernel(
     const __half* __restrict__ cos_cache,
     const __half* __restrict__ sin_cache,
     int seq_len, int num_heads, int head_dim,
-    int half_rot, int past_len) {
+    int half_rot, int past_len,
+    const int* __restrict__ seqlens_k) {
   int idx = blockIdx.x * blockDim.x + threadIdx.x;
   const int batch_idx = __builtin_amdgcn_readfirstlane(blockIdx.y);
+  // When seqlens_k is provided, read past_len from device memory
+  // (ORT convention: seqlens_k[b] = total_tokens - 1, so past_len = total - sq).
+  // All threads read the same address; hardware coalesces into a broadcast.
+  int eff_past_len = seqlens_k ? (seqlens_k[batch_idx] + 1 - seq_len) : past_len;
   int elems = seq_len * num_heads * half_rot;
   if (idx >= elems) return;
 
@@ -203,7 +208,7 @@ __global__ void rope_kernel(
   int s = remaining / num_heads;
   if (s >= seq_len) return;
 
-  int pos = past_len + s;
+  int pos = eff_past_len + s;
   int base = ((batch_idx * seq_len + s) * num_heads + h) * head_dim;
 
   float x1 = __half2float(input[base + d]);
@@ -265,8 +270,10 @@ __global__ void transpose_mid_dims_kernel(
 __global__ void kv_cache_append_kernel(
     const __half* __restrict__ src,
     __half* __restrict__ cache,
-    int sq, int G, int d, int present_seq, int past_len) {
+    int sq, int G, int d, int present_seq, int past_len,
+    const int* __restrict__ seqlens_k) {
   const int batch_idx = __builtin_amdgcn_readfirstlane(blockIdx.y);
+  int eff_past_len = seqlens_k ? (seqlens_k[batch_idx] + 1 - sq) : past_len;
   int idx = blockIdx.x * blockDim.x + threadIdx.x;
   int total_per_batch = sq * G * d;
   if (idx >= total_per_batch) return;
@@ -277,7 +284,7 @@ __global__ void kv_cache_append_kernel(
   int s = remaining / G;
 
   int src_idx = ((batch_idx * sq + s) * G + g) * d + di;
-  int dst_idx = ((batch_idx * G + g) * present_seq + (past_len + s)) * d + di;
+  int dst_idx = ((batch_idx * G + g) * present_seq + (eff_past_len + s)) * d + di;
   cache[dst_idx] = src[src_idx];
 }
 
@@ -553,7 +560,8 @@ extern "C" int hip_gqa_rope(
     void* stream_ptr, const void* input, void* output,
     const void* cos_cache, const void* sin_cache,
     int batch_size, int seq_len, int num_heads,
-    int head_dim, int half_rot, int past_len) {
+    int head_dim, int half_rot, int past_len,
+    const void* seqlens_k) {
   hipStream_t stream = static_cast<hipStream_t>(stream_ptr);
   int elems = seq_len * num_heads * half_rot;
   int blocks = (elems + GQA_THREADS - 1) / GQA_THREADS;
@@ -562,7 +570,8 @@ extern "C" int hip_gqa_rope(
       static_cast<__half*>(output),
       static_cast<const __half*>(cos_cache),
       static_cast<const __half*>(sin_cache),
-      seq_len, num_heads, head_dim, half_rot, past_len);
+      seq_len, num_heads, head_dim, half_rot, past_len,
+      static_cast<const int*>(seqlens_k));
   return hipSuccess;
 }
 
@@ -583,14 +592,16 @@ extern "C" int hip_gqa_transpose_mid_dims(
 // Step 4-5: KV cache append (same-buffer)
 extern "C" int hip_gqa_kv_cache_append(
     void* stream_ptr, const void* src, void* cache,
-    int batch_size, int sq, int G, int d, int present_seq, int past_len) {
+    int batch_size, int sq, int G, int d, int present_seq, int past_len,
+    const void* seqlens_k) {
   hipStream_t stream = static_cast<hipStream_t>(stream_ptr);
   int total = sq * G * d;
   int blocks = (total + GQA_THREADS - 1) / GQA_THREADS;
   kv_cache_append_kernel<<<dim3(blocks, batch_size), GQA_THREADS, 0, stream>>>(
       static_cast<const __half*>(src),
       static_cast<__half*>(cache),
-      sq, G, d, present_seq, past_len);
+      sq, G, d, present_seq, past_len,
+      static_cast<const int*>(seqlens_k));
   return hipSuccess;
 }
 
@@ -817,7 +828,8 @@ fused_gqa_decode_kernel(
     const __half* __restrict__ Vcache,
     __half* __restrict__ O,
     int H, int G, int d, int skv, int max_seq,
-    float scale)
+    float scale,
+    const int* __restrict__ seqlens_k)
 {
     constexpr int NUM_WAVES = D / WAVE_SIZE;
 
@@ -827,6 +839,10 @@ fused_gqa_decode_kernel(
     const int tid     = threadIdx.x;
     const int wave_id = tid / WAVE_SIZE;
     const int lane_id = tid % WAVE_SIZE;
+
+    // When seqlens_k is provided, total_seq = seqlens_k[batch] + 1
+    const int eff_skv = seqlens_k
+        ? __builtin_amdgcn_readfirstlane(seqlens_k[batch] + 1) : skv;
 
     __shared__ float smem[NUM_WAVES];
 
@@ -841,7 +857,7 @@ fused_gqa_decode_kernel(
     float l = 0.0f;
     float O_acc = 0.0f;
 
-    for (int t = 0; t < skv; ++t) {
+    for (int t = 0; t < eff_skv; ++t) {
         float dot = Q_val * __half2float(Kcache[kv_base + (size_t)t * d + tid]);
 
         dot += __shfl_xor(dot, 16);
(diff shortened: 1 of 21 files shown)


aaronn-xilinx:
Hi, few questions:
- why is DirectML EP needed? This also changes OGA to depend on https://github.com/AMDmoore/onnxruntime-genai. What is AMDmoore? We should not add dependency on non-standard OGA.
- the KVcache alloc changes in lib/Runtime/real/gqa.cpp -- were these taken from somewhere else (e.g. some hip/ROCm repo)? Or was all this code created from scratch?

amd-mingw:
> * why is DirectML EP needed? This also changes OGA to depend on https://github.com/AMDmoore/onnxruntime-genai. What is AMDmoore? We should not add dependency on non-standard OGA.

-----Hi @aaronn-xilinx  AMDmoore is Shoucai Ren from the team. add DirectML EP is we want to test dml ep OGA perf for comparision and OGA not support it currently.

lib/Runtime/hipdnn_ep_runtime_state.cpp
```diff
@@ -225,121 +245,159 @@ int hipdnn_ep_state_init_with_fs(RuntimeState **out_state, void *fs,
...
+          state->shared_constants_view = smeta;
+          fprintf(stderr, "[SHARED_CONSTANTS] Published new blob (%zu bytes)\n",
+                  total_size);
+        } else {
+          CloseHandle(new_map);
+        }
+      }
     }
-
-    TIMING_LOG("[Session] hipMemcpy H2D: %.3fs (%zu bytes)\n",
```
fhanuman, line 337: Why are we removing these instrumentation hooks? they are not enabled by default anyways? Are you noticing any performance issues due to these instrumentation hooks?

lib/Runtime/real/miopen.cpp
```diff
@@ -139,19 +139,23 @@ int wrap_miopenConvolutionForward(
...
-
-  // Output: [N, K, H', W']
-  MIOPEN_CHECK(miopenSet4dTensorDescriptor(output_desc, miopenFloat, input_n,
-                                           weights_k, output_h, output_w));
+  // Set tensor descriptors with explicit NCHW layout (float32).
+  // miopenSet4dTensorDescriptor leaves layout as UNKNOWN which triggers
+  // warnings in MIOpen 7.12+.
+  {
+    int in_dims[] = {(int)input_n, (int)input_c, (int)input_h, (int)input_w};
+    MIOPEN_CHECK(miopenSetNdTensorDescriptorWithLayout(
```
fhanuman, line 147: Will this change have any impact on performance?

lib/Runtime/hipdnn_ep_runtime_state.cpp
```diff
@@ -225,121 +245,159 @@ int hipdnn_ep_state_init_with_fs(RuntimeState **out_state, void *fs,
...
+  // by reusing the first model's blob via process-wide named shared memory.
+  bool constants_shared = false;
+
+#ifdef _WIN32
+  char shm_name[128];
+  snprintf(shm_name, sizeof(shm_name), "Local\\hipdnn_const_%lu_%zu",
+           (unsigned long)GetCurrentProcessId(), total_size);
+
+  {
+    void *existing_map = OpenFileMappingA(SHM_FILE_MAP_ALL_ACCESS, 0, shm_name);
```
fhanuman, line 284: I understand the motivation. Just making sure CreateFileMappingA + named shared memory is the right mechanism here. hipdnn_ep_runtime_state.cpp gets linked into every model DLL right? So each session has its own private copy, which I think is why a static singleton won't work. Would it make sense for the EP host to pass a shared-state pointer into hipdnn_ep_state_init_with_fs (same way we pass fs)? Host owns the registry, stays in-process by construction, and the key-collision / cross-process concerns below go away.

lib/Runtime/hipdnn_ep_runtime_state.cpp
```diff
@@ -225,121 +245,159 @@ int hipdnn_ep_state_init_with_fs(RuntimeState **out_state, void *fs,
...
+    if (existing_map) {
+      auto *smeta = (SharedConstantsMeta *)MapViewOfFile(
+          existing_map, SHM_FILE_MAP_ALL_ACCESS, 0, 0,
+          sizeof(SharedConstantsMeta));
+      if (smeta && smeta->blob_size == total_size && smeta->blob_ptr) {
+        long new_ref = shm_ref_inc(&smeta->ref_count);
+        state->gpu_constants_blob = smeta->blob_ptr;
```
fhanuman, line 291: smeta->blob_ptr is only valid in the publisher's address space right? Works today since both sessions are in the same process, but named SHM with Local\ works across processes in the same session. If this ever runs in separate processes, won't session 2 dereference an invalid pointer? Could we store publisher_pid in SharedConstantsMeta and reject when it doesn't match GetCurrentProcessId()?

3rd-party/custom_kernels/hip/gqa_kernel.hip
```diff
@@ -828,6 +840,10 @@ fused_gqa_decode_kernel(
     const int wave_id = tid / WAVE_SIZE;
     const int lane_id = tid % WAVE_SIZE;
 
+    // When seqlens_k is provided, total_seq = seqlens_k[batch] + 1
+    const int eff_skv = seqlens_k
```
fhanuman: eff_skv = seqlens_k[batch]+1 with no clamp against max_seq right? The attention loop a few lines down runs for (int t = 0; t < eff_skv; ++t) and indexes Kcache[kv_base + t*d + tid], so if seqlens_k is ever larger than the pre-allocated buffer we'd read past the KV slice. This gets more load-bearing with this PR since the device seqlens_k is now the source of truth on the decode hot path. Could we pass present_buf_seq in as well and do eff_skv = min(eff_skv, present_buf_seq) in-kernel, or at minimum validate host-side before launch that seqlens_k[b]+1 <= present_buf_seq?

aaronn-xilinx:
> * the KVcache alloc changes in lib/Runtime/real/gqa.cpp -- were these taken from somewhere else (e.g. some hip/ROCm repo)? Or was all this code created from scratch?

thank you. do you know about the KVcache alloc changes in lib/Runtime/real/gqa.cpp -- were these taken from somewhere else (e.g. some hip/ROCm repo)? Or was all this code created from scratch?

lib/Runtime/hipdnn_ep_runtime_state.cpp
AMDmoore: Yes, you are right. As I reply in teams group, we have the same plan as you said.
The issue is that we lack the communication mechanism among model DLLs. We have planned a solution to create a shared constants registry either in EP side or a new runtime shared dll, md5 will be the key, we prefer shared dll now, because it can be extensible for future shared info, and also it seems unreasonable to hold a gpu mem function by EP. 
The solution is easy to implement, but the reason why the temp solution for perf test is still in PR109 is that we think Tejus's memory management solution may also solve this issue, but if the issue remains after MM solution, we will implement our plan.

lib/Runtime/hipdnn_ep_runtime_state.cpp
AMDmoore: I'm not sure about the scenario, two model benchmark tests with the same model? that will have each session in two separate processes, their shm_name should be different. Have I understood you correctly?

lib/Runtime/real/miopen.cpp
zpye: No, it won't. This change is only to be compatible with the new TheRock SDK.

lib/Runtime/hipdnn_ep_runtime_state.cpp
zpye: They were removed by mistake, and now added back.

zpye:
> > * the KVcache alloc changes in lib/Runtime/real/gqa.cpp -- were these taken from somewhere else (e.g. some hip/ROCm repo)? Or was all this code created from scratch?
> 
> thank you. do you know about the KVcache alloc changes in lib/Runtime/real/gqa.cpp -- were these taken from somewhere else (e.g. some hip/ROCm repo)? Or was all this code created from scratch?

All of the changes was written from scratch.

Inspired by your question, we can look for some optimized and verified kernel implementations from other frameworks in future development, and most of the time, CUDA implementations can also serve as references.

3rd-party/custom_kernels/hip/gqa_kernel.hip
zpye: Good catch! Boundary check was ignored.

To fix the problem, clamp is added in fused_gqa_decode_kernel. Some boundary checks are also added on the host side.

lib/Runtime/hipdnn_ep_runtime_state.cpp
fhanuman: You are right, my bad -- the PID in shm_name scopes this to same-process-only, so cross-process never happens. Please ignore this one.

fhanuman (approved):

PR #119, opened 2026-04-15
feat: add Reciprocal and Sqrt operator support

mingyueliuh:
## Summary

Adds ONNX **Reciprocal** and **Sqrt** end-to-end: ONNX → HIP dialect → LLVM `@wrap_power` → runtime. Both ops share one LLVM/runtime entry point **`wrap_power(state, input, output, num_elements, data_type, alpha, beta, gamma)`**; the runtime dispatches on **`(alpha, beta, gamma)`**: HIP elementwise kernels for reciprocal and sqrt, **MIOpen `miopenActivationPOWER`** only for other tuples.

### Reciprocal (`y = 1/x`)

- **Semantics**: ONNX/IEEE element-wise reciprocal on the **full signed** domain (negative `x` must not be collapsed to zero).
- **Why not MIOpen `miopenActivationPOWER`**: Activation-style numerics do not match ONNX for `γ = -1` on negative `x`. Implemented with **`hip_elementwise_reciprocal`** (f32 / f16 / bf16) in `3rd-party/custom_kernels`.
- **Lowering**: `hip.reciprocal` → `@wrap_power(..., 0.0, 1.0, -1.0)` → HIP reciprocal in `power.cpp`.

### Sqrt (`y = sqrt(x)`)

- **Semantics**: ONNX Sqrt (**negative inputs → NaN**). **Not** MIOpen POWER (which did not match ONNX on negatives / edge cases and showed non-zero EP vs CPU L2 on fp16 single-op tests).
- **Implementation**: **`hip_elementwise_sqrt`** in `3rd-party/custom_kernels` (same dtype support as reciprocal). `wrap_power(..., 0.0, 1.0, 0.5)` dispatches to this HIP path.
- **Lowering**: `hip.sqrt` → `@wrap_power(..., 0.0, 1.0, 0.5)` → HIP sqrt in `power.cpp`.

### Compiler / dialect

- **OnnxToHip**: `PowerConversion.cpp` — `onnx.Reciprocal` → `hip.reciprocal`, `onnx.Sqrt` → `hip.sqrt`; **explicit ranked-tensor checks** (`isa` + `notifyMatchFailure`) before `cast` (review feedback).
- **HipToLLVM**: `PowerLowering.cpp` — both lower to `@wrap_power` with the right `α, β, γ`; **`dyn_cast<MemRefType>`** + `notifyMatchFailure` for the `outs` operand (review feedback).
- **Dialect / docs**: `HipOps.td`, bufferization, **`hipdnn_ep_runtime.h`**, **`README.md`** — Reciprocal and Sqrt listed under **Custom HIP**; runtime comments describe HIP vs MIOpen split.

### Tests

- Lit: onnx→hip, hip→llvm, e2e for Reciprocal and Sqrt; `test_reciprocal.mlir` / `test_sqrt.mlir` cover `@wrap_power` LLVM signature.

### Files (high level)

- `lib/Runtime/real/power.cpp` — `wrap_power`: reciprocal HIP, sqrt HIP, other `(α,β,γ)` → MIOpen POWER cache + `miopenActivationForward`.
- `3rd-party/custom_kernels/hip/elementwise_reciprocal_kernel.hip`, **`elementwise_sqrt_kernel.hip`** (+ CMake + `hip_custom_kernels.h`).
- `lib/Runtime/mock/mock_gpu.cpp` — `wrap_power` stub.
- Conversion sources, `README.md`, lit tests as above.


mingyueliuh:
image

mingyueliuh:
image

lib/Conversion/OnnxToHip/PowerConversion.cpp
```diff
@@ -0,0 +1,82 @@
...
+SqrtToHip::matchAndRewrite(mlir::Operation *op,
+                           mlir::PatternRewriter &rewriter) const {
+  auto ctxOrFailure = getContextArg(op, rewriter);
+  if (mlir::failed(ctxOrFailure))
+    return mlir::failure();
+  mlir::Value context = *ctxOrFailure;
+
+  mlir::Location loc = op->getLoc();
+  mlir::Value input = op->getOperand(0);
+  auto resultType =
```
fhanuman, line 77: I understand that `Unranked` tensors don't show up in our case. However, it would be good to add validation checks to make that explicit. Same comment in other places too.

include/hip/Dialect/IR/HipOps.td
```diff
@@ -762,6 +762,66 @@ def Hip_SoftplusOp : Hip_DpsOp<"softplus"> {
...
+  }];
+}
+
+def Hip_SqrtOp : Hip_DpsOp<"sqrt"> {
+  let summary = "Element-wise square root: y = sqrt(x)";
+  let description = [{
+    Computes the square root of each element.
+    Formula: Y = sqrt(X) = X^0.5
+
+    Negative inputs return NaN (per ONNX Sqrt specification).
```
fhanuman, line 801: From the runtime implementation (lib/Runtime/real/power.cpp → MIOpen POWER), we clamp small/negative inputs to 0 instead of returning NaN — won't this cause correctness issues?

fhanuman (approved):

lib/Conversion/OnnxToHip/PowerConversion.cpp
mingyueliuh: Thanks -  we switched to **isa+ notifyMatchFailure** for ranked inputs and results

include/hip/Dialect/IR/HipOps.td
mingyueliuh: You’re right to call this out: miopenActivationPOWER with (α,β,γ)=(0,1,0.5) does not match strict ONNX Sqrt semantics for negative inputs (ONNX: NaN), and the dialect text should not promise ONNX behavior if the runtime does not.

We’ve changed the implementation so hip.sqrt / wrap_power(0,1,0.5) is executed via a HIP elementwise sqrt kernel (hip_elementwise_sqrt), same idea as reciprocal vs MIOpen. That restores negative inputs → NaN and aligns much more closely with ORT CPU / IEEE-style sqrt on the finite domain.

HipOps.td, power.cpp, hipdnn_ep_runtime.h, README, and lowering comments are updated to describe HIP sqrt, not MIOpen, for this path. We re-ran the Sqrt L2 harness (same seed, EP vs CPU); finite elements match as expected; NaN payloads can still differ bitwise between EP and CPU, which is why we still rely on the runner’s non-finite handling for comparisons.

Thanks again for the careful review.

mingyueliuh:
Sqrt kernel changed to use Hip Custom kernel. 
### L2 test result :
image

### small/negative test
Idx | input | ORT CPU EP |  Our EP
-- | -- | -- | --
0 | -64.625 | nan | nan
1 | 151.0 | 12.2890625 | 12.2890625
2 | 229.875 | 15.1640625 | 15.1640625
3 | -162.25 | nan | nan
4 | 118.0625 | 10.8671875 | 10.8671875
5 | 142.375 | 11.9296875 | 11.9296875
6 | 49.90625 | 7.0625 | 7.0625
7 | 49.0 | 7.0 | 7.0
8 | -176.25 | nan | nan
9 | -28.171875 | nan | nan
10 | -176.25 | nan | nan
11 | -204.875 | nan | nan
12 | -226.375 | nan | nan
13 | -21.328125 | nan | nan
14 | 186.625 | 13.6640625 | 13.6640625
15 | -85.5 | nan | nan

PR #120, opened 2026-04-15
fix: use fp32 accumulation for GQA Score GEMM to prevent NaN

zz002:
## Summary
- DeepSeek-R1-Distill-Llama-70B `GroupQueryAttention_seq256` prefill produces NaN because fp16 Score GEMM outputs overflow to inf, then softmax computes inf - inf = NaN
- Switch Score GEMM output to fp32 and add `causal_mask_f32_kernel` + `softmax_f32_to_f16_kernel` so the entire score path stays in fp32 until the final fp16 conversion
- Decode path (fused kernel, sq==1) is unaffected

## Test plan
- [ ] CI windows-build passes (LIT + E2E tests)
- [ ] GPU E2E tests 32/32 pass (especially previously failing `test_single_layer_decode`)
- [ ] DeepSeek-R1-Distill-Llama-70B GQA seq256 output is NaN-free
- [ ] Perf comparison shows no significant regression

### DIFF
--- a/3rd-party/custom_kernels/hip/gqa_kernel.hip
+++ b/3rd-party/custom_kernels/hip/gqa_kernel.hip
@@ -423,6 +423,29 @@ __global__ void causal_mask_kernel(
 }
 
 // -------------------------------------------------------------------------
+// Step 9a (fp32 variant): Causal Mask on fp32 Score matrix
+// -------------------------------------------------------------------------
+
+__global__ void causal_mask_f32_kernel(
+    float* __restrict__ S,
+    int skv, int sq, int batch_stride, int past_len,
+    int local_window_size) {
+  const int head = __builtin_amdgcn_readfirstlane(blockIdx.y);
+  int idx = blockIdx.x * blockDim.x + threadIdx.x;
+  int total = skv * sq;
+  if (idx >= total) return;
+
+  int k = idx % skv;
+  int q = idx / skv;
+  int abs_q = past_len + q;
+  if (k > abs_q) {
+    S[(size_t)head * batch_stride + (size_t)q * skv + k] = -INFINITY;
+  } else if (local_window_size > 0 && k < abs_q - local_window_size + 1) {
+    S[(size_t)head * batch_stride + (size_t)q * skv + k] = -INFINITY;
+  }
+}
+
+// -------------------------------------------------------------------------
 // Step 9b: Column-wise Softmax (numerically stable, in-place)
 // -------------------------------------------------------------------------
 // Per column q: softmax(S[:,q]) = exp(S[:,q]-max) / sum(exp(S[:,q]-max))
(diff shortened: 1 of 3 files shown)


3rd-party/custom_kernels/hip/gqa_kernel.hip
```diff
@@ -422,6 +422,29 @@ __global__ void causal_mask_kernel(
   }
 }
 
+// -------------------------------------------------------------------------
+// Step 9a (fp32 variant): Causal Mask on fp32 Score matrix
+// -------------------------------------------------------------------------
+
+__global__ void causal_mask_f32_kernel(
```
zpye: Could you consider using templates to implement fp16 and fp32 kernels?

3rd-party/custom_kernels/hip/gqa_kernel.hip
```diff
@@ -643,6 +680,96 @@ extern "C" int hip_gqa_causal_mask(
...
+
+  // Pass 3: normalize and write fp16 output
+  float inv_sum = 1.0f / sum_val;
+  for (int i = tid; i < rows; i += blockDim.x) {
+    float e = exp2f((in_col[i] - max_val) * LOG2E);
+    out_col[i] = __float2half(e * inv_sum);
+  }
+}
+
+extern "C" int hip_gqa_softmax_f32_to_f16(
```
zpye, line 747: Is it possible to implement this kernel in an inplace method?

zpye (approved):
Most of it looks good to me.

zpye (commented):
Does this change affect performance?

fhanuman (approved):

3rd-party/custom_kernels/hip/gqa_kernel.hip
zz002: Good suggestion! Refactored to a single template kernel `causal_mask_kernel_impl<T>` with explicit specializations of `causal_mask_fill_value<T>()` for `__half` and `float`. The two launchers now just instantiate the template. Done in the latest push.

3rd-party/custom_kernels/hip/gqa_kernel.hip
zz002: Unfortunately not — the input is fp32 (Score matrix after GEMM) but the output must be fp16 (for the subsequent Value GEMM which expects fp16 B matrix). Since the element sizes differ (4 bytes vs 2 bytes), we cannot reuse the same buffer in-place. The fp32 Score buffer (S_f32) and fp16 softmax output buffer (S_fp16) are both allocated in the workspace.

zz002:
Re: performance impact — comparing CI results:

| | main (PR #116, `6b63d5e`) | This PR (`9f29c6d`) | Delta |
|---|---:|---:|---:|
| **GQA seq128 QPS** | 1219.09 | 1217.97 | -0.09% |
| **full_model QPS** | 7.33 | 7.24 | -1.2% |
| **matmul_down QPS** | 598.91 | 587.02 | -2.0% |

GQA QPS is within noise (< 0.1% difference). The full_model and matmul_down differences are also within normal CI run-to-run variance. This change only affects the GQA decomposed pipeline (prefill path, sq > 1); the decode path (sq == 1) and all non-GQA ops are completely unchanged.

PR #126, opened 2026-04-15
feat: add  CausalConvWithState operator support

amd-genmingz:
## Motivation

Gated DeltaNet (Qwen3.5) and Mamba-based models (Jamba, FalconMamba) use a causal depthwise convolution with carry state for incremental decoding. In ONNX, this is represented as a com.microsoft.CausalConvWithState custom op. Without native support, the runtime falls back to a 3-op pattern (Concat + Conv + Slice), which is suboptimal. This PR adds first-class support for it throughout the entire compilation pipeline.

## Technical Details

Full-stack implementation of hip.causal_conv_with_state:

* Dialect definition (HipOps.td): New DPS op with optional bias/past_state inputs, output/present_state outputs, and activation/ndim attributes.
* ONNX → HIP conversion (CausalConvWithStateConversion.cpp): Matches onnx.Custom with function_name = "CausalConvWithState" and domain_name = "com.microsoft", creates DPS init tensors, and forwards attributes.
* HIP → LLVM lowering (CausalConvWithStateLowering.cpp): Extracts shape info (batch, channels, seq_len, kernel_size) from memref types, converts activation string to enum, and emits an LLVM call to the runtime wrapper.
* GPU runtime (causal_conv_with_state.cpp): Builds a virtual input buffer by prepending past_state to input, runs MIOpen grouped convolution (group=channels for depthwise), adds optional bias via miopenOpTensor, and applies optional SiLU activation via miopenActivationForward + elementwise multiply. Extracts present_state from the tail of the virtual buffer.
* Bufferization & dialect support: Registered bufferizable interface; implemented getDpsInitsMutable() and memory effects.

## Test Plan

onnx-to-hip
- [x] LIT tests: 4 cases (basic with all inputs, no optional inputs, bias-only)

- [x] hip-to-llvm
LIT tests: 3 cases (full op, no optional inputs)

- [x] Verify runtime correctness with end-to-end model inference on a supported AMD GPU

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/lib/Runtime/real/causal_conv_with_state.cpp
+++ b/lib/Runtime/real/causal_conv_with_state.cpp
@@ -1,0 +1,328 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+#include "../debug_log.h"
+#include "../hipdnn_ep_runtime.h"
+#include "runtime_types.h"
+
+#include <cassert>
+#include <cmath>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+
+//===----------------------------------------------------------------------===//
+// CausalConvWithState: Stateful Causal Depthwise Convolution
+//
+// Used by Gated DeltaNet (Qwen3.5) and Mamba (Jamba, FalconMamba).
+// Replaces the 3-op pattern (Concat + Conv + Slice) with a single fused op.
+//
+// Algorithm (1D, ndim=1):
+//   1. Build virtual input by prepending past_state to input:
+//      virtual = [past_state | input]  shape: (batch, channels, k-1 + L)
+//   2. Depthwise causal convolution over virtual input:
+//      For each output position t in [0, L):
+//        output[b,c,t] = sum_{j=0}^{k-1} weight[c,0,j] * virtual[b,c,t+j]
+//      Plus optional bias: output[b,c,t] += bias[c]
+//   3. Extract present_state = last (k-1) positions of virtual input:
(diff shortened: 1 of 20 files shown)


lib/Runtime/real/causal_conv_with_state.cpp
```diff
@@ -0,0 +1,321 @@
...
+      fprintf(stderr, "wrap_causal_conv_with_state: hipMalloc failed for "
+                      "sigmoid buffer\n");
+      ret = -1;
+      goto cleanup;
+    }
+
+    // Compute sigmoid(output) -> sigmoid_buf
+    miopenActivationDescriptor_t actDesc = nullptr;
+    CAUSAL_MIOPEN_CHECK(miopenCreateActivationDescriptor(&actDesc));
+    CAUSAL_MIOPEN_CHECK(miopenSetActivationDescriptor(
```
fhanuman, line 264: if `miopenSetActivationDescriptor` fails, `CAUSAL_MIOPEN_CHECK` jumps to `cleanup`, but `sigmoid_buf` and `actDesc` are local to this block and never get freed. Same thing in the bias block -- `biasDesc` would leak if `miopenSet4dTensorDescriptor` fails. Could we move these to function scope and add them to the cleanup section?

fhanuman (approved):

lib/Runtime/real/causal_conv_with_state.cpp
amd-genmingz: I mistakenly overlooked these memory management needs, which have already been fixed in  d4cb316

amd-genmingz:
image

PR #130, opened 2026-04-16
fix: EP context model crashes on load in onnx.Custom path

BoarQing:
Bumps morphizen submodule to pick up [ROCm/MorphiZen#206](https://github.com/ROCm/MorphiZen/pull/206).

Loading an EP context model currently aborts with exit code 3 inside MLIR's onnx.Custom op construction because `function_name` and `domain_name` get added to the OperationState twice: once programmatically by add_node, once again from the serialized ONNX attributes. The submodule fix checks whether those two attributes are already present before adding them.

**Verification (single_layer_seq128 on gfx1150):**

| Model | Before | After |
|---|---|---|
| EP context model | abort (exit code 3) | session 1.7s, inference OK |
| Normal model | session 10.7s | session 9.5s |


BoarQing:
1. precommit 2. remove all unnecessary logs. if needed, use my log

PR #132, opened 2026-04-17
feat: unified constants emit path with pre-initialized GPU upload for ORT EP

zz002:
## 1. User workflows and the problem

PR132 serves two ORT EP user workflows:

**Workflow A — EPContext export + import (deployment)**

1. One-time export: `onnxruntime_perf_test ... -C "ep.context_enable|1 ep.context_embed_mode|0 ep.context_file_path|foo_ctx.onnx" -I model.onnx`
   - Produces `foo_ctx.onnx` (ONNX metadata shell) + `foo_ctx.onnx_MORPHIZEN.bin` (tar containing `model.dll` + `model.constants.bin`)
2. Deploy / re-launch: `onnxruntime_perf_test ... -I foo_ctx.onnx`
   - Skips compilation, mmap's the sidecar from the tar, single `hipMemcpy` to GPU

**Workflow B — In-process direct compile + run (development / one-shot)**

1. EP ingests the original `model.onnx` directly: `onnxruntime_perf_test ... -I model.onnx`
2. Compile + upload + infer on the fly; no persisted artifact

**The problem is in Workflow B** — two related but independent issues:

**B-1: slow session creation (primary pain point)**

On baseline, to hand constants over to the runtime, each `onnx.Constant`'s raw bytes are written one by one into an intermediate constants file during compilation; then at session-init the runtime reads that file back and copies it to GPU. The same N GB of data goes through the system **three times**: read original `weights.data` → write intermediate file → read intermediate file → hipMemcpy to GPU. For a 15 GB model, session creation takes **41.20 s** (measured on main `bfba96df` in CI) — users pay this cost on every cold start.

**B-2: high peak memory (amplification risk)**

Baseline runtime reads the whole constants blob into host RAM and issues a single `hipMemcpy` to GPU. At the peak moment, the host holds an N GB source while the GPU already has an N GB destination allocated; adding OS overhead the total is roughly **2N + 17 GB**. On a 48 GB APU, the 15 GB model measured at 15.22 GB Peak WS (near saturation); a 30 GB+ model scales linearly past 48 GB and OOMs.

**Workflow A's export stage exhibits the same compile-then-sidecar pattern** (baseline 53.37 s). A is not the primary scope of this PR — its purpose is to persist a portable sidecar for cross-machine deployment — but it picks up a ~12% export speedup (53.37 s → 46.91 s) as a side effect. Import (baseline 7.67 s, this PR 7.43 s) is unchanged.

### Architecture

```mermaid
flowchart TD
    subgraph compileBox [Compile time: OnnxToHip pass]
        direction LR
        Walk["Walk: classify each constant"] --> Decide{"Source kind"}
        Decide -->|"file-ref / splat"| Bake["Bake descriptor<br/>into model.dll metadata"]
        Decide -->|"mem-addr (inline / OrtValue alias)"| Pack["Pack bytes into<br/>partial sidecar +<br/>SidecarSource{sidecar_offset}"]
    end

    Bake --> Artifact["model.dll<br/>(+ partial sidecar<br/>when any mem-addr present)"]
    Pack --> Artifact

    Artifact --> Choose{"Any per-entry source?"}

    subgraph runtimeBox [Runtime: session-init]
        direction LR
        Choose -->|"Yes (Splat / FileRef / Sidecar mix)"| Stream["Per-tensor streaming<br/>through one staging buffer"]
        Choose -->|"No (Workflow A import)"| Bulk["Bulk: read full sidecar<br/>+ single hipMemcpy"]
    end
```

1. **OnnxToHip no longer `fwrite`s the whole constants blob at compile time**: the walk phase only records per-constant source descriptors. Splat and external file-ref entries become tiny in-metadata records; mem-addr entries (DenseAttr raw / OrtValue alias) get packed into a *partial* sidecar that holds only their bytes.
2. **All source info is baked directly into the build artifact**: via the `ConstantSource` union in `HipModelMetaInfo` (`SplatSource` / `FileRefSource` / `SidecarSource`), traveling with `model.dll`. The runtime dispatches per descriptor without probing for file existence.
3. **Runtime streams per descriptor**: at session-init, for each constant in order, fetch from its source (splat tile / file-ref fread / partial-sidecar fread) → `hipMemcpy` → reuse the same staging buffer. Peak host usage equals the size of the single largest tensor rather than the whole model.
4. **Hybrid sidecar for the mem-addr fallback**: previously, any mem-addr constant forced the entire pass to fall back to writing a full N GB sidecar — losing the streaming win for the file-ref + splat majority. The hybrid path keeps mem-addr entries on a compact 64B-aligned partial sidecar (typically MB-class for inline norms / biases), while file-ref and splat constants stay on the streaming path. The runtime per-entry loop handles all three source kinds through the same staging buffer.
5. **Bulk sidecar path preserved for Workflow A import**: when `__metadata_blob` carries no per-entry source (the EPContext export case), runtime still does the single `hipMemcpy` of the full sidecar — unchanged from main.
6. **Compatible with OGA pipeline shared cache**: integrates with the process-wide GPU constants cache introduced in main's PR #109 so prefill + decode can reuse the same GPU blob.

### Path dispatch

**Workflow B — in-process direct compile + run**

| Scenario                     | Trigger                                                                | Constants upload path                                                                                                | User-visible win                                                                |
| ---------------------------- | ---------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------- |
| Streaming path (typical LLM) | All constants are external `weights.data` (file-ref) or splat          | Per-tensor streaming driven by metadata descriptors (**primary win**)                                                | Session time -84%, host peak WS -92%                                            |
| Hybrid path (mixed sources)  | Some constants are mem-addr (inline raw_data / ORT-resolved OrtValue)  | mem-addr packed into a small partial sidecar with `SidecarSource{offset}`; file-ref + splat keep streaming           | Same staging-buffer profile as streaming; sidecar shrunk to mem-addr bytes only |

**Workflow A — EPContext export + import**

| Scenario | Trigger                                | Constants upload path                                     | User-visible win                                         |
| -------- | -------------------------------------- | --------------------------------------------------------- | -------------------------------------------------------- |
| Export   | `ep.context_enable=1` + `embed_mode=0` | Compile-time writes full sidecar into `_MORPHIZEN.bin` tar | Self-contained artifact; cross-machine independent       |
| Import   | Load `_ctx.onnx` directly              | Session-init reads the bulk sidecar from inside the tar   | Same behavior as main                                    |

### Performance (`full_model_seq128.onnx`, 15 GB external data, measured in CI)

Peak WS values below use the same GiB convention as the CI `Publish perf results` step (PowerShell `1GB = 1024³`); MB values from the perf comment are divided by 1024.

Data sources:

- baseline: PR #142 (cherry-picks just the EPContext perf CI step onto origin/main), commit `60cc9d2`, CI run [24882748786](https://github.com/ROCm/onnx-hipdnn-ep/actions/runs/24882748786)
- this PR: commit `fa30493`, CI run [24881732447](https://github.com/ROCm/onnx-hipdnn-ep/actions/runs/24881732447)

**Workflow B — in-process direct compile + run (primary battleground)**

| Metric            | baseline (60cc9d2) | this PR (fa30493) | Delta    | Problem addressed  |
| ----------------- | ------------------ | ----------------- | -------- | ------------------ |
| Session creation  | 41.96 s            | **6.23 s**        | **-85%** | B-1 time (primary) |
| Peak WS           | 15.23 GB           | **1.21 GB**       | **-92%** | B-2 memory         |
| First inference   | 240 ms             | 244 ms            | ±2%      | (no regression)    |
| QPS               | 6.97               | 6.88              | -1%      | (no regression)    |

**Source of the B-1 speedup**: constants no longer round-trip through an intermediate file. The runtime reads the right offset in the original `weights.data` directly per metadata descriptors and `hipMemcpy`s to GPU. One pass through the filesystem instead of three — session time drops from 41.96 s to 6.23 s for the 15 GB model.

**Source of the B-2 Peak WS drop**: the runtime no longer slurps the entire constants blob into host RAM in one shot. It reuses a single staging buffer sized to the largest tensor (≈ 1.05 GB — the embedding weight — for `full_model_seq128`). Host usage scales with the largest single tensor rather than the total model size, so a 48 GB APU can now run 30 GB+ models (GPU VRAM is a separate constraint).

**Workflow A — EPContext export + import (deployment)**

| Metric            | baseline export | this PR export       | baseline import | this PR import       |
| ----------------- | --------------: | -------------------: | --------------: | -------------------: |
| Session creation  |         53.54 s |              48.23 s |          6.76 s |               6.80 s |
| Peak WS           |        15.22 GB |             15.22 GB |        15.26 GB |             15.26 GB |
| First inference   |          221 ms |               218 ms |          226 ms |               239 ms |
| QPS               |            7.32 |                 7.28 |            7.31 |                 7.32 |

A is flat overall: export picks up ~10% (one fewer detour in the sidecar-write path), Peak WS is byte-level unchanged (sub-MB diff vs baseline — within page-cache noise), and import is within noise (sub-50 ms session delta on 6.8 s baseline). The sidecar bake-and-mmap pattern is preserved for Workflow A — the value of this PR is making Workflow B substantially faster and cheaper **without disturbing A's artifact format or behavior**.

**OGA pipeline (end-to-end LLM, prefill + decode sharing the constants cache)**

| Model         | TTFT baseline | TTFT this PR | TPS baseline | TPS this PR | Peak Mem baseline | Peak Mem this PR | Mem Delta |
| ------------- | ------------: | -----------: | -----------: | ----------: | ----------------: | ---------------: | --------- |
| GPT-OSS-20B   |       146.7 ms |     146.7 ms |         56.3 |        55.8 |           12.06 GB |     **1.32 GB**   | **-89%**  |
| Llama-3.1-8B  |       150.8 ms |     153.4 ms |         10.1 |        10.0 |           15.28 GB |     **1.21 GB**   | **-92%**  |

OGA spawns separate prefill + decode model.dlls. With the per-tensor streaming path each one would still pay an N GB hipMemcpy on its own; the process-wide shared constants cache (PR #109, integrated here) lets the second model attach to the first's GPU blob, so OGA Peak Mem collapses to a single staging-buffer allocation rather than 2 × N GB. TTFT and TPS unchanged within noise — same kernels, same scheduling, only the constants-upload path differs.

### Architectural side effects

- **`HipModelMetaInfo` schema extension**: `ConstantInfo.source` becomes a three-way union of `SplatSource` / `FileRefSource` / `SidecarSource`, so the runtime dispatches on metadata semantics instead of probing for file existence.
- **Diff vs main**: 23 files, +1166/−424 (rebased on origin/main `bfba96df`).

### CI verification

- Windows Build (full pipeline): success
- Windows Build (Mock): success
- pre-commit: success
- gpu-test: all perf + L2 + OGA checks pass
- **New EPContext export+import perf step** (scoped to `full_model_seq128`), with two additional tables added to the "Publish perf results" step

### Dependencies

- MorphiZen #207 (external initializer info) — included via submodule pointer
- MorphiZen #206 (EP context attr-leak fix) — included via submodule pointer

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -581,7 +581,7 @@ jobs:
               --plugin_eps "MorphiZenExecutionProvider" `
               --plugin_ep_options "config_file|..\morphizen_config.json" `
               -C "session.disable_cpu_ep_fallback|1 ep.context_enable|1 ep.context_embed_mode|0 ep.context_file_path|$ctxPath" `
-              -t 5 -c 1 -s `
+              -t 5 -c 1 -s -I `
               $m.FullName 2>&1 | Tee-Object -FilePath $exportFile
             if ($LASTEXITCODE -ne 0) {
               Add-Content $exportFile "[ERROR] EPContext export failed"
@@ -601,7 +601,7 @@ jobs:
               --plugin_eps "MorphiZenExecutionProvider" `
               --plugin_ep_options "config_file|..\morphizen_config.json" `
               -C "session.disable_cpu_ep_fallback|1" `
-              -t 30 -c 1 -s `
+              -t 30 -c 1 -s -I `
               $ctxPath 2>&1 | Tee-Object -FilePath $importFile
             if ($LASTEXITCODE -ne 0) { $failed = 1 }
 
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -533,6 +533,84 @@ jobs:
           )
           exit /b %FAILED%
 
+      # ----------------------------------------------------------------------
+      # EPContext export + import perf test (PR #132 sidecar-in-tar path).
+      # Limited to full_model_seq128 to bound runtime: each model triggers
+      # one export pass (~60 s + N GB sidecar write) plus one import perf
+      # pass (-t 30). The _ctx.onnx + _MORPHIZEN.bin are deleted right after
+      # import to keep runner disk usage bounded.
+      # ----------------------------------------------------------------------
+      - name: Run EPContext export/import perf test (GPU)
+        if: always()
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = 'Continue'
+          $PSNativeCommandUseErrorActionPreference = $false
+          $env:THEROCK_DIST = "${{ runner.workspace }}\therock-dist"
+          $pkgBin = Join-Path $PWD "gpu-test-package\bin"
+          $env:PATH = "${{ runner.workspace }}\therock-dist\bin;$pkgBin;$env:PATH"
+          $env:LIB = (Join-Path $PWD "gpu-test-package\lib") + ";${{ runner.workspace }}\therock-dist\lib"
+
+          $modelDir = "${{ runner.workspace }}\model"
+          if (-not (Test-Path $modelDir)) {
+            Write-Host "[SKIP] MODEL_DIR not found: $modelDir"
+            exit 0
+          }
+          $models = @(Get-ChildItem $modelDir -Recurse -Filter "full_model_seq128.onnx" -ErrorAction SilentlyContinue)
+          if ($models.Count -eq 0) {
+            Write-Host "[SKIP] No full_model_seq128.onnx in $modelDir"
+            exit 0
+          }
+
+          New-Item -ItemType Directory -Force -Path "gpu-test-package\results-epctx-export" | Out-Null
+          New-Item -ItemType Directory -Force -Path "gpu-test-package\results-epctx-import" | Out-Null
+          $cacheDir = Join-Path $PWD "gpu-test-package\epctx-cache"
+          New-Item -ItemType Directory -Force -Path $cacheDir | Out-Null
+
+          $failed = 0
+          Push-Location gpu-test-package\bin
+          foreach ($m in $models) {
+            $exportFile = "..\results-epctx-export\$($m.Name).txt"
+            $importFile = "..\results-epctx-import\$($m.Name).txt"
+            $ctxPath = Join-Path $cacheDir "$($m.BaseName)_ctx.onnx"
+            Remove-Item "$ctxPath*" -ErrorAction SilentlyContinue
+
+            Write-Host "=== EPContext export: $($m.Name) ==="
+            & .\onnxruntime_perf_test.exe `
+              --plugin_ep_libs "MorphiZenExecutionProvider|onnxruntime_morphizen_ep.dll" `
+              --plugin_eps "MorphiZenExecutionProvider" `
+              --plugin_ep_options "config_file|..\morphizen_config.json" `
+              -C "session.disable_cpu_ep_fallback|1 ep.context_enable|1 ep.context_embed_mode|0 ep.context_file_path|$ctxPath" `
+              -t 5 -c 1 -s `
+              $m.FullName 2>&1 | Tee-Object -FilePath $exportFile
+            if ($LASTEXITCODE -ne 0) {
+              Add-Content $exportFile "[ERROR] EPContext export failed"
+              $failed = 1
+              Remove-Item "$ctxPath*" -ErrorAction SilentlyContinue
+              continue
+            }
+            if (-not (Test-Path $ctxPath)) {
+              Add-Content $exportFile "[ERROR] _ctx.onnx not produced"
+              $failed = 1
+              continue
+            }
+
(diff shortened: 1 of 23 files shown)


fhanuman (approved):
Thanks for making this change! I skimmed through the changes and got the overall idea; however, I couldn't look into them as deeply as I wanted due to the time constraints we have for the release. I will go ahead and approve to unblock the submission. I will go over the changes thoroughly again after the 0.1 release, and we will probably need to revisit some parts.

PR #136, opened 2026-04-20
feat: Unified Memory Manager (Phase 1 + 2)

tejus26:
## Summary

- Introduce a hardware-abstracted Unified Memory Manager (UMM) that replaces all direct `hipMalloc`/`hipFree`/`hipMemcpy` calls in the runtime with a layered allocator architecture
- HAL abstraction layer with HIP and CPU backends, pluggable via function-pointer table
- Size-class arena bump allocator (8 classes, lock-free atomic fetch-add) with BFC fallback for large allocations
- Static pool manager for compiler-planned weight tensors (single HAL alloc + O(1) offset lookup)
- Paged KV cache block allocator with lock-free free lists, slab allocation, and Copy-on-Write fork for beam search/speculative decoding
- Block table builder with contiguous materialization bridge for existing GEMM kernels
- Fully wired into the runtime hot path: all tensor I/O, pool allocation, and workspace management route through MM
- 62 GoogleTest unit tests, 20/21 E2E model tests passing (unchanged baseline)

## Architecture

```
Existing Runtime (inference_init/compute/cleanup)
        ↓
    mm_* C API Surface (mm_runtime_bridge.h)
        ↓
┌──────────┬──────────────┬────────────┐
│ Static   │ Dynamic      │ KV Cache   │
│ Pool     │ Arena + BFC  │ (Paged)    │
├──────────┴──────────────┴────────────┤
│       Handle Table (ID registry)     │
├──────────────────────────────────────┤
│   Hardware Abstraction Layer (HAL)   │
├──────────┬───────────────────────────┤
│ CPU/Host │ ROCm/HIP                 │
└──────────┴───────────────────────────┘
```

## Test plan

- [x] 62 GoogleTest unit tests pass (HAL, handle table, lifecycle, pool, arena, core, free list, KV block, KV fork, block table)
- [x] 20/21 E2E model tests pass (`test_qmoe_model` is pre-existing failure)
- [x] Full build succeeds with MSVC + Clang bitcode pipeline
- [x] MM debug logging verified via `HIPDNN_EP_DEBUG=1` env var (shows init, pool, tensor alloc/free, workspace, shutdown)
- [ ] CI pipeline validation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,5 +153,8 @@ if(BUILD_HIP_TOOLS)
 
     # E2E integration tests (compile MLIR → DLL → execute)
     add_subdirectory(test/e2e)
+
+    # Memory Manager unit tests (GoogleTest)
+    add_subdirectory(test/unit/mm)
   endif()
 endif()
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -3,6 +3,7 @@
 # ** Licensed under the MIT License.
 ##
 
+add_subdirectory(MemoryManager)
 add_subdirectory(Dialect)
 option(BUILD_HIPDNN_GRAPH "Build hipDNN fused-graph support" OFF)
 set(HIPDNN_AVAILABLE OFF)
--- a/lib/MemoryManager/CMakeLists.txt
+++ b/lib/MemoryManager/CMakeLists.txt
@@ -17,6 +17,7 @@ set(MM_SOURCES
     mm_free_list.cpp
     mm_kv.cpp
     mm_block_table.cpp
+    mm_kv_cache_state.cpp
 )
 
 set(MM_HEADERS
--- a/lib/MemoryManager/CMakeLists.txt
+++ b/lib/MemoryManager/CMakeLists.txt
@@ -1,0 +1,54 @@
+##
+# ** Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+# ** Licensed under the MIT License.
+##
+
+# Unified Memory Manager library
+
+set(MM_SOURCES
+    mm_hal.cpp
+    mm_hal_host.cpp
+    mm_config.cpp
+    mm_core.cpp
+    mm_pool.cpp
+    mm_arena.cpp
+    mm_handle_table.cpp
+    mm_runtime_bridge.cpp
+    mm_free_list.cpp
+    mm_kv.cpp
+    mm_block_table.cpp
+)
+
+set(MM_HEADERS
+    mm_types.h
+    mm_hal.h
+    mm_config.h
+    mm_core.h
+    mm_pool.h
+    mm_arena.h
+    mm_internal.h
+    mm_handle_table.h
+)
+
+add_library(HipDnnEpMemoryManager STATIC ${MM_SOURCES} ${MM_HEADERS})
+
+target_include_directories(HipDnnEpMemoryManager
+    PUBLIC  ${CMAKE_CURRENT_SOURCE_DIR}
+)
+
+# HIP backend (conditional on real runtime)
+if(NOT BUILD_MOCK_RUNTIME)
+    find_package(hip QUIET)
+    if(TARGET hip::host)
+        target_sources(HipDnnEpMemoryManager PRIVATE mm_hal_hip.cpp)
+        target_compile_definitions(HipDnnEpMemoryManager PUBLIC MM_HAS_HIP)
+        target_link_libraries(HipDnnEpMemoryManager PRIVATE hip::host)
+    else()
+        message(WARNING "Memory Manager: hip::host not found, building without HIP backend")
+    endif()
+endif()
+
+# C++17 required for std::shared_mutex, structured bindings
+target_compile_features(HipDnnEpMemoryManager PUBLIC cxx_std_17)
+
+message(STATUS "Memory Manager: enabled (mock=${BUILD_MOCK_RUNTIME})")
--- a/lib/MemoryManager/mm_arena.cpp
+++ b/lib/MemoryManager/mm_arena.cpp
@@ -177,16 +177,19 @@ void *ArenaAllocator::alloc(size_t size, size_t alignment) {
   if (cls >= num_classes_ || size >= kBfcThreshold)
     return bfc_.alloc(size, alignment);
 
-  /* Atomic bump allocation (lock-free) */
+  /* Atomic bump allocation (lock-free via CAS loop) */
   size_t aligned_size = (size + alignment - 1) & ~(alignment - 1);
   Arena &arena = arenas_[cls];
 
-  size_t old_offset =
-      arena.offset.fetch_add(aligned_size, std::memory_order_relaxed);
-  if (old_offset + aligned_size > arena.capacity) {
-    /* Arena exhausted — fall back to BFC */
-    arena.offset.fetch_sub(aligned_size, std::memory_order_relaxed);
-    return bfc_.alloc(size, alignment);
+  size_t old_offset = arena.offset.load(std::memory_order_relaxed);
+  while (true) {
+    if (old_offset + aligned_size > arena.capacity) {
+      /* Arena exhausted — fall back to BFC */
+      return bfc_.alloc(size, alignment);
+    }
+    if (arena.offset.compare_exchange_weak(
+            old_offset, old_offset + aligned_size, std::memory_order_relaxed))
+      break;
   }
 
   return static_cast<char *>(arena.base) + old_offset;
--- a/lib/MemoryManager/mm_arena.cpp
+++ b/lib/MemoryManager/mm_arena.cpp
@@ -10,99 +10,98 @@ namespace mm {
 BfcAllocator::~BfcAllocator() { shutdown(); }
 
 int BfcAllocator::init(const mm_hal_t *hal, int device_id, size_t capacity) {
-    if (!hal || capacity == 0)
-        return MM_ERROR_INVALID_ARG;
+  if (!hal || capacity == 0)
+    return MM_ERROR_INVALID_ARG;
 
-    hal_ = hal;
-    device_id_ = device_id;
-    capacity_ = capacity;
+  hal_ = hal;
+  device_id_ = device_id;
+  capacity_ = capacity;
 
-    base_ = hal_->raw_alloc(device_id_, capacity_, kDefaultAlignment);
-    if (!base_)
-        return MM_ERROR_OUT_OF_MEMORY;
+  base_ = hal_->raw_alloc(device_id_, capacity_, kDefaultAlignment);
+  if (!base_)
+    return MM_ERROR_OUT_OF_MEMORY;
 
-    /* Start with one large free block covering the entire slab */
-    blocks_.push_back({0, capacity_, false});
-    used_bytes_ = 0;
-    return MM_OK;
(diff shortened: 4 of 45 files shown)


wcy123 (changes requested):
## Review: Unified Memory Manager (PR #136)

The HAL abstraction layer and the unit test infrastructure are genuine improvements and worth landing. The KV cache subsystem and the "everything through MM" mandate have fundamental architectural problems that need to be addressed before merge.

---

### 1. The bitcode-merge model makes cross-DLL memory sharing impossible

This is the most critical issue. The runtime is not a shared library — it is LLVM bitcode embedded in the EP DLL, merged per-model at compile time, and fully inlined by O2. After optimization, each `model.dll` contains its own private copy of every runtime function as inline instructions. There is no shared runtime instance, no shared globals, and no shared singleton state across DLL boundaries. The architecture doc states this explicitly: *"No shared runtime state across models. Each compiled model.dll is independent."*

The MM's `MmCore` singleton, the global arena, the global KV block pool, and the global handle table are all DLL-local. They cannot be shared between model sessions. This makes the paged KV block allocator architecturally inert for the primary use case it claims to support.

---

### 2. KV cache is an application-level concern; the EP should not own it

KV cache lifecycle — which blocks to allocate, reuse, evict, fork for beam search — is the responsibility of the serving runtime above the EP (OGA, vLLM, TGI). The EP's contract is: receive tensors, execute a graph, return tensors.

The PR detects KV cache reuse by watching host pointer identity across `prepare_input` / `finalize_output` call pairs. This is reverse-engineering an application-level pattern from pointer aliasing, not an explicit API contract. It breaks when:

- OGA changes its internal buffer aliasing strategy (not a public API guarantee)
- The serving layer uses a prefill/decode split model (two separate ONNX graphs → two separate `model.dll` instances with separate `RuntimeState`; the prefill DLL writes KV, the decode DLL reads it — the MM has no mechanism to bridge this)
- Sequence lengths vary between calls (the persistent entry size was registered at first call)
- Multiple concurrent sessions share the same host address space

The correct fix for redundant H2D/D2H on `past_present_share_buffer` is an explicit `mark_persistent` / `unmark_persistent` API agreed with ORT/OGA, not implicit inference from pointer reuse.

---

### 3. The static pool layer duplicates compiler work that already exists

The compiler's `MemoryPoolingPass` already performs graph-coloring-based static buffer layout. Pool size and all buffer offsets are computed at compile time and emitted directly into the generated `inference_init` call:

```
hipdnn_ep_pool_init(state, pool_size, buffer_offsets, num_buffers)
```

After bitcode merge and O2 inlining, `hipdnn_ep_get_buffer_from_pool` becomes a direct `load + GEP` with zero overhead. Routing this through `mm_bridge_pool_init` / `mm_bridge_pool_get_buffer` adds runtime indirection on top of a problem the compiler already solved statically.

---

### 4. The dual-path fallback makes production behavior non-deterministic

The MM initializes with a silent fallback to legacy paths on failure:

```cpp
if (mm_err == MM_OK) {
    state->mm_initialized = true;
} else {
    // Non-fatal: fall back to legacy allocation paths
}
```

Production code now runs one of two different memory management implementations depending on runtime conditions. This makes profiling, debugging, and reasoning about memory behavior significantly harder. If the MM is correct, the legacy path should be removed. If it is not ready, it should not be wired into the hot path.

---

### 5. CI results show active correctness regressions

The L2 accuracy numbers from the automated run are not acceptable:

| Model | L2 Error | Threshold |
|---|---|---|
| GroupQueryAttention_seq256 | 25.43 | 0.01 |
| MatMulNBits_o_seq128 | 198,298 | 0.01 |
| GroupQueryAttention_seq128 | **failed** | — |

These are not pre-existing failures. The GQA and MatMulNBits models are directly affected by the KV cache routing changes in `hipdnn_ep_runtime_tensor.cpp`. The PR should not merge with these regressions unresolved.

---

### 6. MM debug logging does not survive the optimization pipeline

The `HIPDNN_EP_DEBUG=1` logging in `mm_bridge_*` functions is cited in the test plan as verified. However, after bitcode merge and O2, all `mm_bridge_*` calls are inlined away. The logging that was verified in unit tests does not exist in the optimized `model.dll`. The test plan is testing a code path that is not the production code path.

---

### Recommendation

**Accept:** `mm_hal.h` / `mm_hal_hip.cpp` / `mm_hal_host.cpp` — the HAL abstraction is clean, testable, and useful. The unit test infrastructure is also a net positive.

**Reject (defer to a separate PR with design review):**
- `mm_kv.cpp`, `mm_kv_cache_state.cpp`, `mm_block_table.cpp` and all `mm_bridge_kv_*` functions — KV cache management does not belong in the EP
- The wiring of `mm_bridge_tensor_alloc` / `mm_bridge_tensor_release` into `pool_alloc` / `pool_release` — redundant with MemoryPoolingPass; introduces dual-path risk
- The `kv_cache_state` field in `RuntimeState` and all `prepare_input` / `finalize_output` / `tensor_free_input` heuristic detection logic

Before resubmitting the KV cache work, a design document should answer:
1. What is the explicit API contract with ORT/OGA for marking tensors persistent (not host-pointer heuristics)?
2. How does KV state cross the prefill/decode model.dll boundary?
3. If each DLL is independent (by design), what is the actual scope of what the MM can manage?

wcy123 (dismissed):
## Review: Unified Memory Manager (PR #136)

The HAL abstraction layer and the unit test infrastructure are genuine improvements and worth landing. The KV cache subsystem and the "everything through MM" mandate have fundamental architectural problems that need to be addressed before merge.

---

### 1. The bitcode-merge model makes cross-DLL memory sharing impossible; the interface makes intra-session beam management unactionable

The runtime is not a shared library — it is LLVM bitcode embedded in the EP DLL, merged per-model at compile time, and fully inlined by O2. After optimization, each `model.dll` contains its own private copy of every runtime function as inline instructions. There is no shared runtime instance, no shared globals, and no shared singleton state across DLL boundaries. The architecture doc states this explicitly: *"No shared runtime state across models. Each compiled model.dll is independent."*

The MM's `MmCore` singleton, the global arena, the global KV block pool, and the global handle table are all DLL-local. The correct implication of this is narrower than "paged KV is useless everywhere": the MM cannot be designed as a cross-model or cross-DLL shared service layer. A paged KV block pool with CoW fork is not inherently wrong for an intra-session use case such as beam search — that is a real, per-session concern.

However, intra-session beam search is also not actionable through this PR's mechanism. ORT/OGA orchestrates beam search *above* the EP: it calls `inference_compute(state, span_t* inputs, span_t* outputs)` once per beam per step, passing host-visible KV tensors between calls. The EP never receives beam topology. The CoW fork in `mm_kv_fork_block` can only be useful if the EP knows which hypothesis branches survive each step — information that is not present in the `span_t` interface. For the paged block allocator to deliver its stated value, the EP would need to orchestrate beam search internally, which requires a fundamentally different interface design and is not proposed here. The PR should either scope itself to the intra-session case and define the interface changes needed to support it, or drop the paged KV subsystem.

---

### 2. KV tensor mechanics belong in the EP; the mechanism in this PR does not

There is a meaningful distinction between two layers of KV ownership:

- **Policy** — which context to keep, evict, or schedule across users and requests — belongs in the serving layer (OGA, vLLM, TGI).
- **Mechanics** — given ORT I/O pointers for `past_*` / `present_*` tensors, managing device residency, avoiding redundant copies, and arranging layout for compiled attention kernels — is legitimately the EP's responsibility.

The goal of keeping KV tensors GPU-resident across decode steps and skipping redundant H2D/D2H is correct for an EP to pursue. The mechanism in this PR is not.

The PR detects KV reuse by watching host pointer identity across `prepare_input` / `finalize_output` call pairs — reverse-engineering an application-level pattern from pointer aliasing behavior that is not a documented ORT/OGA contract. This is fragile in multiple ways:

- OGA's buffer aliasing strategy is an implementation detail, not a public API guarantee
- The persistent entry's size is registered at first call; variable sequence lengths will silently use a stale size
- The prefill/decode split (two separate ONNX graphs → two separate `model.dll` instances with separate `RuntimeState`) means the prefill DLL writes KV and the decode DLL reads it — there is no cross-DLL handoff mechanism
- Multiple concurrent sessions sharing the same host address space can collide on pointer identity

The correct approach is an explicit API agreed with ORT/OGA — a tensor attribute or a dedicated EP call — that marks specific buffers as persistent without requiring the EP to infer intent from runtime pointer behavior.

---

### 3. The static pool layer duplicates compiler work that already exists

The compiler's `MemoryPoolingPass` already performs graph-coloring-based static buffer layout. Pool size and all buffer offsets are computed at compile time and emitted directly into the generated `inference_init` call:

```
hipdnn_ep_pool_init(state, pool_size, buffer_offsets, num_buffers)
```

After bitcode merge and O2 inlining, `hipdnn_ep_get_buffer_from_pool` becomes a direct `load + GEP` with zero overhead. Routing this through `mm_bridge_pool_init` / `mm_bridge_pool_get_buffer` adds runtime indirection on top of a problem the compiler already solved statically.

---

### 4. The dual-path fallback makes production behavior non-deterministic

The MM initializes with a silent fallback to legacy paths on failure:

```cpp
if (mm_err == MM_OK) {
    state->mm_initialized = true;
} else {
    // Non-fatal: fall back to legacy allocation paths
}
```

Production code now runs one of two different memory management implementations depending on runtime conditions. This makes profiling, debugging, and reasoning about memory behavior significantly harder. If the MM is correct, the legacy path should be removed. If it is not ready, it should not be wired into the hot path.

---

### 5. CI results show active correctness regressions

The L2 accuracy numbers from the automated run are not acceptable:

| Model | L2 Error | Threshold |
|---|---|---|
| GroupQueryAttention_seq256 | 25.43 | 0.01 |
| MatMulNBits_o_seq128 | 198,298 | 0.01 |
| GroupQueryAttention_seq128 | **failed** | — |

These are not pre-existing failures. The GQA and MatMulNBits models are directly affected by the KV cache routing changes in `hipdnn_ep_runtime_tensor.cpp`. The PR should not merge with these regressions unresolved.

---

### 6. MM debug logging does not survive the optimization pipeline

The `HIPDNN_EP_DEBUG=1` logging in `mm_bridge_*` functions is cited in the test plan as verified. However, after bitcode merge and O2, all `mm_bridge_*` calls are inlined away. The logging that was verified in unit tests does not exist in the optimized `model.dll`. The test plan is testing a code path that is not the production code path.

---

### Recommendation

**Accept:** `mm_hal.h` / `mm_hal_hip.cpp` / `mm_hal_host.cpp` — the HAL abstraction is clean, testable, and useful. The unit test infrastructure is also a net positive.

**Reject (defer to a separate PR with design review):**
- `mm_kv.cpp`, `mm_kv_cache_state.cpp`, `mm_block_table.cpp` and all `mm_bridge_kv_*` functions — KV cache management does not belong in the EP
- The wiring of `mm_bridge_tensor_alloc` / `mm_bridge_tensor_release` into `pool_alloc` / `pool_release` — redundant with MemoryPoolingPass; introduces dual-path risk
- The `kv_cache_state` field in `RuntimeState` and all `prepare_input` / `finalize_output` / `tensor_free_input` heuristic detection logic

Before resubmitting the KV cache work, a design document should answer:
1. What is the explicit API contract with ORT/OGA for marking tensors persistent (not host-pointer heuristics)?
2. How does KV state cross the prefill/decode model.dll boundary?
3. If each DLL is independent (by design), what is the actual scope of what the MM can manage?

wcy123 (commented):
hide the AI autogenerated review comment.

PR #140, opened 2026-04-23
feat: add support Range Op

mingyueliuh:
## Summary

This PR adds end-to-end ONNX `Range` support (`onnx.Range -> hip.range -> wrap_range`) and finalizes runtime error propagation for `delta == 0` without adding `hipStreamSynchronize` on the hot path.

## Main changes

- ONNX -> HIP lowering
  - Added `onnx.Range` lowering to `hip.range` in DPS style.
  - Kept dynamic length computation and compile-time length handling.

- HIP -> LLVM + runtime integration
  - Added `hip.range` lowering to `llvm.call @wrap_range`.
  - Integrated custom HIP range kernel/runtime wrapper path.

- `delta == 0` handling (ORT parity + no hot-path sync)
  - Added compile-time rejection when `delta` is a constant zero.
  - Added device-side detection in range kernel for dynamic `delta == 0`.
  - Added runtime device error flag in `RuntimeState` and propagated status through interface-level checks.
  - Removed Range-specific main_graph customization and centralized error handling in generated interface flow.

- Test updates
  - Updated/added Range conversion and e2e checks (including dynamic declarations/order-safe checks where needed).

## Validation
- LIT :
  - Result: **101 passed**

- Range L2 model validation:
  - Model set: `D:\Develop\m\models\Qwen3.5-35B-A3B_int4_rtn_128gs_cuda\single_op\Range`
  - Result: **28 / 28 passed**, combined L2 = **0** for all cases.

## Notes

- The branch history was rebased onto `main` and force-pushed with lease.
- `3rd-party/morphizen` submodule pointer update is intentionally **not** part of this PR commit history.

### DIFF
--- a/lib/Runtime/real/range.cpp
+++ b/lib/Runtime/real/range.cpp
@@ -21,6 +21,7 @@ int wrap_range(RuntimeState *state, void *start, void *limit, void *delta,
       "[REAL] wrap_range: output_num_elements=%lld, hip_dtype=%lld\n",
       (long long)output_num_elements, (long long)hip_dtype);
 
+  void *deviceErrorFlag = hipdnn_ep_state_get_error_flag_device_ptr(state);
   return hip_range(stream, start, limit, delta, output, output_num_elements,
-                   hip_dtype);
+                   hip_dtype, deviceErrorFlag);
 }
--- a/lib/Runtime/real/range.cpp
+++ b/lib/Runtime/real/range.cpp
@@ -15,10 +15,21 @@ static int readDeviceScalar(void *stream, const void *src, T &dst) {
   auto hip_stream = static_cast<hipStream_t>(stream);
   hipError_t err =
       hipMemcpyAsync(&dst, src, sizeof(T), hipMemcpyDeviceToHost, hip_stream);
-  if (err != hipSuccess)
+  if (err != hipSuccess) {
+    fprintf(stderr, "[REAL] readDeviceScalar: hipMemcpyAsync failed: %s\n",
+            hipGetErrorString(err));
     return -1;
+  }
+  // Synchronization is required here: we branch on host value `delta` to
+  // preserve ORT's runtime delta==0 error behavior before launching hip_range.
   err = hipStreamSynchronize(hip_stream);
-  return err == hipSuccess ? 0 : -1;
+  if (err != hipSuccess) {
+    fprintf(stderr,
+            "[REAL] readDeviceScalar: hipStreamSynchronize failed: %s\n",
+            hipGetErrorString(err));
+    return -1;
+  }
+  return 0;
 }
 
 int wrap_range(RuntimeState *state, void *start, void *limit, void *delta,
@@ -50,11 +61,15 @@ int wrap_range(RuntimeState *state, void *start, void *limit, void *delta,
   }
   case HIP_DTYPE_FLOAT32: {
     float d = 0.0f;
+    // Keep exact zero check for ONNX/ORT parity.
+    // ORT CPU Range rejects delta only when it is exactly zero.
     deltaCheck = readDeviceScalar(stream, delta, d) || (d == 0.0f);
     break;
   }
   case HIP_DTYPE_FLOAT64: {
     double d = 0.0;
+    // Keep exact zero check for ONNX/ORT parity.
+    // ORT CPU Range rejects delta only when it is exactly zero.
     deltaCheck = readDeviceScalar(stream, delta, d) || (d == 0.0);
     break;
   }
@@ -74,5 +89,5 @@ int wrap_range(RuntimeState *state, void *start, void *limit, void *delta,
       (long long)output_num_elements, (long long)hip_dtype);
 
   return hip_range(stream, start, limit, delta, output, output_num_elements,
-                   static_cast<int>(hip_dtype));
+                   hip_dtype);
 }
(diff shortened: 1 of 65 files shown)


mingyueliuh:
### Accuracy tests  
test models in sharedlab : \\WINAMD-RSCPOIV6\work\Qwen3.5-35B-A3B_int4_rtn_128gs_cuda_space\text\single_op\Range\gen  
```
Model                                    Status L2 Msg
-----                                    ------ -- ---
Range_gen_f32_empty_same.onnx            PASS   0     
Range_gen_f32_forward_frac.onnx          PASS   0     
Range_gen_f32_neg_desc.onnx              PASS   0     
Range_gen_f32_tiny_step.onnx             PASS   0     
Range_gen_f64_basic.onnx                 PASS   0     
Range_gen_f64_empty.onnx                 PASS   0     
Range_gen_f64_neg.onnx                   PASS   0     
Range_gen_i16_empty_neg_0_5_m1.onnx      PASS   0     
Range_gen_i16_empty_pos_5_2_1.onnx       PASS   0     
Range_gen_i16_forward_0_10_1.onnx        PASS   0     
Range_gen_i16_large_step_0_1000_100.onnx PASS   0     
Range_gen_i16_neg_desc_10_0_m2.onnx      PASS   0     
Range_gen_i16_single_3_4_1.onnx          PASS   0     
Range_gen_i16_step2_2_10_2.onnx          PASS   0     
Range_gen_i32_empty_neg_0_5_m1.onnx      PASS   0     
Range_gen_i32_empty_pos_5_2_1.onnx       PASS   0     
Range_gen_i32_forward_0_10_1.onnx        PASS   0     
Range_gen_i32_large_step_0_1000_100.onnx PASS   0     
Range_gen_i32_neg_desc_10_0_m2.onnx      PASS   0     
Range_gen_i32_single_3_4_1.onnx          PASS   0     
Range_gen_i32_step2_2_10_2.onnx          PASS   0     
Range_gen_i64_empty_neg_0_5_m1.onnx      PASS   0     
Range_gen_i64_empty_pos_5_2_1.onnx       PASS   0     
Range_gen_i64_forward_0_10_1.onnx        PASS   0     
Range_gen_i64_large_step_0_1000_100.onnx PASS   0     
Range_gen_i64_neg_desc_10_0_m2.onnx      PASS   0     
Range_gen_i64_single_3_4_1.onnx          PASS   0     
Range_gen_i64_step2_2_10_2.onnx          PASS   0     


Total=28 Pass=28 Fail=0
````

File: lib/Runtime/real/range.cpp [hunk 1]
```diff
@@ -0,0 +1,93 @@
...
+  hipError_t err =
+      hipMemcpyAsync(&dst, src, sizeof(T), hipMemcpyDeviceToHost, hip_stream);
+  if (err != hipSuccess) {
+    fprintf(stderr, "[REAL] readDeviceScalar: hipMemcpyAsync failed: %s\n",
+            hipGetErrorString(err));
+    return -1;
+  }
+  // Synchronization is required here: we branch on host value `delta` to
+  // preserve ORT's runtime delta==0 error behavior before launching hip_range.
+  err = hipStreamSynchronize(hip_stream);
```
MaheshRavishankar: This is going to be a performance bottleneck. Do we really need to check for this behavior? or can we check this at compile time?

MaheshRavishankar (commented):
I am fine with this cause of how EP is today, but really we should be using hipDNN here cause there is really no reason this needs to be handled as a real operation.

cc @edelaye @fhanuman

lib/Runtime/real/range.cpp
mingyueliuh: Thanks for calling this out  — agreed that the current host-side check can become a bottleneck.
**Why check delta == 0 ?**
    To stay consistent with ORT semantics, an error is raised at runtime when `delta == 0`.

**Why not check in compile time ?** 
   Compile-time check alone is not sufficient for dynamic delta.
   `delta` may be either a constant or the output of an upstream operator. When it comes from an upstream operator, its value cannot be determined at compile time, so it cannot be handled solely through compile-time checks.
   
In the current implementation, delta is often a runtime device scalar (not guaranteed to be a compile-time constant), so if we branch on host before launching hip_range, we must do D2H + hipStreamSynchronize to read a valid value on the same stream.
The sync is currently required for the host-branching design, but we agree it is not ideal for performance.

**Improvement plan**
- [ ] Compile-time fast path
        Reject delta == 0 during lowering when delta is constant (no runtime cost).
- [ ] Remove per-op host sync on hot path
       Move zero-check to device-side kernel path for dynamic delta, so normal execution does not do host D2H/sync per Range.
- [ ] Error propagation to upper layer
     Add a device error flag and plumb it through main_graph/inference_compute return path, so we still report non-zero status to ORT instead of only logging.
 
 @MaheshRavishankar  Do you think this approach would work?

mingyueliuh: @MaheshRavishankar   we have now replaced the hot-path host sync design with the new approach. Please review again. 
Validation status on this branch:
- Range single-op L2 suite: 28/28 pass, all combined L2 = 0.

amd-genmingz:
rebase to main branch , Range single-op L2 suite: 28/28 pass, all combined L2 = 0.

amd-mingw (approved):
approved and merged it to unblock Qwen3.5 model test, if have more comments, will with a follow-up PR to address them.

PR #143, opened 2026-04-23
fix: treat RotaryEmbedding attributes as optional per ONNX spec

AMDmoore:
## Summary

- Treat `interleaved`, `num_heads`, and `rotary_embedding_dim` as optional attributes (default 0) per the [com.microsoft.RotaryEmbedding spec](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#commicrosoftrotaryembedding)
- Previously, models omitting these attributes caused `ConvertOnnxToHipPass` pattern match failure, falling back to CPU execution
- When absent, infer values from tensor shapes: `rotary_dim = cos_cache.shape[-1] * 2`, `num_heads = input.shape[-1] / rotary_dim`

## Test plan

- [x] Verified with `RotaryEmbedding_q_rotary_seq128.onnx` model via `onnxruntime_perf_test` - compilation succeeds and runs on GPU
- [x] Pre-commit checks passed (trailing-whitespace, lintrunner, licenseheaders, etc.)


amd-mingw (approved):
fix a attr bug.

PR #144, opened 2026-04-23
fix(gqa): apply causal mask during decode when sliding-window is active

zpye:
## Summary

The causal-mask kernel launch in `gqa_forward_hipblaslt` was gated on
`sq > 1`, which only fires during prefill. During decode (`sq == 1`)
with a configured `local_window_size`, keys older than
`total_seq - local_window_size` must still be masked out; without the
mask those tokens leak into the softmax and degrade generation quality.

Extend the gate to `sq > 1 || local_window_size > 0` so sliding-window
decode runs the same causal/window mask path as prefill.
`hip_gqa_causal_mask_f32` already honors `past_len` and
`local_window_size` internally, so no kernel change is required.

## Change

- `lib/Runtime/real/gqa.cpp` — single-line gate extension (+1/-1).

## Test plan

- [x] Short-prompt `p128` generation on GPT-OSS-20B no longer shows the
      "leaked past-window tokens" pattern in decode output when sliding
      window is active.
- [ ] CI / perf jobs.


amd-mingw (approved):
fix a bug.

PR #147, opened 2026-04-24
feat: add Where operator support

amd-genmingz:
## Motivation
Needed to run the Qwen3.5 model end-to-end on the HIP EP. `Where` is one
of the ONNX ops used by that model and is currently missing, so we add
it here to unblock the bring-up.

## Details
- Define `hip.where` in the HIP dialect using destination-passing style:
  `ins(condition, x, y)`, `outs(output)`.
- Lower `onnx.Where` -> `hip.where` (`OnnxToHip/WhereConversion.cpp`),
  resolving dynamic output dims across all three operands to honor
  ONNX multidirectional (NumPy-style) broadcasting.
- Lower `hip.where` -> runtime call `wrap_where` (`HipToLLVM/WhereLowering.cpp`),
  passing a per-operand `(shape_ptr, rank)` pair plus `data_type` so no fixed
  layout is assumed and any rank up to `HIP_WHERE_MAX_RANK = 8` is supported.
- Add runtime wrapper `wrap_where` (`lib/Runtime/real/where.cpp`) that
  dispatches to the new HIP kernel based on dtype.
- Add N-D broadcast HIP kernel `hip_elementwise_where`
  (`3rd-party/custom_kernels/hip/elementwise_where_kernel.hip`):
  left-pads operand shapes to the output rank, precomputes broadcast
  strides (stride = 0 on broadcast axes), and launches a templated
  kernel over the linear output.
- Supported X/Y/output dtypes: FP32, FP16, BF16, INT32, INT64;
  `condition` is `INT8` in onnxruntime.
- Register new sources in CMake and expose the C APIs in the
  custom-kernels and runtime headers.

## Test Plan
- [x] Build passes on the full EP (custom kernels + runtime + conversions).
- [x] All lit tests pass, including the new
      `test/lit/Conversion/onnx-to-hip/test_where.mlir` and
      `test/lit/Conversion/hip-to-llvm/test_where.mlir`, with no regression
      on existing tests.
- [x] Single-op accuracy check for `Where` against a reference
      implementation


amd-genmingz:
image

amd-mingw (approved):
approved and merged it to support Qwen3.5-35B/9B models, in the future, we can research whether it can be done on compile time.

PR #151, opened 2026-04-27
feat: add slice op

amd-genmingz:
## Motivation
ONNX `Slice` is a commonly used op for selecting a sub-tensor along
multiple axes. Add end-to-end support for it in the HIP execution provider so
models that depend on `Slice` can run on the HIP backend. Slice ops are used 
extensively in the qwen3.5B model.
## Details
- **Dialect**: New `hip.slice` op in `HipOps.td` using `AttrSizedOperandSegments`
  to model the optional `axes` / `steps` operands.
- **OnnxToHip**: New `SliceConversion.cpp` pattern that lowers `onnx.Slice`
  to `hip.slice`, normalizing both absent operands and `onnx.NoValue` placeholders.
- **HipToLLVM**: New `SliceLowering.cpp` that emits an `llvm.call @wrap_slice`,
  stack-allocating i64 shape arrays from the memref descriptors and passing
  `nullptr` for missing optional operands.
- **Runtime**: Added `wrap_slice` in `hipdnn_ep_runtime.h` with both the real
  implementation (`real/slice.cpp`) and a mock stub (`mock/mock_gpu.cpp`).
- **Kernel**: Added `slice_kernel.hip` — one thread per output element,
  reconstructs the per-axis output coordinate from the flat index, applies
  ONNX start/step/axes semantics (including negative indices and reverse
  stepping), and writes the corresponding input element.

## Test Plan
- [ ] all `lit` test passed
- [ ] The single op model cut from qwen3.5 ran successfully.
- [ ] The accuracy meets the requirements.

### DIFF
--- a/3rd-party/custom_kernels/CMakeLists.txt
+++ b/3rd-party/custom_kernels/CMakeLists.txt
@@ -39,6 +39,7 @@ hip_add_library(hip_custom_kernels STATIC
     hip/reduce_sum_kernel.hip
     hip/matmul_nbits_kernel.hip
     hip/qmoe_kernel.hip
+    hip/slice_kernel.hip
     INCLUDE_DIRECTORIES
         ${CMAKE_CURRENT_SOURCE_DIR}/include
 )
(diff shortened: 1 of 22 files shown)


amd-genmingz:
Slice has been added to main branch https://github.com/ROCm/onnx-hipdnn-ep/pull/216

PR #161, opened 2026-04-28
feat: enable GPU memory pipeline for OGA m=16384 KV cache

zz002:
## Summary

Wires the EP, the morphizen factory, and the OGA fork together so OGA's KV cache lives in **GPU-mapped pinned memory** instead of host RAM, fixing the 7B / 14B int4 `max_length=16384` TPS regression where every decode step was paying ~4 GB of `hipMemcpyAsync H2D + D2H` for a buffer that didn't change.

- **Mistral-7B int4 m=16384**: `3.94 → 15.92 TPS` (**+304%**)
- **Phi4-14B  int4 m=16384**: `2.12 → 13.18 TPS` (**+522%**)
- `[PERF]` confirms H2D / D2H are ~0 ms / 0 MB on aliased decode steps; the remaining gap is ORT plugin-EP routing + OGA framework overhead (out of scope here).

## Root cause

OGA in `m=16384` mode pre-allocates the KV cache for `max_length` tokens (2 GB on Mistral-7B int4, 5.7 GB on Phi4-14B int4) on its **primary device**. Because OGA didn't know about MorphiZenEP, that primary device defaulted to **CPU**, so every decode token did:

```
prepare_input  : hipMemcpyAsync H2D 2 GB  (past_key/value)
compute        : ~50 ms
finalize_output: hipMemcpyAsync D2H 2 GB  (present_key/value)
```

H2D + D2H accounted for ~82 % of inference time — root-caused via the per-call `[PERF]` telemetry rebuilt in this PR.

## Solution (three-layer fix)

| Layer | Change | Where |
|---|---|---|
| **OGA fork** | Add `DeviceType::MorphiZenEP` so OGA actually allocates the KV cache through our EP allocator | [AMDmoore/onnxruntime-genai#8](https://github.com/AMDmoore/onnxruntime-genai/pull/8) (`OGA_REF` bumped to `c4fc287`) |
| **MorphiZen EP factory** | `hipHostMalloc(Mapped\|Coherent)` `OrtAllocator` + `HipDataTransferImpl`, register the EP as an AMD GPU device by default | [ROCm/MorphiZen#208](https://github.com/ROCm/MorphiZen/pull/208) (submodule bump to `e6ca655`) |
| **EP runtime** *(this PR)* | tensor_t / TensorBuffer ABI gains `memory_type` + `is_aliased`; `prepare_*` aliases caller's GPU buffer; `finalize_output` / `free_input` skip D2H / pool_release when aliased | `lib/Runtime/hipdnn_ep_runtime_tensor.cpp`, `lib/Runtime/hipdnn_ep_runtime.h`, `backend-mlir-compiler/.../custom_op_mlir.hpp`, `backend-mlir-compiler/.../MlirCustomOp.cpp` |

Plus:

- **PERF telemetry fix** — `HIPDNN_EP_PERF` no longer inherits from `HIPDNN_EP_DEBUG` (PERF forces `hipStreamSynchronize` on every inference, which serialises the pipeline and skews measurements). New `RUNTIME_PERF_LOG` macro for the per-inference `[PERF]` line. Dump moved from the dead `hipdnn_ep_stream_sync` hook into `prepare_input(0)`, where it actually fires under the MLIR-emitted `main_graph` (which never calls `stream_sync`). `finalize_output` always re-records `d2h_end` so the window is well-defined when every output is aliased.
- **Build wiring** — `cmake/deps.cmake` forces `morphizen_ENABLE_HIP_GPU_ALLOCATOR=ON` in the parent build (`find_package(hip)` is already a runtime dependency); standalone MorphiZen builds keep the option `OFF`.
- **Test surface** — `hip-test-dll.cpp` explicitly stamps inputs / outputs with `TENSOR_MEMORY_CPU` so it keeps using the legacy host H2D / D2H path unchanged. CPU / FPGA / NPU `memory_type`s also fall through to the legacy path, so `hip-onnx-runner` and any other host-input caller are unaffected.

## Validation

Strix Halo, gfx1151, `m=16384`, `prompt_length=127`, `model_benchmark.exe`:

| Model | wall TPS before | wall TPS after | gain |
|---|---:|---:|---:|
| Mistral-7B int4 | 3.94 | **15.92** | +304 % |
| Phi4-14B  int4  | 2.12 | **13.18** | +522 % |

Sample `[PERF]` line on Mistral-7B int4 m=16384 after the fix:

```
[PERF] #42: H2D 0t/0.0MB/0.00ms | Compute 47.31ms | D2H 0t/0.0MB/0.00ms | Total 47.31ms
```

Compare to the same model before the fix (manual `KV_DBG` instrumentation):

```
H2D ~2048 MB / ~120 ms | Compute ~50 ms | D2H ~2048 MB / ~120 ms | Total ~290 ms
```

i.e. the per-decode 4 GB round-trip is now zero, and the remaining wall-clock cost matches the per-op GEMM theoretical limit.

## Dependencies (do not merge until both land)

- ROCm/MorphiZen#208 — submodule bump target (`e6ca655`)
- AMDmoore/onnxruntime-genai#8 — `OGA_REF` target (`c4fc287`)

CI on this PR will fail until both merge to their respective `main` branches and we re-point the submodule / `OGA_REF` to the post-merge commits. Marking as draft for now.

## Test plan

- [ ] Wait for ROCm/MorphiZen#208 to merge, then re-point submodule `3rd-party/morphizen` to the merge commit on `main`.
- [ ] Wait for AMDmoore/onnxruntime-genai#8 to merge, then re-point `OGA_REF` in `.github/workflows/windows-build.yml` to the merge commit on the OGA fork.
- [ ] CI green on Windows build (existing matrix).
- [ ] Re-run `model_benchmark.exe` Mistral-7B int4 / Phi4-14B int4 m=16384 on Strix Halo and confirm TPS within ~5 % of the numbers in this PR.
- [ ] `hip-test-dll` smoke run (host-input legacy path) on a small model — confirms the alias fast path doesn't regress non-OGA callers.

### DIFF
--- a/backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
+++ b/backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
@@ -42,9 +42,13 @@ enum {
 // Memory ownership: caller-owned. `memory_type` selects the data pointer's
 // placement (see the enum above) and tells the DLL whether to copy or alias.
 //
-// IMPORTANT: this layout is duplicated as `tensor_t` in
-// `lib/Runtime/hipdnn_ep_runtime.h` (extern-C version consumed by the
-// MLIR-compiled model.dll). Keep the two definitions byte-for-byte identical.
+// tensor_t is the wire-protocol ABI between three components that are
+// intentionally kept decoupled (compiler-emitted model.dll, EP runtime
+// DLL, hip-test-dll harness), so we re-declare it here instead of
+// sharing a header. The static_assert block below catches any layout
+// drift at compile time. Sibling copies live at:
+//   * `lib/Runtime/hipdnn_ep_runtime.h`            (extern-C, EP runtime)
+//   * `tools/hip-test-dll/hip-test-dll.cpp`        (test driver, local)
 struct tensor_t {
   void *data;     // Pointer to contiguous tensor data (caller-owned)
   int64_t *shape; // Pointer to shape array of length `rank` (caller-owned)
@@ -53,15 +57,12 @@ struct tensor_t {
   int memory_type;     // One of TENSOR_MEMORY_CPU / _GPU / _FPGA / _NPU
 };
 
-// Compile-time guard against silent layout drift between the three places
-// that duplicate this struct (this header, lib/Runtime/hipdnn_ep_runtime.h,
-// tools/hip-test-dll/hip-test-dll.cpp). The same three asserts live in all
-// three files; if you reorder/add/remove a field in one place you have to
-// touch all three or one of these will fire at build time.
-//
-// We assert per-field offsets, not raw sizeof(tensor_t), because trailing
-// padding after `memory_type` is compiler-defined and not part of what the
-// MLIR-emitted model.dll actually reads.
+// Compile-time guard for the wire-protocol ABI described above. The same
+// three asserts live in each of the three sibling headers; if you reorder
+// / add / remove a field in one copy and forget to mirror it in the others,
+// at least one of them fails to build. Per-field offsets (not raw sizeof)
+// because trailing padding after `memory_type` is compiler-defined and not
+// part of what model.dll actually reads.
 static_assert(offsetof(tensor_t, data) == 0,
               "tensor_t.data must remain the first field");
 static_assert(offsetof(tensor_t, shape) == sizeof(void *),
--- a/backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
+++ b/backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
@@ -13,13 +13,44 @@ namespace mlir_compilation {
 // Data structures — C-compatible, stable ABI
 // ============================================================================
 
+// Memory placement of a tensor's `data` pointer. Carried as an int in
+// tensor_t so the enum is forward-compatible without breaking the on-the-wire
+// ABI between marshalling (onnxruntime_morphizen_ep.dll) and the MLIR-compiled
+// model.dll.
+//
+// IMPORTANT: values are 1:1 with ORT's OrtMemoryInfoDeviceType
+// (onnxruntime_c_api.h). MlirCustomOp can therefore write the ORT value
+// straight into tensor_t.memory_type with no remapping. Today the runtime
+// only special-cases TENSOR_MEMORY_GPU (alias path); CPU / FPGA / NPU all
+// fall through to the legacy host H2D / D2H path, preserving existing
+// behaviour for hip-test-dll, hip-onnx-runner, and any other host-input
+// caller.
+//
+// Must match the matching enum in `lib/Runtime/hipdnn_ep_runtime.h`.
+enum {
+  TENSOR_MEMORY_CPU = 0, // == OrtMemoryInfoDeviceType_CPU
+  TENSOR_MEMORY_GPU = 1, // == OrtMemoryInfoDeviceType_GPU  (alias path; only
+                         // mode optimized today)
+  TENSOR_MEMORY_FPGA =
+      2, // == OrtMemoryInfoDeviceType_FPGA (treated as host today)
+  TENSOR_MEMORY_NPU =
+      3, // == OrtMemoryInfoDeviceType_NPU  (treated as host today)
+};
+
 // Describes a single tensor passed across the DLL boundary.
-// All memory is caller-owned CPU memory; the DLL copies H2D/D2H internally.
+//
+// Memory ownership: caller-owned. `memory_type` selects the data pointer's
+// placement (see the enum above) and tells the DLL whether to copy or alias.
+//
+// IMPORTANT: this layout is duplicated as `tensor_t` in
+// `lib/Runtime/hipdnn_ep_runtime.h` (extern-C version consumed by the
+// MLIR-compiled model.dll). Keep the two definitions byte-for-byte identical.
 struct tensor_t {
-  void *data; // Pointer to contiguous tensor data (CPU memory, caller-owned)
+  void *data;     // Pointer to contiguous tensor data (caller-owned)
   int64_t *shape; // Pointer to shape array of length `rank` (caller-owned)
   size_t rank;    // Number of dimensions (must match the compiled model's rank)
   size_t element_size; // Bytes per element (e.g. 4=float32, 2=float16, 8=int64)
+  int memory_type;     // One of TENSOR_MEMORY_CPU / _GPU / _FPGA / _NPU
 };
 
 // A contiguous array of tensor_t descriptors (inputs or outputs).
--- a/backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
+++ b/backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
@@ -53,6 +53,24 @@ struct tensor_t {
   int memory_type;     // One of TENSOR_MEMORY_CPU / _GPU / _FPGA / _NPU
 };
 
+// Compile-time guard against silent layout drift between the three places
+// that duplicate this struct (this header, lib/Runtime/hipdnn_ep_runtime.h,
+// tools/hip-test-dll/hip-test-dll.cpp). The same three asserts live in all
+// three files; if you reorder/add/remove a field in one place you have to
+// touch all three or one of these will fire at build time.
+//
+// We assert per-field offsets, not raw sizeof(tensor_t), because trailing
+// padding after `memory_type` is compiler-defined and not part of what the
+// MLIR-emitted model.dll actually reads.
+static_assert(offsetof(tensor_t, data) == 0,
+              "tensor_t.data must remain the first field");
+static_assert(offsetof(tensor_t, shape) == sizeof(void *),
+              "tensor_t.shape moved -- update all three tensor_t copies");
+static_assert(offsetof(tensor_t, memory_type) ==
+                  offsetof(tensor_t, element_size) + sizeof(size_t),
+              "tensor_t.memory_type moved -- update all three tensor_t "
+              "copies");
+
 // A contiguous array of tensor_t descriptors (inputs or outputs).
 struct span_t {
   tensor_t *data; // Pointer to array of tensor_t
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -36,7 +36,7 @@ jobs:
       THEROCK_VERSION: therock-dist-windows-gfx1151-7.11.0
       SCCACHE_GHA_ENABLED: "true"
       OGA_REPO: AMDmoore/onnxruntime-genai
-      OGA_REF: 5b547c7ec3be521d8590ae0c5f3e0ac450af0b7a
+      OGA_REF: c4fc2873abe020438e5cf8289b161fb6747554e5
       # ilammy/msvc-dev-cmd and mozilla-actions/sccache-action have no
       # Node.js 24 releases yet; allow Node.js 20 until they are updated.
       ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: "true"
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -36,7 +36,7 @@ jobs:
       THEROCK_VERSION: therock-dist-windows-gfx1151-7.11.0
       SCCACHE_GHA_ENABLED: "true"
       OGA_REPO: AMDmoore/onnxruntime-genai
-      OGA_REF: c4fc2873abe020438e5cf8289b161fb6747554e5
+      OGA_REF: 220b79ff0f19501957b56c40eb2a0089bf3d97c9
       # ilammy/msvc-dev-cmd and mozilla-actions/sccache-action have no
       # Node.js 24 releases yet; allow Node.js 20 until they are updated.
       ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: "true"
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -36,7 +36,7 @@ jobs:
       THEROCK_VERSION: therock-dist-windows-gfx1151-7.11.0
       SCCACHE_GHA_ENABLED: "true"
       OGA_REPO: AMDmoore/onnxruntime-genai
-      OGA_REF: 220b79ff0f19501957b56c40eb2a0089bf3d97c9
+      OGA_REF: c886da303a669377e2dcbb306c06a55da9355731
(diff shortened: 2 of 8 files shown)


wcy123 (approved):
# PR #161 Review: GPU Memory Pipeline for OGA

**PR:** https://github.com/ROCm/onnx-hipdnn-ep/pull/161
**Author:** Zhenze Wang (zz002)
**Reviewer:** Chunye Wang
**Date:** 2026-04-28

---

## Summary

This PR completes the GPU memory pipeline (part 3 of 3-layer fix) enabling zero-copy inference for OGA KV cache.

**Performance:** 4x-6x speedup achieved ✅
- Mistral-7B int4: 3.94 → 15.92 TPS (+304%)
- Phi4-14B int4: 2.12 → 13.18 TPS (+522%)

**Architecture:** Extends tensor ABI with `memory_type` field, adds alias path to skip H2D/D2H copies when caller provides GPU-accessible buffers.

---

## Review Comments

### 1. ✅ Good: ABI Extension Design

**Files:**
- `lib/Runtime/hipdnn_ep_runtime.h:231-238`
- `backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp:104-111`

**Change:** Added `memory_type` field to `tensor_t` struct.

**Analysis:**
- Values match ORT's `OrtMemoryInfoDeviceType` 1:1 (no remapping needed)
- Backward compatible: old code passes uninitialized field → fallback to legacy path
- Comments clearly document the ABI sync requirement between two headers

**Good!** Clean ABI extension with clear documentation.

---

### 2. ⚠️ ABI Sync Maintenance Risk

**Issue:** `tensor_t` layout duplicated in 3 locations:
- `lib/Runtime/hipdnn_ep_runtime.h:231-238`
- `backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp:104-111`
- `tools/hip-test-dll/hip-test-dll.cpp:541-548`

**Problem:** No compile-time check that the layouts stay synchronized. Manual sync required.

**Recommendation:**
Add a static assertion to verify struct layout:

```cpp
// In a common header included by all three files
static_assert(sizeof(tensor_t) == sizeof(void*) + sizeof(int64_t*) +
              sizeof(size_t) * 2 + sizeof(int),
              "tensor_t layout mismatch - check all three definitions");
static_assert(offsetof(tensor_t, memory_type) ==
              offsetof(tensor_t, element_size) + sizeof(size_t),
              "memory_type field offset mismatch");
```

**Severity:** Medium - future maintenance hazard

---

### 3. ✅ Good: Alias Fast Path Logic

**Files:**
- `lib/Runtime/hipdnn_ep_runtime_tensor.cpp:293-305` (prepare_input)
- `lib/Runtime/hipdnn_ep_runtime_tensor.cpp:407-420` (prepare_output)

**Logic:**
```cpp
const bool alias_caller_buffer = (tensor->memory_type == TENSOR_MEMORY_GPU);
if (alias_caller_buffer) {
  gpu_ptr = tensor->data;  // Zero-copy alias
} else {
  gpu_ptr = pool_alloc(size_bytes);  // Legacy path
}
```

**Analysis:**
- Clean separation: GPU memory → alias, others → copy
- Preserves backward compatibility for CPU/FPGA/NPU
- `TensorBuffer.is_aliased` flag correctly tracks ownership

**Excellent!** Clear ownership semantics.

---

### 4. ✅ Fix: DEBUG and PERF Separation

**File:** `lib/Runtime/debug_log.h:161-172`

**Change:** PERF no longer inherits from DEBUG.

**Analysis:**

The fix is correct ✅ - DEBUG and PERF are now independent:
- `HIPDNN_EP_DEBUG=1` → logs only, no sync overhead
- `HIPDNN_EP_PERF=1` → timing measurements with sync cost
- New `RUNTIME_PERF_LOG` macro separates the two

**Design note:**

These are orthogonal concerns that should never have been coupled:
- **DEBUG:** "Show me what's happening" (logging)
- **PERF:** "Measure performance" (timing with sync cost)

The original design coupled them via:
```cpp
// Old (poor design):
return enabled || hipdnn_ep_debug_enabled();
```

This was poor separation of concerns - like requiring profiling whenever you add printf debugging.

**Lesson for future code:**

Keep independent concerns independent. If two features serve different purposes and have different performance characteristics, they should be separately controllable.

**Verdict:** Good fix that improves the design.

---

### 5. ⚠️ PERF Sync Trade-off Needs Documentation

**Location:** `lib/Runtime/hipdnn_ep_runtime_tensor.cpp:319-333`

```cpp
if (g_perf.initialized && g_perf.inference_num > 0) {
  (void)hipStreamSynchronize(static_cast<hipStream_t>(state->stream));
  // Compute timing and print [PERF] line
}
```

**Issue:** The PERF sync behavior needs clearer documentation about the trade-off.

**What's happening:**

PERF mode **must** synchronize to get accurate per-phase timing:
```
hipEventRecord(h2d_start);
hipMemcpyAsync(...);  // H2D async
hipEventRecord(h2d_end);
hipStreamSynchronize();  // ← Required to read event times
hipEventElapsedTime(&h2d_ms, h2d_start, h2d_end);
```

**The trade-off:**

✅ **PERF ON:** Accurate per-phase breakdown (H2D=0ms, Compute=47ms, D2H=0ms)
❌ **PERF ON:** Kills GPU pipelining → artificially low TPS

```
Without PERF (natural pipelining):
  Inference 1: [H2D][Compute][D2H]
  Inference 2:      [H2D][Compute][D2H]  ← overlaps with #1
  → High throughput

With PERF (forced sync every inference):
  Inference 1: [H2D][Compute][D2H] [SYNC - wait]
  Inference 2:                            [H2D][Compute][D2H] [SYNC]
  → Low throughput (no overlap)
```

**What this PR fixed:**

**Before:** `HIPDNN_EP_DEBUG=1` → auto-enables PERF → sync penalty → slow
**After:** `HIPDNN_EP_DEBUG=1` → debug logs only, no sync → fast

Now you can have debug traces without the PERF overhead.

**Recommendation:**

Add comment explaining the trade-off:

```cpp
// PERF measurement requires hipStreamSynchronize before reading event times
// via hipEventElapsedTime. This gives accurate per-phase timing (H2D/Compute/D2H)
// but prevents GPU pipeline overlap across inferences, artificially lowering
// measured TPS.
//
// Use HIPDNN_EP_PERF=1 when you need per-inference breakdown.
// Use HIPDNN_EP_DEBUG=1 (without PERF) to measure natural throughput with logs.
if (g_perf.initialized && g_perf.inference_num > 0) {
```

**Severity:** Low - documentation clarity (behavior is correct)

---

### 6. ✅ Good: Backward Compatibility

**File:** `tools/hip-test-dll/hip-test-dll.cpp:555-563`

```cpp
tensor.memory_type = TENSOR_MEMORY_CPU;  // host buffers, runtime does H2D
```

**Analysis:**
- `hip-test-dll` explicitly stamps CPU memory type
- Falls through to legacy H2D/D2H path
- No behavior change for existing host-input callers

**Excellent!** Preserves backward compatibility.

---

### 7. ⚠️ CMake Workaround for Bad Design in PR #208

**File:** `cmake/deps.cmake:148`

```cpp
if(NOT BUILD_MOCK_RUNTIME)
  set(morphizen_ENABLE_HIP_GPU_ALLOCATOR ON CACHE BOOL ... FORCE)
endif()
```

**Issue:** This PR has to force `MORPHIZEN_ENABLE_HIP_GPU_ALLOCATOR=ON` because PR #208 defaults it to OFF.

**Analysis:**

This is a workaround for a design problem in ROCm/MorphiZen#208:
- The GPU allocator is the **entire purpose** of PR #208 (4x-6x speedup)
- Yet it's a CMake option with default OFF
- This PR must force it ON to actually get the benefit

**The flag shouldn't exist:**
- GPU allocator is not optional - it's core functionality
- Without it, EP falls back to 6x slower CPU memory path
- This is like making CUDA optional in CUDA EP

**See PR #208 review for details:**
- Issue: CMake flag creates broken default configuration
- Recommendation: Remove flag entirely, make GPU allocator mandatory
- Or at minimum: flip default to ON

**For this PR:**

The `FORCE` here is correct given PR #208's current state, but it highlights that PR #208's design is wrong.

**Recommendation:**

After PR #208 merges:
1. Remove `MORPHIZEN_ENABLE_HIP_GPU_ALLOCATOR` flag from MorphiZen entirely
2. Remove this `FORCE` line from deps.cmake
3. GPU allocator becomes mandatory (as it should be)

**Severity:** Low - workaround is correct, but points to upstream design issue

---

### 8. ⚠️ Submodule and OGA_REF Dependencies

**Files:**
- `3rd-party/morphizen` → commit `e6ca655`
- `.github/workflows/windows-build.yml` → `OGA_REF: c4fc287`

**Issue:** PR depends on:
- ROCm/MorphiZen#208 (submodule target)
- AMDmoore/onnxruntime-genai#8 (OGA_REF target)

**PR description says:**
> "CI on this PR will fail until both merge... Marking as draft for now."

**Status:** Both PRs are OPEN (not merged yet).

**Recommendation:**
- Keep as DRAFT until dependencies merge ✅
- Update commit hashes to post-merge commits
- Verify CI green before final review

**Severity:** Blocking - cannot merge before dependencies

---

### 9. ⚠️ Missing: Unit Test for Alias Path

**Observation:** No new test validates the alias fast path.

**Current tests:**
- `hip-test-dll` - explicitly uses CPU memory (legacy path)
- No test exercises `memory_type = TENSOR_MEMORY_GPU`

**Recommendation:**
Add test that:
1. Allocates buffer via `hipHostMalloc(Mapped|Coherent)`
2. Constructs `tensor_t` with `memory_type = TENSOR_MEMORY_GPU`
3. Runs inference
4. Validates PERF shows 0 H2D/D2H bytes

**Severity:** Medium - alias path only tested via OGA integration

---

## Summary

| # | Issue | Severity | Blocking? |
|---|---|---|---|
| 1 | ABI extension design | - | ✅ Good |
| 2 | ABI sync maintenance risk | Medium | No |
| 3 | Alias fast path logic | - | ✅ Good |
| 4 | PERF telemetry fix | - | ✅ Good |
| 5 | PERF sync documentation | Low | No |
| 6 | Backward compatibility | - | ✅ Good |
| 7 | CMake workaround for PR #208 design | Low | No (points to upstream issue) |
| 8 | Dependency on PRs #208, OGA#8 | - | ✅ Blocking (draft) |
| 9 | Missing alias path unit test | Medium | No |

---

## Verdict: **APPROVE (with non-blocking suggestions)**

**Blocking:**
- ✅ Dependencies declared (PRs #208, OGA#8) - keep as DRAFT until they merge

**Should address (non-blocking):**
1. Add static assertions for `tensor_t` layout sync (issue #2)
2. Document PERF sync cost in code (issue #5)
3. Remove `MORPHIZEN_ENABLE_HIP_GPU_ALLOCATOR` flag after PR #208 review (issue #7)
4. Add unit test for alias fast path (issue #9)

**Excellent work:**
- Performance results (4x-6x speedup) ✅
- Clean ABI extension ✅
- Backward compatibility preserved ✅
- PERF telemetry bug fixed ✅
- Clear ownership semantics with `is_aliased` flag ✅

**Core implementation is solid.** The non-blocking issues are defensive programming improvements for long-term maintenance.

---

**Recommendation:**
1. Address non-blocking suggestions in follow-up commits
2. Wait for PRs #208 and OGA#8 to merge
3. Update submodule/OGA_REF to post-merge commits
4. Verify CI green
5. Ready to merge ✅

**Reviewed by:** Chunye Wang
**Date:** 2026-04-28

zz002:
Thanks @wcy123, addressed in `7af3e89`:

| # | Status | Fix |
|---|---|---|
| 2 ABI sync risk | ✅ | `static_assert` trio (data offset 0, shape offset = sizeof(void*), memory_type offset = element_size + sizeof(size_t)) added to all 3 `tensor_t` definitions. Asserts on per-field offsets rather than raw `sizeof` (trailing padding after `memory_type` is compiler-defined). |
| 5 PERF sync trade-off doc | ✅ | Added paragraph above the `hipStreamSynchronize` in `prepare_input(0)` explaining: why sync is required (event timing semantics), the cost (kills inter-inference pipeline overlap, lowers measured TPS), and which env knob to use for which intent (PERF=accurate-but-serialised vs DEBUG=natural-tps). Cross-refs back to `debug_log.h` for the PERF/DEBUG decoupling. |
| 7 CMake `FORCE` workaround | ✅ already done | Picked up in `da63748` of this PR. Morphizen #208 default flipped to ON, so we no longer `set(... ON FORCE)` for real builds; only `OFF FORCE` for mock (which can't link HIP anyway). |
| 9 Alias path unit test | TODO follow-up | Needs a real GPU + real `hipHostMalloc(Mapped\|Coherent)` buffer, which the mock-runtime test infra doesn't have today. Tracked separately. |

Submodule bump to morphizen `a765fdb` is included (carries the matching #208 review fixes + TheRock CI plumbing).

PTAL.

backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
```diff
@@ -13,15 +13,64 @@ namespace mlir_compilation {
...
+// MLIR-emitted model.dll actually reads.
+static_assert(offsetof(tensor_t, data) == 0,
+              "tensor_t.data must remain the first field");
+static_assert(offsetof(tensor_t, shape) == sizeof(void *),
+              "tensor_t.shape moved -- update all three tensor_t copies");
+static_assert(offsetof(tensor_t, memory_type) ==
+                  offsetof(tensor_t, element_size) + sizeof(size_t),
+              "tensor_t.memory_type moved -- update all three tensor_t "
+              "copies");
+
```
tejus26, line 74: Why do we duplicate this struct? Sorry, I may not have all the context.

tejus26: Also, can you please confirm if this PR also avoid h2d and d2h copies for dGPUs where the memory is not unified? Its ok if it is not, but want to confirm. thanks!

zz002: > Also, can you please confirm if this PR also avoid h2d and d2h copies for dGPUs where the memory is not unified? Its ok if it is not, but want to confirm. thanks!

Yes — but with a caveat: on a discrete GPU the implementation avoids the explicit H2D/D2H copies, but it does not avoid the implicit PCIe traversal, so on dGPU it isn't truly zero-cost.

zz002: `tensor_t` is the wire-protocol ABI between three components that are intentionally kept decoupled (compiler-emitted model.dll, EP runtime DLL, hip-test-dll harness). The `static_assert` block in each copy catches any layout drift at compile time. The longer rationale is in the rewritten doc comment in `b658d071`.

PR #162, opened 2026-04-28
 fix: support uint8 packed nibble zero_points in MatMulNBits WMMA/GEMV paths

AMDmoore:
## Summary

* ONNX MatMulNBits models with RTN quantization store zero_points as uint8 packed nibbles (two 4-bit values per byte). The WMMA and GEMV col-major kernel paths were incorrectly casting these uint8 bytes as fp16 pointers, reading garbage zero_point values and producing NaN outputs.
* Added `zp_elem_size` attribute through the full compiler-to-runtime pipeline (OnnxToHip -> HipOps -> HipToLLVM -> Runtime -> GPU kernel) to detect and correctly unpack uint8 packed nibble zero_points to fp16 before use.
* Added compile-time validation to reject unsupported zero_points element types early.

## Test plan

* [x] Qwen2.5-Coder-14B-instruct-rtn-128gs model: MorphiZen EP output matches DML EP baseline (previously produced garbled/NaN output)
* [x] Greedy decoding (top_k=1) produces coherent, contextually correct text for 128 generated tokens
* [x] pre-commit hooks pass

### DIFF
--- a/3rd-party/custom_kernels/hip/matmul_nbits_kernel.hip
+++ b/3rd-party/custom_kernels/hip/matmul_nbits_kernel.hip
@@ -1386,6 +1386,64 @@ __global__ void matmul_nbits_add_bias_rowmajor(
  * Section 4: Host-side launcher (extern "C")
  * ======================================================================= */
 
+/* =======================================================================
+ * Section 4a: uint8 packed nibble zero_points -> FP16 conversion
+ *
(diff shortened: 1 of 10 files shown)


amd-mingw (approved):
this PR fixed Qwen2.5-14B accuracy issue lead by zero_point data type, and has reviewed with Booker and @ZhanxingPu , in order to not block daily test, approved and merged it.

PR #165, opened 2026-04-28
fix(gemm,qmoe): MIOpen broadcast contract + QMoE block_size handling

BoarQing:
## Summary
Two bug fixes in the JIT-compiled runtime that together unblock execution of the gpt-oss-120b single-layer dump:

1. **`broadcastBiasToOutput` (`lib/Runtime/real/gemm.cpp`)** — satisfy MIOpen's `miopenOpTensor` contract that the A‑operand descriptor must match the destination descriptor (only the B operand may broadcast). The old code passed the small bias descriptor (`cDesc`) for both A and B while the destination used the full output descriptor (`outDesc`), so MIOpen returned `A and C Tensors do not match` (status 7) at `gemm.cpp:158` whenever the bias actually needed broadcasting (e.g. `[1,N]` onto `[M,N]`). Crashed any graph in which ORT's `MatMulAddFusion` emitted a `Gemm` with broadcastable bias — most notably the MoE router MatMul.

   ```c
   // Before
   miopenOpTensor(handle, miopenTensorOpAdd,
                  &alpha1, cDesc, C, &alpha2, cDesc, C,   // A & B = bias  (small)
                  &beta_c, outDesc, output);              // dest          (full)
   // After: zero output, then
   miopenOpTensor(handle, miopenTensorOpAdd,
                  &alpha1, outDesc, output,    // A = dest, full shape
                  &alpha2, cDesc,   C,         // B = bias, broadcasts
                  &beta_c, outDesc, output);   // dest = output
   // => output = 1.0 * 0 + beta * broadcast(C) + 0 * 0 = beta * broadcast(C)
   ```

2. **QMoE `block_size` (`lib/Conversion/OnnxToHip/QMoEConversion.cpp` + `lib/Runtime/real/qmoe.cpp`)** — the `com.microsoft.QMoE` `block_size` attribute is documented as optional and is omitted by some quantization tools (e.g. AWQ exports of gpt-oss-120b). The conversion silently defaulted it to `0`, which made `wrap_qmoe` divide by zero in `(hidden_size + block_size - 1) / block_size` and crash with `STATUS_INTEGER_DIVIDE_BY_ZERO` (`0xC0000094`). Derive `block_size` from the FC1 (gate_up_proj) scales tensor shape when the attribute is missing — `scales` has shape `[num_experts, output_features, k_blocks_fc1]` with `k_blocks_fc1 = ceil(hidden_size / block_size)`, so `block_size = ceil(hidden_size / k_blocks_fc1)` gives the exact value for evenly‑split layouts (the common case; gpt-oss-120b yields `block_size = 32`). Fail conversion with a clear MLIR diagnostic when neither the attribute nor the shapes provide it. Add an early-return guard in `wrap_qmoe` so future regressions surface as a readable `fprintf` rather than a Windows DEP exception.

## Test plan

Verified on Strix Halo (gfx1151) with `onnxruntime_perf_test` running the gpt-oss-120b dumps:

### `single_op/QMoE/QMoE_seq*` (8 variants)
| variant | before | after |
|---|---|---|
| `QMoE_seq1` | crash `0xC0000094` in `wrap_qmoe` | **passes**, 1.07 ms |
| `QMoE_seq128` | crash `0xC0000094` | **passes**, 3.43 ms |
| `QMoE_seq256` | crash `0xC0000094` | **passes**, 7.85 ms |
| `QMoE_seq512` | crash `0xC0000094` | **passes**, 11.4 ms |
| `QMoE_seq1024` | crash `0xC0000094` | **passes**, 20.9 ms |
| `QMoE_seq2048` | crash `0xC0000094` | **passes**, 41.8 ms |
| `QMoE_seq3072` | crash `0xC0000094` | **passes**, 64.5 ms |
| `QMoE_dynamic` | EP rejects (`invalid tensor dimension size`) | unchanged (separate, pre-existing bug) |

### `single_layer/single_layer_seq128.onnx`
- **before**: `MIOpen Error: A and C Tensors do not match` at `gemm.cpp:158` → process crash
- **after**: completes inference (`exit=0`), 6.54 ms. Two pre-existing failures still log warnings but are non-fatal:
  - `wrap_group_query_attention: gqa_forward failed (rc=-1)` (pre-existing)
  - `hipBLAS error: hipblasLtMatmulAlgoGetHeuristic ... gemm.cpp:370 (status=3)` for the lm_head shape `M=128 N=201088 K=2880` (pre-existing — already reproducible on `single_layer_seq1` before this PR)

## Notes for reviewers
- Two-file fix for (1) and two-file fix for (2). No public API changes.
- Single-op QMoE tests under `single_op/QMoE/*` now pass, which gives focused coverage for the new code path. The current QMoE LIT/E2E tests don't exercise a model without an explicit `block_size`; consider adding one.
- Issues exposed but not fixed here: (a) `wrap_group_query_attention`'s GQA failure on the gpt-oss-120b layer, (b) `hipblasLtMatmulAlgoGetHeuristic` returning `INVALID_VALUE` for the lm_head shape `M=128 N=201088 K=2880` and the small router shape `M=128 N=128 K=2880`. Tracked separately.

### DIFF
--- a/lib/Conversion/OnnxToHip/QMoEConversion.cpp
+++ b/lib/Conversion/OnnxToHip/QMoEConversion.cpp
@@ -88,9 +88,38 @@ QMoEToHip::matchAndRewrite(mlir::Operation *op,
   auto kIntAttr = op->getAttrOfType<mlir::IntegerAttr>("k");
   auto kAttr = rewriter.getI64IntegerAttr(kIntAttr ? kIntAttr.getSInt() : 1);
 
+  // ms.QMoE's `block_size` attribute is documented as optional (no spec
+  // default) and is omitted by some quantization tools (e.g. AWQ exports of
+  // gpt-oss-120b). Without it `wrap_qmoe` later divides by zero. When the
+  // attribute is absent or non-positive, derive block_size from the
+  // FC1 (gate_up_proj) scales tensor: scales has shape
+  //     [num_experts, output_features, k_blocks_fc1]
+  // with k_blocks_fc1 = ceil(hidden_size / block_size) and the activation
+  // input has shape [..., hidden_size]. So
+  //     block_size = ceil(hidden_size / k_blocks_fc1)
+  // which gives the exact value when the model uses an even split (the
+  // common case).
   auto blockSizeIntAttr = op->getAttrOfType<mlir::IntegerAttr>("block_size");
-  auto blockSizeAttr = rewriter.getI64IntegerAttr(
-      blockSizeIntAttr ? blockSizeIntAttr.getSInt() : 0);
+  int64_t blockSizeValue = blockSizeIntAttr ? blockSizeIntAttr.getSInt() : 0;
+  if (blockSizeValue <= 0) {
+    auto inputType = mlir::dyn_cast<mlir::RankedTensorType>(input.getType());
+    auto scalesType =
+        mlir::dyn_cast<mlir::RankedTensorType>(fc1Scales.getType());
+    if (inputType && scalesType && inputType.getRank() > 0 &&
+        scalesType.getRank() > 0) {
+      int64_t hiddenDim = inputType.getShape().back();
+      int64_t kBlocksDim = scalesType.getShape().back();
+      if (hiddenDim > 0 && kBlocksDim > 0) {
+        blockSizeValue = (hiddenDim + kBlocksDim - 1) / kBlocksDim;
+      }
+    }
+  }
+  if (blockSizeValue <= 0) {
+    return rewriter.notifyMatchFailure(
+        op, "QMoE: missing `block_size` attribute and cannot infer it from "
+            "input/fc1_scales tensor shapes");
+  }
+  auto blockSizeAttr = rewriter.getI64IntegerAttr(blockSizeValue);
 
   auto normIntAttr =
       op->getAttrOfType<mlir::IntegerAttr>("normalize_routing_weights");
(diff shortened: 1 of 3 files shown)


lib/Conversion/OnnxToHip/QMoEConversion.cpp
```diff
@@ -88,9 +88,38 @@ QMoEToHip::matchAndRewrite(mlir::Operation *op,
...
+  if (blockSizeValue <= 0) {
+    auto inputType = mlir::dyn_cast<mlir::RankedTensorType>(input.getType());
+    auto scalesType =
+        mlir::dyn_cast<mlir::RankedTensorType>(fc1Scales.getType());
+    if (inputType && scalesType && inputType.getRank() > 0 &&
+        scalesType.getRank() > 0) {
+      int64_t hiddenDim = inputType.getShape().back();
+      int64_t kBlocksDim = scalesType.getShape().back();
+      if (hiddenDim > 0 && kBlocksDim > 0) {
+        blockSizeValue = (hiddenDim + kBlocksDim - 1) / kBlocksDim;
```
zpye, line 113: If I understand correctly, when `block_size` is not provided, `fc1_scales` will be a 2D tensor, according to the official documentation: [https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#commicrosoftqmoe](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#commicrosoftqmoe)

PR #166, opened 2026-04-28
[Update] onnx-model-splitter: quick chunk export, --all matrix, tiered shape verify

cliu1003:
## Motivation

**`export_chunk_model.py`**  
Per updated product requirements, the chunk exporter should default to a **single** representative sliding-window prefill/decode pair (and matching `genai_config`) instead of always emitting the full sequence-length matrix. Full-matrix export remains available when explicitly requested.

**`verify_extraction.py`**  
Fix **false positives** for some models during extraction verification.

**`README.md`**  
Align user-facing docs with the new export default, optional full-matrix flow, environment setup, and the two-phase verify behavior and flags.

## Technical Details

**`export_chunk_model.py`**

- **By default:** export only the single sliding-window variant tied to **512** (`p512m{max_length}` prefill/decode and matching `genai_config`).
- **Only with `--all`:** export the **full** variant matrix from the built-in sequence-length set.

### DIFF
--- a/tools/onnx-model-splitter/README.md
+++ b/tools/onnx-model-splitter/README.md
@@ -12,114 +12,34 @@ and analysis. Supports LLM (text), Vision, and Embedding models.
 | Script | Description |
 |--------|-------------|
 | `extract_submodels.py` | Main extraction tool — splits an ONNX model into `single_op`, `single_layer`, and `full_model` submodels with multiple dimension variants |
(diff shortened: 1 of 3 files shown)


amd-mingw (approved):
this is a following PR of [PR153](https://github.com/ROCm/onnx-hipdnn-ep/pull/153) to update model pre-process tool, approved and merged it.

PR #168, opened 2026-04-29
fix(gemm,qmoe,gqa,matmul): unblock gpt-oss-120b end-to-end (4 fixes)

BoarQing:
## Summary

Four fixes that together unblock gpt-oss-120b end-to-end through the MorphiZen EP. They walk together because each downstream bug only becomes reachable after the upstream one is fixed: the runtime aborts before fix 2 without fix 1, and so on. After all four, the model executes all 36 transformer layers without aborting and without any cascade-to-NaN inside the inference. (One unfixed accuracy gap remains and is summarized in the [Out of scope](#out-of-scope) section.)

---

### 1. MIOpen broadcast contract — `lib/Runtime/real/gemm.cpp` (commit `3c503f5`, also in #165)

**What went wrong.** `broadcastBiasToOutput` called `miopenOpTensor` with `cDesc` (the small bias) for both A and B, but MIOpen requires the A-operand descriptor to equal the destination descriptor. Result: `A and C Tensors do not match` (status 7) at `gemm.cpp:158` for any broadcastable bias — notably the MoE router MatMul.

**What I did.** Pre-zero the destination, then compute `output = 1.0*output + beta*broadcast(C)` with `aDesc == outDesc == cDesc(dest)` and only the bias broadcasting on the B side.

---

### 2. QMoE `block_size` derivation — `lib/Conversion/OnnxToHip/QMoEConversion.cpp`, `lib/Runtime/real/qmoe.cpp` (commit `3c503f5`)

**What went wrong.** `QMoE`'s `block_size` attribute is documented as optional; AWQ exports of gpt-oss-120b omit it. The conversion pass defaulted it silently to 0, causing `STATUS_INTEGER_DIVIDE_BY_ZERO` (`0xC0000094`) inside `wrap_qmoe`'s `k_blocks` computations.

**What I did.** Derive `block_size = ceil(hidden_size / k_blocks_fc1)` from the FC1 (gate_up_proj) scales tensor shape when the attribute is missing. Add a runtime guard so any future regression surfaces as a readable error instead of a divide-by-zero exception.

---

### 3. GQA prefill sentinel + hipBLASLt heuristic fallback — `lib/Runtime/real/{gqa,gemm,matmul}.cpp` (commit `ecd443f`)

**What went wrong.**
- *GQA prefill:* ORT passes `seqlens_k[b] = -1` on the first prefill step (no past KV). Both `gqa_forward_hipblaslt` branches computed `total_seq = -1+1 = 0` and `past_len = -sq` and rejected with `invalid seqlens_k[0]+1=0`.
- *hipBLASLt heuristic:* `hipblasLtMatmulAlgoGetHeuristic` returns 0 algorithms / status 3 for outlier shapes on gfx1151's Tensile library — MoE router MatMul (`M=128 N=128 K=2880`, very small N) and lm_head MatMul (`M=128 N=201088 K=2880`, very large N).

**What I did.**
- Treat `seqlens_k_val < 0` as a fresh prefill (`past_len=0`, `total_seq=sq`) in both fused-decode and decomposed branches; non-prefill validation still runs for any non-negative value.
- Replace the single `(workspace=256MB, request=1)` heuristic call with a 4-rung escalation ladder; if every rung returns 0 algorithms, store `use_default_algo=true` and pass `algo=nullptr` to `hipblasLtMatmul`. Same change in `wrap_gemm` and `wrap_hipblasLtMatmul`.

| Heuristic rung | Workspace | Request count | Why it can help |
|---|---|---|---|
| 1 | 256 MB | 1 | original behaviour |
| 2 | 0 | 1 | force non-workspace algo (often unblocks narrow-N and huge-`ld` shapes) |
| 3 | 256 MB | MAX | broader candidate sweep |
| 4 | 0 | MAX | non-workspace + broader sweep |
| 5 | (cache fallback) | n/a | `use_default_algo = true`, pass `algo = nullptr` to `hipblasLtMatmul` |

---

### 4. QMoE per-expert zero-point stride — `lib/Runtime/real/qmoe.cpp` (commits `0918fdf`, `42587e0`)

**What went wrong.** ONNX `MatMulNBits` with `bits=4` stores `zero_points` as uint8 packed nibbles (two 4-bit values per byte). `qmoe.cpp`'s per-expert pointer arithmetic for `fc1_zp_e`/`fc2_zp_e` was striding by `N * k_blocks` bytes (the unpacked layout) instead of `N * ceil(k_blocks/2)` bytes. Combined with #162's correct kernel-side unpack, every expert past index 0 read ZPs from a region 2× too large — overshooting into the next expert's bytes. Dequantized weight magnitudes grew ~10× per MoE layer until fp16 overflow; `router_probs` became NaN; `topk_routing` returned -1 for every token; **0 active experts from layer ~3 onward**; garbage logits.

**What I did.** Use `e * N * ((k_blocks + 1) / 2)` for both `fc1_zp_e` and `fc2_zp_e`, matching the kernel's `convertZpToFp16` formula `packed_cols = (groups_k + 1) / 2` and the hard-coded `zp_elem_size = 1` already passed to `hip_matmul_nbits`. Verified locally on gpt-oss-120b prefill: qmoe per-layer output magnitudes are now stable (no exponential growth, no fp16 overflow inside MoE), all 36 MoE layers route healthily (`50–110/128` active per call, was `0/128` from layer 3+).

---

## Out of scope

**fp16 residual stream overflow at deep layers.** After all four fixes, gpt-oss-120b runs all 36 layers without NaN inside `wrap_qmoe`/routing, but generated text is still ASCII noise because `down_proj.scales` carry AWQ outliers that grow with depth (max scale: layer 0 = 0.6 → layer 6 = 12.8 → layer 28 = **548**). The resulting matmul outputs accumulate past fp16's ±65 504 ceiling in the residual stream by layer ~35. `hip_matmul_nbits` is verified correct (EP-vs-CPU-vs-numpy across all gpt-oss shapes and outlier scales up to 100). Recommended fix: switch the residual-path buffers from fp16 to bf16. Detailed analysis + 3-fix comparison + recommended migration plan in the [comment thread](#issuecomment-4342725564).

A separate topk-routing kernel out-of-bounds reachable when `router_probs` contains pathological values (exposed by `hip-onnx-runner`'s random generator) is **resolved by fix 4** — the cascade was caused by upstream NaN, not by `topk_routing` itself.

### DIFF
--- a/lib/Conversion/OnnxToHip/QMoEConversion.cpp
+++ b/lib/Conversion/OnnxToHip/QMoEConversion.cpp
@@ -88,9 +88,38 @@ QMoEToHip::matchAndRewrite(mlir::Operation *op,
   auto kIntAttr = op->getAttrOfType<mlir::IntegerAttr>("k");
   auto kAttr = rewriter.getI64IntegerAttr(kIntAttr ? kIntAttr.getSInt() : 1);
 
+  // ms.QMoE's `block_size` attribute is documented as optional (no spec
+  // default) and is omitted by some quantization tools (e.g. AWQ exports of
+  // gpt-oss-120b). Without it `wrap_qmoe` later divides by zero. When the
+  // attribute is absent or non-positive, derive block_size from the
+  // FC1 (gate_up_proj) scales tensor: scales has shape
+  //     [num_experts, output_features, k_blocks_fc1]
+  // with k_blocks_fc1 = ceil(hidden_size / block_size) and the activation
+  // input has shape [..., hidden_size]. So
+  //     block_size = ceil(hidden_size / k_blocks_fc1)
+  // which gives the exact value when the model uses an even split (the
+  // common case).
   auto blockSizeIntAttr = op->getAttrOfType<mlir::IntegerAttr>("block_size");
-  auto blockSizeAttr = rewriter.getI64IntegerAttr(
-      blockSizeIntAttr ? blockSizeIntAttr.getSInt() : 0);
+  int64_t blockSizeValue = blockSizeIntAttr ? blockSizeIntAttr.getSInt() : 0;
+  if (blockSizeValue <= 0) {
+    auto inputType = mlir::dyn_cast<mlir::RankedTensorType>(input.getType());
+    auto scalesType =
+        mlir::dyn_cast<mlir::RankedTensorType>(fc1Scales.getType());
+    if (inputType && scalesType && inputType.getRank() > 0 &&
+        scalesType.getRank() > 0) {
+      int64_t hiddenDim = inputType.getShape().back();
+      int64_t kBlocksDim = scalesType.getShape().back();
+      if (hiddenDim > 0 && kBlocksDim > 0) {
+        blockSizeValue = (hiddenDim + kBlocksDim - 1) / kBlocksDim;
+      }
+    }
+  }
+  if (blockSizeValue <= 0) {
+    return rewriter.notifyMatchFailure(
+        op, "QMoE: missing `block_size` attribute and cannot infer it from "
+            "input/fc1_scales tensor shapes");
+  }
+  auto blockSizeAttr = rewriter.getI64IntegerAttr(blockSizeValue);
 
   auto normIntAttr =
       op->getAttrOfType<mlir::IntegerAttr>("normalize_routing_weights");
--- a/lib/Runtime/real/gemm.cpp
+++ b/lib/Runtime/real/gemm.cpp
@@ -111,6 +111,18 @@ static bool resolveGemmMiopenType(int64_t typeCode, miopenDataType_t &dt) {
   }
 }
 
+// MIOpen contract for miopenOpTensor: the A-operand descriptor and the
+// destination descriptor must have *identical* shapes; only the B operand may
+// broadcast. The previous implementation passed the small bias descriptor
+// (cDesc) as both A and B while the destination was the full-output descriptor,
+// which MIOpen rejected with "A and C Tensors do not match" (status 7) any time
+// the bias actually needed broadcasting (e.g. a [1,N] bias onto [M,N]).
+//
+// Fix: zero the output buffer, then compute
+//     output = 1.0 * output + beta * broadcast(C)
+// so the A operand and destination both use outDesc and only B (the bias)
+// broadcasts. Pre-zeroing avoids dependency on uninitialized output (which
+// could contain NaNs that 1.0*NaN would propagate).
 static int broadcastBiasToOutput(RuntimeState *state, const void *C,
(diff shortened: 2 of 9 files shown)


BoarQing:
## Why inference is still wrong (after `0918fdf`) and how to fix it

After this PR's three patches land, gpt-oss-120b runs all 36 transformer layers without aborting and without any cascade-to-NaN inside `wrap_qmoe`. Generated text is still single-character ASCII noise because the **fp16 residual stream silently overflows** at layer ~35.

### Root cause — data flow

```text
gpt-oss-120b model file: down_proj.scales has AWQ outliers that grow with depth
    layer  0:  max scale = 0.6     →  matmul out ±50
    layer  6:  max scale = 12.8    →  matmul out ±10,000   ← spike
    layer 22:  max scale = 205     →  matmul out ±100,000s
    layer 28:  max scale = 548     →  matmul out ±100,000s
                       │
                       ▼
   ┌──────────────────────────────────────────────────────────────────┐
   │ residual_stream (fp16)  ← accumulates per-layer matmul outputs   │
   │                                                                   │
   │ layer  0   ±50          (fine)                                    │
   │ layer  6   ±10,500      (already large)                           │
   │ layer 35   ±58,820      (close to fp16 max ±65,504)               │
   │ layer 35   skip = ±33,760                                         │
   │ layer 35   add: 58,820 + 33,760 = 92,580   →  fp16 = +Inf  ← HERE │
   └──────────────────────────────────────────────────────────────────┘
                       │
                       ▼
   miopenT5LayerNormForward(Inf-bearing input)
        Inf · 1/sqrt(mean(Inf²))  =  Inf · 0  =  NaN     ← correctly emits NaN
                       │
                       ▼
   All 36 downstream ops cascade NaN through the residual stream
                       │
                       ▼
   lm_head matmul reads an all-NaN input  →  201,088 NaN logits per token
                       │
                       ▼
   Tokenizer gets a flat/garbage logit distribution  →  random ASCII tokens
```

This is **not** a kernel bug. `hip_matmul_nbits` produces bit-identical results to ORT CPU EP and a numpy oracle across every gpt-oss shape and outlier scales up to 100 (verified standalone). The model is faithfully producing values that exceed fp16's range when summed across 36 layers.

### candidate fix

| # | Fix | Scope | Pros | Cons |
|---|---|---|---|---|
| **1** | **Re-quantize gpt-oss-120b** | Re-run quantization with `bits=8` or AWQ with explicit smoothing nodes baked into the ONNX graph. | EP unchanged. | Out of EP scope. Every future outlier-rich model needs its own re-quant. Larger model file. |

amd-mingw (approved):
this PR fix GPT-120 runtime error, approved and merged it.

PR #171, opened 2026-04-29
perf(gqa): no-expand decode fast path + unified pipeline + amortized workspace

zpye:
## Summary

- Add a group-batched "no-expand" GQA pipeline that reads K/V directly from the BNSD KV cache via hipBLASLt strided-batched GEMM, eliminating the `expand_kv_k` / `expand_kv_v` kernel launches and their `B*H*total_seq*d` fp16 scratch buffers. At `sq == 1` (decode) the Q/O transposes are also skipped because BSHD and BNSD share memory. Gated by `HIPDNN_EP_GQA_NO_EXPAND` (default **on**); A/B against the expand path with `HIPDNN_EP_GQA_NO_EXPAND=0`.
- Unify the expand and no-expand pipelines into a single parameterized path controlled by two orthogonal flags (`use_no_expand`, `need_transpose`), and simplify the activation gate down to the correctness-only minimum (env flag + `present_key`/`present_value` + `sq == 1` or the prefill opt-in).
- Introduce `HIPDNN_EP_GQA_NO_EXPAND_PREFILL` (default **off**) to opt `sq > 1` (prefill) into the no-expand pipeline for A/B experimentation, without changing the shipped prefill behaviour.
- Amortize the shared runtime workspace with a 1.5× growth factor so monotonically-growing decode requests (S buffers sized to `B*H*total_seq` per step) no longer trigger a `hipStreamSynchronize` + `hipFree` + `hipMalloc` cycle on every token. Cold-start keeps the exact requested size.

### `lib/Runtime/real/gqa.cpp`

- **No-expand pipeline**. Score and Value GEMMs use `batch = B * G` with explicit per-operand batch strides:
  - A (K or V) stride = `present_seq * d` — walks one BNSD group matrix per batch, stepping over the pre-allocated cache page so the descriptor stays stable across decode steps.
  - B (Q or S) stride = `HPG * sq * d` (Score) / `HPG * sq * total_seq` (Value) — packs the `HPG` queries of a group into the GEMM's N dimension; the HPG replication is consumed inside the GEMM rather than by duplicating K/V in memory.
  - C (S or O) stride = `HPG * sq * total_seq` (Score) / `HPG * sq * d` (Value) — byte-identical layout to the expand path's BNSD `[B, H, sq, *]` output, so downstream softmax / causal-mask / O-transpose kernels are reused unmodified.
- **`gqa_no_expand_enabled()` helper** reads `HIPDNN_EP_GQA_NO_EXPAND` once via a static initializer; default true when the env var is unset or not `"0"`.
- **`gqa_no_expand_prefill_enabled()` helper** reads `HIPDNN_EP_GQA_NO_EXPAND_PREFILL` once; default false.
- **Unified pipeline**. Replaces the previous dedicated no-expand decode block and the original expand+transpose pipeline with a single path controlled by:
  - `use_no_expand` — selects strided-batched GEMM with `batch = B*G` and the no-expand strides above (`true`) vs. dense `batch = B*H` with expand-kv workspace buffers (`false`).
  - `need_transpose` — true when `sq > 1`, keeping the Q-transpose before Score and the O-transpose after Value; false at `sq == 1` because BSHD and BNSD coincide in memory.
- **Pointer plumbing**. Score A = `use_no_expand ? present_key : d_Kexp`; Score B = `need_transpose ? d_Qtrans : qSrc`; Value A = `use_no_expand ? present_value : d_Vexp`; Value C = `need_transpose ? d_O : output`.
- **Workspace layout** now conditionally allocates buffers — `Qtrans` / `O` only when `need_transpose`, `Kexp` / `Vexp` only when not `use_no_expand`, `S_f32` / `S_fp16` always. RoPE and packed-QKV-split scratch regions are unchanged.
- **Gate simplification**. The previous 9-clause decode guard (`sq==1`, `B==1`, non-null `key`/`value`, `present_key`/`present_value`, `!need_rope`, `!use_smooth_softmax`, null `head_sink`, `!packed_qkv`) collapses to `gqa_no_expand_enabled() && present_key && present_value && (sq == 1 || gqa_no_expand_prefill_enabled())`. Rationale:
  - Packed-QKV split (Step 0) and RoPE (Steps 1–2) already run upstream and redirect `qSrc` / `kSrc` / `vSrc` before the GEMM sees them — they don't interact with the GEMM flavour.
  - Smooth softmax and head-sink run inside the shared softmax kernel on the same BNSD S tensor both flavours produce, so they're orthogonal.
  - `sq == 1` is still required for *skipping the transposes* (enforced by `need_transpose`) but not for correctness of the strided-batched GEMM; `sq > 1` is now supported via the prefill flag.
  - `B == 1` is no longer required — the GEMM batch is now `B * G`.

### `lib/Runtime/hipdnn_ep_runtime_state.cpp`

- `hipdnn_ep_state_ensure_workspace` now grows to `max(needed_size, 1.5 × current)` whenever an existing workspace is being enlarged. Cold-start (no existing workspace) still allocates exactly `needed_size` to avoid silently doubling the first large allocation during warmup. Debug log prints both the allocated size and the requested size.

## Behaviour Summary

| Stage | `HIPDNN_EP_GQA_NO_EXPAND` | `HIPDNN_EP_GQA_NO_EXPAND_PREFILL` | Path taken |
|---|---|---|---|
| Decode (`sq == 1`) | default (on) | — | **no-expand**, no Q/O transpose |
| Prefill (`sq > 1`) | default (on) | default (off) | **expand** + Q/O transpose (unchanged) |
| Decode | `0` | — | expand + Q/O transpose (baseline) |
| Prefill | default (on) | `1` | no-expand + Q/O transpose (opt-in A/B) |

## Test plan

- Build verified.
- Pre-commit hooks run on modified files (trailing whitespace, EOL, mixed line endings, merge conflicts, lintrunner, license headers) — all passed.
- E2E runtime tests executed on Mistral-7B across four configurations, all producing byte-identical logits against the pre-change expand baseline:
  - Pre-change expand (reference).
  - No-expand decode only (`HIPDNN_EP_GQA_NO_EXPAND=1`, `HIPDNN_EP_GQA_NO_EXPAND_PREFILL=0`).
  - Decode-only no-expand post unification (same env, post-refactor binary).
  - Prefill opt-in (`HIPDNN_EP_GQA_NO_EXPAND=1`, `HIPDNN_EP_GQA_NO_EXPAND_PREFILL=1`).
- Workspace amortization verified: steady-state decode no longer invokes `hipFree` / `hipMalloc` on incremental `total_seq` growth (confirmed via operator profiling showing the reallocation stalls only at initial warmup and log-spaced growth points).

### DIFF
--- a/lib/Runtime/hipdnn_ep_runtime_state.cpp
+++ b/lib/Runtime/hipdnn_ep_runtime_state.cpp
@@ -993,6 +993,22 @@ int hipdnn_ep_state_ensure_workspace(RuntimeState *state, size_t needed_size) {
   if (state->workspace_size >= needed_size)
     return 0;
 
+  // Amortize growth: when enlarging an existing workspace, round the new
+  // size up to at least 1.5x the current buffer. Callers whose request
+  // size grows monotonically by a small increment per inference (e.g. the
+  // GQA decode path, which sizes S buffers to B*H*total_seq and adds B*H
+  // elements per token) would otherwise trigger a hipStreamSynchronize +
+  // hipFree + hipMalloc cycle on every decode step; with the 1.5x factor
+  // that drops to O(log N) reallocations over the whole generation.
+  // Cold-start (no existing workspace) keeps the exact requested size so
+  // warmup doesn't silently double large initial allocations.
+  size_t alloc_size = needed_size;
+  if (state->workspace_size > 0) {
+    size_t grown = state->workspace_size + state->workspace_size / 2; // 1.5x
+    if (grown > alloc_size)
+      alloc_size = grown;
+  }
+
   // Grow: free old, allocate new.
   // Sync the stream first to ensure no in-flight kernel is still using the
   // old workspace buffer (prevents use-after-free on async GPU execution).
@@ -1005,17 +1021,18 @@ int hipdnn_ep_state_ensure_workspace(RuntimeState *state, size_t needed_size) {
     state->workspace_size = 0;
   }
 
-  if (hipMalloc(&state->workspace, needed_size) != hipSuccess) {
+  if (hipMalloc(&state->workspace, alloc_size) != hipSuccess) {
     fprintf(
         stderr,
         "hipdnn_ep_state_ensure_workspace: hipMalloc failed for %zu bytes\n",
-        needed_size);
+        alloc_size);
     return -1;
   }
 
-  state->workspace_size = needed_size;
-  RUNTIME_DEBUG_LOG("[workspace] Allocated shared workspace: %zu bytes\n",
-                    needed_size);
+  state->workspace_size = alloc_size;
+  RUNTIME_DEBUG_LOG(
+      "[workspace] Allocated shared workspace: %zu bytes (requested %zu)\n",
+      alloc_size, needed_size);
   return 0;
 }
 
--- a/lib/Runtime/real/gqa.cpp
+++ b/lib/Runtime/real/gqa.cpp
@@ -16,6 +16,7 @@
 #include <cassert>
 #include <cmath>
 #include <cstdio>
+#include <cstdlib>
 #include <cstring>
 #include <functional>
 #include <unordered_map>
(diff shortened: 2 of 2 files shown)


amd-mingw (approved):
this pr removed GQA decode phase transpose and expand kv to improve perf, approved and merged.

PR #177, opened 2026-04-29
feat: add float32/float16 bidirectional cast kernel support

AMDmoore:
## Summary
- Add float32 to float16 and float16 to float32 HIP cast kernels
- Previously only INT64->INT32 was supported, causing runtime failures on models with mixed-precision Cast operations (e.g. Gemma 3 has 276 fp32/fp16 Cast nodes for LayerNorm)
- New kernels use __float2half / __half2float intrinsics

## Test plan
- [x] Gemma 3 4B model: 128 no-chunk benchmark passes with correct output
- [x] Gemma 3 4B model: p512 chunk benchmark passes with correct output
- [x] No unsupported conversion warnings in output
- [ ] CI build and LIT tests pass

### DIFF
--- a/3rd-party/custom_kernels/hip/cast_kernel.hip
+++ b/3rd-party/custom_kernels/hip/cast_kernel.hip
@@ -7,6 +7,7 @@
 #include "debug_log.h"
 
 #include <hip/hip_runtime.h>
+#include <hip/hip_fp16.h>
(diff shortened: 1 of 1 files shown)


amd-mingw (approved):
this PR fixed gemma3-4b CAST kernel issue, approved and merged.

PR #180, opened 2026-04-29
fix(hip): materialize contiguous temporaries for strided memref operands

fhanuman:
## Summary

Adds a new MLIR pass `--hip-promote-strided-operands` that walks every `DestinationStyleOpInterface` op in the `hip` dialect and, for each input memref operand whose layout is non-identity (non-zero offset or non-contiguous strides), inserts a `memref.alloc` + `memref.copy` + `memref.dealloc` sequence to materialize a contiguous temporary.

This closes a silent-corruption gap between MLIR (which can faithfully represent strided memrefs via `memref.subview`) and the HIP runtime call ABI used by `--convert-hip-to-llvm` (which receives only a single bare pointer per memref operand, with no channel for offset or per-dimension strides).

## Why

The HIP runtime wrappers extract just `aligned_ptr` from the LLVM memref descriptor (`extractMemRefPtr` in `HipToLLVMUtils.h`). When a `hip.*` op consumes a strided memref — for example, the result of `memref.subview` produced by `onnx.Split` lowering — the wrapper would otherwise read the parent buffer from its base, ignoring the slice and silently producing wrong results. This was reported on `Split -> Sigmoid` as roughly a 10x L2 error vs CPU.

## What

1. **New pass** at `lib/Dialect/Transforms/PromoteStridedHipOperands.cpp`. Predicate is type-based (`!type.getLayout().isIdentity()`), so it catches strided values regardless of how they were produced — `memref.subview`, `memref.reinterpret_cast` with a non-zero offset, function args carrying a `strided<>` layout, etc.

2. **Pipeline placement**: between `hip-optimize-memrefs` and `hip-pool-allocs` in `Pipelines.cpp`, so the new transient buffers fold cleanly into the existing memory pool — no extra `hipMalloc` per inference.

3. **Helper rename**: `extractMemRefPtr` -> `extractContiguousMemRefPtr` in `HipToLLVMUtils.h` and across 17 `*Lowering.cpp` files, to make the contiguous-only contract explicit at every call site. Also adds an `extractMemRefDescriptor` helper that returns the full descriptor (no callers in this PR; reserved for future per-op zero-copy lowerings).

4. **DPS-init operands are intentionally left alone**. Outputs are guaranteed contiguous by construction in this pipeline (fresh `memref.alloc` from bufferization or function args under `IdentityLayoutMap`), and promoting them would change observable behavior because the consumer writes through them.

This mirrors upstream `linalg::promoteSubViews` in spirit, but with a generalized type-based predicate that catches non-subview-produced strided values and a single-purpose, focused implementation.

## Compiler IR walkthrough

**Full IR dump:** https://gist.github.com/fhanuman/8612bcfdacd24281652613e84cfe5933

The three stages below are the ones that show the new pass at work and confirm it folds cleanly into the existing pool allocation.

<details>
<summary><b>Before <code>--hip-promote-strided-operands</code></b> — strided subviews feed <code>hip.sigmoid</code> directly (this is what would silently corrupt at LLVM lowering)</summary>

```mlir
module attributes {hipdnn.input_count = 1 : i64, ...} {
  func.func @main_graph(%arg0: !hip.context,
                        %arg1: memref<1x128x2048xf16> {onnx.name = "X"},
                        %arg2: memref<1x128x1024xf16> {bufferize.result},
                        %arg3: memref<1x128x1024xf16> {bufferize.result})
      attributes {onnx.graph.name = "split_sigmoid_test"} {
    %subview = memref.subview %arg1[0, 0, 0] [1, 128, 1024] [1, 1, 1]
        : memref<1x128x2048xf16>
          to memref<1x128x1024xf16, strided<[262144, 2048, 1]>>
    %subview_0 = memref.subview %arg1[0, 0, 1024] [1, 128, 1024] [1, 1, 1]
        : memref<1x128x2048xf16>
          to memref<1x128x1024xf16, strided<[262144, 2048, 1], offset: 1024>>
    %cast = memref.cast %subview_0
        : memref<1x128x1024xf16, strided<[262144, 2048, 1], offset: 1024>>
          to memref<1x128x1024xf16, strided<[262144, 2048, 1], offset: ?>>
    hip.sigmoid(%arg0)
      ins(%subview : memref<1x128x1024xf16, strided<[262144, 2048, 1]>>)
      outs(%arg2 : memref<1x128x1024xf16>)
    hip.sigmoid(%arg0)
      ins(%cast : memref<1x128x1024xf16, strided<[262144, 2048, 1], offset: ?>>)
      outs(%arg3 : memref<1x128x1024xf16>)
    return
  }
}
```

Note both `hip.sigmoid` ops receive a `strided<>` memref. At LLVM lowering, `extractMemRefPtr` would extract only `aligned_ptr` from the descriptor and discard the offset/strides — so the second sigmoid would read from offset 0 of `%arg1` instead of offset 1024. That is the bug.

</details>

<details>
<summary><b>After <code>--hip-promote-strided-operands</code></b> — <code>memref.alloc</code> + <code>memref.copy</code> materialize a contiguous temporary; consumer is rewritten to read from it; paired <code>memref.dealloc</code> after</summary>

```mlir
module attributes {hipdnn.input_count = 1 : i64, ...} {
  func.func @main_graph(%arg0: !hip.context,
                        %arg1: memref<1x128x2048xf16> {onnx.name = "X"},
                        %arg2: memref<1x128x1024xf16> {bufferize.result},
                        %arg3: memref<1x128x1024xf16> {bufferize.result})
      attributes {onnx.graph.name = "split_sigmoid_test"} {
    %subview = memref.subview %arg1[0, 0, 0] [1, 128, 1024] [1, 1, 1]
        : memref<1x128x2048xf16>
          to memref<1x128x1024xf16, strided<[262144, 2048, 1]>>
    %subview_0 = memref.subview %arg1[0, 0, 1024] [1, 128, 1024] [1, 1, 1]
        : memref<1x128x2048xf16>
          to memref<1x128x1024xf16, strided<[262144, 2048, 1], offset: 1024>>
    %cast = memref.cast %subview_0
        : memref<1x128x1024xf16, strided<[262144, 2048, 1], offset: 1024>>
          to memref<1x128x1024xf16, strided<[262144, 2048, 1], offset: ?>>

    %alloc = memref.alloc() : memref<1x128x1024xf16>
    memref.copy %subview, %alloc
        : memref<1x128x1024xf16, strided<[262144, 2048, 1]>>
          to memref<1x128x1024xf16>
    hip.sigmoid(%arg0)
      ins(%alloc : memref<1x128x1024xf16>)
      outs(%arg2 : memref<1x128x1024xf16>)
    memref.dealloc %alloc : memref<1x128x1024xf16>

    %alloc_1 = memref.alloc() : memref<1x128x1024xf16>
    memref.copy %cast, %alloc_1
        : memref<1x128x1024xf16, strided<[262144, 2048, 1], offset: ?>>
          to memref<1x128x1024xf16>
    hip.sigmoid(%arg0)
      ins(%alloc_1 : memref<1x128x1024xf16>)
      outs(%arg3 : memref<1x128x1024xf16>)
    memref.dealloc %alloc_1 : memref<1x128x1024xf16>

    return
  }
}
```

Each `hip.sigmoid` now reads from a fresh contiguous buffer. `memref.copy` lowers via `mlir_memref_copy` runtime which honors the source strides correctly. The `memref.subview` ops survive but become inputs to `memref.copy` only — they never reach a HIP runtime call.

</details>

<details>
<summary><b>After <code>--hip-pool-allocs</code></b> — the new <code>memref.alloc</code>s fold into the pool as <code>memref.view</code>s; zero extra <code>hipMalloc</code> per inference</summary>

```mlir
module attributes {hipdnn.buffer_count = 2 : i64,
                   hipdnn.buffer_offsets = [0, 0],
                   hipdnn.pool_size = 262144 : i64, ...} {
  func.func @main_graph(%arg0: !hip.context,
                        %arg1: memref<1x128x2048xf16> {onnx.name = "X"},
                        %arg2: memref<1x128x1024xf16> {bufferize.result},
                        %arg3: memref<1x128x1024xf16> {bufferize.result})
      attributes {onnx.graph.name = "split_sigmoid_test"} {
    %subview = memref.subview %arg1[0, 0, 0] [1, 128, 1024] [1, 1, 1]
        : memref<1x128x2048xf16>
          to memref<1x128x1024xf16, strided<[262144, 2048, 1]>>
    %subview_0 = memref.subview %arg1[0, 0, 1024] [1, 128, 1024] [1, 1, 1]
        : memref<1x128x2048xf16>
          to memref<1x128x1024xf16, strided<[262144, 2048, 1], offset: 1024>>
    %cast = memref.cast %subview_0 : ... to memref<1x128x1024xf16, strided<[262144, 2048, 1], offset: ?>>

    %c262144 = arith.constant 262144 : index
    %0 = hip.get_pool(%arg0, %c262144) : memref<?xi8>
    %c0_1 = arith.constant 0 : index

    %view = memref.view %0[%c0][] : memref<?xi8> to memref<1x128x1024xf16>
    memref.copy %subview, %view
        : memref<1x128x1024xf16, strided<[262144, 2048, 1]>>
          to memref<1x128x1024xf16>
    hip.sigmoid(%arg0) ins(%view : memref<1x128x1024xf16>) outs(%arg2 : memref<1x128x1024xf16>)

    %view_2 = memref.view %0[%c0_1][] : memref<?xi8> to memref<1x128x1024xf16>
    memref.copy %cast, %view_2
        : memref<1x128x1024xf16, strided<[262144, 2048, 1], offset: ?>>
          to memref<1x128x1024xf16>
    hip.sigmoid(%arg0) ins(%view_2 : memref<1x128x1024xf16>) outs(%arg3 : memref<1x128x1024xf16>)
    return
  }
}
```

The two `memref.alloc`s introduced by the new pass have become `memref.view %pool[offset]` ops sharing the existing 256 KiB pool (offsets 0 and 0 — the lifetimes don't overlap). The deallocs are gone (pool views are managed by `hip.get_pool`). At inference time, this is a single `hip.get_pool` call followed by view arithmetic — no `hipMalloc` overhead per call.

</details>

## Notes for reviewers

- The pass is a safety net: zero cost on models that don't need it (e.g., the Llama-3-8B export above), strictly correct on models that do.
- The 17-file rename of `extractMemRefPtr` -> `extractContiguousMemRefPtr` is mechanical and intentional. It enforces the "contiguous-only" contract at every call site so future lowerings can't silently regress on strided inputs.
- The new pass file is ~180 LOC of focused logic; full rationale is in the file header block.

### DIFF
--- a/include/hip/Dialect/Transforms/Passes.td
+++ b/include/hip/Dialect/Transforms/Passes.td
@@ -103,6 +103,43 @@ def OptimizeMemRefsPass : Pass<"hip-optimize-memrefs", "mlir::func::FuncOp"> {
   ];
 }
 
+def PromoteStridedHipOperandsPass
+    : Pass<"hip-promote-strided-operands", "mlir::func::FuncOp"> {
+  let summary = "Materialize contiguous temporaries for non-canonical memref "
+                "operands of hip.* ops";
+  let description = [{
+    Walks every DestinationStyleOpInterface op in the `hip` dialect.  For
+    each input (DPS-input) operand whose memref type has a non-identity
+    layout (non-zero offset or non-contiguous strides), inserts:
+
+    ```mlir
+    %tmp = memref.alloc() : memref<...identity-layout...>
+    memref.copy %strided, %tmp : ..., ...
+    // consumer is rewritten to read from %tmp instead of %strided
+    memref.dealloc %tmp : memref<...identity-layout...>
+    ```
+
+    DPS-init (output) operands are left untouched: in this pipeline they are
+    always either fresh `memref.alloc` results or function arguments under
+    `IdentityLayoutMap`, so they are guaranteed contiguous by construction.
+
+    The pass exists to bridge the contract gap between the IR (which can
+    represent strided memrefs faithfully via memref.subview) and the HIP
+    runtime call ABI (which receives a single bare pointer with no offset /
+    strides channel).  Without this promotion, hip.* runtime calls silently
+    read the base of the parent buffer instead of the intended slice.
+
+    Recommended pipeline position:
+      hip-optimize-memrefs -> hip-promote-strided-operands -> hip-pool-allocs
+
+    This placement ensures the new transient allocations are routed through
+    the HIP memory pool instead of producing extra hipMalloc calls.
+  }];
+  let dependentDialects = [
+    "mlir::memref::MemRefDialect"
+  ];
+}
+
 def PoolAllocsPass : Pass<"hip-pool-allocs", "mlir::func::FuncOp"> {
   let summary = "Pool memref.alloc ops into a grow-on-demand i8 byte buffer";
   let description = [{
--- a/lib/Conversion/HipToLLVM/ActivationLowering.cpp
+++ b/lib/Conversion/HipToLLVM/ActivationLowering.cpp
@@ -43,8 +43,8 @@ lowerMiopenActivation(OpType op, typename OpType::Adaptor adaptor,
 
   // Extract pointers
   Value statePtr = adaptor.getCtx();
-  Value inputPtr = extractMemRefPtr(adaptor.getX(), rewriter, loc);
-  Value outputPtr = extractMemRefPtr(adaptor.getY(), rewriter, loc);
+  Value inputPtr = extractContiguousMemRefPtr(adaptor.getX(), rewriter, loc);
+  Value outputPtr = extractContiguousMemRefPtr(adaptor.getY(), rewriter, loc);
 
   auto outputType = cast<MemRefType>(op.getY().getType());
 
@@ -158,8 +158,8 @@ struct GeluOpLowering : public ConvertOpToLLVMPattern<GeluOp> {
 
     // Extract pointers
     Value statePtr = adaptor.getCtx();
-    Value inputPtr = extractMemRefPtr(adaptor.getInput(), rewriter, loc);
-    Value outputPtr = extractMemRefPtr(adaptor.getOutput(), rewriter, loc);
+    Value inputPtr = extractContiguousMemRefPtr(adaptor.getInput(), rewriter, loc);
+    Value outputPtr = extractContiguousMemRefPtr(adaptor.getOutput(), rewriter, loc);
 
     auto outputType = cast<MemRefType>(op.getOutput().getType());
 
@@ -236,8 +236,8 @@ struct SiluOpLowering : public ConvertOpToLLVMPattern<SiluOp> {
       return failure();
 
     SmallVector<Value> args = {
-        adaptor.getCtx(), extractMemRefPtr(adaptor.getInput(), rewriter, loc),
-        extractMemRefPtr(adaptor.getOutput(), rewriter, loc)};
+        adaptor.getCtx(), extractContiguousMemRefPtr(adaptor.getInput(), rewriter, loc),
+        extractContiguousMemRefPtr(adaptor.getOutput(), rewriter, loc)};
 
     LLVM::CallOp::create(rewriter, loc, *funcOp, args);
     rewriter.eraseOp(op);
@@ -279,8 +279,8 @@ struct MiopenSoftmaxOpLowering
                                  inputDesc.size(rewriter, loc, i));
 
     SmallVector<Value> args = {
-        adaptor.getCtx(), extractMemRefPtr(adaptor.getInput(), rewriter, loc),
-        extractMemRefPtr(adaptor.getOutput(), rewriter, loc), rows, cols};
+        adaptor.getCtx(), extractContiguousMemRefPtr(adaptor.getInput(), rewriter, loc),
+        extractContiguousMemRefPtr(adaptor.getOutput(), rewriter, loc), rows, cols};
 
     LLVM::CallOp::create(rewriter, loc, *funcOp, args);
     rewriter.eraseOp(op);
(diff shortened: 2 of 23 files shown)


MaheshRavishankar (approved):
Ok, I can stamp this, but this is really not the approach to use. Ideally we should just pass the strides along to the runtime call and have the kernel handle strides.

This has the potential of having a performance impact by introducing copies when we dont need it. But since this is a correctness guard we should maybe land this first.

fhanuman:
> Ok, I can stamp this, but this is really not the approach to use. Ideally we should just pass the strides along to the runtime call and have the kernel handle strides.
> 
> This has the potential of having a performance impact by introducing copies when we dont need it. But since this is a correctness guard we should maybe land this first.

Thanks, Mahesh — appreciate the feedback. Sorry, I didn't add this clarification to the PR description. I thought about this, and you are right that descriptor-passing is the proper design. However, I wanted to minimize the source changes due to the `ROCm EP 0.1` release time constraint: every `hip.*` op uses the same single-pointer ABI, so a proper fix touches every `*Lowering.cpp`, every runtime wrapper, and the custom kernels themselves. Also, most of the models I looked at have zero copies due to this issue, so we don’t run into performance issues often.

I do agree that the copies become wasteful once Mingyue's Split work lands. I will have follow-up PRs to address this. Stride-aware lowerings can opt in one op at a time, and the promote pass becomes a fallback.

PR #191, opened 2026-05-05
perf: GQA smart-dispatch + seqlens_k cache + per-Compute timing infra

fhanuman:
## Summary

Three decode-perf changes that close ~70% of the asym Llama-3.1-8B AWQ gap to llama.cpp Vulkan on Strix Halo (gfx1151). The two GQA changes ship default-on; the perf instrumentation is opt-in.

- **GQA seqlens_k cache** (default on; `HIPDNN_EP_GQA_CACHE_SEQLENS=0` to disable). Caches `seqlens_k_val` once per forward pass: a 32-layer Llama decode issues 1 D2H instead of 32, eliminating ~30-45 ms/token of pipeline stalls. New `hipdnn_ep_runtime_begin_compute()` hook is exported from `model.dll` and called from `MlirCustomOp::Compute()` to invalidate the cache. Older `model.dll` without the hook triggers a `LOG(WARNING)` at session creation directing the user to set `=0`.
- **GQA smart-dispatch** (default on; `HIPDNN_EP_GQA_FUSED_DECODE_MAX_T=N` to tune, default 256). Routes decode to the decomposed hipBLASLt path when `total_seq > 256`. The fused decode kernel serializes over time with cross-wave reductions (~12x slower than GEMM-based decomposed at total_seq=2048). Short prompts stay on the fused fast path. `HIPDNN_EP_GQA_DISABLE_FUSED_DECODE=1` retained as a manual A/B override.
- **Per-`Compute()` perf instrumentation** (`HIPDNN_EP_PERF=1`, default off). CPU + GPU bracket timing around `inference_state_->compute()`. Emits `[PERF]` per call and `[PERF SUMMARY]` at unload, splitting wall / marshal / compute_cpu / gpu / fence_residual. ~25 us/call when on; zero overhead when off.

## Validation (Strix Halo gfx1151)

- Asym Llama-3.1-8B AWQ p=2048 m=16384: **9.66 -> 28.77 tok/s (+197%)** with this PR's defaults plus `OGA_USE_IO_BINDING=1`.
- Sym Llama-3.1-8B AWQ p=2048: +59.3% TPS from the OGA IoBinding optimization alone.
- `ctest` E2E: pass; no GQA fixture regressions.

The OGA IoBinding optimization lives in `fhanuman/onnxruntime-genai @ 36acb4a` and is pinned via the workflow `OGA_REPO`/`OGA_REF` (commented inline). Once the change lands upstream the pin reverts to the canonical AMD fork.

We need the following OGA changes to reproduce the performance numbers here:

https://github.com/AMDmoore/onnxruntime-genai/pull/10

Also, please see below confluence page for the benchmarking numbers:

https://amd.atlassian.net/wiki/spaces/AISW/pages/1667157596/Performance+improvements

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -35,14 +35,8 @@ jobs:
       GTEST_VERSION: "1.15.0"
       THEROCK_VERSION: therock-dist-windows-gfx1151-7.11.0
       SCCACHE_GHA_ENABLED: "true"
-      # OGA fork carrying the IoBinding decode optimization that routes
-      # session.Run through OrtIoBinding -- required for the asym
-      # Llama-3.1-8B AWQ p=2048 9.66 -> 28.77 tok/s number reported in the
-      # PR body. Once the IoBinding change lands upstream
-      # (microsoft/onnxruntime-genai), revert OGA_REPO to AMDmoore (or the
-      # canonical AMD fork) and OGA_REF to the merged commit.
-      OGA_REPO: fhanuman/onnxruntime-genai
-      OGA_REF: 36acb4a9afd76273c180a6b45f3029e3ed419d54
+      OGA_REPO: AMDmoore/onnxruntime-genai
+      OGA_REF: 2615301864b4d2397231d443865cb96111cc5fc2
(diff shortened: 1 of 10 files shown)


amd-mingw (approved):
LGTM

PR #194, opened 2026-05-07
Fixes QWen accuracy issues and OP functionality.

amd-genmingz:
## Motivation

This PR combines the changes required for Qwen 3.5 into one document, specifically for testing.

**transpose**: Add support for 4-D transpose
**reduce_sum**: support datatype fp16
**matmulnbits**: add naive bits=8 path
**elemwise broadcast fix**: 
MIOpen miopenOpTensor requires A.shape == C.shape and only allows broadcasting B on dim 1, so we swap lhs/rhs before calling MIOpen when the broadcast source is not aligned with the output shape.
**split fix**: 
Fix pitch derivation for multi-dimensional strided sources in copy by replacing the hardcoded stride[rank-2] logic with dynamic stride-break detection, enabling correct copies of non-last-dim splits (e.g. fused QKV / fused gate-up) into contiguous temporary buffers.
**rope fix**:
Previously the kernel only supported BSNH [B, S, N, H]; now an is_bnsh flag is added: 0 for BSNH (also compatible with 3D [B, S, N*H] row-major layout), non-zero for BNSH [B, N, S, H], the default 4D layout of ONNX


amd-genmingz:
**Qwen3.5-35B/9B accuracy**:
http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hipdnn_ep_oga&id=6a0030c1745cce2015e36d9b
| model | Perplexity | 
| ---- | ---- | 
| Qwen3.5-35B-A3B_int4_rtn_128gs_cuda_space_text | 10.4464 |
| Qwen3.5-9B-rtn-int4-int8-128gs-fp16-onnx-gpu_text | 12.5305 |

all models accuracy(PPL): http://xsjvitisaijenkins:8080/view/ROCm/job/daily_hipdnn_ep_oga/310/

Qwen3.5-35B/9B perf: http://xsjvitisaijenkins:8080/view/ROCm/job/daily_hipdnn_ep_oga/309/
all models perf: http://xsjvitisaijenkins:8080/view/ROCm/job/daily_hipdnn_ep_oga/311/

amd-mingw (approved):
from the above CI test, there is no side effect on other models about perf and accuracy(PPL), so I approved and merged it firstly to do a daily test today before branch 0.1 release tomorrow.
BTW, there are some small issues, I have told @amd-genmingz offline, will with a follow-up PR to fix them.

PR #195, opened 2026-05-07
feat: update transpose op

amd-genmingz:
## Motivation

This is an extension to the existing `transpose` property in the code. Previously, `transpose` only supported 3D and only two dimensions. Now, the definition of `transpose` has been expanded to fully support the functionality of `onnx.Transpose`.

## Test Plan

- [x] lit test passed
- [x] single op accuracy test pased


amd-genmingz:
close because pr 194 contains this change

PR #196, opened 2026-05-07
fix: restore stream_sync call in inference_compute

AMDmoore:
## Summary
- The Range Op commit (c0a8d5e, #140) accidentally removed the \hipdnn_ep_stream_sync\ call from \GenerateInterface.cpp\, replacing it with only \hipdnn_ep_state_read_and_clear_error_flag\.
- This removed the only call site for \op_profile_resolve_and_print()\, causing per-operator GPU profiling data (\HIPDNN_EP_PERF=1\) to silently disappear -- events were recorded but never resolved or printed.
- Fix: restore \stream_sync\ before \
ead_error_flag\ so both profiling output and error flag checking work correctly.

## Changes
- \lib/Dialect/Transforms/GenerateInterface.cpp\: Add \hipdnn_ep_stream_sync\ to runtime func declarations and emit the call in \inference_compute\ between finalize_output and read_error_flag.
- \	est/lit/Transforms/GenerateInterface/test_basic_interface.mlir\: Add CHECK lines for \stream_sync\ and \
ead_and_clear_error_flag\ calls.

## Test plan
- [ ] LIT test \	est_basic_interface.mlir\ passes (verifies stream_sync is emitted)
- [ ] Run with \HIPDNN_EP_PERF=1\ and confirm per-op profiling data is printed after each inference

### DIFF
--- a/lib/Dialect/Transforms/GenerateInterface.cpp
+++ b/lib/Dialect/Transforms/GenerateInterface.cpp
@@ -446,6 +446,7 @@ private:
         {"hipdnn_ep_tensor_buffer_get_rank", i64, {ptr}},
         {"hipdnn_ep_tensor_buffer_get_size_bytes", i64, {ptr}},
         {"hipdnn_ep_state_init_with_fs", i32, {ptr, ptr, ptr, i64}},
+        {"hipdnn_ep_stream_sync", i32, {ptr}},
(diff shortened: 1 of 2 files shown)


amd-mingw (approved):
fix a bug

PR #197, opened 2026-05-08
feat(matmul_nbits): add naive bits=8 path

zpye:
## Summary

- Add a naive `bits=8` path to `MatMulNBits` so models that mix `bits=4`
  and `bits=8` ops in the same graph (e.g. Qwen3.5-35B-A3B
  `int4_rtn_128gs`, where 150 of the `MatMulNBits` ops use `bits=8`)
  no longer fail at runtime with the previous `bits != 4` reject.
- The `bits=4` fast paths (WMMA / GEMV / hipBLASLt) are completely
  untouched: `bits=8` is dispatched in an early-exit block at the top
  of `hip_matmul_nbits`, so existing `bits=4` models keep their
  autotuned behaviour bit-for-bit.
- Naive-first by design: ONNX `bits=8` stores `B` as unpacked `uint8
  [N, K]` with default zero point 128, which doesn't match any of the
  packed-nibble fast-path layouts. A dedicated naive int8 kernel is
  the smallest correct change that unblocks Qwen3.5-style mixed-bit
  models; faster int8 paths can layer on top later.

## Changes Overview

`3rd-party/custom_kernels/hip/matmul_nbits_kernel.hip`
- New `matmul_nbits_kernel_i8` mirroring `matmul_nbits_kernel`
  (the existing int4 naive kernel): row-major, any-alignment,
  batched via `blockIdx.z`, fp32 accumulation, optional bias,
  optional per-block `uint8` zero points (default 128 when
  `zero_points` is null). `B` layout is `[N, k_blocks, block_size]`
  `uint8` with no nibble packing.
- Relax the launcher gate from `bits != 4` to
  `bits != 4 && bits != 8`.
- New `if (bits == 8)` early-exit block at the top of
  `hip_matmul_nbits` that:
  - Validates `zp_elem_size == 1` when `zero_points` is provided.
  - Launches `matmul_nbits_kernel_i8` with the same `kBlockDim`
    grid layout the int4 fallback uses.
  - Logs a `[custom_kernels] hip_matmul_nbits: int8 naive ...` line
    so the dispatch path is visible during runtime debugging.

`3rd-party/custom_kernels/include/hip_custom_kernels.h`
- Update the `MatMulNBits` API header doc to spell out the dual
  `bits=4` / `bits=8` contract: differing `B` layout (packed vs
  unpacked) and differing default zero point (8 vs 128).

## Test plan

- Build verified.
- Numeric precision tests passed: 8 new `bits=8` cases pinning the
  four canonical Qwen3.5 8-bit shapes (router 2048→32,
  expert_up 2048→8192, attn_out 2048→4096, attn_qkv 4096→2048, all
  `block_size=128`) at `seq_len ∈ {1, 128}` (small-M / wide-M).
  All 8 cases match the CPU reference with cosine ≥ 0.999992.
  (Tests live in `onnx-numeric-tests` on the sibling
  `feat/matmul-nbits-int8` branch.)
- Regression: 24 of 25 existing `bits=4` `MatMulNBits` numeric
  precision tests continue to pass; the one pre-existing flaky
  `bits=4` WMMA case (`router_gpt_oss_shape[128]`) is unaffected
  by this change (runtime log confirms it never enters the int8
  path).

## Follow-ups (out of scope)

- GEMV-style int8 fast path for decode (`M=1`).
- WMMA / hipBLASLt int8 path for prefill (`M ≥ 16`).
- Autotuning + caching for int8 like int4 already has.

### DIFF
--- a/3rd-party/custom_kernels/hip/matmul_nbits_kernel.hip
+++ b/3rd-party/custom_kernels/hip/matmul_nbits_kernel.hip
@@ -101,6 +101,67 @@ __global__ void matmul_nbits_kernel(
 }
 
 /* =======================================================================
+ * Section 1a: Naive int8 fallback kernel (row-major, any alignment, batched)
+ *
+ * ONNX MatMulNBits with bits=8 layout (no nibble packing):
+ *   B      : uint8 [N, K]                     (1 byte per weight)
+ *   scales : fp16  [N, num_groups_k]
+ *   zeros  : uint8 [N, num_groups_k]          (or NULL, default zp=128)
+ *
+ * Dequant per element:
+ *   w_fp = (uint8_w - zp) * scale
+ *
+ * Each thread computes one C[m, n] across all batches via blockIdx.z.
+ * Mirrors matmul_nbits_kernel (int4) so behaviour and debug shape are
+ * symmetric; chosen as the first int8 path because it has no alignment
+ * or shape constraints and immediately unblocks Qwen3.5-style models
+ * that mix bits=4 and bits=8 in the same graph.
+ * ======================================================================= */
+
+__global__ void matmul_nbits_kernel_i8(
+    const __half* __restrict__ A,
+    const uint8_t* __restrict__ B,
+    const __half* __restrict__ scales,
+    const uint8_t* __restrict__ zp,
+    const __half* __restrict__ bias,
+    __half* __restrict__ output,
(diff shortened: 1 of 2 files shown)


zpye:
Superseded: re-pushing the branch directly to ROCm/onnx-hipdnn-ep and opening a new PR from there instead of from the zpye fork.

PR #207, opened 2026-05-11
fix(matmul_nbits): correct WMMA swizzle block-to-tile mapping for non-aligned n_tiles

AMDmoore:
## Problem

The WMMA path in `GemmFp16U4Impl` (`matmul_nbits_kernel.hip`) used a swizzle strategy to improve L2 cache locality by reordering how thread blocks are assigned to output tiles. The block-to-tile mapping was **non-bijective** when `n_tiles % sw != 0`:

- Some tiles were computed **twice** (collision — overwritten)
- Some tiles were **never computed** (missing — left with stale data from a prior kernel launch)

The subsequent bias-add kernel performed read-modify-write on the stale output, producing non-deterministic QMoE results. This caused the router `M` counts (tokens-per-expert) to differ across iterations for identical inputs, which forced autotune to re-run every prefill layer and inflated **TTFT by ~8x** on the 120B model.

## Root Cause

Old code fell back to row-major only when `bx >= n_tiles`, which is too late — collisions and missing tiles had already been assigned incorrectly before that check.

## Fix

Split the block grid into two regions:
- **Main region**: `(n_tiles / sw) * sw` columns — full `sw`-wide swizzle groups apply correctly
- **Tail region**: remaining `n_tiles % sw` columns — plain row-major

Also replaced `__builtin_nontemporal_store` with regular stores to restore cache coherence between the WMMA output write and the subsequent bias-add kernel read.

## Verification

- Exhaustive mapping test across **8,000 `(n_tiles, sw)` combinations** — bijection confirmed for all cases
- 120B model perf comparison (2024-token prompt, gfx1151):

| Metric | Baseline (main) | With Fix | Change |
|--------|----------------|----------|--------|
| TTFT | 51.1 s | 6.2 s | **8.2x faster** |
| Decode TPS | 38.1 tok/s | 37.7 tok/s | same |
| E2E | 52.8 s | 7.9 s | **6.7x faster** |

## Note

This fixes the **root cause** of the non-determinism. The `skip_autotune` workaround from PR #198 (which treated a symptom) has been reverted; autotune now runs correctly and caches results as intended.


amd-mingw:
all models perf: http://xsjvitisaijenkins:8080/view/ROCm/job/daily_hipdnn_ep_oga/329/
all models accuracy(PPL): http://xsjvitisaijenkins:8080/view/ROCm/job/daily_hipdnn_ep_oga/330/

amd-mingw (approved):
this PR fixed a bug to improve GPT-OSS-20B/120B TTFT, and from above CI test, not has side effect on other models, approved and merged it.

PR #212, opened 2026-05-11
feat: dynamic sequence length — compile once, run any shape

akholodnamdcom:
End-to-end enablement of ONNX models with dynamic batch/sequence dimensions (compile-once / run-any-shape), plus per-model test coverage for 8 new models and a generic `ModelSpec` / `BaseORTTests` / `BaseOGATests` framework with golden-cache CPU references.

  Pinned to morphizen `akholodn/dynseqlen` branch: dynamic-dim propagation + ONNX `image_features` graph-input/initializer tolerance.

  ## Compiler / runtime — core dynseqlen

  - `pass_main.cpp`: `DimSource` metadata resolves dynamic output dims at runtime from whichever input first defines the symbolic name (`total_sequence_length` → `attention_mask`); compile-time error when a dynamic dim has no resolvable symbolic name; new `resolved` field in `DimSource` proto so `{0,0}` no longer aliases a real source.
  - `MlirCustomOp.cpp::marshal_output_tensors`: resolves dynamic dims via `DimSource`; shape override for `present.N.{key,value}` reads actual shape from the matching `past_key_values.N` input (with a `>` guard so it only fires for shared-buffer mode). `present → past` mapping precomputed in the constructor (O(1) per decode step). New `inference_state_->begin_compute()` runtime hook invalidates per-Compute caches.
  - `RopeLowering`: dynamic batch/seq read from the runtime memref descriptor via `getMemRefDimSize()`; static-only for `num_heads` / `head_dim`.
  - `GqaLowering`: compile-time assert `query_hidden % (H + 2*G) == 0`; reads `seq_len` from the memref descriptor for dynamic dims.
  - `PoolAllocs`: hoist full def chain of dynamic alloc sizes via `SetVector` worklist (fixes SSA dominance for chains >1 hop); pool-grow path now `hipStreamSynchronize`s before `hipFree` of the in-flight buffer; Phase-4 hoist switched to DFS post-order topological sort; `isHoistable` whitelist now includes `arith::DivUIOp` / `DivSIOp`.
  - `ShapeConversion`: emit `tensor<0xi64>` constant for empty start/end ranges per ONNX spec.

  ## gfx1151 host-scalar SEGV fix

  gfx1151's `hipMalloc` returns true device memory (gfx1150 accidentally returned UMA-mapped host-accessible memory, masking the bug). Bufferized `tensor.from_elements` for shape arithmetic (`seqlens_k` etc.) emitted `memref.alloc` + `memref.store` that wrote into device memory from the host once `hip-pool-allocs` absorbed the alloc.

  - `GatherShapeFold` pass: folds `Gather(Shape(x), const_idx)` → `tensor.from_elements(tensor.dim)` before constants are externalized, collapsing the shape-arithmetic chain.
  - `hip-materialize-host-scalars` pass: redirects tiny static int `memref.alloc` ops with host-store users into a per-function host-mapped scratch buffer via the new `hip.get_host_scratch` op. Runtime helper uses `hipHostMalloc(Mapped)`, grow-on-demand, never shrinks — zero per-inference host allocations at steady state.

  ## SHARED_CONSTANTS content fingerprint

  Cache key was `(pid, total_size)`. Two compiled DLLs of one ONNX (fixed-shape decode + dynamic-shape) frequently produce constants blobs of the **same** `total_size` but **different** content because the MLIR compiler folds shape-dependent constants differently. Without a content key the second-to-load DLL silently attached the first one's blob → wrong `W_k` at the first layer whose constants diverged → `present.20.key cosine=0.728` → NaN cascade.

  Fix: extend the cache key with an FNV-1a fingerprint over per-constant **metadata descriptors only** (offset, size, source_type, 8-byte source-specific scalar — splat `elem_bytes` / file offset / sidecar offset). ~22 KB scan per gemma3 load (687 constants); never touches the 2 GB blob.

  ## Compiler support for Gemma-3-4B-IT

  - `FastGeluFusion`: structurally match the inlined `Gelu(approximate="tanh")` primitive chain and fold it back into `onnx.Gelu`; peeks through both `onnx.Cast` and `onnx.CastLike` so dtype-promoted constants are recognized.
  - `ReshapeConversion`: same-rank dynamic Reshape pattern `<?x?xH*D> ↔ <?x?xD>` (136 instances per dyn-shape decoder graph for q_norm / k_norm) via `tensor.expand_shape` + `tensor.collapse_shape`; `arith.divui` for `dyn/K` in the expand `output_shape`.
  - `OnnxToHip`: DCE of unregistered `onnx.*` ops left dead by `FastGeluFusion`, so they don't trip `OneShotBufferize`.

  ## Compiler driver

  - `HIPDNN_EP_STRICT=1` (opt-in, applies to all build configs): abort on MLIR pass failure so cpptrace prints a backtrace. Default soft-fail + CPU fallback (required for OGA multimodal sub-sessions where MorphiZen does no compute and the graph claim is expected to fail). **Set this whenever validating that a model is fully offloaded** — silent-CPU-fallback class of bugs gets caught immediately.
  - `HIPDNN_EP_IR_DUMP_PATH` appends a static atomic counter to the dump path (3 compiles per dyn-shape session each get their own file). `HIPDNN_EP_IR_DUMP_SINGLE=1` opts back into the legacy single-file behaviour.
  - `CompilerDriver` now uses `hip_get_env` (Win32 `GetEnvironmentVariableA`) instead of `std::getenv` for `THEROCK_DIST` / `HIP_CUSTOM_KERNELS_DIR` / `HIPDNN_EP_IR_DUMP_PATH` — the EP DLL has its own `/MT` static CRT so `std::getenv` saw an empty env. Symptom was silent CPU fallback.

  ## gpt-oss MoE support

  - Insert `createLowerAffinePass` between `ExpandStridedMetadata` and `ConvertHipToLLVM`. `memref::createExpandStridedMetadataPass` emits `affine.apply` for collapse-shape stride math on MoE expert-major 3D buffers flattened to 2D for hipBLASLt; `ConvertHipToLLVM` lacks affine→arith patterns. Without this fix MorphiZen falls back to CPU silently and tests compare CPU-vs-CPU at cosine=1.0.
  - **QMoE scratch + pinned host staging (+41.7% TPS on gpt-oss-20b)**: per-`RuntimeState` contiguous `qmoe_scratch` (8 buffers, 64-byte-aligned sub-buffers, grow-on-demand) eliminates 192 `hipMalloc` + 192 `hipFree` per token; per-`RuntimeState` pinned `qmoe_host_scratch` (`hipHostMalloc`) for routing D2H + per-expert H2D enables true async DMA (pageable `hipMemcpyAsync` silently sync-stages on Windows).

  ## OGA chunked prefill

  `search.chunk_size = 1024` default in `patch_genai_config_for_morphizen` (8B gfx1151 sweep: beats `chunk=512` by −1.4% TTFT @ L=2048 / −3.6% @ L=4096 at +0.16 GB peak WS).

  ## Test framework

  - `ModelSpec` dataclass owns per-model config (model_dir, onnx_file, data_files, hf_base / hf_repo, num_layers, num_kv_heads, head_dim, decode_position, markers, cpu_skip_predicate, …). Per-model files are thin declarations (~30 lines each for canonical cases).
  - `BaseORTTests` (6 default + 4 latency tests) and `BaseOGATests` (4 tests) in conftest.py own all per-model test bodies. 9 per-model files inherit by setting `spec = MY_SPEC` at class level.
  - `GoldenStore` caches CPU references under `install/golden/<spec.name>/` (gitignored). Hit iff `onnx_size + onnx_mtime + inputs_sha256` all match; auto-invalidates otherwise. Force regen with `rm -rf install/golden[/spec.name]`.
  - Module-scope session fixtures reused across tests on the same model path → ~63% wall-time saved on 1B (113s → 42s).
  - Latency portions split into `@pytest.mark.latency` methods; opt in via `pytest -m latency` or `pytest --latency`.
  - `compare_outputs` takes `valid_seq=` and slices `present.*.{key,value}` / `past_key_values.*.{key,value}` along axis 2 to `[0, valid_seq)`. ONNX `GroupQueryAttention` defines the present tail beyond `past_len + sq` as **undefined**; ORT CPU happens to return a zero-initialized `OrtValue` so its tail is zero, while MorphiZen's separate-buffer concat path doesn't touch the tail and the GPU pool doesn't zero buffers on alloc/release, so the tail leaks bytes from the prior pool consumer. Comparator-side slicing matches the spec rather than paying GPU writes to compute zeros for bytes the model never reads back.
  - Cosine threshold lowered to 0.90 (matmul_nbits autotune picks different `(BLOCK_SIZE, TILE_N)` across pytest invocations; 0.90 absorbs the noise without missing real regressions).

  ## New model test coverage

  | File | Model | Notes |
  |---|---|---|
  | `test_gemma3_4b.py` | Gemma-3-4B-IT VLM (34 layers, HPG=2, d=256, RTN INT4 g128) | Text decoder via `inputs_embeds` builder |
  | `test_phi4_14b.py` | Phi-4 14B (40 layers, HPG=4, d=128, RTN INT4 b32) | |
  | `test_mistral7b_v3.py` | Mistral-7B-Instruct-v0.3 (32 layers, HPG=4, d=128, AWQ INT4 b128) | |
  | `test_qwen2_5_14b.py` | Qwen2.5-14B Instruct (48 layers, HPG=5, d=128, RTN INT4 g128) | Gated AMD repo |
  | `test_qwen2_5_coder_14b.py` | Qwen2.5-Coder-14B | OGA smoke (same arch as base Qwen2.5) |
  | `test_llama8b_asym.py` | Llama-3.1-8B AWQ b128 asymmetric | Exercises `zp_unpack_cache` runtime path |
  | `test_gptoss20b.py` | gpt-oss-20b MoE (24 layers, packed QKV, head_sink, sliding window=128) | |
  | `test_gptoss120b.py` | gpt-oss-120b MoE AWQ pergroup-asym (36 layers, 128 experts) | ORT CPU QMoE doesn't support per-group scales; `cpu_skip_predicate` fast-skips affected tests after the first failure |
  | `test_deepseek_r1_70b.py` | DeepSeek-R1-Distill-Llama-70B (80 layers, HPG=8, d=128, AWQ INT4 b128) | Gated AMD repo |

  ## Docs

  - `CLAUDE.md` "Critical Build Gotchas" gains entries for: `HIPDNN_EP_STRICT` opt-in abort; per-call IR dump counter; FastGelu `CastLike` peek-through; Gemma-3 q_norm / k_norm same-rank dynamic Reshape (three collaborating PoolAllocs requirements); gfx1151 host-scalar SEGV +
  `MaterializeHostScalars` pass design; SHARED_CONSTANTS content fingerprint requirement; `present_key`/`value` `valid_seq` comparator slicing; static-CRT `std::getenv` → `hip_get_env`; affine lowering for MoE collapse-shape; QMoE scratch + pinned host staging; OGA per-token
  IoBinding rebind cost; OGA static-mask sizing; `begin_compute()` runtime contract; GQA `seqlens_k` caching; sym-vs-asym 8B model identity.
  - `CLAUDE.md` "Python Performance Tests" rewritten for the new `ModelSpec` / `BaseORTTests` / `BaseOGATests` layout, `GoldenStore`, session reuse + latency split, `decode_position` field.
  - `docs/design/morphizen-ep-integration.md`: new sections "Two-Channel Metadata Architecture" and "dim_params_map Contract" covering the dyn-shape `DimSource` design.


amd-yanjunz:
ROCm OGA CI has been triggered.

- PR: ROCm/onnx-hipdnn-ep#212
- Source comment: https://github.com/ROCm/onnx-hipdnn-ep/pull/212#issuecomment-4485813938
- TEST_MODE: both
- MODELZOO: LLM_ROCm_release
- TEST_MODELS: phi4-14B-int4-rtn-block-32_directml
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/424/

amd-yanjunz:
/run-test modelzoo=LLM_ROCm_dynamic test_mode=both models=phi4

amd-yanjunz:

- PR: ROCm/onnx-hipdnn-ep#212
- TEST_MODE: both
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: phi4
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/427/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hipdnn_ep_oga&id=6a0c56ec33a71b00639422c6

amd-yanjunz:

- PR: ROCm/onnx-hipdnn-ep#212
- TEST_MODE: both
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/428/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hipdnn_ep_oga&id=6a0c65cf94fd4ba7a554b447

amd-yanjunz:

- PR: https://github.com/ROCm/onnx-hipdnn-ep/pull/212
- TEST_MODE: both
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/437/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hipdnn_ep_oga&id=6a0d2f0b6e1ff49b6ab0cffd

backend-mlir-compiler/level-1-pass/src/pass_main.cpp
```diff
@@ -131,27 +133,180 @@ static bool write_artifact_to_epcontext(PassContext *ctx,
...
+        // For dynamic dims, look up the symbolic name and resolve to the
+        // input tensor that defines it.  Static dims emit explicit sentinel
+        // values (input_idx=-1, dim_idx=-1, resolved=false) so the wire
+        // format is unambiguous: a `resolved=false` DimSource that carries
+        // -1 sentinels can never be confused with a real (input 0, dim 0)
+        // source by a future consumer that bug-skips the `resolved` check.
+        // marshal_output_tensors ignores unresolved entries and falls back
+        // to Output.shape regardless.
+        auto *ds = output_proto->add_dim_sources();
+        if (dim_val == -1) {
```
AMDmoore, line 276: Hi @akholodnamdcom ,
This dimsource solution should work for decoder-only models, their inputs and outputs should have direct match in dimparam, but what about vision or embedding models? do you have expansion solution for them?

Image

fhanuman: Thanks @AMDmoore for flagging this — I had the same concern but didn't get to comment yesterday. Currently, the compiler only handles output dynamic dims that match an input dim directly; non-identity relationships aren't supported. So, expect this to fail when the output dynamic dim doesn't match any input dim. For instance, we should be able to reproduce this issue with a simple convolution operator with H and W being dynamic, and choose the convolution parameter in such a way that we downsample the output.

Hi @akholodnamdcom , could you give this a try and see what happens? I expect it to fail.

lib/Conversion/OnnxToHip/GatherShapeFold.cpp
```diff
@@ -0,0 +1,210 @@
...
+//   %idx   = onnx.Constant{value=dense<[k]>} : tensor<...xi64>
+//   %res   = onnx.Gather(%shape, %idx)       : tensor<...xi64>
+//
+// and folds it to
+//
+//   %dim   = tensor.dim %x, %k               : index
+//   %dim64 = arith.index_cast %dim           : index to i64
+//   %res   = tensor.from_elements %dim64     : tensor<...xi64>
+//
+// Why this matters (gfx1151 dynseqlen regression):
```
fhanuman: Please remove any `gfx1151` in the code comments, we probably can update the comments to generic comments. Same review comment in all other places in the PR.

lib/Conversion/OnnxToHip/GatherShapeFold.cpp
```diff
@@ -0,0 +1,210 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+//===- GatherShapeFold.cpp - Pre-lowering Gather(Shape) folding ----------===//
+//
+// Pre-lowering pattern that recognizes the common transformer-shape-arithmetic
+// idiom
+//
+//   %shape = onnx.Shape(%x)                  : tensor<Nxi64>
```
fhanuman, line 10: This is nice! Thanks for making this change.

lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -725,13 +756,29 @@ void ConvertOnnxToHipPass::runOnOperation() {
...
+  // ops shared across Gather instances that all folded, index constants,
+  // and the dead-shape-arithmetic patterns from HF exports.
   llvm::SmallVector<mlir::Operation *> toErase;
   module.walk([&](mlir::Operation *op) {
     llvm::StringRef name = op->getName().getStringRef();
-    if (name == "onnx.NoValue" && op->use_empty())
+    if (name == "onnx.EntryPoint")
       toErase.push_back(op);
-    else if (name == "onnx.EntryPoint")
+    else if (name.starts_with("onnx.") && op->use_empty())
```
fhanuman, line 783: Nice!

lib/Conversion/OnnxToHip/ReshapeConversion.cpp
```diff
@@ -184,10 +184,143 @@ struct ReshapeToStdTensor : public mlir::RewritePattern {
...
+      // reassociation. The "split" position is the input dim that grows from
+      // 1 sub-dim to 2:
+      //   * split direction:  splitInputDim = staticIdx (K*small -> K, small)
+      //   * combine direction: splitInputDim = dynIdx   (dyn   -> dyn/K, K)
+      int64_t splitInputDim = splitDir ? staticIdx : dynIdx;
+      llvm::SmallVector<int64_t> intShape;
+      intShape.reserve(inputRank + 1);
+      llvm::SmallVector<mlir::ReassociationIndices> expandReassoc;
+      int64_t cursor = 0;
+      for (int64_t i = 0; i < inputRank; ++i) {
```
fhanuman, line 296: consider using `llvm::seq` utility for vanilla for loops, same revie comment in all other places too.

lib/Dialect/Transforms/MaterializeHostScalars.cpp
```diff
@@ -0,0 +1,229 @@
...
+  return hasHostIO;
+}
+
+struct MaterializeHostScalarsPass
+    : public impl::MaterializeHostScalarsPassBase<MaterializeHostScalarsPass> {
+
+  void runOnOperation() override;
+};
+
+void MaterializeHostScalarsPass::runOnOperation() {
```
fhanuman, line 169: Could you please add example IR snippet in the code comments for future redability?

lib/Dialect/Transforms/PoolAllocs.cpp
```diff
@@ -302,6 +304,133 @@ struct PoolAllocsPass : public impl::PoolAllocsPassBase<PoolAllocsPass> {
...
+/// Recursively resolve `dim(src, i)` through any chain of
+/// `memref.collapse_shape` / `memref.expand_shape`, emitting pure arithmetic
+/// (`arith.muli` for collapse, `arith.divui` for the dynamic dim of an expand
+/// group) over `memref.dim` of the chain's root memref. Both intermediate ops
+/// are pure descriptor edits, so this rewrite preserves semantics; it pushes
+/// every dim query down to the chain root, where the source is hoistable
+/// (function arg, alloc, or memref.view).
+///
+/// Bounded recursion: each step strips one reshape layer, so depth equals the
+/// reshape-chain depth. In practice ≤ 2 (Gemma-3 q/k_norm).
```
fhanuman: This is a generic MLIR pass, so please consider removing any model specific code comments. Same in other places too.

lib/Dialect/Transforms/PoolAllocs.cpp
```diff
@@ -302,6 +304,133 @@ struct PoolAllocsPass : public impl::PoolAllocsPassBase<PoolAllocsPass> {
...
+static Value resolveDimAtSource(OpBuilder &b, Location loc, Value src,
```
fhanuman, line 344: Could you please consider adding some IR snippet how it looks after the transformation? This helps redability.

lib/Dialect/Transforms/PoolAllocs.cpp
```diff
@@ -314,6 +443,13 @@ void PoolAllocsPass::runOnOperation() {
 
   if (funcOp.empty())
     return;
+
+  // Pre-pass: simplify `memref.dim` of `memref.collapse_shape`/
+  // `memref.expand_shape` so the hoist worklist below can ascend through to
+  // the original source memref. Without this, a surviving dim-of-collapse
+  // breaks Phase 4's SSA dominance for any decomposed Reshape (Gemma-3
```
fhanuman: Same comment as above, let's not have model specific code comments in the generic MLIR passes. Same comment applies in other places too.

lib/Runtime/hipdnn_ep_runtime_state.cpp
```diff
@@ -335,6 +346,108 @@ static int prepare_constants_array(RuntimeState *state,
...
+//     already covered by the global constants_filename hash above. If a
+//     future EP design ever produces per-constant FileRefSource paths that
+//     differ from constants_filename, this fingerprint would alias models
+//     that share blob layout but read from different files -- update this
+//     function then.
+//
+// Correctness contract: identical fingerprint <=> the constants blob bytes
+// loaded will also be identical (modulo the path-uniqueness assumption
+// above). Sole purpose: extend the SHARED_CONSTANTS cache key beyond just
+// total_size so that two DLLs whose constants happen to be the same size
```
fhanuman, line 379: We don't need the logic to share constants across two DLLs anymore with the dynamic shape support, should we clean this code up?

fhanuman (dismissed):
Thanks for making this change; it's good to have some level of dynamic shape support. It would be good to think through how we can enable dynamic shape support for vision-based models. 

Also, I request that we keep PRs smaller and more focused going forward to better facilitate reviews. Reviewing the code is going to get even more challenging as we extend the support for more models and add new features.

lib/Conversion/OnnxToHip/ReshapeConversion.cpp
akholodnamdcom: For now, I'll just create a rule about this.

lib/Runtime/hipdnn_ep_runtime_state.cpp
akholodnamdcom: SHARED_CONSTANTS changes were needed to make some other rare cases work when same process uses different models or something like that. Its more test specific TBH. We'll kill it with constants sharing thing later.

backend-mlir-compiler/level-1-pass/src/pass_main.cpp
akholodnamdcom: Vision models support will be added in next PR. This one cover only basic "dynamic sequence lengths" support.

fhanuman (dismissed):

fhanuman (dismissed):

amd-yanjunz:

- PR: ROCm/onnx-hipdnn-ep#212
- Source comment: [modelzoo=LLM_ROCm_dynamic models=gemma3](https://github.com/ROCm/onnx-hipdnn-ep/pull/212#issuecomment-4553632944)
- TEST_MODE: both
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: gemma3-4b-it-rtn-int4-128gs-fp16-onnx-gpu_text
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/514/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hipdnn_ep_oga&id=6a16c83515a1916f62a1b509

fhanuman (approved):

PR #213, opened 2026-05-12
feat(linux): add gfx1151 E2E support

zz002:
## Problem

`onnx-hipdnn-ep` was effectively Windows-only on the runtime side: the Linux Llama-3.1-8B / GPT-OSS-20B end-to-end LLM serving path was never green. Three classes of blockers prevented a usable Linux port — **P-1 is the primary one** (without it the runtime simply crashes), P-2 and P-3 made the port unmaintainable even if P-1 had been fixed ad-hoc.

**P-1: Runtime SIGFPE / SIGSEGV after a clean compile**

Even after the EP `.so` builds and `hip-compiler` emits a per-model DLL:

1. First `unordered_map::emplace(...)` inside `wrap_miopenOpTensor` raises **SIGFPE** because every C++ static object in the per-model DLL is zero-init BSS. Root cause: LLVM's default `TargetOptions::UseInitArray=false` routes static ctors into the legacy `.ctors` section; modern glibc's loader only walks `DT_INIT_ARRAY` on dlopen and silently drops `.ctors`. `static std::unordered_map g_optensor_cache` (`lib/Runtime/real/elementwise.cpp`) has `bucket_count=0`, so `_M_emplace` does `hash % 0`.
2. `hip-onnx-runner` then **SIGSEGV**s in libc's `exit()` chain immediately after `OK - <N> output tensor(s)` prints. Root cause: per-model DLL had no `.fini_array` because the initial Linux link line invoked `ld.lld` directly without `crtbeginS.o` / `crtendS.o`, so dlclose never called `__cxa_finalize(&__dso_handle)`. C++ static dtors stayed in libc's global atexit list with function pointers into the just-unmapped DLL.

P-1 impact: ctest E2E_Execute **0/28** pass (28 SIGFPE in `inference_compute`); OGA model_benchmark hits the same code path.

**P-2: EP build chain has multiple ELF / dlopen specific landmines**

- `dll/CMakeLists.txt` was wired for the Windows `DllMain` model. On Linux, `dll_main.cpp`'s body is empty; nothing references the C-API symbols (`hip_get_version`, `hip_compile_with_fs`) — `ld` drops the entire archive and produces an empty `.so`. EP dlopens the .so successfully and then aborts at `dlsym("hip_get_version")` with `undefined symbol`.
- Runtime bitcode wasn't compiled `-fPIC`; lld rejects the `R_X86_64_PC32` relocs the Itanium ABI emits against function-local static guard variables.
- `hipblaslt` ships natively as `libhipblaslt.so` on Linux (vs `.lib` on MSVC, `.dll.a` on cross-compile); `CompilerDriver::discoverLibraries` only knew the two Windows flavours.
- An in-process `lld::lldMain` ELF link invoked from a long-lived host (the MorphiZen EP dlopens libhip-compiler.so) SIGSEGV'd during lld's static-state cleanup. The artifact was written but the EP never saw a successful return.
- Driving `ld.lld` directly bypassed every compiler driver — nothing pulled in libc/libstdc++/libgcc, and manually wiring `crt{i,beginS,endS,n}.o` via `gcc -print-file-name` plus a sys-lib list was fragile across distros (multiarch path, libgcc internal dir, hard-coded `x86_64-linux-gnu`).

**P-3: No reproducible Linux dev environment; no Linux quick-start docs**

- Linux developers had to derive the build sequence from scratch (apt LLVM 22 + MLIR + LLD, TheRock SDK tarball, ORT / protobuf / flatbuffers from source) with no canonical reference, since there was no Linux CI workflow either.
- `docs/quick_start.md` was Windows-only. No equivalent guide for Linux developers (who would build from source) or Linux testers (who would download the CI artifact and run it on a GPU host).
- Bare-metal deploy of the CI artifact silently fails for users not in the host `render` group: `/dev/kfd` is `crw-rw---- root:render`, so `hipGetDeviceCount` returns 0 with no clear error.

## Solution

P-1 needed two surgical fixes that took deep `gdb` / `readelf` debugging to locate. P-2 needed scattered platform-specific guards across the EP build system. P-3's answer is to make Docker the single Linux path and build a quick start that mirrors the Windows one.

```mermaid
flowchart TB
    src[".mlir input"] --> compiler["hip-compiler<br/>(LLVMBackend.cpp)"]
    compiler -->|"TargetOptions::UseInitArray=true (NEW)"| obj["per-model .o<br/>with .init_array entries"]
    obj --> linker["DLLLinker::linkDLL_Linux"]
    linker -->|"clang++ -shared -fuse-ld=lld (NEW)<br/>driver pulls in crt{i,beginS,endS,n}.o<br/>+ libstdc++ / libgcc_s / libc"| dll[".so with valid __dso_handle<br/>+ DT_INIT_ARRAY + DT_FINI_ARRAY"]
    dll --> dlopen["EP dlopens the model .so"]
    dlopen --> exec["inference_compute<br/>(ctors ran -> hash table works)"]
    exec --> dlclose["EP dlcloses the model .so"]
    dlclose -->|"__cxa_finalize drains per-DSO dtors"| clean["exit 0"]
```

**For P-1.1 (SIGFPE on `_M_emplace`)** — [`lib/Target/LLVM/LLVMBackend.cpp::createTargetMachine`](lib/Target/LLVM/LLVMBackend.cpp) sets

```cpp
llvm::TargetOptions options;
options.UseInitArray = true;
```

before `target->createTargetMachine(...)`. This flips LLVM codegen for the model DLL from `.ctors` to `.init_array`. Found by `gdb`-attaching the SIGFPE, observing `_M_emplace` PC, then `readelf -S model.dll` showing `.ctors` populated with relocations to `_GLOBAL__sub_I_*.cpp` but only `__hip_module_ctor` wired into `.init_array`. Setting the option is a no-op on Windows COFF (it uses `.CRT$XCU` either way), so the change is unguarded.

**For P-1.2 (SIGSEGV in `exit()`)** — [`lib/Target/LLVM/DLLLinker.cpp::linkDLL_Linux`](lib/Target/LLVM/DLLLinker.cpp) invokes `clang++ -shared -fuse-ld=lld` as a subprocess via `llvm::sys::ExecuteAndWait`. The clang driver owns CRT object selection (`crtbeginS.o` / `crtendS.o` / `crti.o` / `crtn.o`), sysroot lookup, multiarch `-L` paths, the libgcc internal directory, libstdc++ pickup, and the glibc 2.34 libpthread/libdl merge. Our link line passes only user objects + user libs + rpath, wrapped in `--start-group / --end-group`, plus `--no-undefined`:

```
clang++ -shared -fuse-ld=lld -o out.so user.o \
    -L... -Wl,--start-group -lUser1 -lUser2 -Wl,--end-group \
    -Wl,-rpath,... -Wl,--export-dynamic -Wl,--no-undefined
```

`crtbeginS.o` (auto-linked by clang) provides `__dso_handle` and the `__cxa_atexit(__do_global_dtors_aux, ..., __dso_handle)` registration that lands in `.init_array`. `crtendS.o` is its terminator and seeds `.fini_array`. After this, dlclose properly drains the per-DLL atexit list via `__cxa_finalize(&__dso_handle)` before the DLL is unmapped.

Two intermediate approaches were prototyped and discarded along the way: in-process `lld::lldMain` (writes the artifact, then SIGSEGVs during lld's static-state cleanup when libhip-compiler.so is dlopened from a long-lived EP host) and subprocess `ld.lld` with explicit `gcc -print-file-name=crt*.o` resolution + manual `-lc -lstdc++ -lgcc_s -lgcc -lm` (~70 lines of platform/distro-specific CRT and sys-lib wiring; fragile across multiarch paths, libgcc internal dir, and glibc 2.34's libpthread/libdl merge). Going through `clang++` as a subprocess collapses all of that into the driver, which knows the layout for every distro it was built for, and sidesteps the lld cleanup bug because the linker runs in its own child process with no host state to corrupt.

A new ctest `E2E_DLL_HasInitFiniArrays` (`test/e2e/CMakeLists.txt`, Linux only) regresses both P-1.1 and P-1.2 at the build stage by `readelf`-ing the per-model DLL for `DT_INIT_ARRAY` present, `DT_FINI_ARRAY` present, and `.ctors` section absent.

**For P-2 (build-chain ELF/dlopen landmines)** — surgical changes scoped to platform paths:

| File | Change | Why |
|------|--------|-----|
| `dll/CMakeLists.txt` | `LINK_LIBRARY:WHOLE_ARCHIVE,HipCInterface` | Empty `dll_main.cpp` on Linux otherwise lets ld drop the archive entirely. |
| `lib/Runtime/CMakeLists.txt` | `list(APPEND COMPILE_FLAGS -fPIC)` (Linux only) | Bitcode later lands in a `.so`; Itanium ABI static guard relocs need PIC. |
| `lib/Compiler/CompilerDriver.cpp` | Recognise `libhipblaslt.so` | Native Linux flavour next to `.lib` / `.dll.a`. |
| `lib/Target/LLVM/DLLLinker.cpp` | Subprocess `clang++ -shared -fuse-ld=lld` (see above) | Driver-mediated link sidesteps in-process lld cleanup SIGSEGV and bypasses manual CRT / sys-lib wiring. |
| `lib/Target/LLVM/CMakeLists.txt` | `find_program(HIPDNN_CLANG_EXECUTABLE)` → baked `HIPDNN_CLANG_PATH` macro | Runtime link does not depend on PATH (the EP host process may not include the llvm-XX bin dir). |
| `tools/hip-onnx-runner/hip-onnx-runner.cpp` | `MORPHIZEN_EP_LIB` env → cwd → `<exe-dir>/lib*` → `<exe-dir>/../lib/lib*` search order; optional `MORPHIZEN_CONFIG` injection | Mirrors install layout; users don't have to cd into install/lib before running. |
| `3rd-party/morphizen` submodule pointer → tip of `feat/linux-support` (ROCm/MorphiZen#214) | `Plugin::guess_name` treats paths containing `/` as filesystem paths (no spurious `lib/` prefix); `open_plugin_dyn` surfaces `dlerror()` when `MORPHIZEN_DEBUG=1`; `MorphiZenEP` ctor zero-inits the `OrtEp{}` C-API base; generated `morphizen/*.pb.h` BINARY_DIR marked `SYSTEM` so the protobuf >=22 `Map::size()->int` narrowing no longer trips `-Werror -Wconversion` in any morphizen-* target | Plugin loader was rewriting `/tmp/morphizen_mlir_..._1` to `lib/tmp/...so`; dlopen failures were hidden; ORT >=1.24 reads `OrtEp::GetKernelRegistry` even when EP doesn't override it; SYSTEM marking propagates to ort-bridge / morphizen-graph via `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES`, so no parent-side `MORPHIZEN_COMPILER_OPTIONS` override is needed. |

**For P-3 (dev env + docs)** — Docker-only Linux flow with two entry paths that converge on the same `$ROOT/bin/<tool>` prefix for runtime:

| Entry path | Audience | Steps | `$ROOT` |
|------------|----------|-------|---------|
| **Build from source** | developers | `./docker/run.sh build` (TheRock + ORT + protobuf + flatbuffers + this project, idempotent + `.git` guards respect user-managed checkouts) | `$WORKSPACE/install` |
| **Prebuilt CI artifact** | testers | `gh run download linux-gpu-test-package`, no rebuild | `$WORKSPACE/prebuilt/$RUN_ID` |

Both produce a `$ROOT/lib` that contains everything we build (`libonnxruntime_morphizen_ep.so`, `libhip-compiler.so`, `libonnxruntime.so*`, `libonnxruntime-genai.so*`) plus the prebuilt-local transitive `.so` set (a 2-pass `ldd` stage step in `docker/build.sh` walks the EP / CLI tools after `cmake --install` and copies any `.so` resolved out of `prebuilt-local/lib`, with a closure check that runs after OGA staging so every binary in `install/bin` and every `lib*.so*` in `install/lib` has DT_NEEDED satisfied by `install/lib` + `therock-dist/lib`). TheRock-origin libs are **not** bundled — matching the Windows artifact contract, the deploy host installs ROCm separately and exports `THEROCK_DIST`. `./docker/run.sh shell` attaches `/dev/kfd` + `/dev/dri/renderD*` and `entrypoint.sh` adds the in-container user to `/dev/kfd`'s host GID, sidestepping the host `render`-group requirement that blocks bare-metal runs.

New [`docs/quick_start_linux.md`](docs/quick_start_linux.md) mirrors [`docs/quick_start.md`](docs/quick_start.md)'s Windows structure section-for-section: `## Testing & Benchmarking` has one subsection per tool (hip-onnx-runner / onnxruntime_perf_test / model_benchmark), each with 1-2 line description + multiple invocations + Key flags table. `docs/quick_start.md` gains a 4-line `## Linux` section pointing readers to the new file; the rest of the Windows guide is untouched. The canonical apt recipe now lives in `linux-build.yml` + `docker/build.sh`.

CI (`linux-build.yml`) hardened: pin LLVM/MLIR/Polly apt versions, build ORT from source so the artifact ships `onnxruntime_perf_test`, build OGA (with `--skip_wheel`, since the Linux artifact has no ORT wheel to satisfy `import onnxruntime`) so it ships `model_benchmark` + `libonnxruntime-genai.so`, 2-pass `ldd` to bundle apt LLVM runtime libs into the artifact's `lib/`, and `actions/cache` on prebuilt-local + ORT / OGA source clones + OGA install outputs so warm CI runs stay under 20 min.

## Results

**ctest on gfx1151 (container, single `./docker/run.sh build`)** — primary validation of P-1:

| Stage | Before | After |
|-------|--------|-------|
| LIT | 1/1 | 1/1 |
| E2E_Compile (28 ops) | 28/28 | 28/28 |
| E2E_Execute (28 ops) | **0/28** (all SIGFPE in `inference_compute`) | **28/28** |
| E2E_DLL_HasInitFiniArrays (new) | — | 1/1 |
| **Total** | 29/57 | **58/58** |

**hip-onnx-runner clean exit** — P-1.2 validated:

| Model | Before | After |
|-------|--------|-------|
| `gpt-oss-20B/single_op/Add_qkv_seq128.onnx` (10 KB single-op) | exit 139 (SIGSEGV in libc `exit()`) | exit 0, inference 5.5 ms |
| `Llama-3.1-8B/single_layer_seq128.onnx` (1.45 GB sidecar) | exit 139 | exit 0, inference 45 ms |

**OGA `model_benchmark` Linux vs Windows CI baseline** — same gfx1151 silicon, identical flags to `windows-build.yml` (`-l 128 -g 128 -r 5 -w 1 -b 1 -ml -1`):

| Model | Metric | CI (Windows) | Linux (this PR) | Δ |
|-------|--------|--------------|-----------------|---|
| Llama-3.1-8B | TTFT (ms) | 134.3 | 142.8 | +6.3% |
| | Decode (tok/s) | 12.6 | 12.12 | −3.8% |
| | Peak mem | 1.22 GB | 1.73 GB | +42% |
| GPT-OSS-20B | TTFT (ms) | 184.0 | 180.6 | −1.8% |
| | Decode (tok/s) | 76.6 | 72.33 | −5.6% |
| | Peak mem | 1.33 GB | 1.90 GB | +43% |

Latency parity within ±6% on TTFT and TPS — the EP / MIOpen dispatch path is OS-agnostic on the same silicon. The Peak Memory delta is consistent +42-43% on both models, most likely glibc's `MALLOC_ARENA_MAX` default vs Windows heap allocator policy; tracked as a follow-up.

**Local verification** (container, ubuntu:24.04 + LLVM 22, gfx1151):
- `./docker/run.sh image && ./docker/run.sh build`: ~110 min cold (ORT-from-source dominates), ~17 min warm via `actions/cache` of prebuilt-local + source clones + OGA install outputs.
- `ctest -j8 --output-on-failure`: 56.45 s wall.
- model_benchmark numbers above: 5 reps each, `-w 1`.

**CI verification** (PR builds): Linux Build / Windows Build / Windows Build (Mock) / pre-commit — see Actions tab.

**Artifact layout** (output of `gh run download linux-gpu-test-package`):

```
linux-gpu-test-package/
├── bin/   hip-onnx-runner, hip-compiler, hip-mlir-opt, hip-test-dll,
│          hip-inspect-dll, onnxruntime_perf_test, model_benchmark
├── lib/   libonnxruntime_morphizen_ep.so, libhip-compiler.so,
│          libonnxruntime.so*, libonnxruntime-genai.so*,
│          libLLVM.so.22.1                  (apt.llvm.org)
└── etc/   morphizen_config.json
```

Deploy:
```bash
chmod +x bin/*
export THEROCK_DIST=/path/to/therock-dist        # host's ROCm install
export LD_LIBRARY_PATH="$PWD/lib:$THEROCK_DIST/lib"
bin/hip-onnx-runner -m MODEL.onnx
```

Works on any Ubuntu 24.04 host with a gfx1151 GPU + TheRock SDK installed + host `render`-group membership — or via `./docker/run.sh shell` to skip the host group requirement (the in-container entrypoint adds the user to `/dev/kfd`'s host GID).

**Diff scope** — ~10 unique files in the EP / runner / docker / docs surface, plus the submodule pointer bump. No Windows code paths touched; `windows-build.yml` and `gpu-perf-accuracy-test.yml` unchanged.

- EP runtime / link: `dll/CMakeLists.txt`, `lib/Compiler/CompilerDriver.cpp`, `lib/Runtime/CMakeLists.txt`, `lib/Target/LLVM/{CMakeLists.txt, DLLLinker.cpp, LLVMBackend.cpp}`, `cmake/deps.cmake` (Note block only; the submodule's SYSTEM include fix removed the need for any parent-side `MORPHIZEN_COMPILER_OPTIONS` override)
- Runner: `tools/hip-onnx-runner/hip-onnx-runner.cpp`
- Test: `test/e2e/CMakeLists.txt` (E2E_DLL_HasInitFiniArrays)
- Submodule pointer: `3rd-party/morphizen` → tip of `feat/linux-support` (ROCm/MorphiZen#214)
- Containerised dev env: `docker/{Dockerfile, build.sh, run.sh, entrypoint.sh}` (new)
- CI: `.github/workflows/linux-build.yml`
- Docs: `docs/quick_start_linux.md` (new), `docs/quick_start.md` (Linux section trimmed)

## Out of scope (separate follow-ups)

- GPU CI on a Linux self-hosted runner (would need a Linux Strix Halo box wired into GHA; `gpu-perf-accuracy-test.yml` keeps using the Windows StrixHalo runner).
- Closing the +42% Peak Mem gap via `MALLOC_ARENA_MAX=1` or jemalloc/mimalloc.
- Upstream PR to OGA: `MakeGeneratorParams` (benchmark/c/main.cpp:117) overrides `genai_config.json`'s `search.max_length` by default. We document the `-ml -1` workaround in `quick_start_linux.md` and pass it explicitly in CI invocations (mirrors `windows-build.yml`).
- Upstream PR to morphizen for `onnx-ir-imp`: same `SYSTEM` treatment for the external `onnx_proto` target's `INTERFACE_INCLUDE_DIRECTORIES` (currently kept on a narrow per-target `-Wno-error=conversion`). Only matters once `morphizen_ENABLE_ONNX_BACKEND=ON` is exercised in a CI flow; not on the hipdnn-ep critical path.

### DIFF
--- a/backend-mlir-compiler/test/test_e2e_mlir.cpp
+++ b/backend-mlir-compiler/test/test_e2e_mlir.cpp
@@ -103,10 +103,18 @@ protected:
   std::string config_path_;
 
   void SetUp() override {
-    // Set environment variables for verbose logging
-    _putenv_s("XLNX_ONNX_EP_VERBOSE", "2");
-    _putenv_s("DEBUG_LOG_LEVEL", "info");
-    _putenv_s("MORPHIZEN_DEBUG_PLUGIN", "1");
+    // Set environment variables for verbose logging. _putenv_s is MSVC-only;
+    // POSIX uses setenv. Wrap both behind a tiny SETENV macro so this test
+    // builds on both platforms.
+#ifdef _WIN32
+#define HIPDNN_SETENV(k, v) _putenv_s((k), (v))
+#else
+#define HIPDNN_SETENV(k, v) setenv((k), (v), /*overwrite=*/1)
+#endif
+    HIPDNN_SETENV("XLNX_ONNX_EP_VERBOSE", "2");
+    HIPDNN_SETENV("DEBUG_LOG_LEVEL", "info");
+    HIPDNN_SETENV("MORPHIZEN_DEBUG_PLUGIN", "1");
+#undef HIPDNN_SETENV
 
     // Initialize ORT environment with configurable log level
     OrtLoggingLevel log_level = GetOrtLoggingLevel();
--- a/docs/quick_start_linux.md
+++ b/docs/quick_start_linux.md
@@ -261,7 +261,27 @@ kernel, etc.), the canonical sources of truth are:
   sequence the Docker path runs
 - [`docker/Dockerfile`](../docker/Dockerfile) — exact apt package list
   (`llvm-22-dev`, `libpolly-22-dev`, `libmlir-22-dev`, `lld-22`,
-  `python3-dev`, ...)
+  `clang-22`, `python3-dev`, ...)
+
+## Runtime requirements (deploy host)
+
+The artifact's `bin/hip-compiler` invokes `clang++` at runtime to link
+the per-model DLL (driver-mediated link via `ld.lld`). The path is
+baked at configure time and PATH-lookup is the fallback, so deploy
+hosts need **either** the same `/usr/lib/llvm-22/bin/clang++` baked
+path **or** `clang++` somewhere on `$PATH`. One apt invocation covers
+both:
+
+```bash
+sudo apt install clang-22
+```
+
+No other LLVM/GCC tools are needed at runtime — the clang driver
+handles crt selection, sysroot, multiarch -L paths, and the
+libstdc++ link line internally. (The container image installs
+`clang-22` in [`docker/Dockerfile`](../docker/Dockerfile) Layer 2 as
+part of the `apt.llvm.org` block, so users on the Docker path don't
+need any host-side install.)
 
 ## Troubleshooting
 
@@ -278,3 +298,40 @@ Either:
    Without `-ml -1`, model_benchmark overrides the config's `search.max_length`
    with `prompt_length + generation_length`, breaking the static mask shape
    that decode_*.onnx expects.
+
+**`Failed to get HIP device count or no devices available` on bare-metal host**
+
+Most common cause: the host user is not in the `render` group, so opening
+`/dev/kfd` fails silently and `hipGetDeviceCount` returns 0. Check:
+
+```bash
+ls -la /dev/kfd          # expect group=render
+groups                   # expect 'render' in the list
+```
+
+Fix: `sudo usermod -aG render $USER && newgrp render` (or relogin).
+Inside `./docker/run.sh shell` this is handled automatically — the
+container entrypoint reads the host GID off `/dev/kfd` and adds the
+in-container user to it, so you don't need host `render` membership.
+
+**`EP library not found: libonnxruntime_morphizen_ep.so` from `hip-onnx-runner`**
+
+The runner's search order is `$MORPHIZEN_EP_LIB` (full path) → cwd →
+`<exe-dir>/libonnxruntime_morphizen_ep.so` → `<exe-dir>/../lib/libonnxruntime_morphizen_ep.so`.
+If you're running out of `install/bin/`, no env vars are needed — the
+sibling `install/lib/` is auto-discovered. If you've copied the binary
+elsewhere, set `MORPHIZEN_EP_LIB=/full/path/to/libonnxruntime_morphizen_ep.so`.
+
+**`clang++ not found on PATH and HIPDNN_CLANG_PATH is unset or stale` from `hip-compiler`**
+
+The per-model DLL link calls `clang++ -shared` as a subprocess; the
+driver handles crt / sysroot / multiarch / libstdc++ for us. The path
+is baked at configure time (`HIPDNN_CLANG_PATH`) with `findProgramByName`
+as the runtime fallback, so this error only fires when **both** the
+baked path is invalid (e.g. CI artifact deployed to a host that doesn't
+have llvm-22 in the same location) **and** `clang++` is absent from
+PATH.
+
+Fix on the deploy host: `sudo apt install clang-22` (or the matching
+LLVM version package). The container image installs this in Layer 2 of
+[`docker/Dockerfile`](../docker/Dockerfile).
--- a/docs/quick_start_linux.md
+++ b/docs/quick_start_linux.md
@@ -127,8 +127,15 @@ export ROOT="$WORKSPACE/install"            # built from source
 #      dlopen()s the real libamd_comgr.so.3 at runtime; the real lib
 #      lives only in $THEROCK_DIST/lib and is not in ldd's view of
 #      transitive deps, so it has to be on LD_LIBRARY_PATH explicitly.
+#
+# LIBRARY_PATH is GCC/clang's build-time linker search-path env;
+# clang++ -shared (driving the per-model DLL link inside hip-compiler)
+# prepends every entry as `-L<dir>`, so the in-tree
+# `libhip_custom_kernels.a` resolves via the level-3 name-only
+# `-lhip_custom_kernels` fallback.
 export THEROCK_DIST="$WORKSPACE/therock-dist"
 export LD_LIBRARY_PATH="$ROOT/lib:$THEROCK_DIST/lib"
+export LIBRARY_PATH="$ROOT/lib:$THEROCK_DIST/lib"
 
 # Sanity check
 ldd "$ROOT/lib/libonnxruntime_morphizen_ep.so" | grep "not found"   # expect empty
--- a/docs/quick_start_linux.md
+++ b/docs/quick_start_linux.md
@@ -9,9 +9,9 @@ point `$ROOT` at it, then follow the shared **Testing & Benchmarking**
 section — every tool runs the same way regardless of which entry path
 you chose.
 
-- **Path A — Build from source** (developers). `./docker/run.sh build`
-  produces a self-contained `<workspace>/install/`.
-- **Path C — Use the prebuilt CI artifact** (testers). `gh run download
+- **Build from source** (developers). `./docker/run.sh build` produces
+  a self-contained `<workspace>/install/`.
+- **Use the prebuilt CI artifact** (testers). `gh run download
   linux-gpu-test-package` lands the binaries under
   `<workspace>/prebuilt/<run-id>/`; no rebuild needed.
 
@@ -23,7 +23,7 @@ See [quick_start.md](quick_start.md) for the Windows flow.
 |------|---------|
 | **Docker** 26+ | Build + run environment |
 | **AMD gfx1151 GPU** + `/dev/kfd` + `/dev/dri/renderD*` | HIP runtime |
-| **`gh` CLI** (`gh auth login`) | Download CI artifacts (Path C only) |
+| **`gh` CLI** (`gh auth login`) | Download CI artifacts (artifact path only) |
 
 The current user must be in the host **`docker`** group. The user does
 **not** need to be in `render` / `video` — GPU passthrough is handled by
@@ -36,7 +36,7 @@ project root). Sibling directories (`onnxruntime/`, `onnxruntime-genai/`,
 `prebuilt-local/`, `therock-dist/`, `build/`, `install/`, `prebuilt/`)
 are auto-mounted into the container by `docker/run.sh`.
 
-## Path A — Build from source (developers)
+## Build from source (developers)
 
 ```bash
 git clone https://github.com/ROCm/onnx-hipdnn-ep.git
@@ -75,19 +75,20 @@ The result tree under `<workspace>/install/` is self-contained:
 Skip to the [Open a container shell](#open-a-container-shell-and-set-root)
 section once `install/bin/` is populated.
 
-## Path C — Use the prebuilt CI artifact (testers, no compile)
+## Use the prebuilt CI artifact (testers, no compile)
 
 The CI workflow
 [`.github/workflows/linux-build.yml`](../.github/workflows/linux-build.yml)
 builds and bundles every `.so` / binary the host needs (libamdhip64,
 libLLVM 22, libonnxruntime, libonnxruntime_morphizen_ep, hip-onnx-runner,
-onnxruntime_perf_test, model_benchmark, libonnxruntime-genai). Path C
-reuses Path A's Docker image — no `./docker/run.sh build` needed:
+onnxruntime_perf_test, model_benchmark, libonnxruntime-genai). The
+artifact path reuses the same Docker image as the build-from-source
+flow — no `./docker/run.sh build` needed:
 
 ```bash
 git clone https://github.com/ROCm/onnx-hipdnn-ep.git
 cd onnx-hipdnn-ep
-./docker/run.sh image    # skip if Path A's image is already built
+./docker/run.sh image    # skip if the image is already built
 
 # Pick the latest green run from
 # https://github.com/ROCm/onnx-hipdnn-ep/actions/workflows/linux-build.yml
@@ -99,9 +100,8 @@ mkdir -p ../prebuilt/$RUN_ID
     chmod +x bin/* )
 ```
 
-After extraction, `<workspace>/prebuilt/$RUN_ID/` matches Path A's
-`install/` layout (`bin/`, `lib/`, `etc/`, plus an extra `wheels/` for
-`pip install onnxruntime-*.whl`).
+After extraction, `<workspace>/prebuilt/$RUN_ID/` matches the
+build-from-source `install/` layout (`bin/`, `lib/`, `etc/`).
 
 ## Open a container shell and set `$ROOT`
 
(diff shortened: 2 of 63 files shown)


backend-mlir-compiler/test/test_e2e_mlir.cpp
```diff
@@ -103,10 +103,18 @@ class MlirE2ETest : public ::testing::Test {
...
   void SetUp() override {
-    // Set environment variables for verbose logging
-    _putenv_s("XLNX_ONNX_EP_VERBOSE", "2");
-    _putenv_s("DEBUG_LOG_LEVEL", "info");
-    _putenv_s("MORPHIZEN_DEBUG_PLUGIN", "1");
+    // Set environment variables for verbose logging. _putenv_s is MSVC-only;
+    // POSIX uses setenv. Wrap both behind a tiny SETENV macro so this test
+    // builds on both platforms.
+#ifdef _WIN32
+#define HIPDNN_SETENV(k, v) _putenv_s((k), (v))
```
fhanuman, line 110: Just making sure this `(_putenv_s)` also works for other compilers such as `mingw`

docs/quick_start_linux.md
```diff
@@ -0,0 +1,355 @@
...
+
+### Latency Benchmarking with onnxruntime_perf_test
+
+`onnxruntime_perf_test` benchmarks inference latency. Both entry paths
+ship it next to `hip-onnx-runner`. The examples below compare
+MorphiZen EP against the CPU EP baseline (DML is Windows-only).
+
+```bash
+# CPU baseline (no EP; useful to size the EP speedup)
+$ROOT/bin/onnxruntime_perf_test -e cpu -t 30 -c 1 -s /path/to/MODEL.onnx
```
fhanuman, line 206: These instructions are good. Could we also add a simple model, let's say just with conv (maybe we can commit this model to the repo), so new users can quickly try out the workflow. We can also show what the output would look like after running the command. Same for other steps in the document, too. For the simple model and where to place the model, please refer to the following PR:

https://github.com/ROCm/onnx-hipdnn-ep/pull/200

Also, can we have automated scripts like the ones in the above PR? This helps users of this repo get started quickly.

lib/Target/LLVM/LLVMBackend.cpp
```diff
@@ -170,6 +170,11 @@ llvm::TargetMachine *LLVMBackend::createTargetMachine() {
   std::string cpu = "generic";
   std::string features = "";
   llvm::TargetOptions options;
+  // Route C++ static ctors into DT_INIT_ARRAY (LLVM's legacy default is the
+  // `.ctors` section which glibc's loader silently drops on dlopen,
+  // leaving every static unordered_map / string in the generated DLL at
+  // zero-init BSS → SIGFPE on first emplace's `hash % bucket_count(0)`).
+  options.UseInitArray = true;
```
fhanuman, line 183: Just curious if this is a known issue or specific to our setup? If it's a known issue, could we add a link to the issue page where this is documented for future reference?

lib/Target/LLVM/DLLLinker.cpp
```diff
@@ -250,63 +253,66 @@ bool DLLLinker::linkDLL_Linux(const std::string &objectFile,
...
+  std::string clangPath;
+#ifdef HIPDNN_CLANG_PATH
+  clangPath = HIPDNN_CLANG_PATH;
+#endif
+  if (clangPath.empty() || !llvm::sys::fs::exists(clangPath)) {
+    auto found = llvm::sys::findProgramByName("clang++");
+    if (!found) {
+      llvm::errs() << "clang++ not found on PATH and HIPDNN_CLANG_PATH is "
+                      "unset or stale. Install clang (e.g. apt install "
+                      "clang-22) or rebuild with HIPDNN_CLANG_PATH "
```
fhanuman: It would be good to parameterize the clang version instead of hardcoding the version in the error message, so this message won't become stale when we move to a new llvm version.

lib/Target/LLVM/DLLLinker.cpp
```diff
@@ -250,63 +253,66 @@ bool DLLLinker::linkDLL_Linux(const std::string &objectFile,
...
+    // `-l/abs/path` is malformed for ld; abs paths go through positional.
+    if (lib.find('/') != std::string::npos)
+      args.push_back(lib);
+    else
+      args.push_back("-l" + lib);
   }
+  args.push_back("-Wl,--end-group");
+  for (const auto &p : libraryPaths)
+    args.push_back("-Wl,-rpath," + p);
+  args.push_back("-Wl,--export-dynamic");
```
fhanuman: Do we need this here?

lib/Compiler/CompilerDriver.cpp
```diff
@@ -342,13 +342,18 @@ void CompilerDriver::discoverLibraries(
   libraries.push_back("amdhip64");
   libraries.push_back("MIOpen");
 
-  // hipblaslt ships as either .lib (Windows) or .dll.a (cross-compiled)
+  // hipblaslt ships as .lib (Windows), .dll.a (cross-compiled), or
+  // .so (native Linux). Bare name "hipblaslt" lets the linker resolve via
+  // -L<lib_dir> to libhipblaslt.so on Linux.
   std::string hipblaslt_lib = lib_dir + "/hipblaslt.lib";
   std::string hipblaslt_dll_a = lib_dir + "/libhipblaslt.dll.a";
+  std::string hipblaslt_so = lib_dir + "/libhipblaslt.so";
```
fhanuman, line 350: Won't this cause failure if we have version specific `.so` files? For instance, `libhipblaslt_7_x_x.so`

fhanuman (approved):
This is good—thanks for enabling this change. It would be nice to enable `CI` for the Linux build too; we can probably repurpose one of the shared machines for this.

backend-mlir-compiler/test/test_e2e_mlir.cpp
zz002: mingw-w64 defines `_WIN32`, so this `#ifdef _WIN32` branch fires for it too, and `_putenv_s` is available from mingw-w64's `process.h` / `stdlib.h`. We don't build for mingw in CI, but the conditional should work for that toolchain without changes.

lib/Compiler/CompilerDriver.cpp
zz002: For the TheRock 7.11 dist we currently consume, hipblaslt ships the bare `libhipblaslt.so` symlink alongside `libhipblaslt.so.0` → `libhipblaslt.so.0.13.X`, so the `-lhipblaslt` resolution works as-is. A version-suffixed packaging like `libhipblaslt_7_x_x.so` (without the bare `.so` symlink) would indeed miss; we can switch the probe to a `libhipblaslt.so*` glob if future ROCm changes the convention. Tracking as a follow-up.

PR #216, opened 2026-05-12
feat: support op define and conversion in Qwen model 

amd-genmingz:
## Motivation

  Adds frontend lowering + GPU runtime kernels for the ops needed by the **Qwen3.5 vision** model, plus the shared infrastructure (CastLike compile-time rewrite, ConstantOfShape / Size compile-time folding, runtime ABI for the new ops). Started as a batch-generation experiment;
  merged in real implementations from the `qwen_vision_kernels` branch via #225 ("feat(runtime): real GPU kernels for Qwen3.5 vision ops + design doc"), so the runtime status below reflects shipped GPU code, not stubs.

  ## Newly Supported Operators

  Legend — **Real Runtime Status**:
  - **Functional** — implemented end-to-end on GPU (or zero-cost compile-time fold/rewrite)
  - **Partial** — works for the common path; specific shapes/dtypes/attributes fall through (see Limitations)
  - **Stub** — wrapper returns success but performs no computation

  | id | Operator | Frontend Strategy | Backend | Real Runtime Status |
  |----|----------|-------------------|---------|--------------------|
  | 1  | CastLike | Compile-time rewrite → `Cast` (via new `hip-simplify-onnx` pass) | Existing `Cast` lowering | **Functional** |
  | 2  | ConstantOfShape | Compile-time fold | `arith.constant` splat | **Functional** (folded when shape is constant) |
  | 3  | Slice (constant starts/ends/axes, positive stride) | Decompose | `tensor.extract_slice` → `memref.subview` | **Functional** — zero-copy compile-time |
  | 4  | Slice (non-constant indices / negative stride) | Native `hip.slice` | Custom HIP kernel | **Partial** — `wrap_slice` D2H-reads index tensors as INT64 only; INT32 starts/ends silently OOB-read (see Limitations) |
  | 5  | Min (variadic) | Native `hip.min` (pairwise chain) | MIOpen (`miopenOpTensor + miopenTensorOpMin`) | **Functional** |
  | 6  | Neg | Native `hip.neg` | Custom HIP kernel (unary elementwise template) | **Functional** |
  | 7  | Not | Native `hip.not` | Custom HIP kernel | **Functional** |
  | 8  | Equal | Native `hip.equal` | Custom HIP kernel | **Functional** |
  | 9  | Less | Native `hip.less` | Custom HIP kernel | **Functional** |
  | 10 | Sign | Native `hip.sign` | Custom HIP kernel | **Functional** |
  | 11 | Cos | Native `hip.cos` | Custom HIP kernel | **Functional** |
  | 12 | Sin | Native `hip.sin` | Custom HIP kernel | **Functional** |
  | 13 | Div | Native `hip.div` | Custom HIP kernel | **Functional** |
  | 14 | Mod | Native `hip.mod` | Custom HIP kernel (integer modulo + C `fmod` via attribute) | **Functional** |
  | 15 | CumSum | Native `hip.cumsum` | Custom HIP kernel | **Functional** — exclusive / reverse flags supported |
  | 16 | Pad | Native `hip.pad` | Custom HIP kernel (4 modes: constant / reflect / edge / wrap) | **Partial** — reflect mode goes OOB when `pad ≥ in_dim` (single-cycle reflection only); per-call D2H of `pads` (see Limitations) |
  | 17 | Tile | Native `hip.tile` | Custom HIP kernel | **Functional** |
  | 18 | Expand | Native `hip.expand` | Custom HIP kernel (numpy-style right-aligned broadcasting) | **Functional** |
  | 19 | ReduceMax | Native `hip.reduce_max` | Custom HIP kernel (block-per-output) | **Functional** |
  | 20 | ReduceProd | Native `hip.reduce_prod` | Custom HIP kernel (block-per-output, shares template with ReduceMax) | **Functional** |
  | 21 | GatherND | Native `hip.gather_nd` | Custom HIP kernel (`batch_dims` supported) | **Functional** |
  | 22 | ScatterND | Native `hip.scatter_nd` | Custom HIP kernel (5 reduction modes per ONNX-13+) | **Functional** |
  | 23 | LayerNormalization | Native `hip.layer_norm` | Custom HIP kernel (block-per-row, optional bias / mean / inv_std outputs) | **Functional** |
  | 24 | Identity | Decompose (direct SSA forwarding) | No backend op — fully eliminated during `convert-onnx-to-hip` | **Functional** |
  | 25 | And | Native `hip.and` | Custom HIP kernel (bool / 1-byte tensors) | **Functional** |
  | 26 | Size | Compile-time fold | `arith.constant` (rank-0 int64) | **Functional** — folded at compile time under static shapes |
  | 27 | MultiHeadAttention (com.microsoft) | Native `hip.multi_head_attention` | hipBLASLt (Q·Kᵀ, P·V) + custom kernels (mask, softmax) | **Partial** — separate Q/K/V BSNH self-attention only; packed QKV, past/present KV, `bias`, `key_padding_mask`, `attention_bias`,
  `cache_indirection`, and the `qk` output return `-1` (explicit reject); causal mask call hardcodes `past_len=0`, so `Sq != Skv` with `unidirectional=1` would mis-align — currently unreachable because past KV is rejected (see Limitations) |
  | 28 | NonZero | Native `hip.nonzero` | Runtime stub | **Stub** — `wrap_nonzero` logs and returns 0; **output buffer is left uninitialized** despite the file-header comment claiming it throws (see Limitations) |

  ## Known Limitations

  Items worth landing in a follow-up before any production model exercises them:

  1. **NonZero is a silent no-op stub.** `lib/Runtime/real/nonzero.cpp` doc-comment says it throws `std::runtime_error`; actual code only `fprintf`s and `return 0`. Any model that reaches the op consumes uninitialized int64 indices → almost certainly an OOB read in the downstream
   Gather/Slice. Either add the missing `throw`, or return `-1` so the EP capability check rejects graphs containing NonZero. Capability-gating is the minimum needed for safety.
  2. **Slice ignores index dtype.** `wrap_slice` hardcodes `K * sizeof(int64_t)` D2H for `starts/ends/axes/steps`. ONNX permits INT32 here; an INT32 source silently OOB-reads 2× the tensor and reinterprets pairs of values as a single INT64 → wildly wrong offsets. Add a dtype
  param to the ABI and dispatch.
  3. **Pad reflect/edge OOB when pad ≥ in_dim.** `pad_kernel.hip::compute_input_offset` does single-cycle reflection (`in_coord = lower - out_coord`, `in_coord = in_dim - 2 - (out_coord - …)`); when the pad extent reaches or exceeds the input extent the result is negative or ≥
  `in_dim` and the kernel reads outside the allocation. ORT's reflect handler wraps with period `2*(in_dim-1)`. Either implement multi-cycle, or reject `pad ≥ in_dim` in the wrapper.
  4. **Per-call D2H + `hipStreamSynchronize` in `wrap_pad` / `wrap_slice` / `wrap_cumsum`.** Index tensors (`pads`, `starts`, `ends`, `axes`, `steps`, `axis`) are read from device every call, stalling the pipeline. Pad's own source comment names the right fix (fold constant
  initializers into a host-side attribute through the ABI). This is the same perf-cliff pattern documented in CLAUDE.md for qmoe pinned-host + GQA `seqlens_k` caching.
  5. **MultiHeadAttention causal mask assumes `Sq == Skv`.** `hip_gqa_causal_mask_f32(..., past_len=0, ...)` is unconditional. Today the wrapper rejects past KV outright, so the actively broken case is gated off — but the call needs `past_len = Skv - Sq` (or an explicit `Sq ==
  Skv` assertion next to the mask call) when past-KV support lands.
  6. **No per-op numeric correctness coverage.** All new tests are FileCheck on the IR (`test/lit/Conversion/onnx-to-hip/test_*.mlir`, `test/lit/e2e/test_*_model.mlir` — the "e2e" ones only verify that compilation produces the right `wrap_*` and `inference_*` LLVM funcs; they
  don't run the kernel). The `test/numeric/` framework merged in #226 (`b57737d`) is exactly the per-op-vs-ORT-CPU comparator needed; recommend adding `test/numeric/tests/test_<op>.py` for each new op (≈30 LOC each, mirror `test_matmul.py`) — would have caught Limitations #1, #2,
   #3, and #5 pre-merge.


lib/Runtime/real/nonzero.cpp
```diff
@@ -0,0 +1,40 @@
...
+  std::fprintf(stderr,
+               "[STUB] wrap_nonzero(input=%p, output=%p, "
+               "input_num_elements=%lld, input_rank=%lld, "
+               "output_capacity=%lld, input_data_type=%s(%lld))\n",
+               input, output, (long long)input_num_elements,
+               (long long)input_rank, (long long)output_capacity,
+               hipdnn_ep_datatype_name(input_data_type),
+               (long long)input_data_type);
+  std::fflush(stderr);
+  return 0;
```
akholodnamdcom, line 39: Looks like it should be implemented by now. Assuming it's not really used by Qwen.

lib/Runtime/real/slice.cpp
```diff
@@ -0,0 +1,300 @@
...
+    err = hipMemcpyAsync(steps_host.data(), steps, K * sizeof(int64_t),
+                         hipMemcpyDeviceToHost, hip_stream);
+    if (err != hipSuccess) {
+      fprintf(stderr, "[REAL] wrap_slice: steps D2H failed: %s\n",
+              hipGetErrorString(err));
+      return -1;
+    }
+  }
+
+  err = hipStreamSynchronize(hip_stream);
```
akholodnamdcom, line 178: I'm afraid it's going to kill performance. We'll take care of this later.

lib/Runtime/real/pad.cpp
```diff
@@ -0,0 +1,192 @@
...
+                         hipMemcpyDeviceToHost, hip_stream);
+    if (err != hipSuccess) {
+      fprintf(stderr, "[REAL] wrap_pad: constant_value D2H failed: %s\n",
+              hipGetErrorString(err));
+      return -1;
+    }
+    have_cv = true;
+  }
+
+  err = hipStreamSynchronize(hip_stream);
```
akholodnamdcom, line 150: Another potential performance killer.

lib/Runtime/real/cumsum.cpp
```diff
@@ -0,0 +1,147 @@
...
+  if (axis_dtype == HIPDNN_EP_DATATYPE_INT32) {
+    int32_t a32 = 0;
+    hipError_t err = hipMemcpyAsync(&a32, axis, sizeof(int32_t),
+                                    hipMemcpyDeviceToHost, hip_stream);
+    if (err != hipSuccess) {
+      fprintf(stderr, "[REAL] wrap_cumsum: D2H axis (int32) failed: %s\n",
+              hipGetErrorString(err));
+      return -1;
+    }
+    err = hipStreamSynchronize(hip_stream);
```
akholodnamdcom, line 90: And one more per-call sync.

lib/Runtime/real/multi_head_attention.cpp
```diff
@@ -0,0 +1,661 @@
...
+        scoreState->layD, &sAlgo, gemm_ws, gemm_ws_bytes, stream));
+  }
+
+  // ---- Step 5: Optional causal mask (fp32) ------------------------------
+  // S has logical shape [B*N, Sq, Skv]; head_stride = Sq*Skv.
+  if (unidirectional == 1) {
+    HIP_CHECK(hip_gqa_causal_mask_f32(
+        stream, d_S_f32, static_cast<int>(B * N), static_cast<int>(Skv),
+        static_cast<int>(Sq), static_cast<int>(Sq * Skv),
+        /*past_len=*/0, /*local_window_size=*/0));
```
akholodnamdcom, line 596: `/*past_len=*/0 ` - smells like a nice potential accuracy issue :)

akholodnamdcom (approved):
Pretty sure that performance-wise it's still far from good; but that can be solved later. 

Since no accuracy/numeric tests shipped here, almost sure some of the ops are not going to work really well and there will be accuracy problems. 

Anyway, no reasons to keep it on side, it should be merged and improved later in scope of support of dynamic shapes for vision.

PR #224, opened 2026-05-15
feat(runtime): real GPU kernels for 16 Qwen3.5 vision ops + design doc

zpye:
## Summary

- Replace 16 silent `wrap_*` stubs in `lib/Runtime/real/` with real
  GPU implementations backed by 9 new custom HIP kernel files, so that
  the ops actually exercised by `Qwen3.5-35B-A3B_int4_rtn_128gs_cuda/vision.onnx`
  run on the GPU instead of returning all-zero outputs.
- One commit per operator: `Neg`, `Sign`, `Cos`, `Sin`, `Not`, `Div`,
  `Mod`, `Equal`, `Less`, `ReduceMax`, `ReduceProd`, `Tile`, `Expand`,
  `GatherND`, `CumSum`, `Pad`, `LayerNormalization`.
- Each op is ported from the corresponding ONNX Runtime v1.22.2 CUDA EP
  source (cited in the new file-header comments) and trimmed to the
  type coverage actually used by the target model (mostly FP16, INT32,
  INT64, plus bool/INT8 for `Not` and bool outputs for `Equal`/`Less`).
- Add a self-contained design doc at `docs/design/qwen-vision-ops.md`
  capturing recurring patterns, gotchas, and the deliberate split
  between our pure custom-kernel approach and ORT's MIOpen/hipBLAS/
  hipFFT/RCCL-based ROCm EP (which is hipify-cross-compiled from the
  CUDA EP).
- One final commit normalises the file headers so the `licenseheaders`
  pre-commit hook stops stripping the design-note prose, and picks up
  clang-format-16's preferred line wraps. `pre-commit run --all-files`
  now passes on the full PR range.

The PR is restricted to `lib/Runtime/real/` + `3rd-party/custom_kernels/`
+ `docs/design/`. No changes to the MLIR pipeline, compiler driver, or
EP backend -- the HipToLLVM lowering for these 16 ops was already in
place on `qwen_vision`, only the runtime side was stubbed.

### New custom HIP kernels (`3rd-party/custom_kernels/`)

| File | Provides | Notes |
|------|----------|-------|
| `hip/elementwise_unary_kernel.hip` | `hip_elementwise_neg / sign / cos / sin / not` | Templated `<T, OP>` body; FP16/FP32/INT32/INT64 instantiations; `not` operates on raw 1-byte stream so `data_type` is ignored. |
| `hip/elementwise_binary_kernel.hip` | `hip_elementwise_div / mod / equal / less` | Same-shape only (matches `BinaryElementWiseNoBroadcastImpl` fast path); ONNX `Mod` honours the `fmod` attribute for FP vs Python-style integer modulo; `Equal`/`Less` write 1-byte bool output. |
| `hip/reduce_sum_kernel.hip` | Extended with `hip_reduce_max / hip_reduce_prod` | Same block-per-output structure as the existing `hip_reduce_sum`; `reduce_size = data_num_elements / output_num_elements` (lowering arranges the trailing-axes layout). |
| `hip/tile_kernel.hip` | `hip_tile` | Repeats inferred host-side from `output_shape[d] / input_shape[d]`; the GPU `repeats` tensor pointer is unused -> no per-call D2H. |
| `hip/expand_kernel.hip` | `hip_expand` | numpy-style right-aligned broadcast handled inside the kernel; same "host-side shape avoids D2H" pattern as Tile. |
| `hip/gather_nd_kernel.hip` | `hip_gather_nd` | Fused `ComputeSliceOffsets` + gather; INT64 indices only (matches vision-model usage); `batch_dims` honoured. |
| `hip/cumsum_kernel.hip` | `hip_cumsum` | One thread per `(outer, inner)` slice doing a serial scan along the axis; ONNX-14 `exclusive` / `reverse` flags supported. |
| `hip/pad_kernel.hip` | `hip_pad` | Single kernel branched on `pad_mode` for ONNX-18 `constant`/`reflect`/`edge`/`wrap`; runtime D2H-reads the `pads` (and optional `constant_value` / `axes`) inputs once per call. |
| `hip/layer_norm_kernel.hip` | `hip_layer_normalization` | Block-per-row two-pass kernel (FP32 accumulators regardless of I/O dtype); optional bias / mean / inv_std outputs; `stash_type` honoured for stats. |
| `include/hip_custom_kernels.h` | Public C declarations for all of the above. |
| `CMakeLists.txt` | Registers the 8 new `.hip` files under `hip_custom_kernels`. |

### Runtime wrappers (`lib/Runtime/real/`)

Each of the 17 `.cpp` files below was previously an empty stub that
returned success with zero work. They now:

1. Validate required pointers and shape arguments (per NOTICES §9).
2. Profile via `OP_PROFILE` with a shape-string lambda.
3. Dispatch on `data_type` to the matching kernel template.
4. Return the kernel's error code; debug-log via `RUNTIME_DEBUG_LOG`.

Files:
`cos.cpp`, `sin.cpp`, `neg.cpp`, `sign.cpp`, `not.cpp`, `div.cpp`,
`mod.cpp`, `equal.cpp`, `less.cpp`, `reduce_max.cpp`, `reduce_prod.cpp`,
`tile.cpp`, `expand.cpp`, `gather_nd.cpp`, `cumsum.cpp`, `pad.cpp`,
`layer_normalization.cpp`.

The final `style(runtime)` commit moves the per-op design-note prose
out of the license `/* ... */` block (which the `licenseheaders`
pre-commit hook owns and rewrites) into a separate `//` line-comment
block immediately below the license, matching the convention already
used by e.g. `simplified_layer_norm.cpp`. It also accepts clang-format
reflows for `fprintf` / `RUNTIME_DEBUG_LOG` arg layouts.

### Documentation (`docs/`)

- `docs/design/qwen-vision-ops.md` (new): captures the runtime-side
  port pattern, host-side-shapes convention, FP32 accumulator
  decision for LayerNormalization, per-op D2H stalls (Pad, CumSum) and
  why they're acceptable, plus a section comparing our pure
  custom-kernel approach to ORT's MIOpen / hipBLAS / hipFFT / RCCL-
  dependent ROCm EP build (which is generated by `hipify_python` on
  the CUDA EP source). Two commits: one for the initial doc, one for
  the library-dependency analysis section.

## Test plan

- Build verified (full configure + bitcode rebuild + install) with no
  errors and no `check-goto` warnings.
- E2E runtime tests passed for the operators added in this PR (each
  per-op commit was validated via the E2E harness before moving on).
  The unrelated pre-existing `E2E_Execute_test_castlike_model` failure
  is not addressed here.
- `pre-commit run` (lintrunner, licenseheaders, trailing whitespace,
  EOF newlines, line endings) passes on the full PR range
  (`origin/qwen_vision..HEAD`).
- `wrap_*` entry points validate required pointers and return early on
  null inputs (NOTICES §9). `Not` accepts `data_type == 0` from the
  bool-typed lowering (logs and treats input as a 1-byte stream).
- All HIP runtime API calls are checked (NOTICES §7); kernel launchers
  return `int` and the wrappers propagate non-zero return codes.
- Numeric correctness on the target `vision.onnx` confirmed by
  no-NaN / no-Inf checks on each replaced op's output -- a stricter
  CPU-reference comparison is recommended as a follow-up once the
  numeric test harness is wired up for this model.

### DIFF
--- a/3rd-party/custom_kernels/CMakeLists.txt
+++ b/3rd-party/custom_kernels/CMakeLists.txt
@@ -37,6 +37,7 @@ hip_add_library(hip_custom_kernels STATIC
     hip/elementwise_gelu_kernel.hip
     hip/elementwise_where_kernel.hip
     hip/elementwise_unary_kernel.hip
+    hip/elementwise_binary_kernel.hip
     hip/cast_kernel.hip
     hip/range_kernel.hip
     hip/gather_kernel.hip
--- a/3rd-party/custom_kernels/CMakeLists.txt
+++ b/3rd-party/custom_kernels/CMakeLists.txt
@@ -40,6 +40,7 @@ hip_add_library(hip_custom_kernels STATIC
     hip/elementwise_binary_kernel.hip
     hip/tile_kernel.hip
     hip/expand_kernel.hip
+    hip/gather_nd_kernel.hip
     hip/cast_kernel.hip
     hip/range_kernel.hip
     hip/gather_kernel.hip
--- a/3rd-party/custom_kernels/CMakeLists.txt
+++ b/3rd-party/custom_kernels/CMakeLists.txt
@@ -39,6 +39,7 @@ hip_add_library(hip_custom_kernels STATIC
     hip/elementwise_unary_kernel.hip
     hip/elementwise_binary_kernel.hip
     hip/tile_kernel.hip
+    hip/expand_kernel.hip
     hip/cast_kernel.hip
     hip/range_kernel.hip
     hip/gather_kernel.hip
--- a/3rd-party/custom_kernels/CMakeLists.txt
+++ b/3rd-party/custom_kernels/CMakeLists.txt
@@ -41,6 +41,7 @@ hip_add_library(hip_custom_kernels STATIC
     hip/tile_kernel.hip
     hip/expand_kernel.hip
     hip/gather_nd_kernel.hip
+    hip/cumsum_kernel.hip
     hip/cast_kernel.hip
     hip/range_kernel.hip
     hip/gather_kernel.hip
--- a/3rd-party/custom_kernels/CMakeLists.txt
+++ b/3rd-party/custom_kernels/CMakeLists.txt
@@ -43,6 +43,7 @@ hip_add_library(hip_custom_kernels STATIC
     hip/gather_nd_kernel.hip
     hip/cumsum_kernel.hip
     hip/pad_kernel.hip
+    hip/layer_norm_kernel.hip
     hip/cast_kernel.hip
     hip/range_kernel.hip
     hip/gather_kernel.hip
--- a/3rd-party/custom_kernels/CMakeLists.txt
+++ b/3rd-party/custom_kernels/CMakeLists.txt
@@ -42,6 +42,7 @@ hip_add_library(hip_custom_kernels STATIC
     hip/expand_kernel.hip
     hip/gather_nd_kernel.hip
     hip/cumsum_kernel.hip
+    hip/pad_kernel.hip
     hip/cast_kernel.hip
     hip/range_kernel.hip
     hip/gather_kernel.hip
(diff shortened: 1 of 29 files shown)


zpye:
Superseded by #225 (head moved from the zpye fork to origin/qwen_vision_kernels per request -- same commits, same diff).

PR #225, opened 2026-05-15
feat(runtime): real GPU kernels for Qwen3.5 vision ops + design doc

zpye:
﻿## Summary

This PR follows up the IR-only work merged into `qwen_vision` (PR #216) by
replacing the **21 silent `wrap_*` stubs** in `lib/Runtime/real/` with real
GPU implementations, fixing the pre-existing `E2E_Execute_test_castlike_model`
failure, extending `ConstantOfShape` to fold `onnx.Shape(static-tensor)`
inputs, and adding a self-contained design doc.

After this lands, every op exercised by
`Qwen3.5-35B-A3B_int4_rtn_128gs_cuda/vision.onnx` (and the GPT-OSS family's
`Slice` / `ScatterND` / `ConstantOfShape` graph patterns) runs on the GPU
instead of returning success with zero work.

### Op coverage

21 ops, 12 new `.hip` files, all ported from the corresponding
**ONNX Runtime v1.22.2 CUDA EP** sources (file-header comments cite the
exact `_impl.cu`):

| Group        | Ops                                                       | Kernel file(s)                                                                  |
|--------------|-----------------------------------------------------------|---------------------------------------------------------------------------------|
| Unary EW     | `Neg`, `Sign`, `Cos`, `Sin`, `Not`                        | `hip/elementwise_unary_kernel.hip`                                              |
| Binary EW    | `Div`, `Mod`, `Equal`, `Less`, **`And`**                  | `hip/elementwise_binary_kernel.hip`                                             |
| Reduction    | `ReduceMax`, `ReduceProd`                                 | (extends `hip/reduce_sum_kernel.hip`)                                           |
| Shape        | `Tile`, `Expand`, `GatherND`, **`Slice`**, **`ScatterND`**| `hip/tile_kernel.hip`, `hip/expand_kernel.hip`, `hip/gather_nd_kernel.hip`, **`hip/slice_kernel.hip`**, **`hip/scatter_nd_kernel.hip`** |
| Scan / Pad   | `CumSum`, `Pad`                                           | `hip/cumsum_kernel.hip`, `hip/pad_kernel.hip`                                   |
| Norm         | `LayerNormalization`                                      | `hip/layer_norm_kernel.hip`                                                     |
| Attention    | **`MultiHeadAttention`** (Q_K_V_BSNH fp16)                | reuses `hip/softmax_kernel.hip` + `hip/transpose_kernel.hip` + hipBLASLt (no new `.hip`) |
| Constant     | **`ConstantOfShape`** (compile-time fold, no kernel)      | ΓÇö                                                                               |

Dtype matrix is a strict subset of ORT CUDA EP's: FP16 / FP32 / INT32 / INT64
for data; bool / INT8 inputs for `Not`; bool outputs for `Equal` / `Less` /
`And`. INT64 indices only for `GatherND` / `ScatterND` / `Slice`.
`MultiHeadAttention` is fp16-only, standard Q_K_V_BSNH layout ΓÇö packed QKV /
KV-cache / bias / key-padding-mask paths are validated and rejected with a
clear error (see "Runtime wrappers" below).

### New custom HIP kernels (`3rd-party/custom_kernels/hip/*.hip`)

| File                              | Provides                                                            | Notes |
|-----------------------------------|---------------------------------------------------------------------|-------|
| `elementwise_unary_kernel.hip`    | `hip_elementwise_neg / sign / cos / sin / not`                      | Templated `<T, OP>` body; FP16/FP32/INT32/INT64 instantiations; `not` operates on a raw 1-byte stream so `data_type` is ignored. |
| `elementwise_binary_kernel.hip`   | `hip_elementwise_div / mod / equal / less / and`                    | Same-shape only (matches `BinaryElementWiseNoBroadcastImpl` fast path); ONNX `Mod` honours `fmod` for FP vs Python-style integer modulo; `Equal` / `Less` / `And` write 1-byte bool output. `And` is bool-only (uint8 stream), no dtype dispatch. |
| `reduce_sum_kernel.hip` (M)       | adds `hip_reduce_max / hip_reduce_prod`                             | Same block-per-output structure as the existing `hip_reduce_sum`; `reduce_size = data_num / output_num` (lowering arranges the trailing-axes layout). |
| `tile_kernel.hip`                 | `hip_tile`                                                          | Repeats inferred host-side from `output_shape[d] / input_shape[d]`; the GPU `repeats` tensor pointer is unused ΓåÆ no per-call D2H. |
| `expand_kernel.hip`               | `hip_expand`                                                        | numpy-style right-aligned broadcast handled inside the kernel; same "host-side shape avoids D2H" pattern as Tile. |
| `gather_nd_kernel.hip`            | `hip_gather_nd`                                                     | Fused `ComputeSliceOffsets` + gather; INT64 indices only; `batch_dims` honoured. |
| `slice_kernel.hip`                | `hip_slice` (**new**)                                               | Services the runtime fallback ΓÇö graph-constant + positive-step Slice still folds to `tensor.extract_slice` upstream. One thread per output element; supports negative steps and non-constant indices; rank Γëñ 8. |
| `scatter_nd_kernel.hip`           | `hip_scatter_nd` (**new**)                                          | All five reductions (`none` / `add` / `mul` / `min` / `max`). Native `atomicAdd` for f32/i32; `atomicAdd(unsigned long long*)` for i64; native `atomicMin` / `atomicMax` for i32; everything else (`mul` for all types, `min`/`max` for f32/i64/f16, `add` for f16) uses CAS emulation with full IEEE-754 NaN propagation. fp16 packed-CAS rewrites the 32-bit word containing the target half (relies on standard 2-byte fp16 alignment). OOB indices clamp to `[0, dim-1]` (matches ORT CUDA EP). |
| `cumsum_kernel.hip`               | `hip_cumsum`                                                        | One thread per `(outer, inner)` slice doing a serial scan along the axis; ONNX-14 `exclusive` / `reverse` flags supported. |
| `pad_kernel.hip`                  | `hip_pad`                                                           | Single kernel branched on `pad_mode` for ONNX-18 `constant` / `reflect` / `edge` / `wrap`; runtime D2H-reads `pads` (and optional `constant_value` / `axes`) once per call. |
| `layer_norm_kernel.hip`           | `hip_layer_normalization`                                           | Block-per-row two-pass kernel (FP32 accumulators regardless of I/O dtype); optional `bias` / `mean` / `inv_std` outputs; `stash_type` honoured for stats. |
| `include/hip_custom_kernels.h` (M)| public C declarations for all of the above                          | All new entry points are inside the existing `extern "C"` block per the runtime-bitcode-linkage contract in `CLAUDE.md`. |
| `CMakeLists.txt` (M)              | registers the new `.hip` sources under `hip_custom_kernels`         | ΓÇö |

### Runtime wrappers (`lib/Runtime/real/`)

Each of the 21 `.cpp` files below was a silent `return 0` stub on the PR base.
Each one now:

1. Validates required pointers and shape arguments (per NOTICES ┬º9).
2. Profiles via `OP_PROFILE` with a shape-string lambda (zero overhead when `HIPDNN_EP_PERF=0`).
3. Dispatches on `data_type` (and where applicable, on `pad_mode` / `reduction`) to the matching kernel template.
4. Returns the kernel's error code; debug-logs via `RUNTIME_DEBUG_LOG`.

Files:
`cos.cpp`, `sin.cpp`, `neg.cpp`, `sign.cpp`, `not.cpp`, `div.cpp`, `mod.cpp`,
`equal.cpp`, `less.cpp`, **`and.cpp`**, `reduce_max.cpp`, `reduce_prod.cpp`,
`tile.cpp`, `expand.cpp`, `gather_nd.cpp`, `cumsum.cpp`, `pad.cpp`,
`layer_normalization.cpp`, **`slice.cpp`**, **`scatter_nd.cpp`**,
**`multi_head_attention.cpp`**.

`slice.cpp` is the only wrapper that performs a runtime D2H + `hipStreamSynchronize`
of the index tensors (`starts` / `ends` / optional `axes` / `steps`) and resolves
the per-axis `(start, step)` host-side per ONNX-13+ clamping rules (negative-index
normalisation, step-sign-aware clamping into `[0, dim]` for positive step and
`[0, dim-1]` / `[-1, dim-1]` for negative step). `scatter_nd.cpp` keeps indices
on the device and lets each kernel thread do its own OOB clamp inline.

**`multi_head_attention.cpp`** implements the standard
`com.microsoft.MultiHeadAttention` `Q_K_V_BSNH` fp16 path
(`query`/`key`/`value` = `[B, S, N*H]`, `output` = `[B, S_q, N*H]`,
attributes `num_heads`, `scale` (`0.0` ΓåÆ `1/sqrt(H)`), `unidirectional`).
The algorithm mirrors GQA's decomposed hipBLASLt pipeline ΓÇö simpler because
`N_q == N_kv` so no KV head expansion ΓÇö across 6 steps:

1. Q/K/V transpose BSHD ΓåÆ BNSH via `hip_transpose_qkv_*` (skipped when the swapped dim is 1).
2. Score GEMM `S_f32 = (Q @ Kß╡Ç) * scale`, batched `B*N`, `transA` on K, via hipBLASLt.
3. Optional causal mask (`unidirectional == 1`) via `hip_gqa_causal_mask_f32`.
4. Softmax fp32 ΓåÆ fp16 via `hip_gqa_softmax_f32_to_f16`.
5. Value GEMM `O_BNSH = P @ V`, batched `B*N`, via hipBLASLt.
6. O transpose BNSH ΓåÆ BSHD via `hip_transpose_bnsh_to_bshd_fp16`.

All other MHA layouts and optional inputs (packed QKV, KV cache `past_key`/
`past_value`, `attention_bias`, `key_padding_mask`, `relative_position_bias`,
non-fp16 dtypes) are validated and return `-1` with a clear `stderr` message
instead of silently producing wrong output.

### Runtime state additions (`lib/Runtime/`)

- `runtime_state_internal.h`: new `void *mha_gemm_cache` field on `RuntimeState`.
  Same shape-keyed hipBLASLt descriptor + heuristic cache pattern as
  `gqa_gemm_cache`, kept distinct so concurrent MHA + GQA sessions do not
  contend on a single map.
- `hipdnn_ep_runtime.h`: new `void hipdnn_ep_mha_gemm_cache_destroy(void *cache)`
  declaration (`extern "C"` block) so the state-cleanup path can free it from
  generated code without taking a dependency on `MhaGemmCache`'s C++ type.
- `hipdnn_ep_runtime_state.cpp`: initialises `mha_gemm_cache = nullptr` in
  `initialize_state_handles`, destroys it in `hipdnn_ep_state_cleanup`
  (immediately after the equivalent `gqa_gemm_cache` block).

The cache itself lives in `lib/Runtime/real/multi_head_attention.cpp` and is
keyed on `(m, n, k, batch, strideA, strideB, strideC, transA, outputFp32)` so
the `hipblasLtMatmulAlgoGetHeuristic` call runs only on cache miss. Steady
state is `unordered_map` lookup ΓåÆ reused descriptors + algorithm.

#### CastLike ΓÇö new standalone MLIR pass (`hip-simplify-onnx`)

The pre-existing `E2E_Execute_test_castlike_model` failure was caused by
`onnx.CastLike`'s second operand ΓÇö a *type donor* whose data is never read.
The existing pattern correctly ignored the operand, but the donor was still
wired up as a `func.func` argument, and `generateModuleMetadata` (which
captures inputs *before* patterns fire) recorded it as a real input. At
runtime the EP marshaled a shape-`[0]` placeholder for it, and
`hipdnn_ep_tensor_prepare_input` rejected the null-data input.

The fix is a new standalone MLIR pass ΓÇö **`hip-simplify-onnx`** ΓÇö that runs
at the head of the onnx-to-hip pipeline (**before** `hip-add-context-arg`,
which lets it stay in the original ONNX function-arg index space and drop
its dependency on `mlir::hip::HipDialect` entirely). For every non-
declaration function it:

1. Snapshots which function arguments are live BEFORE any rewrite.
2. Runs registered per-op simplifiers. Currently:
   - `simplifyCastLike`: rewrites `onnx.CastLike` ΓåÆ `onnx.Cast` (the shared
     `CastConversion` pattern then takes it to `hip.cast`; the ONNX dtype
     enum is derived from the result type so no `to` attribute is needed).
     Identity casts short-circuit by forwarding the input.
3. Erases function arguments that were live before the rewrite and are now
   use-empty (i.e. whose last use was an eliminated type donor). The
   "live-before" snapshot is the key restraint: args that were already dead
   in the input IR stay, so the existing
   `test/lit/Pipeline/module-metadata.mlir` "metadata captures the original
   signature" assertion continues to hold.

**Extension point.** Future "type donor" / "shape donor" ops (e.g.
`onnx.EyeLike`, `onnx.RandomNormalLike`) follow the same shape: add a
`static void simplify<Op>(func::FuncOp)` helper in
`lib/Conversion/OnnxToHip/SimplifyOnnx.cpp` and one line to
`SimplifyOnnxPass::runOnOperation` between the snapshot and the sweep. The
dead-arg invariants are enforced centrally.

`lib/Conversion/OnnxToHip/CastLikeConversion.cpp` was removed ΓÇö subsumed by
the standalone pass + the shared `CastConversion`.

#### ConstantOfShape fold extension

Two changes in `lib/Conversion/OnnxToHip/ConstantOfShapeConversion.cpp`
and `OnnxToHip.cpp` to make `ConstantOfShape` work end-to-end:

1. `getCompileTimeConstantTensor` now also accepts `onnx.Shape(static-tensor)`
   as a recognised compile-time-constant input, honouring the optional
   `start` / `end` slicing attributes from ONNX-15 `Shape`. This covers the
   canonical `Shape ΓåÆ ConstantOfShape` pattern transformer graphs emit for
   zero-initialised KV / mask buffers.

2. The pattern is now run as a dedicated **pre-fold pass** in
   `ConvertOnnxToHipPass::runOnOperation` immediately **before**
   `lowerOnnxConstants`. The externaliser otherwise rewrites the shape input
   to a `memref.global` with null `initial_value` (data moves to
   `constants.bin`), at which point the fold can no longer reach the dense
   bytes. Fixes the `E2E_Compile_test_constant_of_shape_model` "op was not
   bufferized" error.

### Documentation

- `docs/design/qwen-vision-ops.md` (new): self-contained design doc
  capturing port pattern, host-side-shapes convention, FP32-accumulator
  rule for FP16 reductions/normalizations, per-op D2H stalls (Slice,
  Pad, CumSum) and why they're acceptable, the ┬º3.7 "Slice and ScatterND
  ΓÇö host-side indices" rationale, the ┬º3.8 "ConstantOfShape and the
  pre-fold ordering" rationale, and a comparison of our pure custom-
  kernel approach against ORT's MIOpen / hipBLAS / hipFFT / RCCL-
  dependent ROCm EP build (which is generated by `hipify_python` on the
  CUDA EP source). Also lists ONNX corner cases handled per op (e.g.
  Slice's `end = -(dim+1)` sentinel needed to express "go to the bottom"
  with `step < 0`, ScatterND's clamp-not-throw OOB policy, the
  duplicate-index UB rule, fp16 add/mul not implemented by ORT CPU).


zpye:
Promoted the CastLike fix from an inline helper into a standalone `hip-simplify-onnx` MLIR pass (commit `433fa91a`). The pass slots into the pipeline between `hip-add-context-arg` and `convert-onnx-to-hip`, is testable in isolation via `hip-mlir-opt --hip-simplify-onnx`, and has a clean extension point for future type-donor / shape-donor ops (e.g. `onnx.EyeLike`, `onnx.RandomNormalLike`). Adds one focused LIT test (`test_simplify_onnx.mlir`) and updates the existing `test_castlike.mlir` RUN line. Re-verified: **152/152 LIT, 94/94 E2E, pre-commit clean.**

zpye:
Follow-up refactor pushed (commit `9a0a637`):

The `hip-simplify-onnx` pass is now `simplify-onnx`, has zero HIP-dialect dependency, and runs at the head of the pipeline — **before** `hip-add-context-arg` instead of between it and `convert-onnx-to-hip`.

**Why**

The pass only does pure ONNX-dialect work (`onnx.CastLike → onnx.Cast` + drop dead type-donor function arguments). The only thing tying it to the HIP dialect was a single guard in the dead-arg sweep that skipped the `!hip.context` argument injected by `hip-add-context-arg`. By running upstream of context injection, that guard becomes unnecessary — there simply is no `!hip.context` argument to skip — and the pass collapses to a reusable, pure-ONNX transform.

**What changed**

| Surface | Before | After |
|---|---|---|
| CLI flag | `--hip-simplify-onnx` | `--simplify-onnx` |
| C++ class | `HipSimplifyOnnxPass` | `SimplifyOnnxPass` |
| Factory | `createHipSimplifyOnnxPass()` | `createSimplifyOnnxPass()` |
| Register | `registerHipSimplifyOnnxPass()` | `registerSimplifyOnnxPass()` |
| Pipeline | `hip-add-context-arg → hip-simplify-onnx → convert-onnx-to-hip` | `simplify-onnx → hip-add-context-arg → convert-onnx-to-hip` |
| TableGen `dependentDialects` | `mlir::hip::HipDialect`, `mlir::func::FuncDialect` | `mlir::func::FuncDialect` only |
| `SimplifyOnnx.cpp` HIP includes | `hip/Dialect/IR/HipDialect.h` for `ContextType` guard | none |

Tests, pipelines, and `InitAllPasses.h` updated to match. Comments in `test_castlike.mlir` and `test_castlike_model.mlir` describe the new ordering.

**Verified**

- 152/152 LIT pass (incl. `test_simplify_onnx.mlir`, `test_castlike.mlir`, `e2e/test_castlike_model.mlir`, `Pipeline/module-metadata.mlir`)
- 94/94 E2E pass
- pre-commit clean (`lintrunner`, license headers, all whitespace/EOL hooks)

The file still lives in `lib/Conversion/OnnxToHip/SimplifyOnnx.cpp` (most of our ONNX-aware code already lives there), but it is now name- and dependency-honest about being a pure ONNX pre-conversion cleaner.

akholodnamdcom (approved):

PR #226, opened 2026-05-15
fix(onnx-to-hip): drop dead CastLike type-donor function arguments

zpye:
## Summary

Fixes the pre-existing `E2E_Execute_test_castlike_model` failure (visible since the kernels PR's E2E run) without touching the kernel work.

The bug: ONNX `CastLike` takes a *type donor* as its second operand — a tensor whose data is never read; only its element type matters, and that type is already statically known from the result type. The existing `CastLikeConversion` pattern correctly ignored the operand at lowering time, **but** the donor was still wired up as a function argument, and `generateModuleMetadata` (called before any pattern fires) faithfully recorded it as a real input. At run time the EP marshaled a shape-`[0]` placeholder for it, `hipdnn_ep_tensor_prepare_input` rejected the null-data input, and the inference failed.

## Approach

Implement option (A) from the prior discussion: drop the dead second operand at lowering. Mechanically the change is:

1. Add a `preMetadataSimplifyCastLike` step in `convert-onnx-to-hip` that runs **before** `generateModuleMetadata`. It walks every function, rewrites each `onnx.CastLike` to a plain `onnx.Cast` (the shared `CastConversion` pattern then takes it to `hip.cast` — it derives the ONNX dtype enum from the result type, so no `to` attribute is needed), and drops any non-context function argument that was **live before** the rewrite and is now use-empty.
2. The "live before the rewrite" snapshot is the key restraint: arguments that were already dead in the input IR are preserved. The existing contract that `generateModuleMetadata` captures the original signature — asserted by `test/lit/Pipeline/module-metadata.mlir` — stays intact.
3. `lib/Conversion/OnnxToHip/CastLikeConversion.cpp` is now redundant. Delete it, drop its `populate*Patterns` declaration and registration, and update the two CastLike LIT tests to expect the simplified signature.

## Files

- `lib/Conversion/OnnxToHip/OnnxToHip.cpp` — add `preMetadataSimplifyCastLike`, call it before metadata capture; drop `populateCastLikeConversionPatterns`.
- `lib/Conversion/OnnxToHip/OnnxToHipUtils.h` — drop the `populateCastLikeConversionPatterns` declaration.
- `lib/Conversion/OnnxToHip/CMakeLists.txt` — drop `CastLikeConversion.cpp`.
- `lib/Conversion/OnnxToHip/CastLikeConversion.cpp` — deleted (subsumed by the pre-metadata simplification + the existing `CastConversion`).
- `test/lit/e2e/test_castlike_model.mlir` — expect `hipdnn.input_count = 1` and document the new lowering path.
- `test/lit/Conversion/onnx-to-hip/test_castlike.mlir` — update `CHECK-SAME` lines to expect the dropped `%target` argument.

Diff size: **+120 / −115**, six files.

## Test plan

- [x] `ctest -R MorphizenMLIRLitTests` — **151/151 pass** (including `test/lit/Pipeline/module-metadata.mlir`, which asserts dead args are preserved when they were already dead in the input IR).
- [x] Full E2E suite (`test_output\_test.bat e2e`) — **94/94 pass**, including the previously-failing `E2E_Execute_test_castlike_model`.
- [x] `E2E_Compile_test_castlike_model` still passes.

## Notes / alternatives considered

- **Alternative B**: drop the operand at the run-time interface layer (`hipdnn_ep_tensor_prepare_input`) by special-casing `wrap_cast` callers. Rejected: it would have leaked an op-specific concept ("the second input is a type donor") into a generic input-marshaling routine, and would have left the compiled DLL claiming an input it doesn't actually consume — a worse contract.
- **Alternative C**: indiscriminately drop *all* dead function arguments before metadata. Rejected: it silently changes the metadata contract for graphs with unrelated dead args (and breaks `test/lit/Pipeline/module-metadata.mlir`). The implemented approach drops only args whose last live use was the eliminated `CastLike` type donor.
- The `!hip.context` argument is always preserved (it's required by the runtime calling convention and is added by `hip-add-context-arg`, not by the user's ONNX graph).

This is a follow-up to #225 (the 16 Qwen vision-ops kernels PR) and targets the same `qwen_vision` integration branch, but is otherwise independent of the kernel work.


zpye:
Folding this into #225 per request; same diff will land on the qwen_vision_kernels branch.

PR #230, opened 2026-05-17
feat(loop): ONNX Loop operator support via outlined body + counted/dynamic runtime driver

fhanuman:
## Summary

Adds support for ONNX `Loop`. A new pass `--onnx-loop-outline` rewrites each `onnx.Loop` into a `hip.loop` op plus a sibling `func.func` carrying the body, so body ops flow through the same `ConvertOnnxToHip` -> bufferize -> HipToLLVM pipeline as `main_graph`. HipToLLVM emits a per-loop trampoline that dispatches to one of two runtime drivers: a counted fast path when the body identity-yields `cond_in`, or a dynamic path that reads cond each iteration.

Closes the previous "model contains `onnx.Loop`, fall back to CPU EP" gap at MLIR ingestion.

## Why

Outlining the body into a sibling `func.func` is the cleanest way to reuse the existing per-op DPS / bufferize / pool-allocation infrastructure. The alternative -- lowering the region in-place -- would require special-casing every pass that walks function bodies and would leave the body's `onnx.*` ops on a separate code path from `main_graph`.

## What

1. **New `hip.loop` op** (`HipOps.td`, `HipDialect.cpp`). DPS op with the loop-carried operands as DPS inits, so writes land in `v_init` in place. Explicit `Write` memory effects prevent post-bufferize DCE; `verify` checks that `body_func` resolves to a `func::FuncOp`.

2. **New `--onnx-loop-outline` pass** (`LoopOutline.cpp`). Clones each `onnx.Loop` body into an outlined func, lifts implicit SSA captures to explicit args, and sets `cond_is_passthrough` when `cond_out` SSA-equals `cond_in`. Folds 0-D constant trip-count and cond_init into `arith.constant` to avoid host loads from GPU-resident memrefs.

3. **HipToLLVM lowering** (`LoopLowering.cpp`). Emits one per-loop `<body>_trampoline` LLVMFuncOp at module scope that rebuilds rank-0 memref descriptors for iter/cond inline and aliases `v_in_i` with `v_out_i` at the body call.

4. **Runtime drivers** (`hipdnn_ep_runtime_loop.cpp`). Single `runLoopImpl<CondPolicy>` template specialized on `CountedCondPolicy` and `DynamicCondPolicy`. Iter and cond live in pinned host-mapped memory, so the counted path has no per-iter D2H sync. Nested loops are hard-failed by `LoopNestingGuard`.

5. **Pipeline placement**: outlining runs after `HipAddContextArgPass` and before `ConvertOnnxToHip` in `Pipelines.cpp`, so the body's `onnx.*` ops are visited like any other top-level `func.func`.

## Compiler IR walkthrough

<details>
<summary><b>Input IR</b> -- ONNX dialect, fresh out of the importer; <code>onnx.Loop</code> carries its body inline as a region with implicit captures</summary>

```mlir
module {
  func.func @main_graph(%arg0: tensor<?x16x16xf32> {onnx.name = "A"}, %arg1: tensor<16x16xf32> {onnx.name = "B"}) -> (tensor<16x16xf32> {onnx.name = "v_final"}) attributes {onnx.graph.name = "loop_add_v1"} {
    %0 = "onnx.NoValue"() {value} : () -> none
    %1 = "onnx.Constant"() {node.outputs = ["squeeze_axes"], value = dense<0> : tensor<1xi64>} : () -> tensor<1xi64>
    %2 = "onnx.Constant"() {node.outputs = ["cond_init"], value = dense<1> : tensor<ui8>} : () -> tensor<ui8>
    %3 = "onnx.Constant"() {node.outputs = ["acc_init"], ...} : () -> tensor<16x16xf32>
    %4 = "onnx.Shape"(%arg0) {end = 1 : si64, start = 0 : si64, ...} : (tensor<?x16x16xf32>) -> tensor<1xi64>
    %5 = "onnx.Squeeze"(%4, %1) {...} : (tensor<1xi64>, tensor<1xi64>) -> tensor<i64>
    %6 = "onnx.Loop"(%5, %2, %3) ({
    ^bb0(%arg2: tensor<i64>, %arg3: tensor<ui8>, %arg4: tensor<16x16xf32>):
      %7  = "onnx.NoValue"() {value} : () -> none
      %8  = "onnx.Gather"(%arg0, %arg2) {axis = 0 : si64, ...} : (tensor<?x16x16xf32>, tensor<i64>) -> tensor<16x16xf32>
      %9  = "onnx.Add"(%8, %arg1) {...} : (tensor<16x16xf32>, tensor<16x16xf32>) -> tensor<16x16xf32>
      %10 = "onnx.Add"(%arg4, %9) {...} : (tensor<16x16xf32>, tensor<16x16xf32>) -> tensor<16x16xf32>
      %11 = "onnx.Identity"(%arg3) {...} : (tensor<ui8>) -> tensor<ui8>
      "onnx.Yield"(%11, %10) : (tensor<ui8>, tensor<16x16xf32>) -> ()
    }) {...} : (tensor<i64>, tensor<ui8>, tensor<16x16xf32>) -> tensor<16x16xf32>
    "onnx.Return"(%6) : (tensor<16x16xf32>) -> ()
  }
}
```

</details>

<details>
<summary><b>After <code>--onnx-loop-outline</code> + <code>ConvertOnnxToHip</code> + bufferize</b> -- ready for <code>HipToLLVM</code></summary>

```mlir
module attributes {hipdnn.input_shapes = [...], hipdnn.pool_size = 2048 : i64, ...} {
  func.func @main_graph(%arg0: !hip.context, %arg1: memref<?x16x16xf32> {onnx.name = "A"}, %arg2: memref<16x16xf32> {onnx.name = "B"}, %arg3: memref<16x16xf32> {bufferize.result}) attributes {onnx.graph.name = "loop_add_v1"} {
    %true = arith.constant true
    %c2_i64 = arith.constant 2 : i64
    %0 = hip.get_constant(%arg0, %c2_i64) : memref<16x16xf32>
    %dim = memref.dim %arg1, %c0 : memref<?x16x16xf32>
    memref.copy %0, %arg3 : memref<16x16xf32> to memref<16x16xf32>
    hip.loop(%arg0, %dim, %true) iter_args(%arg3 : memref<16x16xf32>)
        captures(%arg1, %arg2 : memref<?x16x16xf32>, memref<16x16xf32>)
        body @main_graph_loop_body_n0
        {cond_is_passthrough, num_loop_carried = 1 : i32}
    return
  }
  func.func private @main_graph_loop_body_n0(
      %arg0: !hip.context,
      %arg1: memref<i64>, %arg2: memref<ui8>,                     // iter, cond_in
      %arg3: memref<16x16xf32>,                                   // v_in (loop-carried)
      %arg4: memref<?x16x16xf32>, %arg5: memref<16x16xf32>,       // captures (A, B)
      %arg6: memref<16x16xf32> {bufferize.result}) {              // v_out (DPS, aliases v_in)
    %c1024 = arith.constant 1024 : index
    %c0    = arith.constant 0    : index
    %c2048 = arith.constant 2048 : index
    %0     = hip.get_pool(%arg0, %c2048) : memref<?xi8>
    %view  = memref.view %0[%c0]   [] : memref<?xi8> to memref<16x16xf32>
    hip.gather(%arg0) ins(%arg4, %arg1 : memref<?x16x16xf32>, memref<i64>) outs(%view : memref<16x16xf32>)
    %view_0 = memref.view %0[%c1024][] : memref<?xi8> to memref<16x16xf32>
    hip.add(%arg0) ins(%view, %arg5 : memref<16x16xf32>, memref<16x16xf32>) outs(%view_0 : memref<16x16xf32>)
    hip.add(%arg0) ins(%arg3, %view_0 : memref<16x16xf32>, memref<16x16xf32>) outs(%arg6 : memref<16x16xf32>)
    return
  }
}
```

The body region is gone; its ops are now in `@main_graph_loop_body_n0` and have lowered to `hip.gather` / `hip.add` via the same path as any other op. Trip count is unboxed (`onnx.Shape -> onnx.Squeeze` folded to `memref.dim`). `cond_is_passthrough` is set on `hip.loop`. The loop-carried slot is bufferized in place (`acc_init` copied to `%arg3` once before the loop; body's `%arg6 {bufferize.result}` aliases it).

</details>

## Performance

<details>
<summary>22x over DML EP on a 16-iter counted-loop body (fp32; <code>onnxruntime_perf_test</code>, 1024 iter/inference)</summary>

| Metric            | **This PR**        | DML EP[1]         | Ratio          |
|---|---:|---:|---:|
| Inferences/sec    | **217.6**          | 9.75              | **22.3x**      |
| Mean latency      | **4.60 ms**        | 102.5 ms          | **22.3x**      |
| P50 / P99 latency | **4.57 / 5.05 ms** | 102.9 / 106.0 ms  | 22.5x / 21.0x  |
| Per-iter cost     | **~4.5 us**        | ~100 us           | **22x**        |

DML can't take `onnx.Loop` and falls back to CPU -- each iteration is a CPU->DML->CPU round trip (~100 us floor independent of body work). The counted fast path keeps the loop inside the compiled GPU pipeline (single hot stream, iter counter in pinned host-mapped memory, no per-iter D2H sync), reaching ~4.5 us/iter -- close to the HIP kernel-launch floor.

[1] End-to-end run additionally requires `onnx.Shape` lowering (PR #212) and a morphizen submodule bump (separate PR); loop runtime performance itself is attributable to this PR.

</details>

## Notes for reviewers

- The trampoline assumes the body's `func::FuncOp` has not yet been converted to `LLVMFuncOp` when `LoopOpLowering` runs. Fails with a clear diagnostic if it has.
- The aliased `v_in == v_out` invariant is documented but not enforced. Safe for bodies composed of kernels that touch each cell once per launch; cross-cell read-after-write (e.g. Conv with halos) will need materialized out buffers + post-body copy -- follow-up.
- Nested loops are hard-failed; lifting requires per-depth iter/cond stacks.
- **Numerics**: Combined L2 = 0 vs ORT CPU EP on the counted fast path. Dynamic-cond path validation is follow-up.


fhanuman:
Hi @amd-bqiao, Just added you to the PR since you were curious, however, please feel free to review if you would like.

fhanuman:
I've verified all changes in the PR and am marking it ready for review so the reviewer can take a pass. I'll take another pass in the morning with a fresh eyes to make sure I haven't missed anything.

There might be additional optimizations we can do here. We will reach out to the ROCm runtime/libraries team once we have more profiling data from Qwen.

zz002:
Status update on the two MorphiZen-EP-side follow-ups this PR depends on for end-to-end `loop_add_v1_N16`:

- MorphiZen [ROCm/MorphiZen#222](https://github.com/ROCm/MorphiZen/pull/222) (Loop/If/Scan implicit-capture propagation) merged: [`cf41db64`](https://github.com/ROCm/MorphiZen/commit/cf41db64d0dd9ea12862c816f91e9764704e9d3b)
- ORT-side companion patch opened: [microsoft/onnxruntime#28608](https://github.com/microsoft/onnxruntime/pull/28608) (makes `MakeComputeCapability` include `ImplicitInputDefs()` in fused subgraph boundary)

zz002 (approved):

amd-mingw (approved):
Have you researched/compared what's the advantages and disadvantages about this PR's approach and "scf-based lowering + optimization" approach? from the long-term view and short-term view? from the architecture view?
but I would approved it firstly because of the timeline.

fhanuman:
> Have you researched/compared what's the advantages and disadvantages about this PR's approach and "scf-based lowering + optimization" approach? from the long-term view and short-term view? from the architecture view? but I would approved it firstly because of the timeline.

Thanks, @amd-mingw , for the review.

Yes, I have thought about this. We could lower it to scf.for / scf.while, however, with our current pipeline, doing so doesn't give us much. Our outer loop is sequential — per iteration we still need to dispatch the inner hip.* kernels to the GPU one by one through the HIP runtime, and the upstream scf → gpu lowering is parallelism-oriented (it maps loop iterations to GPU threads), which is the opposite of what we need here. So even with onnx.Loop → scf.for, the scf.for would stay on the host, and we would still be doing per-iteration HIP kernel launches the same way this PR does — just with an extra abstraction layer in between.

As I briefly mentioned about this in the ROCm EP offsite discussion, ideally, we want to lower some of these to MLIR core dialects (like scf.for, and other core dialects), and from there to the GPU dialect, and from there to the AMD GPU dialect, for some of the OPs that we don't care about performance. I will spend some time with this once all the critical items are out of the way. We can also meet with the folks from Stella's team, as they are involved in supporting this. In the long term, I feel this is more scalable than writing custom kernels for every new OP we encounter.

Additionally, onnx.Loop semantics is more complex than a simple for loop — cond_out per-iteration termination, optional M (cond-only loops), and most importantly scan outputs (per-iteration outputs concatenated into a result tensor) that don't have a clean equivalent operations in the scf dialect. We can probably represent this using decomposition and other MLIR core dialects; however, in terms of performance, this doesn't give us much with our current pipeline.

PR #232, opened 2026-05-18
Add Memory Manager subsystem for GPU memory allocation

tejus26:
## Summary

- Adds the Memory Manager (MM) subsystem — a GPU memory allocator specialized for LLM inference with separate budget pools for weights, activations, and KV cache blocks
- Folds MM into `runtime.bc`: model DLLs call MM functions directly with zero FFI overhead (no separate `mm.lib` needed)
- Integrates MM into the runtime: `hipMalloc`/`hipFree` for constants, workspace, and qmoe_scratch replaced with `mm::alloc`/`mm::free` using typed hints (`MemoryClass`, `Lifetime`)
- Enables MM unit tests in CI (real-build ctest step with TheRock DLLs on PATH)

### Memory Manager (`lib/MemoryManager/`, `include/mm/`)
- **Public API** (`mm_api.h`): `mm::init`, `mm::alloc`, `mm::free`, `mm::get_ptr`, `mm::query`, `mm::metrics`
- **Type system** (`mm_types.h`): `MemoryClass` (Generic/Weight/Activation/KvCache/Scratch), `Lifetime` (Static/Request/Step/Transient), `AllocHints`, `AllocInfo`
- **Core** (`mm_core.cpp`): singleton coordinator — dispatches to subsystems, tracks global metrics, enforces total budget
- **HAL** (`mm_hal.h/cpp`): hardware abstraction wrapping `hipMalloc`/`hipFree`
- **Handle table** (`mm_handle_table.h/cpp`): opaque 64-bit handle → `AllocInfo` registry
- **Activation arena** (`mm_activation.h/cpp`): segregated free-list arena with 8 size classes (1 KB → 4 MB + fallback)
- **KV manager** (`mm_kv_manager.h/cpp`): block manager for paged KV cache with refcounted CoW fork
- **Config** (`mm_config.h/cpp`): configuration struct + `config_default()`
- **KV API** (`mm_kv.h`): `kv_alloc_block`, `kv_free_block`, `kv_fork_block`

### Runtime Integration
- `lib/Runtime/CMakeLists.txt`: `compile_mm_to_bitcode` macro compiles all 6 MM sources to LLVM bitcode, merged into `runtime.bc` via `llvm-link`
- `lib/Runtime/hipdnn_ep_runtime_state.cpp`: constants allocation uses `mm::alloc` with `MemoryClass::Weight`/`Lifetime::Static`; workspace and qmoe_scratch use `mm::alloc`/`mm::free` with `MemoryClass::Activation`/`Lifetime::Request`; shared constants cache key simplified to size-only
- `lib/Runtime/runtime_state_internal.h`: `mm::handle_t` fields for gpu_constants, workspace, qmoe_scratch
- `lib/Runtime/hipdnn_ep_runtime.h`: updated `hipdnn_ep_get_pool_base` signature (removed `needed_size` parameter)

### Build System
- `CMakeLists.txt` / `lib/CMakeLists.txt`: `add_subdirectory(MemoryManager)` guarded by `NOT BUILD_MOCK_RUNTIME`

### Tests / CI
- `test/mm/CMakeLists.txt` + `mm_tests.cpp`: unit tests for MM subsystem
- `.github/workflows/windows-build.yml`: MM unit test step + `mm_tests.exe` packaging in GPU test package

### Documentation
- `docs/design/memory-manager.md`: architecture doc covering file layout, layer responsibilities, allocation/free call trees, budget initialization, and build integration

## Test plan
- [x] pre-commit passes clean
- [x] Build succeeds (mock + real)
- [x] LIT tests pass (zero regressions)
- [x] MM unit tests pass on GPU (ctest step in CI)
- [x] OGA benchmark runs successfully
- [x] L2 accuracy tests pass


amd-yanjunz:
ROCm OGA CI finished with result: FAILURE.

- PR: ROCm/onnx-hipdnn-ep#232
- Source comment: [models=all](https://github.com/ROCm/onnx-hipdnn-ep/pull/232#issuecomment-4497855163)
- TEST_MODE: both
- MODELZOO: LLM_ROCm_release
- TEST_MODELS: (all)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/449/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hipdnn_ep_oga&id=6a0d9aaa46df6f9d15dc75ff

amd-yanjunz:

- PR: ROCm/onnx-hipdnn-ep#232
- Source comment: [models=all](https://github.com/ROCm/onnx-hipdnn-ep/pull/232#issuecomment-4498171590)
- TEST_MODE: both
- MODELZOO: LLM_ROCm_release
- TEST_MODELS: (all)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/450/

lib/Runtime/real/paged_attention.cpp
```diff
@@ -0,0 +1,338 @@
...
+        size_t copy_bytes =
+            static_cast<size_t>(total_seq) * head_dim * element_size_bytes;
+
+        char *ks = static_cast<char *>(k_contiguous);
+        char *vs = static_cast<char *>(v_contiguous);
+        char *kd = static_cast<char *>(paged_kv.key_buffers[layer_idx]);
+        char *vd = static_cast<char *>(paged_kv.value_buffers[layer_idx]);
+
+        for (int h = 0; h < static_cast<int>(kv_num_heads); ++h) {
+          hipMemcpyAsync(kd + h * head_dst_stride, ks + h * head_src_stride,
```
fhanuman: Could we wrap these two calls in `HIP_CHECK` like you did in other places?

lib/Runtime/hipdnn_ep_runtime.h
```diff
@@ -557,6 +557,26 @@ int wrap_group_query_attention(
     int64_t batch_size, int64_t seq_len_q, int64_t seq_len_kv,
     int64_t past_buf_seq, int64_t head_dim, int64_t element_size_bytes);
 
+// PagedAttention operation wrapper (com.microsoft.PagedAttention v1)
+// Block-based paged attention for long-context LLM inference.
+// KV cache uses block layout: [num_blocks, block_size, kv_num_heads, head_dim]
+// Block table maps logical blocks to physical blocks.
+int wrap_paged_attention(
```
fhanuman: Would it be possible to split the PR into the following to make it easier to review or do you see any complications?
1. Memory manager
2. Paged attention operator support

include/hip/Dialect/IR/HipBufferize.h
```diff
@@ -99,6 +99,8 @@ registerHipBufferizableOpInterfaceModels(DialectRegistry &registry) {
     WhereOp::attachInterface<HipDstBufferizableModel<WhereOp>>(*ctx);
     LinearAttentionOp::attachInterface<
         HipDstBufferizableModel<LinearAttentionOp>>(*ctx);
+    PagedAttentionOp::attachInterface<
```
fhanuman: We need to make similar change to attach the interface in `tools\hip-mlir-opt\hip-mlir-opt.cpp`.

```
    mlir::hip::PagedAttentionOp::attachInterface<
        HipDstBufferizableModel<mlir::hip::PagedAttentionOp>>(*ctx);
```

The duplication is not related to your changes, but now that I think about it, we can probably unify these two places to avoid having to keep them in sync.

lib/Conversion/HipToLLVM/PagedAttentionLowering.cpp
```diff
@@ -0,0 +1,187 @@
...
+    Value doRotary = createI64Const(op.getDoRotary());
+    Value rotaryInterleaved = createI64Const(op.getRotaryInterleaved());
+    Value localWindowSize = createI64Const(op.getLocalWindowSize());
+
+    // === Extract shape info from memref types ===
+
+    // query: [num_tokens, hidden_size]
+    auto queryType = cast<MemRefType>(op.getQuery().getType());
+    auto queryShape = queryType.getShape();
+    int64_t numTokens = queryShape[0];
```
fhanuman: Do we need to update this logic to handle dynamic shapes? (#212 is updating the other lowerings via getMemRefDimSize().)

include/mm/mm_api.h
```diff
@@ -0,0 +1,32 @@
+/*
```
fhanuman, line 1: I am trying to build a mental model of the MM stack from the code so I can review the changes properly. Could we add a short architecture overview — call tree, layer responsibilities, and one worked example — somewhere like lib/MemoryManager/README.md or docs/design/memory-manager.md? It would help with this review and the next person who touches it.

fhanuman (commented):
Some initial review comments. Will continue to review.

lib/Runtime/real/paged_attention.cpp
tejus26: Done — wrapped both `hipMemcpyAsync` calls in `HIP_CHECK` to match the rest of the file.

lib/Runtime/hipdnn_ep_runtime.h
tejus26: I considered splitting but the MM-into-runtime.bc change is architecturally coupled to PagedAttention: the paged_attention runtime code calls MM functions that are now linked via runtime.bc (previously mm.lib). Splitting would create a state where either (a) the MM bitcode integration has no consumer and can't be tested, or (b) PagedAttention still depends on a separate mm.lib that this PR removes. The CI ctest step for mm_tests validates the MM independently regardless.

That said, the review burden is real — I've added `docs/design/memory-manager.md` (see comment 5 response) to make the MM stack easier to follow.

include/hip/Dialect/IR/HipBufferize.h
tejus26: Good catch — added `PagedAttentionOp::attachInterface` to `hip-mlir-opt.cpp`'s local `registerHipBufferizableOpInterfaceModels`. Agree the duplication is unfortunate; unifying the two registration sites would be a good follow-up.

lib/Conversion/HipToLLVM/PagedAttentionLowering.cpp
tejus26: Not in this PR — our compiler currently requires fully static shapes (documented constraint). PR #212 is the dynseqlen effort that introduces `getMemRefDimSize()` across all lowerings. PagedAttention's lowering will need the same treatment when #212 lands, but adding it now would create a dependency on unmerged code and diverge from every other lowering in main. I'll flag this as a follow-up for the dynseqlen integration.

include/mm/mm_api.h
tejus26: Added `docs/design/memory-manager.md` with the architecture overview: file layout, layer responsibilities, call trees for alloc/free/kv_alloc, budget initialization, build integration (bitcode linking), and a worked example of the paged attention decode path.

include/mm/mm_config.h
```diff
@@ -0,0 +1,57 @@
...
+#include <cstddef>
+#include <cstdint>
+#include <limits>
+
+#include "mm/mm_types.h"
+
+namespace mm {
+
+struct WatermarkConfig {
+  double warn = 0.80;     // Trigger background pressure actions.
```
fhanuman: I didn't understand how these values are chosen. Any rationally for these thresholds? Some comments clarifying the rationale would be good.

include/mm/mm_config.h
```diff
@@ -0,0 +1,57 @@
...
+  std::size_t default_alignment = 256;
+  std::size_t gpu_memory_limit = 0; // 0 = query from HAL.
+  double kv_cache_fraction =
+      0.90; // Portion of free GPU memory to dedicate to KV blocks.
+  std::size_t kv_cache_max_bytes =
+      0; // Explicit cap for KV pool. 0 = derive from fraction.
+  std::uint32_t kv_block_size_tokens = 16;
+  std::size_t kv_bytes_per_token_hint =
+      0; // Optional override when descriptor omits bytes/token.
+  std::array<std::size_t, 8> activation_size_class_upper_bounds = {
```
fhanuman, line 22: I didn't understand how ended up having these block sizes. Could you please elaborate and add clarification as code comments for future reference.

include/mm/mm_config.h
```diff
@@ -0,0 +1,57 @@
...
+  bool enable_cpu = false;
+  bool enable_ssd = false;
+  bool enable_cxl = false;
+};
+
+struct Config {
+  int device_id = 0;
+  std::size_t default_alignment = 256;
+  std::size_t gpu_memory_limit = 0; // 0 = query from HAL.
+  double kv_cache_fraction =
```
fhanuman: Same comment as earlier — would appreciate adding rationale in the other places where we have made similar decisions with some of the constants we have chosen.

include/mm/mm_types.h
```diff
@@ -0,0 +1,125 @@
...
+
+enum class AccessPattern : std::uint8_t {
+  Sequential = 0,
+  Random = 1,
+  WriteOnce = 2,
+  ReadMostly = 3
+};
+
+enum class Device : std::uint8_t {
+  Gpu0 = 0,
```
fhanuman: Clarification: Why do we have different `enum` values to represent `Gpu0`, `Gpu1`, etc. and how does this impact w.r.t memory planning? Could you elaborate and also add the clarification as code comments.

lib/MemoryManager/mm_core.cpp
```diff
@@ -0,0 +1,417 @@
...
+  if (hints)
+    local_hints = *hints;
+
+  const std::size_t alignment = local_hints.alignment
+                                    ? local_hints.alignment
+                                    : g_state.config.default_alignment;
+  void *ptr = nullptr;
+  std::size_t reserved_size = align_up(size_bytes, alignment);
+
+  if (local_hints.mem_class == MemoryClass::Activation) {
```
fhanuman, line 144: Do we need to couple LLM-specific concerns into the core memory manager? Worth considering whether to abstract them out.

Today mm_core.cpp ties the generic allocator and the LLM KV layer tightly: ManagerState owns KvManager directly, init() partitions HBM with kv_cache_fraction = ..., mm::free dispatches on mem_class == KvCache, and mm_api.h directly includes mm_kv.h. The result is that the activation/scratch allocator and the KV manager are harder to reason about, test, and evolve independently — changes to either typically reach into mm_core.cpp.

One direction worth considering: a generic mm core that exposes a Pool interface for allocation strategies, with the KV manager moved into an mm_llm extension that consumes the core via pool_id. The core stops caring about KV cache; the LLM-specific bits stay together and evolve without dragging the allocator API along. As more memory consumers land over time (activation reuse, weight residency, eventually tiering), that coupling cost tends to compound.

lib/MemoryManager/mm_core.cpp
```diff
@@ -0,0 +1,417 @@
...
+}
+
+int is_initialized() {
+  std::lock_guard<std::mutex> lock(g_mutex);
+  return g_state.initialized ? 1 : 0;
+}
+
+handle_t alloc(std::size_t size_bytes, const AllocHints *hints,
+               stream_t stream) {
+  (void)stream;
```
fhanuman: Do we need the stream parameter on mm::alloc / mm::free / kv_alloc_block / kv_free_block? Worth considering whether to drop it.

It's (void)'d in all four definitions in mm_core.cpp and mm_kv_manager.cpp, and the underlying HAL (hipMalloc/hipFree) is synchronous, so it isn't doing anything end-to-end. When multi-stream support is actually designed, the API will likely grow in shape — not sure yet how we would handle cross-stream interactions — rather than just consume this single parameter, so keeping a dead stream_t today doesn't reduce that future work and does invite misuse by callers who reasonably assume it has semantics.

fhanuman (changes requested):
Will continue to review.

include/mm/mm_types.h
fhanuman: Clarification: Is this for multi-GPU support? If yes, the memory manager would need to change significantly anyway right?

So, does it make sense to simplify and keep only what's needed today? My suggestion is to make the memory manager as lean as possible, and extend/scale it as and when needed. (Same concern as the stream parameter — both added ahead of any implementation that would actually use them.)

fhanuman (changes requested):

include/mm/mm_config.h
tejus26: Added inline comments explaining each constant: 0.80/0.95 watermarks (standard memory-pressure thresholds), 256B alignment (HIP minimum + RDNA3 cache-line alignment), 90% KV fraction (LLM inference is KV-dominated), 16 tokens/block (vLLM default, granularity vs overhead tradeoff), size-class progression (power-of-4 from 1 KB to 4 MB), 256 KB slab size (8 classes × 256 KB = 2 MB total pre-commitment).

include/mm/mm_config.h
tejus26: Added comments — the size classes follow a power-of-4 progression (1 KB, 4 KB, 16 KB, ..., 4 MB) to cover the typical activation tensor range with the last class as catch-all for oversized allocations. See the updated mm_config.h.

include/mm/mm_config.h
tejus26: Done — added rationale comments for all config constants (alignment, KV fraction, block size, slab size). See the updated mm_config.h.

include/mm/mm_types.h
tejus26: Agreed — the multi-GPU enum values were speculative. Simplified `Device` to just `Gpu` since we're single-GPU today. Multi-device support will need a significantly different design anyway (per-device allocators, cross-device transfer semantics), so pre-enumerating GPU IDs doesn't buy us anything.

tejus26: Agreed — simplified the `Device` enum to a single `Gpu` value. The multi-GPU entries (Gpu0-Gpu3, Npu, Cpu, Any) were speculative and none were referenced anywhere. When multi-device support is actually designed, the API will likely need per-device allocator instances rather than an enum tag, so keeping these values would have been misleading.

lib/MemoryManager/mm_core.cpp
tejus26: Good call — removed `stream_t` from all four public API functions (`alloc`, `free`, `kv_alloc_block`, `kv_free_block`). It was `(void)`'d in every implementation and the underlying HAL operations (hipMalloc/hipFree) are synchronous. Updated all callers in `hipdnn_ep_runtime_state.cpp` and `mm_tests.cpp`. The HAL still has `memset_async(stream)` internally for future use, but the public API is now stream-free.

lib/MemoryManager/mm_core.cpp
tejus26: Valid concern. The coupling is real but manageable today — there's exactly one consumer (PagedAttention) and the KV manager is a self-contained subsystem behind a clean internal interface (`KvManager` class). Extracting it into a pool-based abstraction is the right long-term direction, but it's a non-trivial refactor (needs a pool registry, lifetime-based eviction policy interface, etc.) that I'd rather do as a dedicated follow-up when a second memory consumer actually lands, rather than speculating on the interface shape now. Filed as a follow-up.

amd-yanjunz:

- PR: ROCm/onnx-hipdnn-ep#232
- Source comment: [models=all](https://github.com/ROCm/onnx-hipdnn-ep/pull/232#issuecomment-4572581181)
- TEST_MODE: both
- MODELZOO: LLM_ROCm_release
- TEST_MODELS: (all)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/533/

fhanuman (approved):

PR #241, opened 2026-05-21
Gemma3 vision encoder: dynamic-shape compile path + ViT compute coverage + Qwen3.5 Vision prep

akholodnamdcom:
## Summary

  HuggingFace ONNX exports for transformer vision encoders (Gemma-3 SigLIP, Qwen3.5 vision, similar) leave intermediate `RankedTensorType`s loose (`<?x?x?x?>`) even when the shape operand uniquely determines them, and the EP↔ORT boundary needs an output shape **before** inference
   runs. This PR lands the compile-time refinement infrastructure, the C ABI + proto channels, the runtime resolution path, and the runtime/lowering bug fixes needed to compile and run such models end-to-end with finite output on the MorphiZen EP — without modifying the ONNX file
   on disk.

  ## What's in the box

  **Compile-time refinement (`lib/Conversion/OnnxToHip/InferOnnxShapes.cpp`)**
  - New MLIR pass that walks `onnx.*` ops in topo order and tightens each result type to the maximum static info derivable from the inputs. `isStrictlyTighter` guarantees the pass never widens, never changes element type, never changes rank.
  - Both the type-inference dispatch and the per-output-dim SSA-origin trace go through a file-static `llvm::StringMap` registry — adding a new op is one insert per rule. Trace coverage: Reshape (dim 0 outer-product-ratio: identity or divide-by-K; dim > 0 strict passthrough),
  Transpose, MatMul (outer-batch right-aligned over the lhsRank-2/rhsRank-2 OUTER slice — the gotcha test in `test_infer_onnx_shapes.mlir` guards against full-rank-alignment regression), Cast, unary same-shape, binary broadcast, Conv/AveragePool/MaxPool, the four Reduce* ops, and
   the shape-arithmetic family ViT exports need: Squeeze, Unsqueeze, Concat, Slice, Gather.
  - Refinement runs in a 4-round quiescence loop inside `convert-onnx-to-hip` — LLMs converge in ~2 rounds, vision in ~3. The cap emits a warning if it ever exits without quiescence so a bouncing pattern surfaces instead of slipping through as silent partial inference.
  - Post-DCE walk emits a hard `emitError + signalPassFailure` for any `onnx.*` op that survives `convert-onnx-to-hip` — replaces the terse `op was not bufferized` from OneShotBufferize that previously masked unconverted ops as silent CPU fallback.
  - Unknown-op producers in the trace path emit a warning once per function per kind ("no trace rule for op X — output dims passing through this op may be unresolvable at runtime"), so the next bring-up sees exactly which rules to add.

  **C ABI + proto channels**
  - New `CompilationOutputs` struct threaded as a trailing param on `hip_compile_with_fs` — caller-allocated `shapes_buf` + `origins_buf` with `*_capacity` / `*_needed` discovery. Single in-process channel, no thread-local stash, no two-call symbol-lookup round trip.
  - The pass attaches refined shapes + per-dim SSA origins as module-level `ArrayAttr`s (`hip.refined_output_shapes`, `hip.refined_output_dim_origins`) inside `InferOnnxShapes` while the function is still `func::FuncOp`. Module attributes survive the later `func.func → llvm.func`
   conversion; `CompilerDriver` reads them after `runMLIRPasses` and the C ABI packs them into the caller's buffer.
  - `metadata.proto::DimSource` carries one of three states: STATIC (`static_value > 0`), RUNTIME-INPUT-LOOKUP (`resolved=true` + `input_idx` + optional `mult` for divide-by-K patterns), UNRESOLVED. `pass_main.cpp::build_metadata_json` populates in 4-priority order: STATIC →
  `dim_param` name match → SSA-traced origin with composed mult → fail loudly. The SSA-traced fallback is what lets the EP serve models whose output `dim_param`s don't match any input — Gemma-3 vision's `num_image_tokens` vs `num_images`, Qwen3.5 vision's `num_logical_patches =
  num_patches/4` — without any on-disk model edit.

  **Runtime resolution**
  - `marshal_output_tensors` reads DimSource per output dim and computes `round(inputs[input_idx].shape[dim_idx] * mult)`.
  - OGA `past_present_share_buffer` override takes the past input's pre-allocated shape as the present output's shape when the past buffer is strictly larger than the DimSource-resolved value — keeps `past_key == present_key` pointer identity for in-place GQA append.
  - Standalone fp16 Softmax runtime + kernel for the vision projector's softmax (text decoders fuse softmax inside `hip.gqa` and don't reach this path).

  **Vision compute coverage**
  - `FastGeluFusion`: new `InlinedFastGeluVisionToGelu` variant for the Gemma-3 SigLIP MLP (Add instead of Sum, `Mul(x, Mul(x, x))` instead of `Pow(x, 3)`, etc.).
  - `ProjectorOpsRewrites`: decomposes `Pow(x, c)`, `ReduceMean`, `AveragePool` (K=stride, no pad, NCHW); rewrites broadcasting `onnx.Div` into `Mul(x, Reciprocal(y))` so the RMSNorm divisor broadcasts via the existing `miopenOpTensor` path. Integer-typed broadcast Div has no
  Reciprocal-based decomposition — the rewrite emits a clear compile-time warning instead of silently letting the runtime OOB-read on batch > 0.

  **Runtime correctness fixes — five orthogonal latent bugs, each is a class the next vision encoder will hit**
  1. **Conv runtime hardcoded `miopenFloat` for all three descriptors.** With fp16 buffers, MIOpen reads with fp32 strides → conv output pinned at fp16-max with ~1.6% NaN/Inf. Fix: thread an `int64_t data_type` parameter, derive it from the output memref element type.
  2. **Conv bias passed in but never applied.** `miopenConvolutionForward` doesn't add bias; `miopenConvolutionForwardBias` only overwrites y. Fix: explicit `miopenOpTensor(miopenTensorOpAdd, ...)` after the conv.
  3. **`hip_reduce_sum` only sums a trailing contiguous suffix.** `ReduceMean` over `[3, 5]` of a 6-D `[B, C, H/K, K, W/K, K]` tensor is non-trailing; the kernel silently summed the wrong block. Fix in the lowering, not the kernel: `AveragePoolToReshapeMean` emits a
  `Transpose(perm=[0,1,2,4,3,5])` between the Reshape and the ReduceMean.
  4. **`hip.div` does not broadcast.** Broadcasting RMSNorm divisor `(?, 256, 1)` divided into `(?, 256, 1152)` walks past the end of the rhs buffer into uninitialised pool memory. Fix: `BroadcastDivToMulReciprocal` rewrite (Reciprocal is element-wise, subsequent Mul broadcasts).
  5. **hipBLASLt batched matmul stride wrong for broadcast B.** Cache unconditionally set `STRIDED_BATCH_OFFSET = K*N` on layA — correct for per-batch `[B, K, N]`, wrong for rank-2 broadcast `[K, N]` (and rank-N `[1, ..., 1, K, N]`). Image 0 correct, image 1+ entirely NaN. Fix:
  new `b_batch_stride : int64` ABI parameter; compiler derives 0 vs K*N from B's leading-dim product (compile time when static, runtime select when dynamic); `MatmulCacheKey` is keyed on the int stride so any future site using a stride other than `{0, K*N}` also gets its own
  cache entry.

  These five had been latent through the entire LLM suite — LLMs have no Conv, no non-trailing-axis ReduceMean, no broadcasting Div, and never combine `batch > 1` with a rank-2 broadcast weight in the same MatMul.

  ## Test plan

  - [x] `lit` suite — 201/201 pass, including:
    - `test_infer_onnx_shapes.mlir` — per-rule type refinement coverage (Reshape, Transpose, MatMul outer-batch alignment, Cast, unary, binary broadcast).
    - `test_infer_onnx_shapes_trace_mult.mlir` — pins K=3 Reshape mult precision (mult=1/3, bit pattern `0x3FD5555555555555`).
    - `test_projector_rewrites.mlir` — pins `AveragePoolToReshapeMean`'s `perm=[0,1,2,4,3,5]` Transpose + `BroadcastDivToMulReciprocal` emitting `hip.reciprocal`.
    - `test_split_invalid.mlir` — `not hip-mlir-opt | FileCheck` on the new surviving-op error message.
    - `test_matmul.mlir` — both rank-2 broadcast B and the rank-3 leading-one B `[1, K, N]` cases assert `b_batch_stride = 0` via `CHECK-NOT: llvm.icmp / llvm.select`.
  - [x] `TestGemma3_4BVisionDynShape` — 6/6 pass, including the new `test_vision_num_images_1_matches_cpu_cosine` (≥ 0.999 cosine vs CPU; measured 0.999987 on gfx1151). The cosine test is the value-vs-CPU floor — bit-equality and shape tests would silently pass for kernels
  producing consistent-but-wrong outputs.
  - [x] `test_qwen3_5_9b.py` — micro-test for Qwen3.5 vision's patch-merger Reshape divide-by-K validates the InferOnnxShapes trace → module attribute → `CompilationOutputs` buffer → `DimSource.mult` → `marshal_output round(input * mult)` chain end-to-end at three input sizes.
  - [x] Per-model regression sweep on gfx1151 — 86 passed / 33 skipped across `test_llama1b`, `test_llama8b`, `test_llama8b_asym`, `test_mistral7b_v3`, `test_phi4_14b`, `test_qwen2_5_14b`, `test_qwen3_5_9b`, `test_gemma3_4b`, `test_gptoss20b`. No regressions vs the LLM baseline.


3rd-party/custom_kernels/hip/softmax_kernel.hip
```diff
@@ -0,0 +1,130 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
```
fhanuman, line 3: Hi @akholodnamdcom , Thanks for making these changes. Please refer to the contribution guidelines on creating large PRs. Can we split this PR into smaller PRs: one for new-operator-support and one for enhancements -- this can include new MLIR passes and any other changes that can be decoupled --  to make the PRs more focused? This will help review the PR better.

fhanuman (changes requested):

PR #242, opened 2026-05-21
feat(build): auto-detect GPU architecture and match TheRock distribution

bcrane-amd:
Enhances build.py to automatically detect the GPU architecture and download the matching TheRock distribution, eliminating the mismatch between hardcoded URL (gfx1151) and detected hardware (gfx1150).

Changes:
- Add THEROCK_DIST environment variable support for custom TheRock paths
- Add GPU detection via wmic command-line tool (no Python dependencies)
- Add GPU detection via WMI Python module (fallback if wmic fails)
- Add TheRock architecture probing (scans TensileLibrary_lazy_gfx*.dat files)
- Add device ID to architecture mapping (gfx1150, gfx1151)
- Validate detected GPU matches TheRock supported architectures
- Add helpful reminder at build completion to set THEROCK_DIST and PATH

Benefits:
- Works out-of-box on fresh Windows install (wmic is built-in)
- Supports multi-arch development (multiple TheRock folders via THEROCK_DIST)
- Prevents "Result too large" errors from architecture mismatches
- Better error messages when GPU/TheRock mismatch detected

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

## Notes

If THEROCK_DIST is not already set, the current version of build.py will download the gfx1151 version of therock.  When on a gfx1150 machine, the build will be successful, but running the binaries will fail.  The changes in this PR will probe for the installed gpu, to download the appropriate version of therock for the machine.  

If THEROCK_DIST is already set, the changes in this PR will probe therock at that location, to determine gfx1150/gfx1151 support. 

TODO: build.py looks for a file ".ok" in therock folder.  Without that file, build.py may decide to remove therock folder and download it again.  I think this applies only when THEROCK_DIST is not set, but I don't remember testing that situation.


akholodnamdcom (dismissed):
We could have kept it here for some time. Later we'd anyway have to fetch a rock distro per every gfx and build a backend.dll per every arch we support. 

Very first backend for reference is here:
https://github.com/ROCm/onnx-hipdnn-ep/pull/223

bcrane-amd:
The Checks are failing, I asked Claude about that, and was recommended to add this comment to the PR:

Note: The CI check fails due to missing MORPHIZEN_DEPLOY_KEY secret (required for private submodule access). This is expected for fork PRs. The build.py changes have been tested locally and work correctly.

amd-mingw:
> The Checks are failing, I asked Claude about that, and was recommended to add this comment to the PR:
> 
> Note: The CI check fails due to missing MORPHIZEN_DEPLOY_KEY secret (required for private submodule access). This is expected for fork PRs. The build.py changes have been tested locally and work correctly.

Hi @bcrane-amd  @akholodnamdcom, about the CI, the root cause is ROCm access limit, the workaround way is(other PRs used currently): not fork the repo, directly create branch on this repo. this is known issue.

amd-mingw:
comment about this PR: not recommend involve hardware detection here, the better solution is this EP build one time for all gpu architecture, the idea is: all the architecture dependency kernel move to therock, and therock as a driver level dependency. this is in-plan.  
https://github.com/ROCm/onnx-hipdnn-ep/pull/237 is the first step, move gpu architecture dependency custom kernel to the 3-party repo, the next step is move it to therock, and therock release is gpu architecture/OS dependency driver level package.

bcrane-amd:
> comment about this PR: not recommend involve hardware detection here, the better solution is ... in-plan. 

True, hardware-detection here is not so good.  But, that happens only when THEROCK_DIST is not set.  
.................................................................
The real issue is that build.py will always download its own copy of TheRock, whereas the Quick Start instructions tell to download TheRock first, then set THEROCK_DIST to that folder.  This means there will be 2 copies of TheRock.  Someone could be building TheRock locally, but onnx-hipdnn-ep would be building against its downloaded-copy of TheRock. 

This PR makes build.py check THEROCK_DIST and avoid downloading its own copy, so onnxruntime + onnxruntime-genai + onnx-hipdnn-ep are all built using the same copy of TheRock. 
.................................................................
The hardware-detection only happens when THEROCK_DIST is not set.  That is the case only when someone wants to download/build/test onnx-hipdnn-ep in isolation.  

In real-world usage, onnx-hipdnn-ep needs to be built using the same copy of TheRock that all other modules are built with, which is indicated by setting THEROCK_DIST.  That means the hardware-detection code will not be used in real-world usage.  
.................................................................
So, the hardware-detection code is really just a convenience, for working with onnx-hipdnn-ep in isolation.  For real-world usage, THEROCK_DIST will be set, and the hardware-detection code will not be used.

bcrane-amd:
Close this, replaced by https://github.com/ROCm/onnx-hipdnn-ep/pull/255

PR #246, opened 2026-05-25
feat(conversion): collapse Range tensor<1xT> operands to rank-0 scalars

cliu1003:
## Motivation

ORT/HF often export Range start/limit/delta as tensor<1xT> while hip.range expects rank-0 tensors. So submitting this PR to support tensor<1xT>.

### lib/Conversion/OnnxToHip/RangeConversion.cpp
Lower tensor<1xT> via tensor.collapse_shape before hip.range and keep compile-time delta==0 checks on the original operands.
### include/hip/InitAllPasses.h
Register arith BufferDeallocationOpInterface so the full bufferization pipeline succeeds after introducing collapse_shape on this path.
### test/lit/Conversion/onnx-to-hip/test_range.mlir & test/lit/e2e/test_range_model_1d.mlir
Add onnx-to-hip and e2e LIT coverage for static 1x1 operands.

## Test Plan
- [x] onnx-to-hip test
- [x] e2e LIT

### DIFF
--- a/lib/Conversion/OnnxToHip/RangeConversion.cpp
+++ b/lib/Conversion/OnnxToHip/RangeConversion.cpp
@@ -187,24 +187,44 @@ struct RangeToHip : public RewritePattern {
     if (!elemTy.isIntOrFloat())
       return rewriter.notifyMatchFailure(op, "unsupported element type");
 
-    for (Value v : op->getOperands()) {
-      auto t = dyn_cast<RankedTensorType>(v.getType());
-      if (!t || t.getRank() != 0 || t.getElementType() != elemTy)
-        return rewriter.notifyMatchFailure(
-            op, "expected 0-D operands matching result element type");
-    }
-
     Location loc = op->getLoc();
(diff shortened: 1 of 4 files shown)


mingyueliuh (approved):

lib/Conversion/OnnxToHip/RangeConversion.cpp
```diff
@@ -187,24 +187,44 @@ struct RangeToHip : public RewritePattern {
     if (!elemTy.isIntOrFloat())
       return rewriter.notifyMatchFailure(op, "unsupported element type");
 
+    Location loc = op->getLoc();
+    SmallVector<Value, 3> operands;
     for (Value v : op->getOperands()) {
+      // Operand normalization: tensor<1xT> is collapsed to 0-D tensor<T> via
```
fhanuman, line 193: Just letting you know,   `tensor.extract + tensor.from_elements` combination will achieve the same

fhanuman (approved):

PR #253, opened 2026-05-26
fix(onnx-to-hip): support onnx.Loop with cond_init=none

zz002:
## Problem

ONNX Loop spec marks operand 1 (`cond_init`) as optional; importers materialize a missing cond_init as `onnx.NoValue` (none type). ONNX semantics when absent: cond stays true forever, only `max_trip_count` terminates the loop.

[`lib/Conversion/OnnxToHip/LoopOutline.cpp`](onnx-hipdnn-ep/lib/Conversion/OnnxToHip/LoopOutline.cpp) initial-version head comment explicitly listed missing-cond as an unsupported limitation, while every downstream layer was already cond=missing-ready:

- `Hip_LoopOp` dialect declares `cond_init` as `Optional<I1>`
- `LoopOpLowering` treats null `condInit` as i1-true
- runtime `hipdnn_ep_run_counted_loop` accepts any bool

Only the outlining-pass entry was the trip-wire.

Reference test case: Qwen3.5-9B vision encoder has 27 `onnx.Loop` nodes, all with `cond_init=missing`. Without this fix `perf_test` aborts at `LoopOutline::unboxCondInit` (`could not unbox cond_init to i1, got 'none'`) for every Loop in the graph.

## Solution

`lib/Conversion/OnnxToHip/LoopOutline.cpp` (~17 lines, no API change):

- Detect `condInitIsNone = isa<NoneType>(condInitTensor.getType())` after the operand binding step
- When cond_init is none, synthesize `arith.constant true : i1` for the hip.loop op's `cond_init` operand and force `cond_is_passthrough` so `LoopOpLowering` picks the fast `hipdnn_ep_run_counted_loop` path (no per-iter cond readback). Matches ONNX spec semantics: missing cond means cond_out is ignored and only `max_trip_count` terminates the loop
- Update file-header `Limitations` comment: missing-cond is now supported; missing-M (while-style) remains unsupported and continues to reject with a clean diagnostic
- LoopOutline call site uses `condI1Val` (new local) instead of `*condI1` so both branches share the downstream `LoopOp::create` call

`test/lit/Conversion/onnx-to-hip/test_loop_outline.mlir`: new Test 4 covering `cond_init=onnx.NoValue`: outlines into hip.loop with cond synthesized to `arith.constant true`, `cond_is_passthrough` UnitAttr set, body func skips the cond return slot. Added in-place to the existing 3-test fixture (i1 / ui8 / dynamic-cond) following the file's per-variant convention.

## Verification

- Full repo LIT: 203/203 PASS, including `test_loop_outline.mlir` 4 sub-tests + `test_loop_outline_invalid.mlir` + `hip-to-llvm/test_loop.mlir` (HipToLLVM Loop lowering)
- Local perf_test on Qwen3.5-9B `vision.onnx` (875 MB, 27 Loop, full dynamic-shape pipeline): the `could not unbox cond_init to i1` error is confirmed gone for all 27 Loop nodes. The next failure layer is unrelated (importer `value_info` fallback on Loop result types; tracked separately in ROCm/MorphiZen#225) and out of scope here

## Test plan

- [ ] Windows build CI green
- [ ] Linux build CI green
- [ ] gpu-perf-accuracy nightly does not regress on already-supported models (this PR's fix only touches the Loop-bearing path; non-Loop graphs go through the unchanged `else` branch in LoopOutline)

### DIFF
--- a/lib/Conversion/OnnxToHip/LoopOutline.cpp
+++ b/lib/Conversion/OnnxToHip/LoopOutline.cpp
@@ -23,9 +23,14 @@
 //   6. Replace the onnx.Loop op with hip.loop and erase it.
 //
 // Limitations of this initial version:
-//   - Requires M (trip count) and cond_init both present.  Missing-M
-//     (while-style infinite loop) and missing-cond (unconditional counted
-//     loop) variants are rejected with a clean diagnostic for now.
+//   - Missing-M (while-style infinite loop) is rejected.  Counted loops
+//     with `cond_init` absent (`onnx.NoValue` / `none` type) are
+//     supported: cond is synthesized as `arith.constant true : i1` and
+//     `cond_is_passthrough` is forced so the runtime takes the fast
+//     `hipdnn_ep_run_counted_loop` path (no per-iter cond readback).
+//     Per ONNX spec, missing cond means the loop runs M times with no
+//     conditional early exit -- the body's `cond_out` (if any) is
+//     ignored.
 //   - Scan outputs (results beyond the loop-carried slot count) are not
 //     yet supported; rejected with a clean diagnostic.
 //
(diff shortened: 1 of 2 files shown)


fhanuman (approved):
LGTM

PR #254, opened 2026-05-26
feat: enable Qwen3.5-9B embedding + text on MorphiZen EP (dynamic shapes)

AMDmoore:
## Summary

Enables the **embedding model** and the **text decoder model** of Qwen3.5-9B (and structurally similar VLMs — Gemma3, etc.) to run end-to-end on the MorphiZen EP with dynamic batch / sequence shapes. This unblocks the multimodal VLM pipeline on top of the dynseqlen base.

Six commits, grouped into three areas:

| Area | Commits | What it does |
|---|---|---|
| NonZero / ScatterND runtime contract | `c0ccb345`, `5770e18d` | NonZero emits a 1-element `count_buf` output; ScatterND backtraces through `Transpose/Cast/Reshape` to find it and uses it as `valid_count` so only real rows are scattered. NonZero output is zero-initialized to make the unused tail well-defined. |
| Embedding model bit-exact (VLM pipeline) | `2bc84f88` | Five independent EP-side correctness fixes (Equal scalar broadcast, ordered NonZero scan, partial-fill Slice, ScatterND `count_ptr` honoring, sidecar shape inputs in hip-onnx-runner). Verified L2 = 0 vs ORT CPU on a six-shape sweep of the full Qwen3.5-9B `embedding.onnx`. |
| Text decoder dynamic-shape compilation | `1f9cce07`, `7e313771`, `2a60c94f` | Three classes of dynamic-shape compilation errors in the OnnxToHip → bufferize → HipToLLVM pipeline, all triggered by Qwen3.5-9B mrope's `Reshape(Range(0, bs*ss, 1), [bs, ss])`. See below. |

## Background

The text decoder of Qwen3.5-9B (and other Qwen3 VLMs) materialises position_ids via `Range(0, batch_size * seq_len, 1)` followed by `Reshape(_, [batch_size, seq_len])`. This is the first model in our zoo that produces:
- a 1-D dynamic `Range` whose start/limit/delta come from rank-1 size-1 tensors fed via `tensor.from_elements` / scratch buffer loads,
- a `Reshape` to a 2-D output where **both** dims are dynamic SSA values inside one reassociation group,
- downstream `memref.copy` ops with **dynamic strides** on the source (per-step strided KV writes).

Each of these tripped a distinct point in the lowering pipeline. Below is the narrative of the three compilation errors and how this PR fixes them.

### 1. `hip-pool-allocs`: "operand #0 does not dominate this use"

**Symptom.** Compilation fails inside `hip-pool-allocs` Phase 4 (hoist allocs to the function prelude). The dynamic-length `memref.alloc` for `onnx.Range`'s output references SSA values that are defined later in the function body than the prelude — typically `memref.load %scratch[%c0]` of `start/limit/delta`.

**Why.** Three independent gaps:
- Range's start/limit/delta inputs are rank-1 size-1 tensors. The old converter extracted them via `tensor.extract`, which bufferizes to `memref.load %scratch[%c0]`. `memref.load` was not on the hoist whitelist, so the count-formula chain (`ceildivsi`, `cmpi`, `select`, …) could not be lifted.
- The Range count formula uses ops that were also not on the hoist whitelist (`cmpi`, `and`/`or`, `select`, `maxsi`/`minsi`, sign-extension casts) — even after the scalars were available, the formula itself couldn't be hoisted.
- `resolveDimAtSource` / `foldDimOfReshape` stopped at `memref.subview` and `memref.cast` chains introduced by packed-QKV split and strided KV outputs, so `memref.dim` ops on those views were never folded to either a constant or a `dim` of the root alloc.

**Fix.**
- **OnnxToHip — hybrid two-phase constant externalisation.** Phase 1 leaves inline `onnx.Constant`s below 64 elements in place so converters can read their `value` attr directly as SSA. Phase 2 (after `convertComputeOps`) externalises whatever survived. Threshold sits between converter literals (≤ 16 elems) and the smallest real weight tile (≥ 1024 elems) — no overlap.
- **RangeConversion.** Peek through value-preserving reshape views to the scalar producer and materialise start/limit/delta as pure-SSA `arith.constant` from the inline `value` attr. Falls back to `tensor.from_elements` → rank-0 tensor → `tensor.extract` in order.
- **PoolAllocs.**
  - New `foldScalarMemrefArith` walks scalar `memref.load`s back through the most recent aliasing writer to produce pure SSA arith. Aliasing detected via `BufferOriginAnalysis`, so multiple `reinterpret_cast`s of one scratch slot are unified.
  - `isHoistable` whitelist expanded with `ceildivsi`, `cmpi`, `and`, `or`, `select`, `maxsi`, `minsi`, `extsi`, `extui`, `trunci`, `memref.get_global`, and `memref.load` of session-immutable globals (carries `constant` or `hip.external_data` attr).
  - `resolveDimAtSource` and `foldDimOfReshape` worklists extended to `memref.subview` (reads i-th size operand, handles rank reduction) and `memref.cast` (recurses into the source).
- **`arith` BufferDeallocation external model** registered in `InitAllPasses.h` and `hip-mlir-opt.cpp` — required when scalar `arith.constant` / `arith.select` survive into bufferized IR.

### 2. `expand-strided-metadata`: "There must be at most one dynamic size per group"

**Symptom.** After (1) is fixed, the next compilation hits an upstream MLIR assertion in `memref::createExpandStridedMetadataPass`:

```
"There must be at most one dynamic size per group"
```

… on the `memref.expand_shape %src [[0, 1]] output_shape [%bs, %ss]` that Reshape now emits for `[bs, ss]`. The op carries explicit `output_shape` operands that fully determine every dim and is otherwise valid — the assertion is a known upstream limitation.

**Why.** This was actually masked by a second bug in `ReshapeConversion`: its `output_shape` builder fell back to "input total size for every dynamic dim in the group", producing the **same** SSA value for both `[bs, ss]`. The resulting expand_shape claimed an `[N, N]` layout backed by an `N`-element buffer — semantically broken, but it happened to have only one distinct dynamic SSA value per group, which appeased upstream's assertion. Once Reshape was corrected to emit two distinct dynamic SSA values, the upstream assertion fired.

**Fix.**
- **ReshapeConversion.** For reassociation groups with 2+ dynamic output dims, resolve each dim from the ONNX `shape` operand directly. Handles `tensor.from_elements`, compile-time constant tensors, and `onnx.Concat` of rank-1 size-1 elements. Single-dyn groups (Unsqueeze, intermediate expand inside Reshape decomposition) keep the existing input-dim formula.
- **New pass `hip-relax-multi-dyn-expand-shape`.** Matches exactly the multi-dyn-per-group `expand_shape` case and rewrites it as

  ```
  memref.reinterpret_cast %src to
      offset: [0], sizes: [%bs, %ss], strides: [%ss, 1]
      : memref<?xi64> to memref<?x?xi64>
  ```

  Sizes/strides come from the `output_shape` operands and the source's static layout. Single-dyn `expand_shape` and all `collapse_shape` are untouched and continue through the upstream `ExpandStridedMetadata` path. Pipeline position: very first sub-pass of `buildHipToLLVMPipeline`, before `createExpandStridedMetadataPass`.
- **Retirement plan.** When upstream `expand-strided-metadata` is upgraded to consult `output_shape`, this pass becomes a no-op and can be deleted.

### 3. `lld-link`: `undefined symbol: memrefCopy`

**Symptom.** After (1) and (2) are fixed, the model compiles through every MLIR pass and emits LLVM IR, but `lld-link` fails:

```
undefined symbol: memrefCopy
```

… and the EP silently falls back to ORT CPU for the entire graph. This is invisible to accuracy tests that compare CPU-vs-CPU.

**Why.** After rewriting multi-dyn `memref.expand_shape` into `memref.reinterpret_cast` (fix 2 above), downstream produces `memref.copy` ops where the source has dynamic strides — e.g. a per-step KV write view of a strided buffer. The upstream `MemRefToLLVM` lowering's dynamic-strided code path falls back to the `memrefCopy` runtime libcall from MLIR runner utilities, which is not linked into our compiled model DLLs.

**Fix — extend `MemRefCopyOpLowering` to lower the dynamic-stride case directly when the destination is a row-major identity layout** (the only destination shape that survives this pipeline for memref.copy today). Dispatch:

| Source shape | Lowering |
|---|---|
| both src/dst contiguous (last-dim stride == 1, dst row-major identity) | single `wrap_hipMemcpyAsync` of total bytes |
| non-contiguous source with contiguous innermost dim | `wrap_hipMemcpy2DAsync` (pitched 2-D copy) |
| non-unit innermost stride | fall through to the upstream pattern (will still fail to link — but no such op survives the pipeline today; left as a guard) |

Both `wrap_hipMemcpyAsync` and `wrap_hipMemcpy2DAsync` already exist in the runtime and are declared `extern "C"`, so generated LLVM IR links cleanly without pulling in `memrefCopy`.

### LIT (compiler)

- `test/lit/Dialect/hip-relax-multi-dyn-expand-shape.mlir` — 2-dyn rewrite, single-dyn passthrough, all-static passthrough, collapse_shape untouched, multi-group with one offending group only rewrites that group.
- `test/lit/Conversion/hip-to-llvm/test_memref_copy.mlir` — static dense, static pitched, dynamic rank-1, dynamic 2-D pitched, the Qwen-style `reinterpret_cast` source pattern, and a negative non-unit-stride passthrough.
- `test/lit/Conversion/onnx-to-hip/test_expand_shape_trace.mlir` — new for the Expand placeholder-shape path.
- `test/lit/Conversion/hip-to-llvm/test_equal.mlir` — extended for scalar-broadcast Equal.
- Full lit suite: **202 / 202 passing**.

### Full embedding model — bit-exact

`cursor_space/_sweep_qwen35_embedding.bat` runs `hip-onnx-runner` on the **full Qwen3.5-9B `embedding.onnx`** (every op, end-to-end — not a stripped sub-graph) across a six-shape sweep (`seq_len ∈ {1, 4, 8, 16, 32, 64}` × placeholders ∈ {0 … 16}). All six shapes report **L2(EP, ORT CPU) = 0** on `output_0_inputs_embeds_fp16` (524288 fp16 elements at the largest shape). The eight expected wrappers dispatch in order: `equal → gather → expand × 2 → nonzero → transpose → slice → scatter_nd`. `HIPDNN_EP_DEBUG=1` confirms the NonZero kernel-side `actual_count` readback matches `placeholders × hidden_dim`.

### Full text decoder model — compile + single-step accuracy

- `hip-onnx-runner` on the **full Qwen3.5-9B text decoder ONNX** (end-to-end, every op): compiles cleanly through every MLIR pass and links into a model DLL, **2783 `[REAL] wrap_*` GPU dispatches per inference**, zero `memrefCopy` link errors, no CPU fallback.
- CPU-vs-MorphiZen EP single-step crosscheck on the same full text decoder: NaN-free, logits L2 = 2.15 over 248K elements (RMS ≈ 0.004), all 65 outputs (logits + 32 layers of present K/V) within expected fp16 drift.

### OGA E2E

The OGA multi-step E2E pipeline (vision + embedding + text decoder, all on MorphiZenEP) currently shows **degraded output quality** vs the pure-ORT-CPU baseline (repetition patterns, occasional garbled tail), despite per-step logits being numerically close. This is being tracked separately and is **out of scope for this PR** — root causes under investigation include `[REAL] unsupported data_type 4 for MIOpen` warnings (uint8 path in MIOpen wrappers) during embedding on MorphiZenEP. The compilation-error fixes in this PR are necessary but not sufficient for OGA E2E bit-fidelity; that work continues on top of this base.

## Known limitations / out of scope

- OGA multi-step decode accuracy regression (see Tests § OGA E2E). To be fixed in a follow-up PR — likely in MIOpen wrappers' uint8 support and/or one of the embedding paths exercised only by OGA (different attention_mask layout, KV alloc).
- `hip-relax-multi-dyn-expand-shape` is a local workaround for an upstream MLIR limitation. When upstream `expand-strided-metadata` learns to consult `output_shape`, the pass and its lit test can be removed without changing behavior.
- The non-unit-innermost-stride `memref.copy` case still falls through to the upstream `memrefCopy` pattern and would fail to link if it ever appeared. No such op is produced by any model in our current zoo; left as an explicit guard so future regressions surface as link errors rather than silent corruption.

## Files changed (high level)

- `lib/Conversion/OnnxToHip/{OnnxToHip,RangeConversion,ReshapeConversion,ExpandConversion,ShapeConversion}.cpp` + `OnnxToHipUtils.h`
- `lib/Conversion/HipToLLVM/{EqualLowering,MemoryLowering}.cpp`
- `lib/Dialect/Transforms/{PoolAllocs,RelaxMultiDynExpandShape,Pipelines}.cpp` + `Passes.td` + `CMakeLists.txt`
- `include/hip/InitAllPasses.h`, `tools/hip-mlir-opt/hip-mlir-opt.cpp`
- `3rd-party/custom_kernels/hip/{elementwise_binary,expand,nonzero}_kernel.hip` + `hip_custom_kernels.h`
- `lib/Runtime/{hipdnn_ep_runtime.h, real/{equal,expand,slice,scatter_nd,nonzero}.cpp, mock/mock_gpu.cpp}`
- `tools/hip-onnx-runner/hip-onnx-runner.cpp` (shape sidecar reading)
- `test/lit/{Dialect/hip-relax-multi-dyn-expand-shape.mlir, Conversion/{hip-to-llvm/test_memref_copy.mlir, hip-to-llvm/test_equal.mlir, onnx-to-hip/test_expand_shape_trace.mlir}}`
- `CLAUDE.md` (consolidated five-fix embedding-bit-exact gotcha)

## Follow-ups (separate PRs)

- OGA E2E accuracy: triage `data_type 4` MIOpen path, multi-step decode repetition pattern.
- Retire `hip-relax-multi-dyn-expand-shape` when upstream is upgraded.
- Documented memref.copy strategy (`docs/design/split-copy-lowering.md`) — drafted locally, to be sent in a docs-only PR.


3rd-party/custom_kernels/hip/nonzero_kernel.hip
```diff
@@ -0,0 +1,166 @@
+/*
```
fhanuman, line 1: Hi @AMDmoore, Thanks for making these changes. Please refer to the contribution guidelines in the comments above. Can we split this PR into smaller PRs: one for new-operator-support and a separate PR for the bug-fixes section, to make it more focused? This will help improve the review of the PR. It would be good to create a separate PR for enhancements as well.

fhanuman (changes requested):

3rd-party/custom_kernels/hip/nonzero_kernel.hip
AMDmoore: Hi @fhanuman, thanks for your comment.
Could you review this PR by separate commit? I remember I split each function part by a commit, first 2 commits for nonzero op support, then 1 commit for enabling embedding model, the last three for text model issue fixes and each for one or two issue, so I think that's better for review.
Because of the remaining accuracy work, it's easier for others to continue the work in one PR, they don't need to collect PR 1, 2, 3..., but I agree and will submit small PRs in the future.

fhanuman: Hi @AMDmoore, thanks for the clarification. There are three PRs (including this one) that I am currently reviewing, with changes totaling over 3000 LOC; it's going to be very hard to review. So I would prefer we split the PR to make it more focused, rather than combining several different changes. Also, collecting stacked PRs isn't much work — and I'd be more than happy to help split this change into multiple PRs if that's useful.

amd-yanjunz:
ROCm OGA CI finished with result: FAILURE.

- PR: ROCm/onnx-hipdnn-ep#254
- Source comment: [modelzoo=LLM_ROCm_dynamic](https://github.com/ROCm/onnx-hipdnn-ep/pull/254#issuecomment-4562249996)
- TEST_MODE: both
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/524/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hipdnn_ep_oga&id=6a18078959066c8ed568fb86

PR #260, opened 2026-05-29
feat(hip): shape inference infrastructure for HIP DPS ops (verifier + reify + propagation pass) -- 1/n

fhanuman:
Landing shape inference infrastructure for HIP DPS ops in a 6-PR cascade (#260-#265). This first PR adds three pieces: a static verifier on `hip.matmul`, a default `reifyResultShapes` body shared by every DPS op via a new in-dialect `HipDpsOpInterface`, and a module-level `--hip-infer-shapes` pass that refines `?` dims and rebuilds `tensor.empty` producers. `hip.matmul` is the worked example for both the bespoke-runtime-reify opt-out (`autoReify=0`) and the auto-emitted `inferReturnTypes` opt-in (`autoInfer=1, declareInfer=1`); the dialect's remaining 51 single-result DPS ops inherit the auto-emitted body when #262 flips the default.

**LOC delta (net): +959 source, +591 doc, +519 test.**

**Example MLIR that previously stayed dynamic** (chained matmul, both `tensor.empty` outs operands sized with dynamic SSA values that `convert-onnx-to-hip` could not statically resolve):

```mlir
// Before --hip-infer-shapes:
%e1 = tensor.empty(%d1, %d2) : tensor<?x?xf16>
%y1 = hip.matmul ins(%a, %b : tensor<2x4xf16>, tensor<4x8xf16>)
                 outs(%e1 : tensor<?x?xf16>) : tensor<?x?xf16>
%e2 = tensor.empty(%d3, %d4) : tensor<?x?xf16>
%y2 = hip.matmul ins(%y1, %c : tensor<?x?xf16>, tensor<8x16xf16>)
                 outs(%e2 : tensor<?x?xf16>) : tensor<?x?xf16>

// After:
%e1   = tensor.empty() : tensor<2x8xf16>          // dyn operands dropped
%y1   = hip.matmul ins(%a, %b : tensor<2x4xf16>, tensor<4x8xf16>)
                   outs(%e1 : tensor<2x8xf16>) : tensor<2x8xf16>
%cast = tensor.cast %y1 : tensor<2x8xf16> to tensor<?x?xf16>  // barrier
%e2   = tensor.empty() : tensor<2x16xf16>
%y2   = hip.matmul ins(%cast, %c : tensor<?x?xf16>, tensor<8x16xf16>)
                   outs(%e2 : tensor<2x16xf16>) : tensor<2x16xf16>
```

The second matmul's M dim is recovered as `2` even though `%cast` keeps its `ins`-edge type at `tensor<?x?xf16>` — `tensor::DimOp::fold` looks through the cast to the static-typed source, so reify recovers static dims without whole-chain narrowing.

**The fix:** four pieces sharing one shape helper for matmul, plus a TableGen-driven default reify for the rest of the dialect.

* `mlir::hip::inferMatmulShape` — shape contract `[..., M, K] @ [..., K, N] -> [..., M, N]`. Used by both `MatmulOp::verify` (driven by `verifyHipOpShape`, which compares the helper's output against the op's `outs` types with `kDynamic` as wildcard) and `MatmulOp::reifyResultShapes`.
* `HipDpsOpInterface` — new in-dialect interface owns the shared default `reifyResultShapes` body (walks `getDpsInits()`, lifts each init via `tensor::getMixedSizes` / `memref::getMixedSizes`). The parameterized `Hip_DpsOp` TableGen base auto-emits per-op dispatchers via `extraClassDefinition`, gated by an `autoReify` bit. 52 of 53 DPS ops gain a working `reifyResultShapes` for free; only `hip.matmul` opts out (`autoReify=0`) to keep its bespoke runtime body in `HipReifyResultShapesImpl.cpp` (recovers M/K/N from operand shapes, lifts each via `reifyDimOrConstant`).
* `MatmulOp::inferReturnTypes` — auto-emitted construction-time body via `autoInfer=1, declareInfer=1` (single opt-in on this PR; #262 flips the default for the remaining 51 single-result ops). The `MatMulConversion.cpp` callsite drops the explicit `resultType` from `Op::create` and uses the inferred-type overload. The two layers are orthogonal: bespoke runtime reify (recovers dims from operand shapes) coexists with auto-emitted construction-time `inferReturnTypes` (the typed outs operand already encodes the result shape, so the body just reads it back).
* `--hip-infer-shapes` — module-level pass after `convert-onnx-to-hip` (`Pipelines.cpp::buildOnnxToHipPipelineTail`). Walks reify-interface ops, refines `?` dims in place, rebuilds `tensor.empty` producers so the DPS contract `result_type == outs_type` stays satisfied. A `tensor.cast` on every non-DPS-init use is the per-op propagation barrier (rationale + pinned test in the [design doc](https://github.com/ROCm/onnx-hipdnn-ep/blob/fhanuman.shape_inference/docs/design/hip-shape-inference.md#per-op-refinement-not-whole-chain)).

LIT split into producer (`hip-matmul-reify-shapes.mlir`, driven by `--resolve-shaped-type-result-dims` — **not** in our pipeline, but the only consumer that materialises reify branches into FileCheck-able IR) vs consumer (`hip-infer-shapes.mlir`, driven by `--hip-infer-shapes` itself), plus a new [`hip-dps-op-interface.mlir`](https://github.com/ROCm/onnx-hipdnn-ep/blob/fhanuman.shape_inference/test/lit/Dialect/hip-dps-op-interface.mlir) that pins the interface-default body on `hip.silu` (static / dynamic / mixed shape). 206/206 LIT pass. Two pre-existing tests (`hip-lower-allocs.mlir`, `hip-optimize-memrefs.mlir`) are tightened where the old verifier accepted matmul shapes the new rank contract correctly rejects.

P.S. Rank-1 ONNX `MatMul` operands are explicitly rejected by the verifier today; the fix belongs in `MatMulConversion.cpp` (`tensor.expand_shape` to rank-2 around the `hip.matmul`). No supported model exercises this case.

**Stack overview** — this PR is the foundation; the rest of the cascade fans out from here. Each PR is independently shippable.

* **#260 (this PR):** `HipDpsOpInterface` + parameterized `Hip_DpsOp` base; verifier + reify + `--hip-infer-shapes` for `hip.matmul`.
* **#261:** `InferTypeOpInterface` on `Hip_LoopOp` (region op, orthogonal scaffolding).
* **#262:** flips `autoInfer=1, declareInfer=1` defaults; introduces `Hip_DpsOp_SameShape` sub-base; migrates `rope` / `rms_norm` and the matmul-shape-contract ops (`gemm`, `qmoe`, `matmul_nbits`) onto the auto-emitted `inferReturnTypes` body.
* **#263:** introduces `Hip_DpsOp_Broadcast` and migrates the 23 broadcast + same-shape elementwise ops.
* **#264:** cleans up the remaining bespoke shape ops (`pad`, `tile`, `expand`, `slice`, `range`, `gather*`, `reduce_*`, `transpose`) and the multi-init / variadic-out ops (`gqa`, `mha`, `linear_attention`, `causal_conv_with_state`, `layer_norm`, `skip_rms_norm`, `hipdnn_graph`); closes the reify surface across all 53 ops.
* **#265:** adds a LIT case exercising the interface-default body through `--hip-infer-shapes` end-to-end.

### DIFF
--- a/lib/Dialect/IR/HipShapeUtils.cpp
+++ b/lib/Dialect/IR/HipShapeUtils.cpp
@@ -128,10 +128,11 @@ LogicalResult mlir::hip::verifyHipOpShape(
     Operation *op,
     function_ref<SmallVector<SmallVector<int64_t>>()> computeExpected,
     bool checkElementType) {
-  auto dpsOp = dyn_cast<DestinationStyleOpInterface>(op);
-  if (!dpsOp)
-    return op->emitOpError(
-        "verifyHipOpShape requires DestinationStyleOpInterface");
+  // Required-by-construction: every op that wires up `verifyHipOpShape`
+  // also implements `DestinationStyleOpInterface` via TableGen. Asserting
+  // cast matches the upstream Linalg pattern in `verifyStructuredOpInterface`
+  // (`cast<LinalgOp>(op)` / `cast<IndexingMapOpInterface>(op)`).
+  auto dpsOp = cast<DestinationStyleOpInterface>(op);
 
   SmallVector<SmallVector<int64_t>> expected = computeExpected();
   // Empty outer vector: the shape helper failed and already issued a
@@ -139,11 +140,16 @@ LogicalResult mlir::hip::verifyHipOpShape(
   if (expected.empty())
     return failure();
 
+  // Programmer-error invariant: each shape helper returns one expected shape
+  // per DPS init operand by construction. Assert in debug builds (matches the
+  // upstream `assert(X.size() == Y.size() && "...")` idiom in
+  // `tensor::CastOp` folding); the `return failure()` keeps release builds
+  // safe by avoiding the out-of-bounds `expected[i]` in the loop below.
   auto inits = dpsOp.getDpsInits();
+  assert(expected.size() == inits.size() &&
+         "shape helper must produce one expected shape per DPS init operand");
   if (expected.size() != inits.size())
-    return op->emitOpError("internal: shape helper produced ")
-           << expected.size() << " expected shape(s) but op has "
-           << inits.size() << " init operand(s)";
+    return failure();
 
   for (auto [i, init] : llvm::enumerate(inits)) {
     auto initType = dyn_cast<ShapedType>(init.getType());
--- a/lib/Dialect/IR/HipShapeUtils.cpp
+++ b/lib/Dialect/IR/HipShapeUtils.cpp
@@ -45,7 +45,7 @@ std::string formatShape(ArrayRef<int64_t> shape) {
 } // namespace
 
 SmallVector<int64_t>
-mlir::hip::inferContractionShape(ArrayRef<int64_t> aShape,
+mlir::hip::inferMatmulShape(ArrayRef<int64_t> aShape,
                                  ArrayRef<int64_t> bShape,
                                  function_ref<InFlightDiagnostic()> emitError) {
   if (aShape.size() < 2) {
--- a/lib/Dialect/IR/HipShapeUtils.cpp
+++ b/lib/Dialect/IR/HipShapeUtils.cpp
@@ -105,9 +105,10 @@ LogicalResult mlir::hip::verifyHipOpShape(
     Operation *op,
     function_ref<SmallVector<SmallVector<int64_t>>()> computeExpected) {
   // Required-by-construction: every op that wires up `verifyHipOpShape`
-  // also implements `DestinationStyleOpInterface` via TableGen. Asserting
-  // cast matches the upstream Linalg pattern in `verifyStructuredOpInterface`
-  // (`cast<LinalgOp>(op)` / `cast<IndexingMapOpInterface>(op)`).
+  // also implements `DestinationStyleOpInterface` via TableGen. Use
+  // asserting `cast<>` to express that contract — a missing interface is
+  // a programmer error in the op's TableGen def, not a user-facing
+  // diagnostic.
   auto dpsOp = cast<DestinationStyleOpInterface>(op);
 
   SmallVector<SmallVector<int64_t>> expected = computeExpected();
@@ -117,10 +118,9 @@ LogicalResult mlir::hip::verifyHipOpShape(
     return failure();
 
   // Programmer-error invariant: each shape helper returns one expected shape
-  // per DPS init operand by construction. Assert in debug builds (matches the
-  // upstream `assert(X.size() == Y.size() && "...")` idiom in
-  // `tensor::CastOp` folding); the `return failure()` keeps release builds
-  // safe by avoiding the out-of-bounds `expected[i]` in the loop below.
+  // per DPS init operand by construction. Assert in debug builds; the
+  // `return failure()` keeps release builds safe by avoiding the
+  // out-of-bounds `expected[i]` in the loop below.
   auto inits = dpsOp.getDpsInits();
   assert(expected.size() == inits.size() &&
          "shape helper must produce one expected shape per DPS init operand");
--- a/lib/Dialect/IR/HipShapeUtils.cpp
+++ b/lib/Dialect/IR/HipShapeUtils.cpp
@@ -45,9 +45,8 @@ std::string formatShape(ArrayRef<int64_t> shape) {
 } // namespace
 
 SmallVector<int64_t>
-mlir::hip::inferMatmulShape(ArrayRef<int64_t> aShape,
-                                 ArrayRef<int64_t> bShape,
-                                 function_ref<InFlightDiagnostic()> emitError) {
+mlir::hip::inferMatmulShape(ArrayRef<int64_t> aShape, ArrayRef<int64_t> bShape,
+                            function_ref<InFlightDiagnostic()> emitError) {
   if (aShape.size() < 2) {
     emitError() << "matmul A must have rank >= 2, got rank " << aShape.size();
     return {};
--- a/lib/Dialect/IR/HipShapeUtils.cpp
+++ b/lib/Dialect/IR/HipShapeUtils.cpp
@@ -15,7 +15,6 @@
 #include "hip/Dialect/IR/HipShapeUtils.h"
 
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
-#include "mlir/Dialect/Traits.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/Interfaces/DestinationStyleOpInterface.h"
 #include "llvm/ADT/STLExtras.h"
@@ -48,18 +47,44 @@ SmallVector<int64_t>
 mlir::hip::inferContractionShape(ArrayRef<int64_t> aShape,
                                  ArrayRef<int64_t> bShape,
                                  function_ref<InFlightDiagnostic()> emitError) {
-  if (aShape.size() < 2) {
-    emitError() << "matmul A must have rank >= 2, got rank " << aShape.size();
+  size_t aRank = aShape.size();
+  size_t bRank = bShape.size();
+  if (aRank < 2) {
+    emitError() << "matmul A must have rank >= 2, got rank " << aRank;
     return {};
   }
-  if (bShape.size() < 2) {
-    emitError() << "matmul B must have rank >= 2, got rank " << bShape.size();
+  if (bRank < 2) {
+    emitError() << "matmul B must have rank >= 2, got rank " << bRank;
     return {};
   }
 
-  int64_t M = aShape[aShape.size() - 2];
+  // Tightened contract: matches what `MatMulConversion.cpp` +
+  // `MatmulLowering.cpp` actually compute correctly today. Codegen
+  // derives result rank and batchCount from `A`, with optional `B`
+  // rank-2 broadcast (one matrix re-used across all batches). Anything
+  // wider — `B`'s rank > `A`'s rank, mixed ranks where `B` is not
+  // exactly rank-2, or per-dim batch broadcasting (`1` vs `>1`) —
+  // would be miscompiled by codegen + runtime, so the verifier rejects
+  // it here. Widening this contract requires matching widening in
+  // codegen and the runtime's strided-batch layout (zero stride for
+  // the broadcast side). Tracked under "Codegen contract" in
+  // `docs/design/hip-shape-inference.md`.
+  if (bRank > aRank) {
+    emitError() << "matmul B's rank (" << bRank << ") exceeds A's rank ("
+                << aRank
+                << "); B-side batch broadcasting is not supported by codegen";
+    return {};
+  }
+  if (aRank > bRank && bRank != 2) {
+    emitError() << "matmul mixed-rank operands require B to be rank-2 (got A "
+                   "rank "
+                << aRank << ", B rank " << bRank << ")";
+    return {};
+  }
+
+  int64_t M = aShape[aRank - 2];
   int64_t Ka = aShape.back();
-  int64_t Kb = bShape[bShape.size() - 2];
+  int64_t Kb = bShape[bRank - 2];
   int64_t N = bShape.back();
 
   // Contraction K must agree (kDynamic on either side is a wildcard).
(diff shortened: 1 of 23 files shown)


zpye:
Hi @fhanuman, thanks for raising these PRs. Currently, the PRs are still under review. I would like to first raise a few open questions on supporting dynamic shapes:

### 1. Data-dependent shape inference (extent only known at runtime)

For ops whose output shape depends on the values of the data rather than just operand shapes/attributes — i.e. the extent can't be computed at compile time — how does this design "complete" shape inference for those dimensions? Does reify simply leave them dynamic (?), and if so, where and how is the real extent computed at runtime and propagated to downstream consumers?

```
Example (illustration of the scenario):

%y = onnx.NonZero(%x)     // %x : tensor<?x4096xi64>
                          // %y : tensor<2x?xi64>
                          // dim 1 = number of non-zero elements — knowable only after
                          // inspecting %x's data, not from shapes/attrs
```

### 2. Is DPS mandatory for every op?

Must every HIP op be destination-passing style? Today each result's shape is tied to an outs operand produced by a tensor.empty, and the result type mirrors that init. For some ops — especially data-dependent or shape-changing ones — could the op instead produce its result directly as a normal op result (no tensor.empty destination), so the shape is defined by the op itself rather than by a pre-allocated init? What's the rationale for requiring DPS uniformly?

```
Example (illustration of the contrast):

// DPS (current): result type comes from the tensor.empty init
%init = tensor.empty(%n) : tensor<?xi64>
%y = hip.range(%ctx) ins(%start, %limit, %delta) outs(%init) -> tensor<?xi64>
// non-DPS (the alternative I'm asking about): op defines the result itself
%y = hip.range(%ctx) ins(%start, %limit, %delta) -> tensor<?xi64>
```

### 3. Allocating the output buffer during compute (once a data-dependent shape is known), at two different scopes

These two are the same underlying idea — defer allocation until the actual output shape is known at runtime, instead of sizing/allocating the buffer before the kernel runs — but they apply at two different scopes, so I'm treating them as separate concerns:

(a) Internal scope — intermediate buffers inside the compiled runtime. For an op with a data-dependent output, instead of taking a destination that's sized up front from the single grow-on-demand GPU pool, could the runtime allocate the buffer mid-compute, once the actual size has been computed? And would that require a memory pool able to hand out separate, independent blocks, rather than the current single contiguous grow-on-demand pool that assumes sizes are known ahead of time?

(b) Boundary scope — graph outputs handed back to ORT. Output OrtValues are currently allocated before inference_compute (the EP resolves dynamic dims via DimSource and calls ctx.GetOutput() with a fully resolved shape). Could we instead pass an ORT allocator down into the runtime and let it allocate the output buffer during compute, once the true shape is known — for models where the exact output shape can't be determined beforehand?

So: (a) concerns internal pool buffers within the compiled model.dll runtime, while (b) concerns the EP↔ORT output boundary and ORT's own allocator — same "allocate when the shape is known" goal, different layers.

```
Example — scope (a), internal pool:

// current: destination sized up front from the pool
%out = <pool buffer, size fixed before the kernel>
hip_nonzero(%x, %out)
// asked-about: size discovered mid-compute, then allocate (a separate block?)
%count = <count non-zeros of %x at runtime>
%out   = <allocate %count * elem_size here>
hip_nonzero_write(%x, %out)
```

```
Example — scope (b), EP↔ORT boundary:

// current: EP resolves dynamic dims (DimSource) and allocates the OrtValue before compute
out = ctx.GetOutput(i, resolved_shape);   // shape must be known here
inference_compute(..., out.data());
// asked-about: pass an allocator; runtime allocates the graph output when the shape is known
inference_compute(..., ort_allocator);     // runtime allocates mid-compute
```

fhanuman:
> Hi @fhanuman, thanks for raising these PRs. Currently, the PRs are still under review. I would like to first raise a few open questions on supporting dynamic shapes:
> 
> ### 1. Data-dependent shape inference (extent only known at runtime)
> For ops whose output shape depends on the values of the data rather than just operand shapes/attributes — i.e. the extent can't be computed at compile time — how does this design "complete" shape inference for those dimensions? Does reify simply leave them dynamic (?), and if so, where and how is the real extent computed at runtime and propagated to downstream consumers?
> 
> ```
> Example (illustration of the scenario):
> 
> %y = onnx.NonZero(%x)     // %x : tensor<?x4096xi64>
>                           // %y : tensor<2x?xi64>
>                           // dim 1 = number of non-zero elements — knowable only after
>                           // inspecting %x's data, not from shapes/attrs
> ```
> 
> ### 2. Is DPS mandatory for every op?
> Must every HIP op be destination-passing style? Today each result's shape is tied to an outs operand produced by a tensor.empty, and the result type mirrors that init. For some ops — especially data-dependent or shape-changing ones — could the op instead produce its result directly as a normal op result (no tensor.empty destination), so the shape is defined by the op itself rather than by a pre-allocated init? What's the rationale for requiring DPS uniformly?
> 
> ```
> Example (illustration of the contrast):
> 
> // DPS (current): result type comes from the tensor.empty init
> %init = tensor.empty(%n) : tensor<?xi64>
> %y = hip.range(%ctx) ins(%start, %limit, %delta) outs(%init) -> tensor<?xi64>
> // non-DPS (the alternative I'm asking about): op defines the result itself
> %y = hip.range(%ctx) ins(%start, %limit, %delta) -> tensor<?xi64>
> ```
> 
> ### 3. Allocating the output buffer during compute (once a data-dependent shape is known), at two different scopes
> These two are the same underlying idea — defer allocation until the actual output shape is known at runtime, instead of sizing/allocating the buffer before the kernel runs — but they apply at two different scopes, so I'm treating them as separate concerns:
> 
> (a) Internal scope — intermediate buffers inside the compiled runtime. For an op with a data-dependent output, instead of taking a destination that's sized up front from the single grow-on-demand GPU pool, could the runtime allocate the buffer mid-compute, once the actual size has been computed? And would that require a memory pool able to hand out separate, independent blocks, rather than the current single contiguous grow-on-demand pool that assumes sizes are known ahead of time?
> 
> (b) Boundary scope — graph outputs handed back to ORT. Output OrtValues are currently allocated before inference_compute (the EP resolves dynamic dims via DimSource and calls ctx.GetOutput() with a fully resolved shape). Could we instead pass an ORT allocator down into the runtime and let it allocate the output buffer during compute, once the true shape is known — for models where the exact output shape can't be determined beforehand?
> 
> So: (a) concerns internal pool buffers within the compiled model.dll runtime, while (b) concerns the EP↔ORT output boundary and ORT's own allocator — same "allocate when the shape is known" goal, different layers.
> 
> ```
> Example — scope (a), internal pool:
> 
> // current: destination sized up front from the pool
> %out = <pool buffer, size fixed before the kernel>
> hip_nonzero(%x, %out)
> // asked-about: size discovered mid-compute, then allocate (a separate block?)
> %count = <count non-zeros of %x at runtime>
> %out   = <allocate %count * elem_size here>
> hip_nonzero_write(%x, %out)
> ```
> 
> ```
> Example — scope (b), EP↔ORT boundary:
> 
> // current: EP resolves dynamic dims (DimSource) and allocates the OrtValue before compute
> out = ctx.GetOutput(i, resolved_shape);   // shape must be known here
> inference_compute(..., out.data());
> // asked-about: pass an allocator; runtime allocates the graph output when the shape is known
> inference_compute(..., ort_allocator);     // runtime allocates mid-compute
> ```


Hi @zpye ,

Excellent questions!

**1. Data-dependent shape inference (extent only known at runtime)**

RE:

This is a tricky scenario. We will not be able to propagate the shape in this case, and the shape inference/shape propagation pass leaves the shape as is, i.e, `?` in this case. However, we will allocate the maximum possible memory, and this is already handled correctly by our compiler. But there is a limitation in one of the upstream passes (not the one we wrote), so the allocated buffer gets hoisted out, causing a bug — need to check how to get around this.

IR after bufferization for the case you mentioned — please see the buffer `%alloc`:

```mlir
module attributes {hipdnn.input_count = 1 : i64, hipdnn.input_element_sizes = array<i64: 8>, hipdnn.input_shapes = [array<i64: -9223372036854775808, 4>], hipdnn.output_count = 1 : i64, hipdnn.output_element_sizes = array<i64: 8>, hipdnn.output_shapes = [array<i64: 2, -9223372036854775808>]} {
  func.func @main_graph(%arg0: !hip.context, %arg1: memref<?x4xi64> {onnx.name = "X"}) -> memref<2x?xi64> attributes {onnx.graph.name = "scenario_1_nonzero_basic"} {
    %c4 = arith.constant 4 : index
    %c1 = arith.constant 1 : index
    %dim = memref.dim %arg1, %c0 : memref<?x4xi64>
    %0 = arith.muli %c1, %dim : index
    %1 = arith.muli %0, %c4 : index
    %alloc = memref.alloc(%1) {alignment = 64 : i64} : memref<2x?xi64>
    %c1_0 = arith.constant 1 : index
    %dim_1 = memref.dim %alloc, %c1_0 : memref<2x?xi64>
    hip.nonzero(%arg0) ins(%arg1 : memref<?x4xi64>) outs(%alloc : memref<2x?xi64>) {input_data_type = 4 : i64}
    return %alloc : memref<2x?xi64>
  }
}
```

---

**2. Is DPS mandatory for every op?**

RE:

Making ops in the HIP dialect DPS was a deliberate choice I made early in this project. Yes for ops that produce a new buffer; no for ops that don't — `onnx.Reshape`/`Squeeze`/`Unsqueeze` aren't HIP ops at all; they lower directly to `tensor.expand_shape`/`collapse_shape`. For kernel-backed compute, uniform DPS is deliberate: `OneShotBufferize` uses `DestinationStyleOpInterface` to place the alloc, `hip-pool-allocs` packs every `memref.alloc` into the grow-on-demand GPU pool, and the EP boundary is caller-allocated (`inference_compute(ctx, in[], out[])`). `tensor.empty` carries the size SSA → bufferization emits the alloc → pool grabs it → `BufferResultsToOutParams` hoists it to a function arg. Drop DPS for one op, and that chain breaks at exactly that op.

You're right that the result type is tied to the `tensor.empty` today (the op's `outs` shape is the result shape) — decoupling the result type from the init while keeping DPS for allocation is a small change if it helps. For data-dependent ops, the `tensor.empty` is upper-bound (see the IR example above for `hip.nonzero`); that's a property of GPU kernels (can't grow a buffer mid-kernel), not really a DPS issue.

---

**3. Allocating the output buffer during compute (once a data-dependent shape is known), at two different scopes**

RE:

This is what we already do today — though we don't handle all the scenarios. See the IR after `hip-pool-allocs` for a simple dynamic-shaped attention block — the body has zero `memref.alloc` ops, just a single `hip.get_pool` call at the start sized from the input's runtime sequence dim, and every intermediate buffer is a cheap `memref.view` at a runtime-computed offset into that pool:

```mlir
%dim    = memref.dim %arg1, %c1                           // seq dim from input
%12     = ...                                             // compute size using %dim
%13     = hip.get_pool(%arg0, %12) : memref<?xi8>         // one runtime alloc, sized from %dim
%view   = memref.view %13[%c0][%dim] : memref<?xi8> to memref<2x?x64xf32>
hip.matmul(...) outs(%view : ...)
%view_3 = memref.view %13[%10][%dim, %dim] : memref<?xi8> to memref<2x?x?xf32>
hip.matmul(...) outs(%view_3 : ...)
hip.matmul(...) outs(%arg2 : memref<2x?x64xf32>)         // graph output, EP-supplied
```

`hip.get_pool` grows the pool on demand and reuses it across inferences, so we are effectively allocating once the actual size is known — just collapsed into a single per-inference call instead of a per-op callback. For scope (a), the internal pool already covers this; a block-style allocator handing out independent blocks mid-compute would do strictly more work for the same result, because shape inference already gives the packer every intermediate's size up front. For scope (b) the EP boundary, dynamic outputs whose shape is resolvable from the input shapes go straight to the EP-supplied `%arg2` (`DimSource` resolves the shape and ORT pre-allocates) — same idea pushed one layer up.

The case we don't handle today is graph outputs that are themselves data-dependent (e.g., a graph returning `NonZero`) — need to look into this.

lib/Dialect/IR/HipShapeUtils.cpp
```diff
@@ -0,0 +1,143 @@
...
+  os << "]";
+  return os.str();
+}
+
+} // namespace
+
+SmallVector<int64_t>
+mlir::hip::inferMatmulShape(ArrayRef<int64_t> aShape, ArrayRef<int64_t> bShape,
+                            function_ref<InFlightDiagnostic()> emitError) {
+  if (aShape.size() < 2) {
```
zpye, line 50: From the official docs, dims = 1 may work.

https://onnx.ai/onnx/operators/onnx__MatMul.html
https://numpy.org/doc/stable/reference/generated/numpy.matmul.html

zpye (dismissed):
Thanks for your reply. I left a few comments in PR #260  #261 #262, please take a look when you have time.

fhanuman:
I am refactoring the changes as discussed in yesterday's meeting to avoid code bloating. So, reverting the PR to draft state.

lib/Dialect/IR/HipShapeUtils.cpp
fhanuman: You're correct that the ONNX spec supports this; the check is intentional. `hip.matmul's` lowering + runtime assume rank ≥ 2 (we index aShape[rank-2] / bShape[rank-1]), so a rank-1 operand would silently misindex into hipBLASLt. The verifier acts as a diagnostic instead. Documented in docs/design/hip-shape-inference.md § "hip.matmul rejects rank-1 operands".

Supporting rank-1 operands would be a converter-side fix: in `OnnxToHip/MatMulConversion.cpp`, expand `rank-1` operand with `tensor.expand_shape` to `2d` and `tensor.collapse_shape` the result to `1d` coming out — keeping `hip.matmul's rank-2`-or-more contract intact for the lowering and runtime. This can be a follow-up; would prefer not to bundle it with these infrastructure changes.

fhanuman:
Hi @zpye, I have made some refactoring — the new commits introduce `HipDpsOpInterface + a parameterized Hip_DpsOp` base so PRs #261–#265 can drop a lot of per-op shape-inference boilerplate. Could you please take another look?

zpye (approved):
LGTM. Thanks for your update.

amd-mingw (approved):
@fhanuman, should resolved CI failed issue before merged, although that CI failed not caused by this PR.

zpye:
Here is a question, won’t block the merge:

Is it possible to use different classes instead of flag bits in HipOps.td?

For example, autoReify and customReifyBody may fused to a single class, and it could be one of:
  1. EmptyReify or [] for no implementation;
  2. ElementReify, BroadcastReify, ReduceReify(axes), ... ;
  3. CustomReifyDefinition(body) or CustomReifyDeclare(implement outside);

fhanuman:
> Here is a question, won’t block the merge:
> 
> Is it possible to use different classes instead of flag bits in HipOps.td?
> 
> For example, autoReify and customReifyBody may fused to a single class, and it could be one of:
> 
> 1. EmptyReify or [] for no implementation;
> 2. ElementReify, BroadcastReify, ReduceReify(axes), ... ;
> 3. CustomReifyDefinition(body) or CustomReifyDeclare(implement outside);

Thanks @zpye, good direction. The flag bits control two independent things — reify and infer. Some ops want custom reify but auto infer (e.g. matmul), so a single class would need to cover all combinations. For the shape-rule families (your point #2), PR #263 adds exactly this — Hip_DpsOp_Broadcast and Hip_DpsOp_Reduction sub-base classes. I kept this refactor for #263 to make #260 easier to review.

fhanuman:
> @fhanuman, should resolved CI failed issue before merged, although that CI failed not caused by this PR.

Hi @amd-mingw, yes the failure is not related to my change -- Tejus's fix for memory manager probably will fix the issue.

PR #261, opened 2026-05-31
feat(hip): InferTypeOpInterface on Hip_LoopOp returning v_init operand types -- 2/n

fhanuman:
This PR adds `InferTypeOpInterface::inferReturnTypes` to `Hip_LoopOp`, returning one `RankedTensorType` per `v_init` operand. The `hip.loop` verifier requires `result_type[i] == v_init[i].getType()`, and this lets future callers construct `hip.loop` ops verifier-clean by construction. Stacks on top of #260.

**Example MLIR that triggers the verifier today:**

```mlir
func.func @loop_v_init_result_mismatch(%ctx: !hip.context,
                                       %M: index,
                                       %cond: i1,
                                       %v: tensor<?x?x?xf16>) -> tensor<f16> {
  // error: 'hip.loop' op result type #0 ('tensor<f16>') must match
  //        v_init type #0 ('tensor<?x?x?xf16>')
  %r = hip.loop(%ctx, %M, %cond)
                 iter_args(%v : tensor<?x?x?xf16>)
                 -> (tensor<f16>)
                 body @loop_body_mismatch
                 {num_loop_carried = 1 : i32, cond_is_passthrough}
  return %r : tensor<f16>
}
```

The dynamic-shape vision-encoder ONNX exports I have been working with hit this — the importer leaves rank-0 placeholders while `v_init` is `tensor<?x?x?xf16>`, and the onnx-loop outliner forwards those under-refined types straight into the `hip.loop` builder.

**The fix:**

* Declare `InferTypeOpInterface` on `Hip_LoopOp` and implement `LoopOp::inferReturnTypes` returning one `RankedTensorType` per `v_init` operand. Memref-mode `v_init` (post-bufferization) produces 0 result types.
* Existing callers that still pass explicit result types are unaffected — the interface is purely additive.
* New LIT pins the verifier rule the impl is engineered to preserve.

P.S. Migrating `OnnxLoopOutlinePass` to the InferType-aware builder is the natural fix for the verifier abort above, but it is coupled work — the outliner has to refine the body func's arg / yield / op types from `v_init` *and* re-run shape inference inside the body so cloned ops verify against the now-refined entry-block args. That coupled change ships in a follow-up PR.

### DIFF
--- a/lib/Dialect/IR/HipDialect.cpp
+++ b/lib/Dialect/IR/HipDialect.cpp
@@ -166,11 +166,11 @@ LogicalResult LoopOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
 //
 // Memref-mode v_init produces 0 result types (DPS post-bufferization
 // convention; the verifier rejects mixed mode anyway).
-LogicalResult LoopOp::inferReturnTypes(
-    MLIRContext *context, std::optional<Location> location,
-    ValueRange operands, DictionaryAttr attributes,
-    OpaqueProperties properties, RegionRange regions,
-    SmallVectorImpl<Type> &inferredReturnTypes) {
+LogicalResult
+LoopOp::inferReturnTypes(MLIRContext *context, std::optional<Location> location,
+                         ValueRange operands, DictionaryAttr attributes,
+                         OpaqueProperties properties, RegionRange regions,
+                         SmallVectorImpl<Type> &inferredReturnTypes) {
   LoopOpAdaptor adaptor(operands, attributes, properties, regions);
   auto vInit = adaptor.getVInit();
   inferredReturnTypes.reserve(vInit.size());
--- a/lib/Dialect/IR/HipDialect.cpp
+++ b/lib/Dialect/IR/HipDialect.cpp
@@ -156,6 +156,35 @@ LogicalResult LoopOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
   return success();
 }
 
+// Result types are mechanically `v_init` operand types (verifier contract,
+// see verify() above and HipOps.td:126-127). Builders without explicit
+// result types pick this up automatically; ensures any new caller is
+// verifier-clean by construction.
+//
+// Status quo callers (notably LoopOutline.cpp's outlining) still pass
+// explicit result types and are unaffected. Migrating those to the
+// InferType-friendly builder requires upstream shape refinement on the
+// outlined body func (so body arg / yield / op types agree with v_init);
+// that work lives in Phase 2.5 / PR #265, which then flips LoopOutline
+// over to this path. Until then, this method only services NEW callers.
+//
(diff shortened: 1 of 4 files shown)


lib/Dialect/IR/HipDialect.cpp
```diff
@@ -156,6 +156,28 @@ LogicalResult LoopOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
...
+// convention; the verifier rejects mixed mode anyway).
+LogicalResult
+LoopOp::inferReturnTypes(MLIRContext *context, std::optional<Location> location,
+                         ValueRange operands, DictionaryAttr attributes,
+                         OpaqueProperties properties, RegionRange regions,
+                         SmallVectorImpl<Type> &inferredReturnTypes) {
+  LoopOpAdaptor adaptor(operands, attributes, properties, regions);
+  auto vInit = adaptor.getVInit();
+  inferredReturnTypes.reserve(vInit.size());
+  for (Value v : vInit) {
```
zpye, line 174: Do "scan outputs" need consideration?

https://onnx.ai/onnx/operators/onnx__Loop.html

zpye (approved):

amd-mingw:
@fhanuman, https://github.com/ROCm/MorphiZen/pull/228 fix a issue about unranked tensors, please ref. cc @zz002

fhanuman:
I am refactoring the changes as discussed in yesterday's meeting to avoid code bloating. So, reverting the PR to a draft state.

lib/Dialect/IR/HipDialect.cpp
fhanuman: Good catch — however, scan outputs aren't part of `hip.loop's` contract currently. The ODS def has a single variadic result (v_final) corresponding to loop-carried values only, and `LoopOutline` `(lib/Conversion/OnnxToHip/LoopOutline.cpp:246-250)` rejects (please see code snippet below for reference) any `onnx.Loop` whose body yield has more than `1 + numLoopCarried` operands. So by the time we build a `hip.loop`, `v_init.size() == num_results`, and `inferReturnTypes` deriving result types from `v_init` is complete. If/when we add scan-output support, both the outliner and `Hip_LoopOp's` ODS will need to grow new operand/result groups; the `inferReturnTypes` extension would be part of that future PR.

```
  Operation *yieldOp = bodyBlock.getTerminator();
  if (yieldOp->getName().getStringRef() != "onnx.Yield")
    return loopOp->emitOpError("onnx.Loop body must end with onnx.Yield, got ")
           << yieldOp->getName().getStringRef();
  if (yieldOp->getNumOperands() != 1 + numLoopCarried)
    return loopOp->emitOpError(
               "onnx.Loop scan outputs not yet supported (expected ")
           << (1 + numLoopCarried) << " yield operands for " << numLoopCarried
           << " loop-carried, got " << yieldOp->getNumOperands() << ")";
 ```

zpye (dismissed):

PR #262, opened 2026-05-31
feat(hip): InferType + reify on five single-result transformer ops -- 3/n

fhanuman:
This PR rolls out the verifier / reify / `inferReturnTypes` infra from #260 + #261 across five single-result DPS ops — `hip.rope`, `hip.rms_norm`, `hip.qmoe`, `hip.matmul_nbits`, and `hip.gemm`. With these wired, `--hip-infer-shapes` refines result types end-to-end across these op kinds — previously only `hip.matmul` boundaries refined; everything else stayed dynamic. Stacks on top of #261.

**Example MLIR that previously stayed dynamic** (the last dim is the integer attribute `N` on the op, but it stayed `?` because reify did not exist for `hip.matmul_nbits`):

```mlir
// Before --hip-infer-shapes:
%e = tensor.empty(%d0, %d1, %d2) : tensor<?x?x?xf16>
%y = hip.matmul_nbits(%ctx)
       ins(%a, %b, %s : tensor<?x?x2880xf16>, tensor<5120x90x16xui8>, tensor<5120x90xf16>)
       outs(%e : tensor<?x?x?xf16>)
       {N = 5120, K = 2880, ...} : tensor<?x?x?xf16>

// After:
%e = tensor.empty(%d0, %d1) : tensor<?x?x5120xf16>            // last dim recovered from N attr
%y = hip.matmul_nbits(%ctx) ins(...) outs(%e : tensor<?x?x5120xf16>) {N = 5120, ...}
       : tensor<?x?x5120xf16>
%c = tensor.cast %y : tensor<?x?x5120xf16> to tensor<?x?x?xf16>   // barrier on non-DPS-init uses
```

The same refinement chains through any downstream reify-aware op, so a `rms_norm -> matmul_nbits -> rope` triple now recovers all three result types in a single pass.

**The fix:**

Each op opts in to the parameterized `Hip_DpsOp` base from #260, mirroring the `hip.matmul` precedent:

* `Hip_RopeOp`, `Hip_RmsNormOp`, `Hip_QMoEOp`, `Hip_MatMulNBitsOp`, `Hip_GemmOp` set `autoInfer=1, declareInfer=1, autoReify=0` on `Hip_DpsOp` (see [`include/hip/Dialect/IR/HipOps.td`](https://github.com/ROCm/onnx-hipdnn-ep/blob/fhanuman.shape_inference.pr262/include/hip/Dialect/IR/HipOps.td)). The base's `extraClassDefinition` auto-emits each `inferReturnTypes` body — push the outs operand's tensor type, the DPS post-bufferize convention — and declares `InferTypeOpInterface` so converters get the inferred-type `Op::create` overload. No per-op `.cpp` boilerplate for InferType.
* `reifyResultShapes` bodies live in [`HipReifyResultShapesImpl.cpp`](https://github.com/ROCm/onnx-hipdnn-ep/blob/fhanuman.shape_inference.pr262/lib/Dialect/IR/HipReifyResultShapesImpl.cpp) next to `MatmulOp::reifyResultShapes`. Shape-preserving ops (`rope`, `rms_norm`, `qmoe`) collapse to a one-line call to a new `reifyElementwiseSameShape` helper in `HipShapeUtils`. `matmul_nbits` takes leading dims from `$A` and the final dim from the static `N` attribute. `gemm` builds `[M, N]` from the `transA` / `transB`-selected dims of `$input_a` / `$input_b`. All five source dims from input operands rather than the outs operand, so refinement propagates from upstream rather than freezing at the destination buffer's allocation-time shape.
* The five `OnnxToHip*Conversion.cpp` callsites for these ops drop the explicit result type from `OpClass::create(...)` and rely on the InferType-aware builder overload generated by ODS.
* [`test/lit/Dialect/hip-infer-shapes.mlir`](https://github.com/ROCm/onnx-hipdnn-ep/blob/fhanuman.shape_inference.pr262/test/lit/Dialect/hip-infer-shapes.mlir) gains one dynamic-shape refinement case per op, covering the static-N (matmul_nbits), shape-preserving (rope / rms_norm / qmoe), and 2D-from-inputs (gemm) reify branches.

P.S. Multi-result ops (`hip.gqa`, `hip.layer_norm`, `hip.skip_rms_norm`) follow the same recipe but with multi-output reify and hand-written `inferReturnTypes` bodies that walk every `outs` operand (the auto-emit body pushes a single result type and is wrong for multi-output ops). Deferred to a follow-up PR so the review window for this PR stays focused on the single-result rollout.


lib/Dialect/IR/HipResultTypeInferenceImpl.cpp
```diff
@@ -0,0 +1,106 @@
...
+                                      SmallVectorImpl<Type> &results) {
+  if (auto t = dyn_cast<RankedTensorType>(out.getType()))
+    results.push_back(t);
+  return success();
+}
+
+} // namespace
+
+//===----------------------------------------------------------------------===//
+// RopeOp
```
zpye: Could these functions share one implementation?

zpye (approved):

fhanuman:
I am refactoring the changes as discussed in yesterday's meeting to avoid code bloating. So, reverting the PR to a draft state.

lib/Dialect/IR/HipResultTypeInferenceImpl.cpp
fhanuman: Fixed.

zpye (dismissed):

PR #265, opened 2026-06-01
feat(hip): refine outlined hip.loop body func types from v_init -- 6/n

fhanuman:
This PR closes the work I promised in #261's P.S.: outlined `hip.loop` body func types are now driven from `v_init`, and `--hip-infer-shapes` gains a Phase 2 catch-up that keeps the contract holding when v_init narrows post-conversion. Vision-encoder `onnx.Loop` subgraphs that previously aborted at the OnnxLoopOutlinePass verifier now build verifier-clean. Stacks on top of #264.

**Failure today (under-refined source body block arg types):**

```mlir
// `--onnx-to-hip-pipeline` aborts at the OnnxLoopOutlinePass verifier:
//   error: 'hip.loop' op result type #0 ('tensor<*xf16>') must match
//          v_init type #0 ('tensor<?x?x?xf16>')
%r = "onnx.Loop"(%M, %cond, %vi) ({
    ^bb0(%iter: tensor<i64>, %c: tensor<i1>, %v: tensor<*xf16>):
      // ^- importer has no per-iter rank for body block args; %v ships
      //    unranked even though %vi is rank-3 dynamic
      %step = "onnx.Concat"(%v, %x) {axis = 1 : si64}
              : (tensor<*xf16>, tensor<?x?x?xf16>) -> tensor<*xf16>
      "onnx.Yield"(%c, %step) : (tensor<i1>, tensor<*xf16>) -> ()
  }) : (tensor<i64>, tensor<i1>, tensor<?x?x?xf16>) -> tensor<*xf16>
```

**The fix:**

* `OnnxLoopOutlinePass` sources the outlined body func's v_carry entry-arg types and `hip.loop`'s own result types from `v_init` — not from the under-refined source body block arg types. `hip.loop`'s verifier requires `result_type[i] == v_init[i].type`, satisfied by construction.

* `--hip-infer-shapes` gains a Phase 2 (`refineLoopSignatures`) that propagates v_init narrowing into the body func's `FunctionType` + entry-block args and into `hip.loop`'s result types after Phase 1's reify walk, then re-walks the body func once. Inert when v_init is tight on entry.

* The under-refined body block arg types are an importer-side defect, fixed in [ROCm/MorphiZen#228](https://github.com/ROCm/MorphiZen/pull/228). Cross-repo contract: [`docs/design/unranked-tensor-handling.md`](https://github.com/ROCm/onnx-hipdnn-ep/blob/fhanuman.shape_inference.pr265/docs/design/unranked-tensor-handling.md).

P.S. Until #228 merges and the `3rd-party/morphizen` submodule is bumped, Loop-heavy models still fail rank-aware conversion on this branch — `TODO(unranked-import-contract)` in `Pipelines.cpp` tracks it. I'll send a separate submodule-bump PR once #228 merges.


zpye (commented):
Could we switch to the Hip dialect as soon as possible (even if it's incorrect) and perform shape inference based on the Hip dialect, otherwise too many Onnx implementations may be introduced.

zpye (approved):

fhanuman:
> Could we switch to the Hip dialect as soon as possible (even if it's incorrect) and perform shape inference based on the Hip dialect, otherwise too many Onnx implementations may be introduced.

Thanks @zpye — shape inference IS still happening in the HIP dialect via `--hip-infer-shapes`. The --onnx-infer-shapes pass + rules library you saw in earlier commits was a workaround for a MorphiZen importer bug (rank-0 emitted for unranked tensors, producing illegal IR). I have cleaned it up; the proper fix on the importer side is [ROCm/MorphiZen#228](https://github.com/ROCm/MorphiZen/pull/228).

Open concern: #228 introduces unranked tensors — UnrankedTensorType . Handling unranked tensors end-to-end (converters, tensor.empty sizing, bufferizer) is non-trivial. I need to check whether we can avoid unranked tensors altogether from the MorphiZen side.

PR #271, opened 2026-06-02
feat(runtime): ck_dsl-generated f32 RMSNorm fast paths for Qwen3.5

BoarQing:
## Summary

Adds two `ck_dsl`-generated **f32 RMSNorm** kernels for gfx1151 as drop-in fast paths in front of the existing MIOpen baseline for the Qwen3.5 residual-stream shape (`f32`, `hidden_dim=4096`). Each kernel is gated by an exact shape/dtype match **and a runtime gfx1151 device check** (the HSACOs are gfx1151-only); anything else returns `-2` and falls back to MIOpen unchanged, so behavior is identical for every other model, shape, or GPU architecture.

## How each kernel is generated

Both HSACOs are built **offline** from `ck_dsl` (no Python/`ck_dsl` dependency at EP build or runtime) and vendored as plain byte-array headers (`xxd`-style hexdump, 16 bytes/line). They were generated against a small upstream `ck_dsl` patch that adds f32 I/O to `rmsnorm2d` / `add_rmsnorm2d_bf16`: ROCm/rocm-libraries#7972.

| Kernel (HSACO symbol) | Generated from | Replaces |
|---|---|---|
| `ck_dsl_rmsnorm2d_fwd_f32_N4096_b256_v4` | `ck_dsl` `rmsnorm2d` (f32 I/O patch), gfx1151 | `miopenT5LayerNormForward` for the residual-stream RMSNorm |
| `ck_dsl_add_rmsnorm2d_bf16_f32_N4096_b256_v4_sr` | `ck_dsl` `add_rmsnorm2d` (f32 I/O patch), gfx1151 | `miopenOpTensor(ADD)`×2 + `miopenT5LayerNormForward` (the 3-launch fused skip-norm chain) |

To regenerate a kernel: rebuild the HSACO via `ck_dsl`, then re-emit the byte array (e.g. `xxd -i <kernel>.hsaco`). The recipe is also noted at the top of each generated header.

## Performance comparison (vs MIOpen)

Isolated microbench on Radeon 8060S (Strix Halo, gfx1151):

| Kernel | Wall speedup vs MIOpen | Max abs diff vs MIOpen |
|---|---|---|
| `rmsnorm2d` f32 / N=4096 | **2.97×** | 4.8×10⁻⁷ |
| `add_rmsnorm2d` (fused skip-norm) f32 / N=4096 | **3.48×** | 2.4×10⁻⁷ |

The win is mostly per-call CPU dispatch overhead, not raw GPU throughput: MIOpen's `T5LayerNorm` (+ `OpTensor(ADD)`×2 for skip-norm) spends ~20-30 µs per call in descriptor/handle dispatch before the kernel launches; the `hipModuleLaunchKernel` path is ~2-3 µs, and the fused skip kernel also collapses 3 launches into 1. On GPU time the f32 norm is faster too (2.7 µs vs 9.3 µs for MIOpen's T5LN).

End-to-end on `Qwen3.5-9B-rtn-int4-int8-128gs-fp16-onnx-gpu`, `model_benchmark.exe -l 128 -g 32 -r 5 -w 4 -ml 256`, 33 paired runs: **+2.13% decode TPS** (95% CI [+0.53%, +3.73%], paired t-test p ≈ 0.009). Small but statistically significant.

## Integration

Each existing MIOpen wrapper grows one `extern "C"` decl plus a pre-MIOpen `if (shape matches && device is gfx1151) { if (ck_dsl(...) == 0) return 0; }` block. No existing MIOpen code path is modified.

### DIFF
--- a/lib/Runtime/real/ck_dsl_simplified_layer_norm.cpp
+++ b/lib/Runtime/real/ck_dsl_simplified_layer_norm.cpp
@@ -63,7 +63,7 @@ const std::unordered_map<std::string, ckdsl::KernelDef> &kernelTableF16() {
 // launch overhead dominates) but loses to MIOpen on prefill-shaped large M, so
 // keep big batches on the MIOpen baseline. Prototype heuristic; tune via a
 // per-M sweep.
-constexpr int64_t kMaxCkDslDecodeRows = 32;
+constexpr int64_t kMaxCkDslDecodeRows = 1;
 
 // ABI matches RMSNorm2DSpec.signature() (natural alignment; the 3 leading
 // pointers are 8-aligned and M/N/eps pack contiguously, so field offsets
--- a/lib/Runtime/real/ck_dsl_skip_simplified_layer_norm.cpp
+++ b/lib/Runtime/real/ck_dsl_skip_simplified_layer_norm.cpp
@@ -78,7 +78,7 @@ constexpr int kHiddenDim = 4096;
 // through ck_dsl. ck_dsl wins on small M (autoregressive decode) but loses to
 // MIOpen on prefill-shaped large M, so keep big batches on the MIOpen baseline.
 // Prototype heuristic; tune via a per-M sweep.
-constexpr int64_t kMaxCkDslDecodeRows = 32;
+constexpr int64_t kMaxCkDslDecodeRows = 1;
 
 // ABI matches AddRMSNorm2DBF16Spec.signature() (natural alignment; the 5
 // leading pointers are 8-aligned and M/N/eps pack contiguously, so field


wcy123 (dismissed):
**LGTM** - Strong perf results (+2.13% decode TPS, p ≈ 0.009), zero impact on non-target shapes/devices.

**Architecture:**

Two runtime functions affected (zero compiler changes):
- `RmsNormOpLowering` → `wrap_miopenT5LayerNormForward` → try `ck_dsl_simplified_layer_norm()`, fallback MIOpen
- `SkipRmsNormOpLowering` → `wrap_skip_simplified_layer_norm` → try `ck_dsl_skip_simplified_layer_norm()`, fallback MIOpen 3-launch chain

Pre-built gfx1151 HSACOs vendored as byte arrays (no ck_dsl/Python dependency). Per-architecture kernel table, extensible to discrete GPUs.

**Performance:**
- Microbench: 2.97-3.48× vs MIOpen (max abs diff ~10⁻⁷)
- End-to-end Qwen3.5-9B: +2.13% decode TPS, 95% CI [+0.53%, +3.73%], p ≈ 0.009
- Win source: CPU dispatch overhead (MIOpen 20-30 µs → hipModuleLaunchKernel 2-3 µs) + fused skip-norm (3 launches → 1)

**Improvement suggestions:**

**1. Public API pollution**

`ck_dsl_simplified_layer_norm()` and `ck_dsl_skip_simplified_layer_norm()` exposed in `hipdnn_ep_runtime.h` but only called internally. Move to private header or forward-declare in callers.

**2. Header-only design**

`inline` functions in `ck_dsl_kernel.h` with static locals. Consider moving to `.cpp` for explicit global state. Current approach works but adds no value (not inlineable due to mutex).

**3. Global kernel cache**

Process-wide static with unbounded growth and global mutex (all sessions contend). Same pattern as PR #319 `g_gemm_algo_cache`, PR #330 `g_hipdnn_ep_current_stream`.

Root cause: No mechanism for op-specific state with proper lifetime management. Current workaround: global statics (never cleaned up, shared across all sessions).

Better approach: Per-session op state owned by `RuntimeState`, isolated per session, auto-cleanup on session end, no mutex needed. Would enable:
```cpp
// Per-session ck_dsl state (not global)
CkDslState *ck_state = get_op_state(state, "ck_dsl");
hipFunction_t fn = ck_state->kernel_cache[symbol];  // No mutex, per-session
// Auto cleanup when RuntimeState destroyed
```

Current approach is acceptable (doesn't worsen existing pattern), but same architectural issue affects multiple PRs.

**Recommendation: APPROVE**

Suggestions above are refinements, not blockers.

PR #273, opened 2026-06-02
perf(profiling): Reduce HIPDNN_EP_PERF profiler self-overhead and add report formatter + bench harness

fhanuman:
## What this PR does

When you turn the profiler on with `HIPDNN_EP_PERF=1`, it was skewing its own numbers. There were two problems:

1. The per-op timing work — recording the HIP events and printing the table — happened *inside* the same window we use to time each `Compute()` call. So the profiler was charging its own cost to the model and reporting throughput lower than the model actually runs at.
2. Every profiling event used a default system fence. That adds a little cost on every single record, and at the margins it could even make a slower kernel look faster than a quicker one.

This PR fixes both, and along the way it adds the same per-op profiling to the output-allocator path (the default for KV-cache models, which until now produced no per-op data at all). It also adds a few small tools that turn the raw `[PERF]` log into a readable report.

With the profiler **off** there is no overhead — none of this code runs. The numbers below are about the cost *when you ask for profiling*.

## Commits (please review in this order)

Only the first commit touches runtime/EP code — that's the one to look at closely. The other three are tooling and have no effect on the build or runtime.

* **`d42d5915` — the profiler fix.** Three things:
  * Create every profiling event with `hipEventDisableSystemFence`. This is safe because we only read the timing with `hipEventElapsedTime` *after* a `hipStreamSynchronize`, so the fence wasn't buying us anything. The reasoning is written out in one place (`op_profile.cpp`) and the other call sites point back to it.
  * Move the "resolve and print the per-op table" step *out* of the timed window. The EP now calls a new runtime hook, `hipdnn_ep_runtime_flush_op_profile`, right after the timing window closes. This is backwards compatible: an older `model.dll` that doesn't export the symbol simply skips the step (the cached function pointer is null), exactly like the existing `begin_compute` hook.
  * Allocate the EP's own timing event pair once per session and reuse it, instead of creating and destroying it on every `Compute()`.
  * Same commit also adds the timing to the output-allocator path and folds the shared event/timing logic into one small `EpPerfTimer` helper that both paths use, so there's no copy-pasted timing code.
* **`2d672ccf`** — `format_perf_report.py`: a standard-library-only Python script that turns a raw `[PERF]` log into a four-section report (headline, decode breakdown, per-op GPU table, per-call distribution).
* **`cfe7af00`** — `run_bench.sh` + `docker_run_bench.sh`: a one-command way to run a benchmark and format the result.
* **`4e4d61c5`** — a benchmark runner and report formatter for multimodal (VLM) models. Tooling only, no runtime impact.

<details><summary>Sample <code>--mode perf</code> report (Llama-3.1-8B)</summary>

```
$ tools/perf-report/docker_run_bench.sh --oga --prompt 128 --gen 128 --reps 3 --warmup 1 --mode perf
[pr212] dynamic-shape support present (commit 6b5cf01 reachable from HEAD)

=== Llama-3.1-8B-awq-g128-int4-asym-fp16-onnx-dml | OGA model_benchmark | dynamic | l=128 g=128 r=3 w=1 ===
    model dir (staged): tools/perf-report/oga_dynamic_dirs/Llama-3.1-8B-awq-g128-int4-asym-fp16-onnx-dml
    log               : tools/perf-report/_perf_logs/Llama-3.1-8B-awq-g128-int4-asym-fp16-onnx-dml_oga_dynamic_p128g128_perf_20260602_155310.log
──────────────────────────────────────────────────────
  wall: 29s   exit=0

  ════════════════════════════════════════════════════════════════════════════
    HIPDNN EP profile  ·  Llama-3.1-8B-awq-g128-int4-asym-fp16-onnx-dml  ·  prompt=128 gen=128  ·  inferences=639
  ════════════════════════════════════════════════════════════════════════════

    § 1  HEADLINE  —  model_benchmark (batch=1, prompt=128, gen=128)
    ──────────────────────────────────────────────────────────────────────────
                            throughput     p50 latency        stddev   samples
      Prefill (TTFT)        743.03 t/s    166367.06 us   13173.46 us   3 * 128 token(s)
      Decode                 39.82 t/s     25080.20 us     178.63 us   381 * 1 token(s)
      Sampling            15290.29 t/s        65.22 us       1.17 us   3 * 1 token(s)
      E2E generation                 —      3353.94 ms      16.99 ms   3
      Peak working set   1.52 GiB (1630154752 bytes)

    § 2  STEADY-STATE DECODE BREAKDOWN  —  1 decode token, median across run
    ──────────────────────────────────────────────────────────────────────────
                                                    latency    share   source
      OGA::GenerateNextToken                      25.080 ms   100.0 %   ◀ § 1 Decode p50
      ├─ OGA + ORT framework overhead              1.015 ms     4.0 %
      │  ├─ Token sampling                         0.065 ms     0.3 %   ◀ § 1 Sampling avg
      │  └─ Other (IoBinding rebind, etc.)         0.950 ms     3.8 %
      └─ EP MlirCustomOp::Compute()               24.065 ms    96.0 %   ◀ § 4 wall_ms
         ├─ Marshal in (input descriptors)         0.006 ms     0.0 %   ◀ § 4 marshal_in_ms
         ├─ Marshal out (output desc.)             0.054 ms     0.2 %   ◀ § 4 marshal_out_ms
         ├─ model.dll inference_compute()         23.995 ms    95.7 %   ◀ § 4 compute_cpu_ms
         │  ├─ GPU: OP_PROFILE kernels (sum)      22.900 ms    91.3 %   ◀ § 3 TOTAL
         │  ├─ GPU: outside § 3 scopes             1.051 ms     4.2 %
         │  └─ CPU: host dispatch + sync poll      0.044 ms     0.2 %
         └─ Trailing fence (hipStreamSync)         0.009 ms     0.0 %   ◀ § 4 fence_residual_ms

      notes:
        - "GPU: outside § 3 scopes" = § 4 gpu_ms − § 3 TOTAL: GPU work not bracketed by any
          OP_PROFILE wrapper. Typical sources: MIOpen / hipBLASLt internal helper
          launches, ops lacking an OP_PROFILE scope, hipMemset / glue between ops. If
          this row grows after a runtime change, OP_PROFILE coverage has regressed.
        - "Other (IoBinding rebind, etc.)" = § 1 Decode p50 − sampling − EP wall: OGA
          framework overhead between Generator steps. Dominated by IoBinding's per-token
          rebind of all input/output tensors (scales with KV buffer size, not seq len).

    § 3  PER-OP GPU BREAKDOWN  —  last Compute() = steady-state decode token
    ──────────────────────────────────────────────────────────────────────────
      ===============================================================
                                     calls  gpu (ms)  cpu (ms)  gpu %
       matmul_nbits                    225      20.1       0.6  87.3%
         m=1,n=14336,k=4096             64      10.0       0.2  43.3%
         m=1,n=4096,k=14336             32       4.5       0.1  19.6%
         m=1,n=4096,k=4096              64       3.2       0.2  14.0%
         m=1,n=128256,k=4096             1       1.3       0.0   5.4%
         m=1,n=1024,k=4096              64       1.1       0.2   4.9%
       skip_layernorm                   64       1.0       0.3   4.4%
         1x4096                         64       1.0       0.3   4.4%
       gqa                              32       1.0       0.3   4.2%
         b=1,sq=1,skv=256,h=32,d=128    32       1.0       0.3   4.2%
       rotary_emb                       64       0.4       0.2   1.5%
         h=32,hd=128,rd=128,bnsh=0      32       0.2       0.1   0.9%
         h=8,hd=128,rd=128,bnsh=0       32       0.1       0.1   0.6%
       elementwise                      64       0.3       0.2   1.2%
         1x1x1x14336                    64       0.3       0.2   1.2%
       activation                       32       0.3       0.1   1.2%
         n=14336                        32       0.3       0.1   1.2%
       layernorm                         1       0.0       0.0   0.0%
         1x4096                          1       0.0       0.0   0.0%
       cast                              2       0.0       0.0   0.0%
         n=1                             2       0.0       0.0   0.0%
       reduce_sum                        1       0.0       0.0   0.0%
         256->1                          1       0.0       0.0   0.0%
       gather                            1       0.0       0.0   0.0%
         n=4096                          1       0.0       0.0   0.0%
       sub                               1       0.0       0.0   0.0%
         n=1                             1       0.0       0.0   0.0%
       TOTAL                                    23.1       1.8
      ===============================================================

    § 4  PER-CALL DISTRIBUTION  —  EP MlirCustomOp::Compute() over 639 invocations (all ms)
    ──────────────────────────────────────────────────────────────────────────
                                min     median        p99          max
      wall_ms                23.738     24.065     25.185     8005.980
      compute_cpu_ms         23.672     23.995     25.038     8000.806
      gpu_ms                 23.632     23.951     24.994     8000.736
      marshal_in_ms           0.006      0.006      0.011        0.017
      marshal_out_ms          0.048      0.054      0.117        5.152
      fence_residual_ms       0.008      0.009      0.011        0.046
      notes:
        - `max` includes first-Compute cold start (kernel autotune + pool grow);
          § 2 uses median to skip those outliers.
        - this count covers ALL EP calls (warmup + verbose-display + timed reps);
          § 1's per-block `samples` column is the OGA-timed subset only.

  ════════════════════════════════════════════════════════════════════════════
    [OGA] prefill=743.03 tok/s  TTFT=172.27 ms  decode=39.82 tok/s  peak_WS=1.52 GiB (1630154752 bytes)
  ════════════════════════════════════════════════════════════════════════════

done. logs in tools/perf-report/_perf_logs
```

</details>

## How much overhead is left when profiling is on

Measured on gfx1151 (Strix Halo, 16 CUs), dynamic-shape single graph, OGA `model_benchmark`, L=128 / G=128, `-r 3 -w 1`. `--mode none` means the profiler is off — that's the real performance. `--mode perf` turns it on and adds the per-op `hipEventRecord` pair on every profiled op, plus the resolve/flush between calls. Quote the **off** numbers as the headline throughput; the **on** numbers are only useful for reading the per-op breakdown.

### Llama-3.1-8B-awq-g128-int4-asym

| Metric | profile OFF (`--mode none`) | profile ON (`--mode perf`) | Δ |
|---|---|---|---|
| Prefill | 702.85 t/s | 689.84 t/s | −1.9% |
| Decode  | 42.66 t/s  | 39.48 t/s  | −7.5% |
| TTFT    | 182.12 ms  | 185.55 ms  | +1.9% |
| Peak working set | 1.52 GiB | 1.52 GiB | ~0 |

### phi4-14B-int4-rtn-block-32

| Metric | profile OFF (`--mode none`) | profile ON (`--mode perf`) | Δ |
|---|---|---|---|
| Prefill | 763.57 t/s | 756.81 t/s | −0.9% |
| Decode  | 24.82 t/s  | 22.94 t/s  | −7.6% |
| TTFT    | 167.63 ms  | 169.13 ms  | +0.9% |
| Peak working set | 1.50 GiB | 1.50 GiB | ~0 |

Decode takes the bigger hit (~7.5%) because it pays the per-op record-and-flush cost on every single-token step, and there are hundreds of those per run (381 here). Prefill spreads the same fixed cost over a handful of large calls, so it's under 2%. Peak memory doesn't move — the event pool is tiny. The two fixes in `d42d5915` (disabling the system fence and moving the resolve/print out of the timed window) are what keep the leftover overhead this small.

### Multimodal / VLM (Qwen3.5-9B, `model_mm` via the new `--mm` path)

Same comparison via the new `--mm` mode (`model_mm`, image `eiffel.jpg` + prompt
"What is in this image?"). Fixed work: **130 new tokens** from an **894-token
prompt** (894 includes image patch tokens). The cap is set with a `--variant`
config (`search.max_length = 894 + 130`) because `model_mm` ignores `-g`.

| Metric | profile OFF (`--mode none`) | profile ON (`--mode perf`) | Δ |
|---|---|---|---|
| Decode | 18.40 t/s | 16.82 t/s | −8.6% |
| TTFT   | 12.29 s   | 12.25 s   | ~0 |
| New tokens generated | 130 | 130 | — |

Command (host wrapper; `--variant perfcap` is the capped-`max_length` config):

```bash
tools/perf-report/docker_run_bench.sh --mm --variant perfcap \
    --model <vlm-dir> --image eiffel.jpg --prompt "What is in this image?" \
    --mode none          # and --mode perf for the profiler-on run
```

P.S. `run_bench.sh --model` currently takes a directory name, not an absolute path — it looks for `$WORKSPACE/oga_models/<name>/`. Happy to extend it to absolute paths in a follow-up if reviewers would prefer.

### DIFF
--- a/backend-mlir-compiler/custom-op-mlir/src/InferenceState.cpp
+++ b/backend-mlir-compiler/custom-op-mlir/src/InferenceState.cpp
@@ -57,19 +57,24 @@ void InferenceState::resolveEntryPoints(const LoadedArtifact &artifact) {
   set_output_allocator_fn_ =
       artifact.get_method<void, void *, const output_allocator_t *>(
           hipdnn::abi::kSetOutputAllocator);
+  flush_op_profile_fn_ =
+      artifact.get_method<void, void *>(hipdnn::abi::kRuntimeFlushOpProfile);
 }
 
 InferenceState::InferenceState(PrivateTag, void *state,
                                std::unique_ptr<LoadedArtifact> artifact)
     : state_(state), artifact_(std::move(artifact)), compute_fn_(nullptr),
       compute_alloc_fn_(nullptr), cleanup_fn_(nullptr),
-      begin_compute_fn_(nullptr), set_output_allocator_fn_(nullptr) {
+      begin_compute_fn_(nullptr), set_output_allocator_fn_(nullptr),
+      flush_op_profile_fn_(nullptr) {
   resolveEntryPoints(*artifact_);
 
   MY_LOG(2) << "begin_compute symbol "
             << (begin_compute_fn_ ? "resolved" : "not exported (no-op)");
   MY_LOG(2) << "set_output_allocator symbol "
             << (set_output_allocator_fn_ ? "resolved" : "not exported (no-op)");
+  MY_LOG(2) << "flush_op_profile symbol "
+            << (flush_op_profile_fn_ ? "resolved" : "not exported (no-op)");
 
   // Without begin_compute, the GQA seqlens_k cache survives across
   // forward passes and decode returns token-1 values for tokens 2..N
@@ -199,6 +204,12 @@ void InferenceState::set_output_allocator(
 void InferenceState::begin_compute() const {
   if (begin_compute_fn_ && state_) {
     begin_compute_fn_(state_);
+  }
+}
+
+void InferenceState::flush_op_profile() const {
+  if (flush_op_profile_fn_ && state_) {
+    flush_op_profile_fn_(state_);
   }
 }
 
--- a/backend-mlir-compiler/custom-op-mlir/src/InferenceState.h
+++ b/backend-mlir-compiler/custom-op-mlir/src/InferenceState.h
@@ -85,6 +85,14 @@ public:
   // warns when the symbol is absent.
   void begin_compute() const;
 
+  // Flush per-op profile (HIPDNN_EP_PERF). Called by the EP AFTER its
+  // wall_ms timing window closes, so the resolve + std::map + fprintf cost
+  // no longer pollutes Compute() latency. No-op when the model.dll predates
+  // the export (per-op PERF block is silently skipped for such DLLs;
+  // inference is unaffected). Symbol resolved once in create() so the call
+  // is a single cached indirect dispatch.
+  void flush_op_profile() const;
+
   // Diagnostic-only accessor: returns the hipStream_t used by
   // inference_compute, as a void*.  Relies on RuntimeState
   // (lib/Runtime/runtime_state_internal.h) keeping hipStream_t as its first
@@ -138,6 +146,13 @@ private:
   // DLLs); only used in output-allocator mode.
   using SetOutputAllocatorFn = void (*)(void *, const output_allocator_t *);
   SetOutputAllocatorFn set_output_allocator_fn_;
+
+  // Cached function pointer for hipdnn_ep_runtime_flush_op_profile. Same
+  // contract as begin_compute_fn_: resolved once at session creation, null
+  // when the symbol is absent (older DLLs), one cached indirect call when
+  // present.
+  using FlushOpProfileFn = void (*)(void *);
+  FlushOpProfileFn flush_op_profile_fn_;
 };
 
 } // namespace mlir_compilation::customop
--- a/backend-mlir-compiler/custom-op-mlir/src/MlirCustomOp.cpp
+++ b/backend-mlir-compiler/custom-op-mlir/src/MlirCustomOp.cpp
@@ -736,11 +736,15 @@ MlirCustomOp::~MlirCustomOp() {
   for (const HostOutputScratch &slot : host_out_scratch_)
     if (slot.ptr)
       (void)hipFree(slot.ptr);
-#endif
-#ifndef BUILD_MOCK_RUNTIME
+
   // Release the lazy-allocated HIPDNN_EP_PERF event pair. Created on first
-  // perf-enabled Compute() and reused for the session lifetime; nullptr
-  // here means perf was never enabled, so nothing to release.
+  // perf-enabled Compute() and reused for the session lifetime; nullptr here
+  // means perf was never enabled, so nothing to release. Guarded by
+  // HIPDNN_EP_LINK_HIP_HOST (not BUILD_MOCK_RUNTIME): hipEvent_t /
+  // hipEventDestroy are only declared when the HIP headers are linked, and
+  // ep_perf_ev_* are only ever assigned by EpPerfTimer, which is itself behind
+  // this same macro -- so in builds without it these handles are always null
+  // and need no cleanup.
   if (ep_perf_ev_start_) {
     (void)hipEventDestroy(static_cast<hipEvent_t>(ep_perf_ev_start_));
     ep_perf_ev_start_ = nullptr;
--- a/backend-mlir-compiler/custom-op-mlir/src/MlirCustomOp.cpp
+++ b/backend-mlir-compiler/custom-op-mlir/src/MlirCustomOp.cpp
@@ -26,6 +26,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <mutex>
+#include <optional>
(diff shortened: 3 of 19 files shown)


fhanuman:
We have issue with dynamic shaped model. See https://github.com/ROCm/onnx-hipdnn-ep/issues/274 for details.

PR #279, opened 2026-06-05
feat(conv): rank-3 Conv via 2D reshape (fp16 + fp32) -- 1/n [Ready]

chizamd:
First of a **5-PR cascade** adding end-to-end **Whisper-large-v3** (speech-to-text) support to the MorphiZen EP — the first encoder-decoder / cross-attention model on the stack. Each PR is independently shippable (compiles + its own tests pass).

This PR adds **1D convolution** end-to-end (the Whisper encoder front-end, but a self-contained path usable by any rank-3 `Conv` — no attention/Whisper dependency).

A 1D conv is just a 2D conv with height = 1, so rank-3 `Conv` reshapes to the **shared** `hip.conv` rather than introducing a dedicated op/kernel.

## What's new

**`ConvConversion.cpp` — lower 1D conv as a 2D conv.** A rank-3 `Conv` is rewritten to:
- reshape input/weights from `[N, C, L]` to `[N, C, 1, L]` (add a height dim of 1)
- run the existing `hip.conv` with height = 1 (`kernel_shape=[1,K]`, `strides=[1,s]`, `pads=[0,b,0,e]`)
- reshape the result back from `[N, C, 1, L']` to `[N, C, L']`

These reshapes are free — they only relabel the tensor's shape, no data is moved and no extra GPU kernels run. One subtlety: I reshape the *output* buffer the same way, not just the conv result. That makes `collapse(expand(init))` fold back to the original buffer, so the compiler recognizes it as the function's real output (written in place) instead of a leftover temporary.

**`PoolAllocs.cpp` — pool a single allocation too.** Tied to the point above: if the output isn't recognized as the real output, it becomes a leftover temporary. Our memory-pooling pass used to skip the case where there's only one such temporary (it only kicked in at 2+). An un-pooled buffer compiles to a runtime symbol that doesn't exist (`hip_device_malloc`), so the model fails to link and silently falls back to CPU. The pass now pools even a single allocation.

**`wrap_miopenConvolutionForward` (`lib/Runtime/real/miopen.cpp`) — reuse a workspace buffer.** The shared 2D MIOpen wrapper now draws its scratch workspace from a per-session pool that grows on demand, instead of doing a `hipMalloc`/`hipFree` on every call. (fp16/fp32 and per-channel bias are already supported in this shared kernel and are exercised by the 1D path.)

## Tests

- LIT `test_conv1d.mlir` — rank-3 `Conv` → `expand_shape` → `hip.conv` → `collapse_shape`. ✅ (full LIT suite 216/216)
- Numeric `test_conv1d.py` — 7 cases, **fp16 + fp32** vs ORT CPU. ✅ (7/7, cosine 1.0)
- Conversion output is byte-identical to the version validated end-to-end on the integration branch (#283): Whisper encoder cosine fp16 0.999979 / fp32 1.000000, e2e JFK quote verbatim, decode fp16 ~48 tok/s.

## Stack overview

Whisper-large-v3 support, landed as 5 independently-shippable PRs:

- **1/5 (this PR, #279):** 1D conv via 2D reshape (encoder front-end).
- **2/5 (#280):** GQA `no_causal` flag + fp32 decomposed attention path (extends the existing GQA op; Llama/gpt-oss byte-identical).
- **3/5 (#281):** `SkipLayerNormalization` → `hip.add` + `hip.layer_norm`.
- **4/5 (#282):** Whisper encoder/decoder attention → `hip.gqa` (depends on #280).
- **5/5 (#283):** end-to-end Whisper (model setup + tests + CLI + docs) (depends on #279–#282).


include/hip/Dialect/IR/HipOps.td
```diff
@@ -537,6 +537,56 @@ def Hip_ConvOp : Hip_DpsOp<"conv", /*traits=*/[], /*outsAccessor=*/"Output",
   }];
 }
 
+def Hip_Conv1dOp : Hip_DpsOp<"conv1d"> {
```
fhanuman: Thanks, Chi — nice work here, the bias handling and scratch-pool reuse are well done.

That said, I don't think we need a separate op, conversion pass, and kernel for 1D. As conv1d.cpp itself notes, this is really a 2D conv with H=1 — it reinterprets NCL as NC1L and calls the same miopenConvolutionForward the existing 2D wrapper uses. A dedicated hip.conv1d op + lowering + kernel (plus the mock/bitcode wiring) is a lot of parallel surface for that.

Can we instead extend the input to 2D, run the existing 2D conv, and collapse the result back to 1D? Unsqueeze/Squeeze already lower to expand_shape/collapse_shape as zero-cost metadata ops, so there's no runtime cost — and this is the usual upstream MLIR approach to avoid proliferating ops/passes. It lets us reuse hip.conv and its lowering directly.

If the shared 2D path is missing anything we need for Whisper (e.g. fp16 or bias), I'd rather we close that gap in the common kernel than fork a 1D-only one — that way both paths benefit.

fhanuman (changes requested):

include/hip/Dialect/IR/HipOps.td
chizamd: Thanks — done exactly as suggested, pushed in `c6196272`. Here's what changed.

**1D conv now reuses the 2D conv**  with height = 1, so instead of a separate op/kernel we just reshape:

- reshape input/weights from `[N, C, L]` to `[N, C, 1, L]` (add a height dim of 1)
- run the existing `hip.conv` with the height set to 1 (`kernel_shape=[1,K]`, `strides=[1,s]`, `pads=[0,b,0,e]`)
- reshape the result back from `[N, C, 1, L']` to `[N, C, L']`


**The whole dedicated 1D pipeline is deleted** (`Hip_Conv1dOp`, its lowering, `conv1d.cpp/.h`, the mock stub, bitcode wiring) — net **−362 LOC**.



**Verified:** LIT 216/216; numeric `test_conv1d.py` 7/7 (fp16 + fp32, cosine 1.0); conversion output byte-identical to the integration branch where the full Whisper end-to-end passes (encoder cosine fp16 0.999979 / fp32 1.000000, JFK quote verbatim).

fhanuman (approved):

PR #282, opened 2026-06-05
feat(onnx-to-hip): Whisper encoder/decoder attention -> hip.gqa -- 4/n [Depends on 2/n]

chizamd:
Fourth of a **5-PR cascade** adding **Whisper-large-v3** support. Routes all three Whisper attention forms through the `no_causal` / fp32 GQA path.

> **Stacked on #280** (GQA `no_causal` + fp32). The diff is against that branch; review/merge #280 first. Functionally also needs #279 (conv1d) + #281 (SkipLayerNorm) for the full encoder, but this PR compiles against #280 alone.

**LOC delta: 12 files, +1542 / −1.**

## What's new

- **`AttentionConversion.cpp`** — encoder fused-QKV `com.microsoft.Attention` → `hip.gqa(no_causal=true)`, accepting the real `[H,3H]` packed-QKV weight layout (no transpose).
- **`MultiHeadAttentionConversion.cpp`** —
  - decoder **cross-attn** (8-input, empty past) → `hip.gqa(no_causal=true)`
  - decoder **self-attn-with-past** (9-input, slot-8 `past_sequence_length`) → `hip.gqa(no_causal=false)`, keyed on the operand trio (`key`/`value`/`pastKey`/`pastValue`/`pastSequenceLength`) rather than a `past_present_share_buffer` attribute (ORT's MHA schema rejects that attr).
- **`HipGqaBuilder.cpp`** — shared `buildHipGqaCall` helper (19-slot `AttrSizedOperandSegments` layout), used by both converters; declared in `OnnxToHipUtils.h`.
- **`OnnxToHip.cpp`** registers `populateAttentionConversionPatterns`.

## Tests

- LIT `test_whisper_encoder_attention.mlir` / `test_whisper_cross_mha.mlir` / `test_whisper_self_mha_with_past.mlir`. ✅
- Numeric `test_whisper_{encoder,cross,self}_attention.py` — 18 cases, fp16 + fp32 vs ORT CPU (the no_causal path). ✅ (18/18)

## Stack overview

- **1/5 (#279):** conv1d op.
- **2/5 (#280):** GQA `no_causal` + fp32 decomposed path.
- **3/5 (#281):** `SkipLayerNormalization` → `hip.add` + `hip.layer_norm`.
- **4/5 (this PR):** Whisper encoder/decoder attention → `hip.gqa` (stacked on #280).
- **5/5:** end-to-end Whisper (model setup + tests + CLI + docs) — depends on #279–#281 + this PR.

### DIFF
--- a/lib/Conversion/OnnxToHip/AttentionConversion.cpp
+++ b/lib/Conversion/OnnxToHip/AttentionConversion.cpp
@@ -1,0 +1,287 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+#include "OnnxToHipUtils.h"
+
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinTypes.h"
+
+namespace mlir {
+namespace hip {
+namespace {
+
+/// onnx.Custom(com.microsoft.Attention) -> hip.gqa (bidirectional self-attn).
+///
+/// Whisper's encoder uses the legacy com.microsoft.Attention op with a single
+/// fused QKV projection weight + bias.  We rewrite each such node to:
+///
+///   1. One QKV projection over the fused weight (no transpose):
+///        qkv_proj  = onnx.MatMul(hidden, qkv_w)         : [B,S,H] @ [H,3H]
+///                                                            -> [B,S,3H]
+///        qkv_biased = onnx.Add(qkv_proj, qkv_b)         : broadcast on last
+///        dim
+///      The ORT com.microsoft.Attention spec stores the fused weight as
+///      [input_hidden, q_hidden+k_hidden+v_hidden] = [H, 3H] (what Whisper
+///      actually exports — `qkv_proj.weight` is [1280, 3840]).  This is the
+///      canonical layout, so the matmul consumes qkv_w directly — no transpose.
+///
+///   2. Three tensor.extract_slice ops splitting the last axis of qkv_biased
+///      into Q / K / V (`qkv_hidden_sizes`-driven offsets).
+///
+///   3. One hip.gqa with `no_causal = true` (encoder is bidirectional),
+///      `num_heads == kv_num_heads` (HPG=1, multi-head not group-query),
+///      and a compile-time `seqlens_k` / `total_sequence_length` constant
+///      equal to the static query sequence length S.  No past_key/past_value
+///      (encoder self-attn has no KV cache); the present_* outputs are
+///      DPS-style empty buffers that just satisfy the hip.gqa signature.
+///
+/// We keep the QKV projection fused (Option B) rather than splitting the
+/// weight into three independent MatMul/Add chains (Option A): the fused
+/// layout matches what the original ONNX graph already has and produces a
+/// single large GEMM (cleaner lowering, fewer hipBLASLt launches).
+struct AttentionToHip : public mlir::RewritePattern {
+  AttentionToHip(mlir::MLIRContext *ctx)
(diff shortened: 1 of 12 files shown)


lib/Conversion/OnnxToHip/HipGqaBuilder.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include "OnnxToHipUtils.h"
+
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinTypes.h"
+
+namespace mlir {
+namespace hip {
+
+mlir::LogicalResult
+buildHipGqaCall(mlir::Operation *op, mlir::PatternRewriter &rewriter,
```
fhanuman, line 29: I have only looked at general conversion pass checks. I am assuming the sematic of the operator is captured properly and e2e test that you have added verified the functionality. Same comment in other places too.

lib/Conversion/OnnxToHip/AttentionConversion.cpp
```diff
@@ -0,0 +1,287 @@
...
+namespace hip {
+namespace {
+
+/// onnx.Custom(com.microsoft.Attention) -> hip.gqa (bidirectional self-attn).
+///
+/// Whisper's encoder uses the legacy com.microsoft.Attention op with a single
+/// fused QKV projection weight + bias.  We rewrite each such node to:
+///
+///   1. One QKV projection over the fused weight (no transpose):
+///        qkv_proj  = onnx.MatMul(hidden, qkv_w)         : [B,S,H] @ [H,3H]
```
fhanuman, line 21: It's nice that we are decomposing these operations to use the existing infra.

fhanuman (approved):

lib/Conversion/OnnxToHip/HipGqaBuilder.cpp
chizamd: Thanks — and good call flagging the semantic side explicitly. You're right that the conversion-pass mechanics being correct (operand segments, DPS inits, types) says nothing about whether the *attention math* is right, so here's what backs that up beyond the e2e test.

Each attention variant has a dedicated **op-level numeric test** that compares the EP output against an ORT CPU reference for that exact ONNX subgraph (not just the full model):

- `test/numeric/tests/test_whisper_encoder_attention.py` — fused-QKV self-attn, `no_causal` path
- `test/numeric/tests/test_whisper_cross_attention.py` — cross-attn (null past KV)
- `test/numeric/tests/test_whisper_self_attention.py` — decoder self-attn with past KV

Tolerances are tight: fp32 `cos >= 0.9999 / atol 2e-3`, fp16 `cos >= 0.999 / atol 2e-2`. So a wrong scale / QKV layout / causal-vs-bidirectional mask would fail these, not silently pass.

Two things worth calling out on the "same comment in other places" point:

1. **The shared builder is the only semantic surface.** Both callers (MultiHeadAttention for decoder self/cross-attn, Attention for encoder fused-QKV) funnel through `buildHipGqaCall`, so the operand-order / segment-table logic you reviewed is exercised by all three numeric tests above — there's no second hand-rolled path to get out of sync.

2. **`no_causal` defaults off, so existing callers are untouched.** The new flag is `DefaultValuedAttr<BoolAttr, "false">`; Llama / gpt-oss GQA produce byte-identical output (pinned by `test/lit/gqa_no_causal_default_off.mlir`). Whisper is the only path that sets it true.

Full whisper suite (encoder / prefill / decode correctness fp16+fp32, e2e greedy transcription) passes on gfx1151 with per-step decode logits at cosine 1.0 vs CPU.

PR #283, opened 2026-06-05
feat(whisper): end-to-end Whisper-large-v3 support (integration) -- 5/n [Depends on 1/n-4/n]

chizamd:
Final, **integration** PR of the cascade adding end-to-end **Whisper-large-v3** (fp16 **and** fp32 encoder-decoder speech-to-text) to the MorphiZen EP — the first encoder-decoder / cross-attention model on the stack.

> **Integration PR (base = `main`).** It bundles the four cascade slices — 1D conv via 2D reshape (#279), GQA `no_causal`+fp32 (#280), SkipLayerNormalization (#281), Whisper attention converters (#282) — **plus** the Python model-build / test / CLI / docs layer, so the whole model runs and is verified end-to-end. The cascade's `#279`–`#282` are the reviewable per-feature slices; this PR is where it all comes together and the e2e gate runs. Merge the four first, then this.

> **Conv lowering (carries #279's reviewed change).** Rank-3 `Conv` reshapes to the **shared** `hip.conv` (NCL → `NC1L` via `tensor.expand_shape`/`collapse_shape`, zero-cost) instead of a dedicated `hip.conv1d` op — the shared 2D MIOpen wrapper gained fp16/fp32 + per-channel bias + a `conv_scratch` workspace pool, and `PoolAllocs` now pools a single alloc so the lone-conv graph doesn't leak an undefined `hip_device_malloc`. Verified here: LIT 216/216, numeric `test_conv1d.py` 7/7 (fp16+fp32, cosine 1.0). See #279 for the per-feature breakdown.

## Pipeline

```
audio ──► feature extraction ──► ENCODER ──► cross-KV
                                                │
          START_TOKENS ──► DECODER-PREFILL (S=4) ─┤──► first token
                                                │
                     ┌────► DECODER-DECODE (S=1) ─┤──► next token ──┐
                     └──────────────(loop until <|endoftext|>)──────┘
```

## Reproducible fp16 + fp32 from one pinned builder

Both precisions are **built locally** from a single pinned **OGA (onnxruntime-genai) DirectML model builder** — there is **no model download**. `python build.py --build-whisper-models` runs the builder for `-p fp16` and `-p fp32` at a pinned HF commit revision, in a **dedicated isolated venv** (`install/whisper-builder-venv/`) so the stock `onnxruntime-genai-directml` the builder needs never shadows the OGA *fork* that `build.py --build-oga` installs. Pinning the builder deps + the HF revision makes both models deterministic.

The OGA fp16 and fp32 graphs share an **identical layout**, so one surgery pipeline (`inject_seqlens_k` + `fix_shapes`) serves both.

**fp16 is the default running precision** — faster on GPU and **argmax-lossless**, because the OGA build keeps `lm_head` in fp32 while the body is fp16. Greedy decode emits the verbatim JFK quote, token-identical to the CPU-fp16 reference (prefill logit cosine 1.0). Pass `--fp32` to opt out.

## What this PR adds on top of #279–#282

- **`scripts/build_whisper_models.py`** *(new)* — the pinned, isolated-venv OGA builder for both precisions (idempotent; revision pinned via the builder's real `from_pretrained` / `hf_hub_download` path).
- **`build.py`** — `--build-whisper-models` flag (opt-in, like `--build-vulkan`).
- **`test/python/conftest.py`** — `setup_whisper_model_dir` / `setup_whisper_fp16_model_dir` are **consume-only** (run `inject_seqlens_k` surgery + `fix_shapes` on the already-built raw model; raise a build hint if absent); `make_whisper_inputs`; LibriSpeech + jfk provisioning (auto-download, **no committed audio**).
- **`test/python/whisper/`** — e2e correctness + perf suite, the shared `whisper_infer` harness (**IOBinding + KV-cache aliasing as the default decode path**, dtype-threaded for fp16/fp32, bit-identical to the numpy fallback), `inject_seqlens_k` unit tests.
- **`scripts/`** — `transcribe_whisper.py` (standalone CLI: **fp16 default**, `--fp32` opt-out, `--metrics` / `--warmup` / `--cpu` / `--compare`, consume-or-instruct), `setup_whisper_model.py` (**fp16 default**, `--fp32`), `build_whisper_vulkan.py`.
- **`docs/whisper_quick_start.md`**, **`CLAUDE.md`** gotchas, **`environment.yml`** (`transformers==4.49.0` test-runtime dep only), **`.gitignore`**.

## Cross-backend × precision perf (jfk.wav ~11 s, gfx1151, rep-averaged steady-state)

| Backend | precision | decode tok/s | encoder ms | transcription |
|---|---|---|---|---|
| MorphiZen EP (GPU) | **fp16 (default)** | **~45** | **~460** | JFK quote ✅ |
| MorphiZen EP (GPU) | fp32 | ~24 | ~1300 | JFK quote ✅ |
| CPU EP | fp32 | ~22 | — | JFK quote ✅ |
| CPU EP | fp16 | ~18 | — | JFK quote ✅ (emulated) |
| DirectML EP | fp16 / fp32 | N/A | — | decoder cross-attn MHA rejected on DML |
| Vulkan (whisper.cpp) | f16 GGUF | ~77 | — | JFK quote ✅ (loose reference) |

fp16 is **both correct and faster** on MorphiZen (~1.9× decode, ~2.8× encoder vs fp32). The in-suite fp16 number (~45) matches the isolated `scripts/transcribe_whisper.py` (~45–48); an earlier perf-test artifact — setting `HIPDNN_EP_DEBUG=1` (which latches a process-wide `static const` debug flag → per-op FD logging for the whole process) plus pytest `capfd` FD capture — under-reported it at ~28 and is now fixed (the test never sets that env var; FD capture is scoped to session-create only). DML cannot run the decoder in either precision (a DML `MultiHeadAttention` op-coverage gap on the cross-attn layout — confirmed by a controlled CPU-vs-DML A/B); skipped by default, opt in via `HIPDNN_WHISPER_PERF_DML=1`. Numbers vary run-to-run; quote as approximate.

## Test Plan (all verified on gfx1151)

Requires `THEROCK_DIST` + `install/{therock,dist}/bin` on PATH (see `docs/whisper_quick_start.md`).

- [x] `python build.py --build-whisper-models` (builds fp16 + fp32 in the isolated pinned venv)
- [x] `python scripts/setup_whisper_model.py` (consume-only: surgery + fix_shapes)
- [x] `ctest -R MorphizenMLIRLitTests` (MLIR conversions) — pass
- [x] numeric per-op (conv1d / layer_norm / encoder+cross+self attention, fp16 + fp32) vs ORT CPU — **46/46**
- [x] `pytest test/python/whisper/test_inject_seqlens_k.py` — **3/3**
- [x] fp32 correctness + e2e (decode per-step cosine 1.0; verbatim JFK quote) — **4/4**
- [x] fp16 correctness (prefill cosine 1.0; verbatim JFK quote) — **2/2**
- [x] LibriSpeech (5 GPU==CPU + WER) + long-30s GPU==CPU verbatim — **11/11** (WER 0–10%, long-clip 3.4%)
- [x] cross-backend × precision perf matrix — pass

## Stack overview

- **1/n (#279):** 1D conv via 2D reshape + shared-kernel fp16/bias/scratch.
- **2/n (#280):** GQA `no_causal` + fp32 decomposed path.
- **3/n (#281):** `SkipLayerNormalization` → `hip.add` + `hip.layer_norm`.
- **4/n (#282):** Whisper encoder/decoder attention → `hip.gqa` (stacked on #280).
- **5/n (this PR):** end-to-end integration — reproducible fp16+fp32 OGA builder (fp16 default) + Python/CLI/docs (base=main).


test/numeric/tests/test_whisper_self_attention.py
```diff
@@ -0,0 +1,255 @@
+#
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+# Licensed under the MIT License.
+#
+
+"""Numeric tests for Whisper-large-v3 decoder SELF-attention (with past KV).
```
fhanuman, line 6: This is a lot of test code in this PR. Is there way we can abstract some of this? `torch-mlir` has `e2e` infra, can we check if there is something we can inspire from there?

lib/Conversion/OnnxToHip/MultiHeadAttentionConversion.cpp
```diff
@@ -114,6 +143,140 @@ mlir::LogicalResult MultiHeadAttentionToHip::matchAndRewrite(
...
+    //
+    // total_seq_len: the runtime infers `(past_seq_len + S)` itself from
+    // seqlens_k + the query's S dim, so we don't need to compute it on the
+    // MLIR side.  We pass `seqlens_k` again as a placeholder (any rank-0 or
+    // rank-1 i32 tensor works since the kernel reads seqlens_k for the actual
+    // KV length).  Actually hip.gqa REQUIRES a scalar total_seq_len, so we
+    // emit a constant tensor of `present_key.shape[2]` — the static buffer
+    // size from the IR (== `past_sequence_length_max + S`, baked in at compile
+    // time).
+    auto presentKType =
```
fhanuman, line 181: I am hoping that we have verified all this generated code, if not, please do so. It's not feasible for reviewers to go through every-line of code and confirm the functionality. Same in other places too.

chizamd: Thanks for flagging this! Quick heads up — after rebasing onto latest `main`, this file (and the attention-conversion code in general) is no longer part of this PR's diff. It actually landed in `main` via #350. So this thread is now pointing at code that's already upstream 🙂

On your underlying point though — totally fair, and worth answering properly. 

The attention math is genuinely checked, not just eyeballed. Each of the three attention forms has its own op-level numeric test comparing the EP output against an ORT CPU reference for that exact subgraph 
-  `test_whisper_{encoder,cross,self}_attention.py` (18 cases, fp16 + fp32, tight tolerances) 
- plus the LIT shape tests. 

And the full Whisper e2e (encoder / prefill / decode correctness fp16+fp32, greedy transcription) passes on gfx1151 with per-step decode logits at cosine 1.0 vs CPU. 

So a wrong scale / QKV layout / causal-vs-bidirectional mask would fail these rather than slip through. 

(Same detail as my reply over on #282 #discussion_r3400072303 if useful.)

test/numeric/tests/test_whisper_self_attention.py
chizamd: Good point, thanks! Same note as below— after the rebase onto `main`, this test file is no longer in this PR's diff; it landed via #350. So the per-op attention tests are upstream now, and what's left in this PR is just the Whisper *integration* layer (e2e harness, build scripts, docs, conftest).

On the abstraction idea — I'm with you, it's worth doing. The op-level tests already share infra (`framework.comparator.compare_outputs`, `framework.onnx_utils.make_model_from_nodes`); most of the per-file size is each attention form building its own distinct ONNX subgraph, which is genuinely different topology.

chizamd: Followed up on the abstraction idea properly — opened #352 to track it.

What I found, in plain terms:

**The reusable machinery already exists.** `test/numeric/framework/` already has the shared pieces — the backend abstraction, the ORT-CPU reference + its disk cache, `compare_outputs`, and `make_model_from_nodes`. Tests already call into these. So the duplication you saw isn't "everyone reimplements the engine" — it's the thin glue around each test (build inputs → run → compare), repeated by hand in every file.

**What torch-mlir adds that we don't have yet** is a declarative layer on top: instead of writing a test class + method, you *declare* a case (its model, its inputs, its reference, its tolerances) and register it, and one shared driver runs them all. That's the part worth borrowing.

**The one tricky requirement.** Most ops can get their reference by just running the same ONNX on ORT CPU. But the attention tests can't — ORT's CPU fp16 attention is numerically off (cosine drifts to ~0.6), so those tests compute their own reference with a small numpy function instead. So the harness has to support *two* kinds of reference equally: "run it on CPU" and "call this python function." If it only supported the CPU one, the attention tests would have to keep bypassing the harness like they do today.

Scoping it as its own incremental PR (build the registry, port one CPU-reference test and one numpy-reference test to prove it out, then convert the rest) rather than folding a suite-wide test-framework change into this integration PR. Details + open questions in #352.

chizamd:
@fhanuman — quick reviewer's guide now that this PR has been rebased onto latest `main`.

**TL;DR:** all the compiler/runtime/op-level changes from this cascade (conv, SkipLayerNorm, GQA `no_causal`, the attention conversions, and their per-op numeric tests) have already merged into `main` via #279 / #280 / #281 / #350. After rebasing, **this PR is now purely the Whisper *integration + end-to-end test* layer** — no new conversion passes, no new kernels, no per-op tests. It's all Python + docs.

### How to quickly review this PR

**1. The model-prep plumbing** 

- `test/python/conftest.py` (+565) — the shared setup helpers. 
    - The one that matters for correctness is `inject_seqlens_k()`: ONNX "surgery" that rewrites the 32 decoder self-attention `MultiHeadAttention` nodes to take a `past_sequence_length` input (slot 8), leaving the 32 cross-attn nodes alone. 
    - Plus `setup_whisper_model_dir` / `_apply_whisper_surgery_and_fix_shapes` (fp32 + fp16 prep) and `make_whisper_inputs`.
- `test/python/whisper/test_inject_seqlens_k.py` (+193) 
    - unit test for *just* that surgery (counts nodes, checks slot-8 wiring, verifies cross-attn untouched). Read it right after the helper — it documents the contract.

**2. The inference harness**
- `test/python/whisper/whisper_infer.py` (+538) — encoder→decoder greedy-decode driver for both CPU and MorphiZen EP (`encoder_cross_kv`, `greedy_decode_cpu`, `greedy_decode_morphizen`, IOBinding path). This is the reusable engine the tests call.

**3. The actual tests** — `test/python/whisper/test_whisper.py` (+1502). Each is GPU-vs-CPU, run for both fp16 and fp32:

| Test | What it checks |
|---|---|
| `test_encoder_correctness` | encoder hidden-states cosine, GPU vs CPU |
| `test_decoder_prefill_correctness` | first decode step (S=4, empty past) logits cosine |
| `test_decoder_decode_correctness` | N sequential S=1 decode steps, per-step logits cosine |
| `test_e2e_transcription_greedy` | full greedy transcription of jfk.wav == CPU quote |
| `test_perf_decode_tps` | cross-backend decode tok/s + accuracy (perf table) |
| `test_librispeech_gpu_vs_cpu` | GPU greedy == CPU greedy verbatim, per LibriSpeech clip |
| `test_librispeech_wer` | WER of GPU transcription vs ground truth (lenient) |
| `test_long_30s_gpu_vs_cpu` | ~24s concatenated clip, GPU greedy == CPU greedy |

**4. Standalone scripts + docs (skim)**
- `scripts/build_whisper_models.py` — reproducible fp32+fp16 ONNX from a pinned OGA builder (isolated venv).
- `scripts/setup_whisper_model.py` / `transcribe_whisper.py` — prep an already-built model / CLI transcribe on GPU.
- `scripts/build_whisper_vulkan.py` — whisper.cpp Vulkan baseline (perf reference only).
- `docs/whisper_quick_start.md`, `CLAUDE.md` (+gotchas), `environment.yml`, `.gitignore`.

### Verification status
Full suite passes on gfx1151 for **both fp16 and fp32**: encoder/prefill/decode correctness (per-step decode logits cosine 1.0 vs CPU), e2e greedy transcription matches the CPU quote verbatim. The conv front-end and attention dispatch were confirmed on GPU via the `[REAL] wrap_*` tripwire (no silent CPU fallback).

The "lots of test code" point from your earlier review is being tracked separately in #352 (declarative numeric-test harness) — that's a `test/numeric/` refactor, independent of this integration PR.

fhanuman (approved):
Thank you for creating the issue to track test code bloating. Appreciate it.

PR #286, opened 2026-06-06
docs(design): pool-allocs memory planning design doc

fhanuman:
## Summary
Stack PR **1/4**. Adds the design document for the dynamic-shape memory-pooling subsystem.

📄 **Rendered:** [docs/design/pool-allocs-memory-planning.md](https://github.com/ROCm/onnx-hipdnn-ep/blob/def33fa37fe1aab61fca0c478075ba99f992970e/docs/design/pool-allocs-memory-planning.md) (pinned to this PR's tip — formatted view with working in-page anchors)

**Why this exists / why now:** `hip-pool-allocs` is a generic algorithm — it packs every `memref.alloc` into a small number of grow-on-demand pools. Its one hard constraint is dominance: every alloc size must be computable above its pool's acquisition point. When a program has dynamic-size allocations whose sizes derive from many independent, scattered SSA def points, pool-allocs cannot cover them with one pool and partitions them into separate **dominance domains**, each backed by its own runtime pool. The number of domains is **unbounded** — the runtime grows its per-domain pool arrays on demand (reallocating, zero-filling new slots, the first time a higher `domain_id` is seen), so opening another domain costs one array slot, never a compile failure. Two cooperating concerns round out the design: pre-/post-bufferize passes that collapse scattered size queries to a common root (keeping the common case at one domain), and a per-domain pool runtime ABI for the residue that genuinely needs more. The algorithm is model-agnostic; reshape-heavy dynamic-shape graphs are simply the workload that first stressed it.

**This doc is the design those passes implement** — it is the source of truth (normative) for the code PRs stacked on top, not a retro-active write-up of already-merged behavior. Reviewers should hold the stacked code PRs (#293, #287, #288) to the design described here — and #287 in particular to its *Compiler-Runtime ABI* and *Module-Attribute Contract* sections (neither #293 nor #288 touches those contracts; they are upstream IR-shaping passes).

Covers: dominance-domain partitioning, the cooperating passes (`hip-infer-shapes`, `hip-resolve-tensor-dims`, `hip-hoist-alloc-size-arith`, `hip-pool-allocs`), the per-domain pool runtime API, failure modes, retirement contracts, and a worked example.

## Reading guide (~875 lines)
- **Mental model:** *Overview* → *Pipeline Position* → *The hip-pool-allocs Algorithm*.
- **Normative contracts (what #287 must match):** *Compiler-Runtime ABI* + *Module-Attribute Contract*.
- **Ground it:** *Worked Example: Two-Domain Partition*.
[...]

PR #287, opened 2026-06-06
feat(hip): per-domain memory pool placement engine + runtime ABI

fhanuman:
## Summary
Stack PR **3/4** (base: #293). Implements the placement engine and per-domain runtime ABI that the #286 design doc specifies. The placement pass is generic — it pools `memref.alloc`s by dominance and offset-packs them; "GPU" only enters on the runtime side, where the pools back device/host-mapped memory via `hipdnn_ep_get_pool_base`. When an alloc's dynamic size can't be computed above an earlier pooled alloc, the pass splits allocs into independent **dominance domains**, one runtime pool each; the domain count is **unbounded** (the runtime grows its pool arrays on demand), so opening a domain costs one array slot, never a compile failure. The `--hip-hoist-alloc-size-arith` pass (#293) runs first to collapse the common case to a single domain; this PR consumes its output.

📄 **Design (normative):** [docs/design/pool-allocs-memory-planning.md](https://github.com/ROCm/onnx-hipdnn-ep/blob/176e58c62ea2ef1bc6d138b1d9d2a44989776c37/docs/design/pool-allocs-memory-planning.md) (pinned to this PR's tip) — hold this PR to its *Compiler-Runtime ABI* and *Module-Attribute Contract* sections.

**What's in this PR (behavioural):**

* **`hip-pool-allocs` per-domain placement.** Emits per-bucket size arithmetic and one `hip.get_pool` per domain at the latest legal single-block position (after every dyn-operand def, before the earliest pooled alloc — no code motion of its own). Static allocs get best-fit offset packing; dynamic allocs are bucketed by a structural byte-size key and bin-packed by lifetime so disjoint lifetimes share an offset.
* **`hip.get_pool` gains an optional `domain_id : i64` attribute** (default 0, elided by the printer — single-domain textual IR is byte-identical to the pre-multi-domain output). `ConvertHipToLLVM` lowers it to `hipdnn_ep_get_pool_base(state : ptr, domain_id : i32, size : i64) -> ptr`.
* **Runtime: unbounded per-domain pools.** `RuntimeState` holds heap `pool_base`/`pool_size` arrays sized by `num_pool_domains`, grown on demand the first time a higher `domain_id` is seen (zero steady-state host allocations; never shrinks; `realloc`-move-safe). There is no fixed cap. A high domain count emits a non-fatal advisory remark (a likely sign that upstream hoisting is missing) rather than failing compilation.
* **IR-dump ergonomics.** `HIPDNN_EP_IR_DUMP_PATH` now inserts the per-compile counter *before* the extension (`ir.mlir` → `ir.0.mlir`), plus `HIPDNN_EP_IR_DUMP_AFTER_ONLY` (suppress the per-pass "before" dump) and `HIPDNN_EP_IR_DUMP_TREE` (one file per pass under a directory).

**Single-domain compatibility:** the common (one-domain) path stays bit-identical to the pre-multi-domain output — one `hip.get_pool` with `domain_id` elided, legacy `hipdnn.pool_size` / `buffer_count` / `buffer_offsets` attributes only. The multi-domain metadata (`hipdnn.domain_count` / `pool_sizes` / `buffer_domains`) is emitted only when a function actually partitions into ≥2 domains. This PR also removes the fixed cap #286 flagged: `RuntimeState` no longer carries `pool_base[8]`/`pool_size[8]`, and `hip-pool-allocs` no longer bails past a domain count.

## Stack
1. #286 — design doc (base: `main`)
2. #293 — `--hip-hoist-alloc-size-arith` pass (base: #286)
3. **#287** — pool placement engine + runtime ABI (base: #293) ← you are here
4. #288 — reshape-dim resolution (base: #287)


fhanuman:
Hi @AMDmoore and @zpye , I will go ahead and merge the changes to unblock both of your submissions — I have reviewed the changes myself. If we run into any issues during the integration of other dynamic shape changes, I will help fix them.

PR #295, opened 2026-06-08
docs(design): add output allocator design doc

zpye:
## Summary

Adds `docs/design/output-allocator-design.md` — a **design** (the *what*, not an implementation plan) for an **in-graph output allocator** that lets the compiled `main_graph` allocate each output **at the point its shape is computed**, pulling the buffer from an EP-supplied allocator. The graph owns *when / what shape*; the EP owns *where the memory comes from*.

Documentation only — no code changes.

## Motivation

Today, outputs are converted to out-params by `BufferResultsToOutParams` and allocated by the EP **before** `main_graph` runs. For dynamic shapes the output extent is only known **inside** the graph (e.g. `M = memref.dim %input`), so the EP must re-derive output shapes outside the graph — duplicating shape computation and making any allocator retrofit into the classic pipeline complex. This doc proposes allocating outputs **in-graph** instead, behind a dedicated pipeline.

## What the doc covers

- **Overview & scope** — shape-derived dynamic outputs; data-dependent outputs (`NonZero`, `Range`) explicitly deferred.
- **Two pipelines** — Classic (out-params, unchanged) vs Allocator, selected by a `use-output-allocator` pipeline option; no conditional logic bolted onto the classic path.
- **Phased design (5 layers):**
  1. **In-graph allocation abstraction** — new `hip.alloc_output` op; `hip-use-output-allocator` pass rewrites output `memref.alloc` → `hip.alloc_output`.
  2. **LLVM lowering** — lower to a `hipdnn_ep_alloc_output` runtime call + memref descriptor.
  3. **Runtime allocator contract** — allocator interface struct + runtime exports; allocator always returns a device pointer.
  4. **Allocator-aware codegen** — `generate-allocator-interface` emits a 2-arg `inference_compute`; `main_graph`'s `return` is left untouched (the `-> i32` entry is synthesized by `convert-hip-to-llvm`, same as the classic path).
  5. **EP bridge & zero-copy** — sequence diagram, EP allocator callback, EP-owned D2H flush, KV-cache shared-buffer handling.

## Status

Draft design for review. Opened as a draft PR.

## Test plan

- [x] Documentation only — no build/test impact.
- [x] Mermaid diagrams render on GitHub.
- [ ] Design review / sign-off on the proposed `hip.alloc_output` op + `use-output-allocator` pipeline.


fhanuman (commented):
Thanks @zpye for the detailed design documentation. Appreciate the effort here.

fhanuman (commented):
Overall proposal sounds good to me. There are few clarifications regarding the compiler pipeline, will check with you after the meeting today.

fhanuman (approved):
LGTM

PR #296, opened 2026-06-08
feat(hip): add hip.alloc_output op + hip-use-output-allocator pass (phase 1)

zpye:
## Summary

Phase 1 of the output-allocator design (`docs/design/output-allocator-design.md`). Standalone building blocks only -- the new pass is registered for `hip-mlir-opt` / LIT but **not** wired into any pipeline yet (placement is a later phase).

- **`hip.alloc_output` op** (`include/hip/Dialect/IR/HipOps.td`, `lib/Dialect/IR/HipDialect.cpp`): obtains a graph-output buffer from the EP output allocator. Carries a generic **Write** effect (intentionally **not** `Allocate`) so the buffer is never pooled or freed by buffer-deallocation / `hip-pool-allocs`, while the op still survives DCE and CSE. Verifier checks the dynamic-size operand count matches the result memref's dynamic dims.
- **`hip-use-output-allocator` pass** (`lib/Dialect/Transforms/UseOutputAllocator.cpp`): an `OpRewritePattern<memref::AllocOp>` run via `applyPatternsGreedily` that rewrites each returned `memref.alloc` into `hip.alloc_output`, sets `out_idx` from the `func.return` operand position, dedupes aliased multi-outputs (same alloc returned twice), and erases any `memref.dealloc` of the returned buffer.
  - **Public-only scope:** only public (graph-entry) functions are rewritten. Private helpers (e.g. outlined `onnx.Loop` bodies) also carry `!hip.context` arg 0 and return `memref.alloc`s, but their results are DLL-internal and must never become EP outputs (rewriting would emit an `out_idx` colliding with real graph outputs). Functions lacking `!hip.context` arg 0 are also skipped. Signature and `func.return` are untouched.
  - **Module marker:** the pass also **unconditionally** stamps the `hipdnn.use_output_allocator` unit attribute on the parent module -- the allocator-mode marker that downstream `convert-hip-to-llvm` / `generate-interface` read to select the allocator ABI. Stamped even when no alloc is rewritten, because the mode is decided by the pass running, not by IR contents (a zero-output graph must still be marked). Same idempotent FuncOp-writes-module pattern as `PoolAllocs` stamping `hipdnn.pool_size`.
- **Build:** adds `/bigobj` to the `HipDialectIR` target -- the `GET_OP_CLASSES` translation unit (`HipDialect.cpp`) now exceeds the MSVC COFF section limit (fatal error C1128) once the new op is added. Documented as a build gotcha in `CLAUDE.md`.

## Test plan

- [x] `--hip-use-output-allocator` is listed by `hip-mlir-opt --help` (registered)
- [x] No reference to the pass in `lib/Dialect/Transforms/Pipelines.cpp` (not in any pipeline)
- [x] New LIT tests pass:
  - `test/lit/Dialect/hip-use-output-allocator.mlir` -- transforms + left-alone (intermediate, passthrough, no-context, dealloc-erase, aliased multi-output); the `hipdnn.use_output_allocator` module-attr stamp; **public-only guard** (`@loop_body`, `@priv_with_dealloc` -- private fns keep their alloc/dealloc and are not rewritten); **`@return_order`** (`out_idx` follows `func.return` position, not definition order); plus a realistic real-op graph `@main_graph` (design-doc Add -> MatMul -> Sigmoid)
  - `test/lit/Dialect/hip-alloc-output.mlir` -- op round-trip + verifier failures (too few / too many dynamic sizes)
  - `test/lit/Dialect/hip-alloc-output-no-dce.mlir` -- memory-effect survival under `--canonicalize` and `--cse`
- [x] Full `MorphizenMLIRLitTests` suite green (214 tests, no regressions)


lib/Dialect/Transforms/UseOutputAllocator.cpp
```diff
@@ -0,0 +1,116 @@
...
+// docs/design/output-allocator-design.md, Phase 1).
+//
+// Before:
+//   func.func @main_graph(%ctx: !hip.context, ...) -> memref<?x?xf16> {
+//     %out = memref.alloc(%M, %N) : memref<?x?xf16>      // returned output
+//     hip.sigmoid(%ctx) ins(%t) outs(%out)
+//     return %out : memref<?x?xf16>
+//   }
+//
+// After:
```
fhanuman, line 43: This is very helps. Thanks

fhanuman (dismissed):

wcy123:
Consider using `OpRewritePattern<memref::AllocOp>` instead of a custom pass with manual walk. This transformation is a good fit for the pattern-based approach:

```cpp
struct ReplaceOutputAllocPattern : public OpRewritePattern<memref::AllocOp> {
  using OpRewritePattern::OpRewritePattern;
  
  LogicalResult matchAndRewrite(memref::AllocOp allocOp,
                                PatternRewriter &rewriter) const override {
    // Find if this alloc is returned (graph output)
    func::ReturnOp returnOp = nullptr;
    int64_t outIdx = -1;
    for (OpOperand &use : allocOp->getUses()) {
      if (auto ret = dyn_cast<func::ReturnOp>(use.getOwner())) {
        returnOp = ret;
        outIdx = use.getOperandNumber();
        break;  // First occurrence for aliased outputs
      }
    }
    if (!returnOp)
      return failure();  // Not a graph output
    
    // Get !hip.context from function arg 0
    auto func = allocOp->getParentOfType<func::FuncOp>();
    if (func.getNumArguments() == 0 || 
        !isa<hip::ContextType>(func.getArgument(0).getType()))
      return failure();
    Value ctx = func.getArgument(0);
    
    // Remove deallocs (EP owns buffer)
    for (Operation *user : llvm::make_early_inc_range(allocOp->getUsers()))
      if (auto dealloc = dyn_cast<memref::DeallocOp>(user))
        rewriter.eraseOp(dealloc);
    
    // Replace with hip.alloc_output
    auto allocOutput = hip::AllocOutputOp::create(
        rewriter, allocOp.getLoc(), allocOp.getType(), ctx,
        allocOp.getDynamicSizes(), 
        rewriter.getI64IntegerAttr(outIdx));
    
    rewriter.replaceOp(allocOp, allocOutput.getResult());
    return success();
  }
};

void UseOutputAllocatorPass::runOnOperation() {
  RewritePatternSet patterns(&getContext());
  patterns.add<ReplaceOutputAllocPattern>(&getContext());
  
  if (failed(applyPatternsAndFoldGreedily(getOperation(), std::move(patterns))))
    signalPassFailure();
//
}
```

**Benefits:**
- Standard MLIR pattern idiom (more maintainable)
- Eliminates manual deduplication bookkeeping
- Composable with other patterns if needed
- Greedy driver handles convergence

Current implementation works fine, but patterns are preferred for local transformations like this.

qianglin-amd (approved):

wcy123 (commented):
## MLIR Isolation Contract Violation

`UseOutputAllocatorPass` is `OperationPass<func::FuncOp>` but modifies parent module (line 137):

```cpp
getOperation()->getParentOfType<ModuleOp>()->setAttr(...)
```

**MLIR contract** ([source](https://mlir.llvm.org/docs/PassManagement/)):
> "Passes must not modify the state of operations other than the operations that are nested under the current operation."

> "Exception: attributes of the **current operation** may be modified freely."

`getOperation()` returns `func::FuncOp` (current). `getParentOfType<ModuleOp>()` returns parent (not current) → **violation**.

**Additional issue**: Pass runs on each function. If module has 3 functions, `runOnOperation()` called 3 times → module attribute set **3 times**.

**Upstream search**: `llvm/llvm-project` for `getParentOfType<ModuleOp>` + `setAttr` from FuncOp pass → **zero results**.

**Tradeoff:**

| Approach | Isolation | Attribute Sets | Parallelization |
|----------|-----------|----------------|----------------|
| FuncOp pass (current) | ❌ Violates | N times (wasteful) | ✓ Automatic |
| ModuleOp pass | ✓ Respects | Once | ❌ Lost (manual `walk()`) |

**Code comment claims**:
> "safe here because the write is an idempotent set of the same UnitAttr"

Thread safety ≠ idempotency. Isolation contract enables parallelization, not just correctness.

**Alternative:**

```cpp
// Passes.td: Change to ModuleOp
def UseOutputAllocatorPass : Pass<"hip-use-output-allocator", "ModuleOp">

struct UseOutputAllocatorPass : OperationPass<ModuleOp> {
  void runOnOperation() override {
    ModuleOp module = getOperation();
    module->setAttr("hipdnn.use_output_allocator", UnitAttr::get(&getContext()));  // Set once
    
    module.walk([&](func::FuncOp func) {
      // Transform functions
    });
  }
};
```

**Recommendation**: Refactor to ModuleOp pass or document why violating isolation is unavoidable.

PR #297, opened 2026-06-08
feat(hip): lower hip.alloc_output to @hipdnn_ep_alloc_output (phase 2)

zpye:
﻿## Summary

Phase 2 of the output-allocator design (`docs/design/output-allocator-design.md`). Rebased onto `main` now that Phase 1 (#296) is merged, so the diff is just the Phase 2 lowering (3 files).

- **Lower `hip.alloc_output` to LLVM.** New `AllocOutputOpLowering` in `lib/Conversion/HipToLLVM/MemoryLowering.cpp` stack-allocates an `i64[rank]` shape array (static dims as `llvm.mlir.constant`, dynamic dims from the op operands, in type order), calls the EP output allocator `@hipdnn_ep_alloc_output(state, out_idx, shape, rank, elem_size) -> ptr`, and wraps the returned device pointer in a standard row-major memref descriptor.
- **EP-owned, address-space aware.** No `hipMalloc`/`hipFree` is emitted (the buffer is a graph output — matches `AllocOutputOp`''s no-Allocate effect). The generic (AS 0) pointer is `addrspacecast` to the result memref''s address space only when they differ (AS 0 results emit no cast).
- **Wiring.** Register the pattern in `populateMemoryLoweringPatterns`; add the `kHipAllocOutput` symbol constant in `HipToLLVMUtils.h`.
- **Tests.** Add `test/lit/Conversion/hip-to-llvm/test_alloc_output.mlir` covering static / fully dynamic / mixed / rank-1 shapes, the AS-0 no-cast path, dynamic sizes sourced from `memref.dim` / `arith.muli` / `memref.load`, and downstream consumption of the result by `hip.sigmoid`.

## Test plan

- [x] Full ctest suite (LIT + e2e compile/execute): 115/115 passing, 0 failed (no regressions)
- [x] `MorphizenMLIRLitTests` (incl. new `test_alloc_output.mlir` + Phase 1 output-allocator LIT tests) passes
- [x] `pre-commit run` on the 3 changed files: all hooks pass (whitespace, license headers, lintrunner/clang-format/ruff)

## Follow-ups (later phases, not in this PR)

- Phase 3: runtime allocator contract (`hipdnn_ep_alloc_output` / `hipdnn_ep_set_output_allocator`)
- Phase 4: allocator-aware codegen (`generate-allocator-interface`)
- Phase 5: EP bridge + zero-copy output binding

PR #298, opened 2026-06-08
feat(runtime): add output allocator runtime contract (phase 3)

zpye:
## Summary
- Adds the Phase 3 **runtime contract** for the output allocator: two entry points bridging the MorphiZen EP and the generated `model.dll`.
  - `hipdnn_ep_set_output_allocator` (exported) installs an allocator onto `RuntimeState` before `inference_compute`.
  - `hipdnn_ep_alloc_output` (internal) forwards the lowered `hip.alloc_output` call to the installed callback, null-guarding when none is installed.
- `hipdnn_output_allocator_t` (`{ self, allocate }`) is a **fixed-layout ABI contract** across the `model.dll` <-> EP boundary — treated exactly like the sibling wire struct `tensor_t`: the layout is locked with `static_assert`s and the EP-side copy is declared identically. There is intentionally **no** self-describing size/version field: a layout change is an ABI break that requires rebuilding the `model.dll`. Total absence of the contract is handled by symbol resolution (`GetProcAddress` returns null -> EP no-ops).
- Kept in its own translation unit (`lib/Runtime/output_allocator.cpp`) so a **GPU-free unit test** can compile it natively against the mock runtime types. Wires the bitcode into both the real and mock runtime aggregates, and exports the setter via `CompilerDriver`'s symbol list so it survives LLVM optimization.
- Docs: the design doc's Phase 3 section is design-level (struct + the two entry points + the ABI decision); the build mechanics (bitcode `DEPENDS` list, the dllexport decl/def C2375 gotcha, the GPU-free unit-test convention) live in `CLAUDE.md`. Comment cleanup: the dllexport rationale is now stated once at the `HIPDNN_EP_RT_EXPORT` macro rather than triplicated.

## Test plan
- [x] Full ctest suite (LIT + e2e compile/execute + unit): **116/116 passing, 0 failed** on the latest `main`.
- [x] `OutputAllocatorUnitTest` (new `test/runtime` CTest target) — 5 cases: arg forwarding, uninstalled null-guard, clear-via-null-allocator, null-`state` setter, null-`state` alloc. **PASS**.
- [x] LIT (`hip-alloc-output.mlir`, `hip-alloc-output-no-dce.mlir`, `hip-use-output-allocator.mlir`) — op contract, Write-effect no-DCE/CSE, pass rewrite. **PASS**.
- [x] **Real** + **mock** (`BUILD_MOCK_RUNTIME=ON`) builds — `output_allocator.cpp` links into both runtime bitcode aggregates; EP + tools link.
- [x] `pre-commit` on changed files — all hooks pass.

Rebased onto `main` (Phase 1 #296 and Phase 2 #297 are merged), so the diff is just the Phase 3 runtime contract. Third PR in the output-allocator stack (phase 1 = op + pass, phase 2 = lowering, phase 3 = runtime contract).


wcy123:
The `struct_size` versioning handles forward/backward compatibility well, but adds complexity (~30 LOC + 7 tests). 

**Question:** Is this complexity justified? EP and model.dll are from the same codebase and typically deployed together. How often do you expect this struct to evolve across cached DLL versions?

If the struct rarely changes, simpler versioning might be sufficient. If plugin architecture or frequent evolution is planned, current design makes sense.

Implementation is correct either way—just questioning necessity.

lib/Runtime/hipdnn_ep_runtime.h
```diff
@@ -165,6 +165,65 @@ typedef struct RuntimeState RuntimeState;
...
+// shape is known (lowered from hip.alloc_output -> hipdnn_ep_alloc_output).
+//
+// ABI: this struct crosses the model.dll <-> EP boundary, so its layout is a
+// contract -- treated exactly like tensor_t below (the other cross-component
+// wire struct): a fixed layout locked by static_asserts, with the EP-side copy
+// declared identically. There is intentionally NO self-describing
+// size/version field: a layout change is an ABI break that requires rebuilding
+// the model.dll, and the model cache is invalidated the same way any other
+// runtime change is -- by deleting stale model.dlls (the cache key is the ONNX
+// hash, not the runtime version; see CLAUDE.md "Stale compiled-model DLLs").
```
fhanuman: We probably don't want to have these kind of references in the code that we are going to make it open source:

```
(the cache key is the ONNX
// hash, not the runtime version; see CLAUDE.md "Stale compiled-model DLLs")
```

lib/Runtime/hipdnn_ep_runtime.h
```diff
@@ -165,6 +165,65 @@ typedef struct RuntimeState RuntimeState;
...
+#else
+#define HIPDNN_EP_RT_EXPORT
+#endif
+
+// EP -> model.dll (exported), installs the allocator before inference_compute.
+// `state` is RuntimeState* to match every other state entry point; the EP side
+// resolves this by name and treats state as an opaque void* (pointer-
+// compatible), exactly like hipdnn_ep_runtime_begin_compute.
+HIPDNN_EP_RT_EXPORT void
+hipdnn_ep_set_output_allocator(RuntimeState *state,
```
fhanuman, line 213: Could you please help add a call-tree somewhere in the code comment or design document? It took me a while to trace to code to make sense of it. Maybe a call-tree similar to below will help readability?

```
MlirCustomOp::Compute(context)                            [MlirCustomOp.cpp:843]
├─ inference_state_->begin_compute()                      [:847]  per-Compute cache reset (GQA seqlens_k)
├─ if (use_output_allocator_)  ← cached at CTOR from metadata_.use_output_allocator() [:738,:849]
│  └─ compute_with_output_allocator(context)             [:771]
│     │
│     ├─ marshal_input_tensors(context, ...)             [:775]  EP-side: OrtValue → inputs.span
│     ├─ OutputAllocatorCtx octx{ctx,outputs,maps,host scratch,allocated[]}     [:778]
│     ├─ output_allocator_t alloc{ self=&octx, allocate=&output_allocate_cb }   [:788]
│     ├─ inference_state_->set_output_allocator(&alloc)  [:792]
│     │     └─ hipdnn_ep_set_output_allocator(state,&alloc)   [InferenceState.cpp:206; FATAL if DLL lacks symbol]
│     │
│     ├─ inference_state_->compute_allocator(&inputs.span)    [:794 → InferenceState.cpp:193]
│     │  └─ ★ inference_compute(state, inputs)   2-arg, NO outputs span   [IR 1814]
│     │     ├─ hipdnn_ep_tensor_prepare_input(state, inputs, 0, 2, &Xbuf)   [IR 1823]  marshal X
│     │     ├─ hipdnn_ep_tensor_prepare_input(state, inputs, 1, 2, &Wbuf)   [IR 1835]  marshal W
│     │     │     (get_gpu_ptr + get_shape_ptr → memref descriptor; IR 1845-1908) → inputs_array %18
│     │     ├─ hipdnn_ep_state_reset_error_flag(state)        [IR 1911]
│     │     ├─ main_graph(state, %18)                         [IR 1919]  (2-arg wrapper, unpacks descriptors)
│     │     │  └─ main_graph_internal(state, <X fields>, <W fields>)        [IR 1665]
│     │     │     ├─ hipdnn_ep_get_pool_base(state, 0, pool_bytes)          [IR 1705]  one grow-on-demand pool
│     │     │     ├─ wrap_miopenOpTensor(...)        Add(X,X)  → pool slot   [IR 1746]
│     │     │     ├─ wrap_hipblasLtMatmul(...)       MatMul    → pool slot   [IR 1769]
│     │     │     ├─ hipdnn_ep_alloc_output(state, out_idx=0, shape=[M,N], rank=2, elem=2)   ★★ OUTPUT ALLOC
│     │     │     │  │                                         [IR 1786: returns device void*]
│     │     │     │  └─ output_allocator.cpp:59  (runtime forwarder)
│     │     │     │     └─ allocator->allocate(self=&octx, 0, [M,N], 2, 2)
│     │     │     │        └─ output_allocate_cb(...)         [MlirCustomOp.cpp:653, noexcept]
│     │     │     │           ├─ present.*→past share-buffer shape override   [:666]
│     │     │     │           ├─ ctx.GetOutput(ort_idx, {M,N})  ← ORT allocates the OrtValue HERE  [:683]
│     │     │     │           ├─ octx.allocated[out_idx] = true               [:688]
│     │     │     │           └─ GPU → return ORT ptr (zero-copy)
│     │     │     │              host → GPU scratch + queue pending_d2h        [:690]
│     │     │     └─ wrap_miopenActivationForward(...) Sigmoid → writes device_ptr   [IR 1805]
│     │     │        return {device_ptr, sizes=[M,N], strides=[N,1]}
│     │     ├─ hipdnn_ep_stream_sync(state)                   [IR 1927]  all GPU writes complete on return
│     │     ├─ hipdnn_ep_state_read_and_clear_error_flag      [IR 1937]
│     │     └─ hipdnn_ep_tensor_free_input(state, X/W) ×2     [IR 1945-1946]
│     │
│     ├─ inference_state_->set_output_allocator(nullptr)      [:798]  clear before octx leaves scope
│     ├─ output-completeness guard: ∀ outputs allocated[i] else FATAL          [:809]
│     └─ #ifndef BUILD_MOCK_RUNTIME: pending_d2h → blocking hipMemcpy(host←GPU) [:823]
└─ return
```

fhanuman (dismissed):

lib/Runtime/hipdnn_ep_runtime.h
zpye: ﻿Good catch — removed the `CLAUDE.md` reference. The comment is now self-contained: it states the substantive point (the compiled-model cache key is the ONNX graph hash, not the runtime version, so a runtime ABI change does not by itself invalidate cached DLLs) without pointing at an internal file.

lib/Runtime/hipdnn_ep_runtime.h
zpye: ﻿Thanks — agreed this needs a call-tree, and your draft is a great template. I''d like to land it in **Phase 5** rather than here, because the tree you sketched roots at `MlirCustomOp::Compute` and walks `compute_with_output_allocator -> output_allocate_cb -> ctx.GetOutput -> pending D2H` — all of which is **EP-side code that doesn''t exist in this PR**. Phase 3 is only the runtime contract (`hipdnn_ep_runtime.h` struct + `output_allocator.cpp` forwarder/setter + wiring); the bridge the tree describes arrives in Phase 5.

Adding it here would (a) forward-reference code not in this diff, and (b) go stale immediately — the draft still shows the `present.*->past` share-buffer override and the old `compute_allocator` name, both of which were dropped/renamed in the Phase 5 work. Putting the tree next to the bridge in Phase 5 lets it be accurate and verifiable against present code.

So for Phase 5 I''ll add the full end-to-end call-tree (in the design doc''s Phase 5 section and/or a header comment), refreshed to match the final code. Flagging this thread to carry over to the Phase 5 PR.

wcy123 (approved):
well done.

PR #259, opened 2026-05-28
[WIP] MS Build Release

akholodnamdcom:

PR #284, opened 2026-06-05
feat(runtime): runtime + custom-kernel slice of PR #259 (mm preserved)

AMDmoore:
## Summary

This PR carves the **runtime + custom-kernel** portion out of the large `release/msbuild` branch (PR #259) so it can be reviewed on its own, rebased on top of current `main`.

Scope is two directories: `lib/Runtime/**` and `3rd-party/custom_kernels/**`. No compiler frontend (`lib/Conversion/OnnxToHip`), backend (`backend-mlir-compiler`), or build-system changes are included.

## MemoryManager (mm) is preserved — zero net change to mm/pool infra

PR #259 branched **before** the MemoryManager subsystem was merged to `main` (commit `6dc93b45`), so a naive `lib/Runtime` copy from `release/msbuild` would silently revert mm. This PR keeps `main`'s versions of the mm/pool infrastructure, introducing **zero net diff** to them:

- `hipdnn_ep_runtime_state.cpp`, `runtime_state_internal.h`, `hipdnn_ep_runtime_loop.cpp`, `lib/Runtime/CMakeLists.txt` — identical to `main` (mm `alloc` for constants/workspace/qmoe stays).
- `hipdnn_ep_runtime.h` — infra declarations reverted to `main` (`get_pool_base` 1-arg, no `push/pop_pool_depth`, no `get_host_scratch_base`, `begin_compute` without `dllexport`). Only per-op wrapper signatures changed.

The dynamic grow-on-demand pool, nested-pool, and host-scalar scratch infra are intentionally **out of scope** — they are compiler-coupled and will land with their MLIR passes.

## ⚠️ KNOWN ISSUE — 8 op-wrapper ABI changes (compiler-coupled)

8 runtime `wrap_*` functions changed signature. These are an **ABI contract with the HipToLLVM lowerings**, and `main` already ships those lowerings (emitting the *old* arg list). Until the matching lowering changes land on `main`, any model that reaches one of these ops will pass the **old** arg list into the **new** wrapper definition → reads extra/garbage args → wrong result (links fine; silently bypasses CPU-fallback accuracy tests). The matching `*Lowering.cpp` + `*Conversion.cpp` changes are deliberately **deferred to per-op feature PRs**.

| Op | `wrap_*` change | main impact |
|----|-----------------|-------------|
| **Gather** | `+axis_size, +inner_size` | **All LLMs** (embedding lookup) — always hit |
| **Div** | `num_elements` → 4D broadcast shape | Models using Div broadcast |
[...]

PR #299, opened 2026-06-08
feat(conversion): NumPy-broadcast init + 4D-shape ABI for Div/Sub/Equal (#259 split 1/N)

AMDmoore:
## Summary

First compiler-side slice of the #259 split, **stacked on PR284** (`feat/runtime-custom-kernels-split`).

- `OnnxToHipUtils`: add `createBroadcastEmptyTensor` -- per-axis selection of the non-broadcasting operand for a dynamic DPS init (e.g. `[?x1] + [1x?] -> [?x?]`, where dim 0 comes from lhs and dim 1 from rhs). The old `createEmptyTensor(resultType, single-source)` picks one operand for all axes and is wrong when the two sides disagree on which supplies a dynamic extent.
- Broadcast elementwise/util conversions use it: Add/Mul/Sub, Div, And, Min, Mod, Equal, Less. main's inferred-type `Op::create` is kept (no explicit result type).
- Lowerings `DivLowering` / `ElementwiseLowering` (Sub) / `EqualLowering` emit the new ABI that matches the runtime functions PR284 introduced: `wrap_div` / `wrap_elementwise_sub` (full 4D operand shapes for `hip_expand` broadcast materialisation) and `wrap_equal` (per-operand element counts). Uses the already-on-main `extractShape4D` / `computeNumElements` helpers.

## Background -- the defect `createBroadcastEmptyTensor` closes (from #259)

The original problem + solution is recorded in the #259 branch (`release/msbuild`), commit `1fca3099` -- *"Land Kokoro op coverage + dynamic-shape robustness"*:

> `BinaryElementwise`, `Elementwise`, `Tier3Compare` (`Where` / `Equal` / `Greater` / ...): replace the lhs-or-rhs source picker with `createBroadcastEmptyTensor`, which walks every candidate input per output dim and falls back for pure scalar broadcasts. **Avoids `tensor.dim` on rank-0 operands.**

Two distinct defects in the old single-source `createEmptyTensor` that this helper closes:

1. **Mixed-dynamic mis-sizing.** When different result dims take their dynamic extent from different operands -- `[?x1] + [1x?] -> [?x?]` (dim 0 from lhs, dim 1 from rhs) -- keying every `tensor.dim` on one operand sizes the wrong axis from a static-`1` extent, so the DPS init buffer is mis-sized. Our elementwise kernels do not broadcast internally (the runtime materialises broadcasts via `hip_expand` into a same-shape buffer before the kernel runs), so a mis-sized init buffer is a real correctness bug, not just slack.
2. **`tensor.dim` on rank-0 operands.** Pure scalar broadcasts (e.g. Kokoro's `scalar op tensor`) made the old picker call `tensor.dim` on a rank-0 value, which is invalid. The helper right-aligns each operand to the result and only considers operands that actually span the axis, so rank-0 / shorter-rank operands are skipped instead of being indexed.

The helper's selection rule per dynamic result axis: pick the first operand that genuinely spans the axis and is not statically `1` there; otherwise fall back to the first spanning operand; fail cleanly (no silent wrong size) if none spans it.

Broadcast-init LIT coverage for this helper landed in #259 (`4c4c33cd`, *"broadcast div/sub"*).

## Merge ordering (important)

PR284 changed the ABI of the **pre-existing** `wrap_div` / `wrap_elementwise_sub` / `wrap_equal`; these lowerings are their callers. **PR284 + this PR form an atomic merge train** -- this PR must merge together with / immediately after PR284, before model-compile CI gates main, otherwise Div/Sub/Equal model compilation breaks on a 6-arg-call vs new-ABI-runtime mismatch.

## Test plan

- [x] `hip-compiler.dll` + `hip-mlir-opt` rebuild clean (Release)
- [x] lit: `test_{div,equal,sub}` (hip-to-llvm) and `test_{div,sub}` (onnx-to-hip) pass
- [x] pre-commit (lintrunner + clang-format + license) clean
- [ ] CI on the stacked train

### DIFF
--- a/lib/Conversion/OnnxToHip/AndConversion.cpp
+++ b/lib/Conversion/OnnxToHip/AndConversion.cpp
@@ -12,10 +12,9 @@ namespace {
 /// onnx.And -> hip.and
 ///
 /// Element-wise logical AND of two boolean tensors with NumPy-style
-/// broadcasting. Mirrors the EqualConversion pattern: pick whichever input
-/// already matches the broadcast output rank as the shape source for
-/// `createEmptyTensor`, so dynamic-shape models still get correct
-/// `tensor.dim` extractions for the DPS init.
+/// broadcasting. Dynamic output dims are resolved via
+/// `createBroadcastEmptyTensor` so each axis picks the non-broadcasting operand
+/// (e.g. `[?x1] & [1x?] -> [?x?]`).
 struct AndToHip : public mlir::RewritePattern {
   AndToHip(mlir::MLIRContext *ctx)
       : RewritePattern("onnx.And", /*benefit=*/1, ctx) {}
@@ -44,15 +43,14 @@ struct AndToHip : public mlir::RewritePattern {
     if (!aType || !bType)
       return rewriter.notifyMatchFailure(op, "expected ranked tensor inputs");
 
-    // Pick the side whose rank already matches the broadcast output so
-    // dynamic dims line up dim-by-dim for createEmptyTensor. When both
-    // operands have the same rank as the result this is just `a`; when
-    // one operand has a smaller rank (e.g. broadcasting a [N] mask against
-    // a [B, N] tensor) we prefer the larger-rank operand.
-    mlir::Value source = (aType.getRank() == resultType.getRank()) ? a : b;
-    mlir::Value init = createEmptyTensor(rewriter, loc, resultType, source);
+    mlir::FailureOr<mlir::Value> initOrFailure =
+        createBroadcastEmptyTensor(rewriter, loc, resultType, {a, b});
+    if (mlir::failed(initOrFailure))
+      return rewriter.notifyMatchFailure(
+          op, "And: no ranked operand spans dynamic result dim");
 
-    auto hipOp = mlir::hip::AndOp::create(rewriter, loc, context, a, b, init);
+    auto hipOp =
+        mlir::hip::AndOp::create(rewriter, loc, context, a, b, *initOrFailure);
     rewriter.replaceOp(op, hipOp->getResult(0));
     return mlir::success();
   }
--- a/lib/Conversion/OnnxToHip/OnnxToHipUtils.h
+++ b/lib/Conversion/OnnxToHip/OnnxToHipUtils.h
@@ -94,6 +94,64 @@ inline mlir::Value createEmptyTensor(mlir::OpBuilder &builder,
                                        resultType.getElementType(), dynSizes);
 }
 
+/// Create a tensor.empty for a DPS init whose shape is the NumPy-style
+/// broadcast of \p operands. Operand shapes are right-aligned with the
+/// result. For each dynamic dimension of \p resultType, the size is taken
+/// from the first operand that truly contributes at that axis -- i.e. whose
+/// corresponding dim is not statically 1. Shorter-rank operands (left-padded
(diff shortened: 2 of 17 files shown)


lib/Conversion/OnnxToHip/AndConversion.cpp
```diff
@@ -44,15 +43,14 @@ struct AndToHip : public mlir::RewritePattern {
...
       return rewriter.notifyMatchFailure(op, "expected ranked tensor inputs");
 
-    // Pick the side whose rank already matches the broadcast output so
-    // dynamic dims line up dim-by-dim for createEmptyTensor. When both
-    // operands have the same rank as the result this is just `a`; when
-    // one operand has a smaller rank (e.g. broadcasting a [N] mask against
-    // a [B, N] tensor) we prefer the larger-rank operand.
-    mlir::Value source = (aType.getRank() == resultType.getRank()) ? a : b;
-    mlir::Value init = createEmptyTensor(rewriter, loc, resultType, source);
+    mlir::FailureOr<mlir::Value> initOrFailure =
```
fhanuman, line 46: Won't this already get addressed as part of the `shape inference+ shape propagation` that we added? Can you quickly try few of these operators in the main branch to see if the issue is still there?

lib/Conversion/OnnxToHip/OnnxToHipUtils.h
```diff
@@ -94,6 +94,64 @@ inline mlir::Value createEmptyTensor(mlir::OpBuilder &builder,
...
+/// from the first operand that truly contributes at that axis -- i.e. whose
+/// corresponding dim is not statically 1. Shorter-rank operands (left-padded
+/// with 1) and statically-1 dims are skipped. If every spanning operand is
+/// statically 1 at the axis, fall back to the first operand that spans it.
+///
+/// Use this for binary/multinary broadcast elementwise ops (Add, Mul, Where,
+/// ...). Do NOT use `createEmptyTensor(resultType, source)` when operands can
+/// disagree on which side supplies a dynamic extent (e.g. `[?x1] + [1x?] ->
+/// [?x?]` -- dim 0 from lhs, dim 1 from rhs).
+inline mlir::FailureOr<mlir::Value>
```
fhanuman, line 109: Same comment as earlier. We should check if we need this -- I think `shape inference + shape propagation` should take care of this.

fhanuman (changes requested):

lib/Conversion/OnnxToHip/AndConversion.cpp
AMDmoore: @fhanuman thanks for your comment.
Do you mean shape inference will change the two inputs of Add/Div/Sub/... to the same shape? if not, I think this is needed, because this is a running issue fix in pr259.

fhanuman: Hi @AMDmoore, Yes, I mean by `shape inference`, could you please confirm with basic tests?

fhanuman: I checked we need this logic, so dropping this comment.

fhanuman (approved):

AMDmoore:
for runtime part, refer to #284

AMDmoore:
all comments resolved, will merge this pr.

PR #301, opened 2026-06-08
feat(hip): allocator-aware codegen + pipeline branch (phase 4)

zpye:
## Summary

Phase 4 of the output-allocator feature: wires the `use-output-allocator` flag end-to-end so the compiler can emit the **allocator pipeline** (graph outputs allocated in-graph via `hip.alloc_output`) alongside the existing **classic** out-param pipeline. The flag defaults **OFF**, and the EP bridge (Phase 5) is not yet wired, so this is inert in production.

The mode is carried by a single `hipdnn.use_output_allocator` **module attribute** (a typed `BoolAttr`, read by value -- absence or `= false` both mean classic), not by a flag threaded through every pass or guessed from `main_graph`'s arity.

- **Pipeline branch** (`lib/Dialect/Transforms/Pipelines.cpp`): `buildOnnxToHipPipelineTail` branches on the flag -- classic runs `buffer-results-to-out-params` (slot 3); allocator runs `hip-use-output-allocator` at **slot 4.5, AFTER `buffer-deallocation`**. The ordering is load-bearing: `hip.alloc_output` has a Write but no Allocate effect, so running it *before* deallocation makes the ownership pass clone the output at the `return` (a per-inference alloc + full-output copy that defeats zero-copy). `buildHipToLLVMPipeline` is mode-agnostic -- it always runs `convert-hip-to-llvm` + `generate-interface`, which read the attribute.
- **Attribute stamp folded into `hip-use-output-allocator`** (`lib/Dialect/Transforms/UseOutputAllocator.cpp`): the pass that rewrites returned `memref.alloc` -> `hip.alloc_output` also stamps the `hipdnn.use_output_allocator` module attribute, unconditionally (a zero-output graph must still be marked). No separate marker pass.
- **Mode + arity check** (`lib/Conversion/HipToLLVM/HipToLLVM.cpp`): `transformMainFunction` reads the attribute to pick classic (3-arg) vs allocator (2-arg) `main_graph`, then uses the unpacked-memref param count only to *verify* the wrapper matches that mode (mode-specific `emitError` on mismatch). A single `expectedParams` with the output scope gated by `if (!allocatorMode)` (no dual count). `allocatorMode` is also required for return-type routing: in allocator mode `main_graph_internal` returns a memref-descriptor struct that is discarded (return 0), else it would be returned from the `i32` wrapper (type mismatch).
- **Interface generator** (`lib/Dialect/Transforms/GenerateInterface.cpp`): a single `GenerateInterfacePass` emits both ABIs, reading the attribute -- allocator mode emits a 2-arg `inference_compute` (no `prepare_output`/`finalize_output`, no `outputs` span). The classic-only output staging is gated on the attribute (`if (!allocatorMode)`) and tagged "delete when classic is removed" so it is mechanically removable once the allocator path is the only one.
- **Options** (`include/hip/Dialect/Transforms/Pipelines.h`): `useOutputAllocator` lives on the ONNX-to-HIP-half structs only -- **not** on `HipToLLVMPipelineOptions`, since the HIP-to-LLVM half reads the module attribute instead of a second flag.

## Pipeline ordering (empirical, gate test)

`hip-use-output-allocator` MUST run *after* `buffer-deallocation`. `test/lit/Pipeline/output-allocator-dealloc.mlir` pins both orderings -- the rejected (before) order produces a `bufferization.clone` at the return; the chosen (after) order returns the `hip.alloc_output` buffer directly with no clone. See `docs/design/output-allocator-design.md` section 2 and the CLAUDE.md gotcha.

## Test plan

- [x] LIT: `test/lit/e2e/test_output_allocator_model.mlir` -- allocator (2-arg `inference_compute` + `hipdnn_ep_alloc_output`, attribute present) vs classic (3-arg + `prepare_output`/`finalize_output`).
- [x] LIT: `test/lit/Dialect/hip-use-output-allocator.mlir` -- rewrite + module attribute stamped.
- [x] LIT: `test/lit/Pipeline/output-allocator-dealloc.mlir` -- both pass orderings (clone present in before-order, absent in after-order).
- [x] LIT: `test/lit/Conversion/hip-to-llvm/test_main_graph_param_mismatch.mlir` -- mode-specific param-count mismatch diagnostic.
- [x] Full `MorphizenMLIRLitTests` LIT suite: **218/218 pass**.
- [x] `OutputAllocatorUnitTest` (GPU-free): pass.
- [x] Full CTest suite: **118/118 pass** on the latest `main` (REAL runtime build).
- [x] `python build.py`: clean build.
- [x] `pre-commit run` on all PR files: all hooks pass.

## Review fixes applied
- Folded the separate `hip-set-output-allocator-attr` marker pass into `hip-use-output-allocator` (per @wcy123 / @fhanuman).
- Removed the `expectedClassic`/`expectedAllocator` dual count in `HipToLLVM.cpp` -> single `expectedParams` gated by `allocatorMode`.
- `llvm::seq` for the input/output param loops.
- Attribute is a typed `BoolAttr` read by value (`= false` and absence both classic), renamed `hipdnn.use_output_allocator`.
- Simplified the phase-4 comments/docs: dropped prose that duplicated the Before/After IR snippet in `transformMainFunction`, removed the slot-4.5 clone-mechanics duplication from the `Pipelines.cpp` header (kept inline at the slot), and condensed the `UseOutputAllocator`/`GenerateInterface` comments -- core ideas retained.

Rebased onto `main` (Phase 1 #296, Phase 2 #297, Phase 3 #298 are merged), so the diff is just the Phase 4 codegen + pipeline branch. Fourth PR in the output-allocator stack (phase 1 = op + pass, phase 2 = lowering, phase 3 = runtime contract, phase 4 = codegen + pipeline).


wcy123:
Why modify `ConvertHipToLLVMPass` at all? The existing implementation already handles both pipelines:

**Current logic:**
- Reads `outputCount` from metadata
- Unpacks `outputCount` outputs from the outputs array
- In allocator pipeline: `outputCount = 0` → loop runs 0 times → no output unpacking

**No changes needed.** The metadata-driven unpacking naturally adapts to both cases.

Adding arity auto-detection and `allocatorMode` branching contradicts the design principle:
> "Each pipeline has single responsibility, no conditional logic, clear semantics. Changes to allocator model don't affect classic path."
> — docs/design/output-allocator-design.md

**Recommendation:** Remove all changes to `ConvertHipToLLVMPass`. It already works for both pipelines.

wcy123:
The shared `GenerateInterfacePassBase` with `allocatorMode` flag mixes both code paths, contradicting the design principle:

> "Each pipeline has single responsibility, no conditional logic, clear semantics. Changes to allocator model don't affect classic path."
> — docs/design/output-allocator-design.md

**Current approach:**
- Shared base class
- `runImpl(module, allocatorMode)` with conditionals throughout
- Both passes coupled through shared implementation

**Recommendation:** TWO completely separate implementations:

- `GenerateInterfacePass` — **unchanged** from main branch (classic 3-arg)
- `GenerateAllocatorInterfacePass` — **new** separate implementation (allocator 2-arg)

Yes, this means code duplication. But that's the trade-off for true separation:
- Classic code can't be broken by allocator changes
- Each pass has single, clear responsibility
- No conditional logic mixing both modes
- Maintainable - change one without affecting the other

Same principle as the dual pipeline architecture: keep them separate, not branched.

wcy123 (changes requested):
Two major design issues that contradict the stated principle of keeping classic and allocator pipelines separate:

## 1. ConvertHipToLLVMPass modifications are unnecessary

The existing implementation already handles both pipelines through metadata-driven logic (`outputCount`). Adding arity auto-detection and `allocatorMode` branching adds complexity without benefit and violates the separation principle.

**Recommendation:** Remove all changes to `ConvertHipToLLVMPass`.

## 2. GenerateInterface shared base class mixes both paths

The `GenerateInterfacePassBase` with `allocatorMode` flag couples both implementations, making them fragile and hard to maintain.

**Recommendation:** Create two completely separate implementations:
- `GenerateInterfacePass` - unchanged from main branch
- `GenerateAllocatorInterfacePass` - new, standalone implementation

---

Both issues violate the design principle:
> "Each pipeline has single responsibility, no conditional logic, clear semantics. Changes to allocator model don't affect classic path."

The pipeline branching architecture is excellent, but the pass implementations should follow the same separation principle.

zpye:
> Two major design issues that contradict the stated principle of keeping classic and allocator pipelines separate:
> 
> ## 1. ConvertHipToLLVMPass modifications are unnecessary
> The existing implementation already handles both pipelines through metadata-driven logic (`outputCount`). Adding arity auto-detection and `allocatorMode` branching adds complexity without benefit and violates the separation principle.
> 
> **Recommendation:** Remove all changes to `ConvertHipToLLVMPass`.
> 
> ## 2. GenerateInterface shared base class mixes both paths
> The `GenerateInterfacePassBase` with `allocatorMode` flag couples both implementations, making them fragile and hard to maintain.
> 
> **Recommendation:** Create two completely separate implementations:
> 
> * `GenerateInterfacePass` - unchanged from main branch
> * `GenerateAllocatorInterfacePass` - new, standalone implementation
> 
> Both issues violate the design principle:
> 
> > "Each pipeline has single responsibility, no conditional logic, clear semantics. Changes to allocator model don't affect classic path."
> 
> The pipeline branching architecture is excellent, but the pass implementations should follow the same separation principle.

reconstruct and update

include/hip/Dialect/Transforms/Passes.h
```diff
@@ -23,7 +23,12 @@ struct CompilationOptionsT;
 /// Creates a pass that generates the C interface for the compiled module.
 /// Transforms @main_graph to produce four C-ABI wrapper functions:
 /// inference_init, inference_compute, inference_cleanup,
-/// inference_get_metadata_json.
+/// inference_get_metadata_json. The classic 3-arg (state, inputs, outputs) vs
+/// 2-arg allocator (state, inputs) ABI is selected from the module's
+/// `hipdnn.output_allocator` attribute (set by hip-set-output-allocator-attr);
```
wcy123: `use_output_allocator`?

include/hip/Dialect/Transforms/Passes.td
```diff
@@ -292,6 +292,22 @@ def UseOutputAllocatorPass
   ];
 }
 
+def SetOutputAllocatorAttrPass
```
wcy123: why we need a new pass

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -93,41 +162,61 @@ struct ConvertHipToLLVMPass
...
       int64_t rank = cast<DenseI64ArrayAttr>(shapeAttr).size();
-      expectedParams += kMemRefPtrs + kMemRefOffset + rank + rank;
+      expectedAllocator += kMemRefPtrs + kMemRefOffset + rank + rank;
     }
+    unsigned expectedClassic = expectedAllocator;
     for (auto shapeAttr : outputShapesAttr) {
       int64_t rank = cast<DenseI64ArrayAttr>(shapeAttr).size();
-      expectedParams += kMemRefPtrs + kMemRefOffset + rank + rank;
+      expectedClassic += kMemRefPtrs + kMemRefOffset + rank + rank;
     }
```
wcy123, line 167: we should avoid these details.

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -93,41 +162,61 @@ struct ConvertHipToLLVMPass
...
 
     unsigned actualParams = mainFunc.getFunctionType().getNumParams();
-    if (actualParams != expectedParams) {
+    // Mode is decided by the `hipdnn.output_allocator` module attribute (set by
+    // hip-set-output-allocator-attr), NOT by param count -- this disambiguates
+    // a zero-output graph, where expectedClassic == expectedAllocator. The
+    // count is then only used to verify @main_graph matches the chosen mode.
+    bool allocatorMode = module->hasAttr("hipdnn.output_allocator");
+    unsigned expected = allocatorMode ? expectedAllocator : expectedClassic;
```
wcy123: we don't need `expectedClassic`, skip iteration on output if `allocatorMode=true`

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -162,23 +250,37 @@ struct ConvertHipToLLVMPass
...
+    // Forward to the internal main_graph and return the i32 status:
+    //   - allocator mode: the internal returns a memref descriptor (by-value
+    //     struct) for the in-graph-allocated output. Call it, DISCARD the
+    //     struct result (the output reaches the EP through the
+    //     hipdnn_ep_alloc_output callback, not the return value), and return 0.
+    //   - classic void internal (real lowering): call, return 0.
+    //   - classic i32 internal (hand-written LIT shape): call, forward result.
     auto internalRetTy = mainFunc.getFunctionType().getReturnType();
-    if (isa<LLVM::LLVMVoidType>(internalRetTy)) {
+    if (allocatorMode || isa<LLVM::LLVMVoidType>(internalRetTy)) {
```
wcy123, line 268: why we need `allocatorMode`

lib/Dialect/Transforms/GenerateInterface.cpp
```diff
@@ -467,6 +478,7 @@ class GenerateInterfacePass
   }
 
   LogicalResult verifyPrerequisites(ModuleOp module) {
+    bool allocatorMode = module->hasAttr("hipdnn.output_allocator");
```
wcy123: check `output_allocator  =true`, minor comment
what if `output_allocator  = false`?

lib/Dialect/Transforms/GenerateInterface.cpp
```diff
@@ -673,16 +695,28 @@ class GenerateInterfacePass
...
+  /// Allocator mode (module carries the `hipdnn.output_allocator` attr): the
+  /// wrapper takes (state, inputs) only; output-tensor handling
+  /// (prepare_output, the output-memref array, finalize_output) and the outputs
+  /// span are all skipped -- graph outputs are allocated in-graph by
+  /// hip.alloc_output (lowered to the hipdnn_ep_alloc_output callback) and
+  /// written directly, so
+  /// @main_graph is called as (state, input_memrefs).
   void generateInferenceCompute(ModuleOp module, ArrayAttr inputShapes,
                                 ArrayAttr outputShapes) {
+    bool allocatorMode = module->hasAttr("hipdnn.output_allocator");
```
wcy123: same

lib/Dialect/Transforms/Pipelines.cpp
```diff
@@ -22,7 +22,30 @@
...
+/// `hip.alloc_output` (EP-owned, allocated in-graph via the output-allocator
+/// callback), then `hip-set-output-allocator-attr` stamps the
+/// `hipdnn.output_allocator` module attr that the HIP-to-LLVM half reads to
+/// select the allocator ABI.
+///
+/// The slot-4.5 placement is load-bearing and was chosen empirically (see the
+/// slot 4.5 comment below and test/lit/Pipeline/output-allocator-dealloc.mlir):
+/// `hip.alloc_output` carries a Write effect but NO Allocate effect, so if it
+/// were created BEFORE buffer-deallocation, the deallocation pass would treat
+/// the returned value as unowned and insert a `bufferization.clone` at the
```
wcy123: ` buffer-deallocation` is not actually needed at the first place.

wcy123 (dismissed):
after resolve the minor issues, we can merge the pull request.

include/hip/Dialect/Transforms/Passes.td
fhanuman: +1 We probably don't need a new pass just to set the attribute, can we do this as part of `hip-use-output-allocator`?

lib/Conversion/HipToLLVM/HipToLLVM.cpp
```diff
@@ -162,23 +250,37 @@ struct ConvertHipToLLVMPass
...
-          builder.create<LLVM::LoadOp>(loc, memrefStructType, outputStructPtr);
-      unpackMemRefStructWithAddrCast(builder, loc, outputMemref, rank,
-                                     mainInternalArgs);
+    // Classic mode forwards each output out-param (unpacked memref struct) to
+    // the internal call. Allocator mode has no output args -- the internal
+    // main_graph allocates and returns the output via hip.alloc_output, so the
+    // outputs pointer is absent from the wrapper signature.
+    if (!allocatorMode) {
+      Value outputsArg = entryBlock->getArgument(2);
+      for (int64_t i = 0; i < outputCount; i++) {
```
fhanuman: Nit: This can be `llvm::seq`

lib/Dialect/Transforms/GenerateInterface.cpp
```diff
@@ -747,11 +792,16 @@ class GenerateInterfacePass
...
     }
-    for (auto i : llvm::seq<size_t>(0, numOutputs)) {
-      (void)i;
-      Value bufferPtr = LLVM::AllocaOp::create(builder, loc, ptrType, i8Type,
-                                               tensorBufferSize, 0);
-      outputBuffers.push_back(bufferPtr);
+    // classic out-param staging - delete when classic is removed. Allocator
+    // mode allocates outputs in-graph via hip.alloc_output, so no per-output
+    // TensorBuffer is staged here.
+    if (!allocatorMode) {
```
fhanuman, line 802: As I mentioned in the teams chat we don't need to maintain separate code paths for with and without `allocator` paths. We should use `allocator` path both for `static` and `dynamic` shapes. Let's do this in a follow-up task after ROCm EP 0.2 release.

fhanuman (dismissed):

lib/Conversion/HipToLLVM/HipToLLVM.cpp
zpye: fixed

lib/Dialect/Transforms/GenerateInterface.cpp
zpye: Okay.

lib/Dialect/Transforms/Pipelines.cpp
zpye: Will not change in this PR.

lib/Dialect/Transforms/GenerateInterface.cpp
zpye: fixed

lib/Dialect/Transforms/GenerateInterface.cpp
zpye: fixed

lib/Conversion/HipToLLVM/HipToLLVM.cpp
zpye: fixed

lib/Conversion/HipToLLVM/HipToLLVM.cpp
zpye: fixed

include/hip/Dialect/Transforms/Passes.td
zpye: fixed

include/hip/Dialect/Transforms/Passes.h
zpye: fixed

lib/Conversion/HipToLLVM/HipToLLVM.cpp
zpye: added in comment:

```
// The `allocatorMode` term is required: in allocator mode internalRetTy is
// the memref descriptor struct (neither void nor i32), so without it this
// would fall into the else branch and return that struct from the i32
// wrapper -- a type mismatch. The attr routes allocator mode to discard +
// return 0 instead.
```

wcy123 (approved):
**Approved** — Excellent refactor addressing all architectural concerns.

## Key improvements:

1. **Module attribute as single source of truth** (`hipdnn.use_output_allocator` BoolAttr) eliminates flag threading and makes mode explicit throughout compilation
2. **`transformMainFunction` simplified** — reads attribute for mode, uses param count only for verification (no inference)
3. **Unified `GenerateInterfacePass`** — cleaner than separate passes, branches on attribute with `// delete when classic is removed` markers
4. **Pass ordering critical fix** — `hip-use-output-allocator` AFTER `buffer-deallocation` prevents clone-at-return (no Allocate effect ⇒ ownership pass clones if run before)
5. **Gate test coverage** — `output-allocator-dealloc.mlir` pins both orderings (clone present/absent)

## Design quality:

The module attribute approach is superior to my original suggestion (separate passes):
- Single source of truth (no sync issues)
- Typed BoolAttr (absence = false, explicit false both work)
- Natural for mode verification in `transformMainFunction`
- Classic-removal markers make migration path clear

The pass ordering landmine (slot 4.5, after deallocation) is well-documented in CLAUDE.md and gated by LIT test.

Excellent work resolving the architectural issues while keeping the implementation clean.

wcy123 (approved):
**Approved** — Excellent refactor addressing all architectural concerns.

## Key improvements:

1. **Module attribute as single source of truth** (`hipdnn.use_output_allocator` BoolAttr) eliminates flag threading and makes mode explicit throughout compilation
2. **`transformMainFunction` simplified** — reads attribute for mode, uses param count only for verification (no inference)
3. **Unified `GenerateInterfacePass`** — cleaner than separate passes, branches on attribute with `// delete when classic is removed` markers
4. **Pass ordering critical fix** — `hip-use-output-allocator` AFTER `buffer-deallocation` prevents clone-at-return (no Allocate effect ⇒ ownership pass clones if run before)
5. **Gate test coverage** — `output-allocator-dealloc.mlir` pins both orderings (clone present/absent)

## Design quality:

The module attribute approach is superior to my original suggestion (separate passes):
- Single source of truth (no sync issues)
- Typed BoolAttr (absence = false, explicit false both work)
- Natural for mode verification in `transformMainFunction`
- Classic-removal markers make migration path clear

The pass ordering landmine (slot 4.5, after deallocation) is well-documented in CLAUDE.md and gated by LIT test.

Excellent work resolving the architectural issues while keeping the implementation clean.

wcy123 (approved):
**Approved** — Excellent refactor addressing all architectural concerns.

## Key improvements:

1. **Module attribute as single source of truth** (`hipdnn.use_output_allocator` BoolAttr) eliminates flag threading and makes mode explicit throughout compilation
2. **`transformMainFunction` simplified** — reads attribute for mode, uses param count only for verification (no inference)
3. **Pass ordering critical fix** — `hip-use-output-allocator` AFTER `buffer-deallocation` prevents clone-at-return (no Allocate effect ⇒ ownership pass clones if run before)
4. **Gate test coverage** — `output-allocator-dealloc.mlir` pins both orderings (clone present/absent)

## Design quality:

The module attribute approach is architecturally superior - single source of truth, no sync issues between passes.

## Minor suggestion for follow-up:

Consider refactoring large functions with repeated `if (allocatorMode)` branches into smaller helper functions:
- Extract classic-specific logic → `generateClassicOutputStaging(...)`
- Extract allocator-specific logic → `generateAllocatorFlow(...)`
- Main function orchestrates: `if (allocatorMode) generateAllocatorFlow() else generateClassicOutputStaging()`

This reduces cognitive load and makes the two paths more obviously separate. Not blocking - can be a follow-up refactor.

Excellent work resolving the architectural issues.

PR #303, opened 2026-06-08
feat(onnx-to-hip): dynamic Reshape conversion + Reshape(_,Shape(x)) fold

AMDmoore:
## Summary

Makes dynamic-shape ONNX `Reshape` patterns compile correctly. Two coordinated pieces (rebased onto current `main`):

1. **`ReshapeConversion` dynamic delta** -- rank-0 scalar wrap; rank-change expand/collapse now *falls through* instead of hard-failing when MLIR can't prove a clean reassociation; new `tensor.reshape` fallback that consumes the runtime shape operand verbatim and resolves a literal `-1` before lowering (`memref.reshape` does not interpret `-1`). main's same-rank decomposition is preserved verbatim (strict superset -- `main` is otherwise the OLD hard-fail version).
2. **`ReshapeShapeFold`** (new pre-lowering populate-pattern, sibling of `GatherShapeFold`) -- rewrites `Reshape(data, Shape(src))` so the shape operand becomes an explicit `tensor.from_elements(tensor.dim(src, *))` that the conversion fallback consumes. Fixes the `[N, N]` mis-sizing of 2-D dynamic reshapes (a `1x1` input passes by coincidence; every asymmetric shape silently corrupts). Canonical trigger: Qwen3.5 mrope `Range -> Reshape([bs, ss])` 2-D `position_ids`.

## Changed in response to review (this branch was force-rebased)

The original push of this PR carried two extra pieces that reviewer feedback flagged as superseded by `main`. Both are now **removed**, and the branch was rebased onto current `main` (which carries the merged pool engine #286 / #287 / #288):

- **`RelaxMultiDynExpandShapePass` -- REMOVED.** Per the `Passes.td` review comment ("not needed" under the new design). Deleted in full:
  - `lib/Dialect/Transforms/RelaxMultiDynExpandShape.cpp` (deleted)
  - `include/hip/Dialect/Transforms/Passes.td` -- pass def removed (back to `main` verbatim)
  - `lib/Dialect/Transforms/Pipelines.cpp` -- `buildHipToLLVMPipeline` wiring removed (back to `main` verbatim)
  - `lib/Dialect/Transforms/CMakeLists.txt` -- source entry removed (back to `main` verbatim)
  - `test/lit/Dialect/hip-relax-multi-dyn-expand-shape.mlir` (deleted)
- **`PoolAllocs.cpp` delta -- REMOVED.** Per the `PoolAllocs.cpp` review comment ("revert this file"). `main`'s modular `HoistAllocSizeArithPass` + `ResolveTensorDimsPass` (the merged pool engine) now own alloc-size / `tensor.dim` resolution, so the original #259 hacks are obsolete. File is back to `main` verbatim -- not re-ported.

Net diff vs `main` is now exactly **6 files** (the two pieces in Summary + their registration + one lit test).

## [WARNING] Test-coverage gaps (read before merging)

**1. The two headline conversion paths have no dedicated lit.** `test_reshape.mlir`'s new cases cover only the rank-0 scalar -> rank-N wrap. The dynamic `tensor.reshape` `-1`-resolving fallback and `ReshapeShapeFold`'s `tensor.from_elements` rewrite (the `[N, N]` fix) are **not** asserted by any FileCheck; they are exercised only by the pending Qwen3.5 e2e below.

**2. Open e2e dependency (the `RelaxMultiDyn` removal bet).** `ReshapeShapeFold` + `ReshapeConversion`'s multi-dyn branch are precisely what **produce** the 2-dynamic-dim `memref.expand_shape` that the now-removed `RelaxMultiDynExpandShape` used to fix up (it dodged upstream `expand-strided-metadata`'s *"at most one dynamic size per group"* assert).

Removing that pass **bets that `main`'s `ResolveTensorDimsPass` resolves those dims pre-bufferize**, so the 2-dyn `memref.expand_shape` never reaches the upstream assert. If the bet is wrong, the dynamic-reshape path (Qwen3.5 mrope) will either hit the upstream assert (compile abort) or silently fall back to CPU.

**lit only exercises the ONNX->HIP conversion stage and cannot prove either point.** The remaining gate is a **Qwen3.5 GPU end-to-end run with `HIPDNN_EP_STRICT=1`** (strict mode turns any silent CPU fallback into a hard failure). Until that passes, treat this PR as build/lit-green but e2e-unverified.

## Test plan

- [x] Full configure + build + install (`build-hipdnn-ep.bat`) -- green
- [x] `--hip-relax-multi-dyn-expand-shape` flag confirmed **gone** from `hip-mlir-opt`
- [x] Conversion + Dialect lit -- **142/142**
- [x] `hip-mlir-opt --convert-onnx-to-hip` + FileCheck on `test_reshape.mlir` -- pass (rank-0 scalar wraps only; see gap #1)
- [x] pre-commit (lintrunner / clang-format / license) -- pass
- [ ] **Qwen3.5 GPU e2e with `HIPDNN_EP_STRICT=1`** -- pending (covers gaps #1 and #2 above)

## Dependencies

Independent -- based directly on `main`, no dependency on the #284 ABI train.

### DIFF
--- a/include/hip/Dialect/Transforms/Passes.td
+++ b/include/hip/Dialect/Transforms/Passes.td
@@ -402,6 +402,51 @@ def LowerAllocsPass : Pass<"hip-lower-allocs", "mlir::func::FuncOp"> {
   ];
 }
 
+def RelaxMultiDynExpandShapePass
+    : Pass<"hip-relax-multi-dyn-expand-shape", "mlir::func::FuncOp"> {
+  let summary = "Rewrite memref.expand_shape ops with multiple dynamic output "
+                "dims per reassociation group into memref.reinterpret_cast";
+  let description = [{
+    Workaround for upstream `expand-strided-metadata` (LLVM as of our
+    prebuilt) asserting "There must be at most one dynamic size per group"
+    when a `memref.expand_shape` carries a reassociation group containing
+    more than one dynamic output dim — even though the op carries explicit
+    `output_shape` operands that fully determine every dim's size.
+
+    This pass matches exactly that case and rewrites the op into a
+    `memref.reinterpret_cast` with explicit sizes/strides computed from the
+    `output_shape` operands and the source's static layout.  The resulting
+    `memref.reinterpret_cast` is already lowerable by the existing
+    downstream LLVM-conversion patterns.
+
+    Single-dyn-per-group `memref.expand_shape` and all `memref.collapse_shape`
+    are left untouched — upstream handles those correctly.
+
+    Canonical IR pattern that triggers a match (ONNX `Range -> Reshape`
+    materialising 2-D position_ids):
+
+        %expand = memref.expand_shape %src [[0, 1]]
+                  output_shape [%bs, %ss]
+                  : memref<?xi64> into memref<?x?xi64>
+
+    After this pass:
+
+        %cast = memref.reinterpret_cast %src to
+                  offset: [0], sizes: [%bs, %ss], strides: [%ss, 1]
+                  : memref<?xi64> to memref<?x?xi64>
+
+    Pipeline placement: very first sub-pass of `buildHipToLLVMPipeline`,
+    immediately before `memref::createExpandStridedMetadataPass()`.
+
+    Retirement: when upstream `expand-strided-metadata` is upgraded to
+    consult `output_shape`, this pass becomes a no-op and can be deleted.
+  }];
+  let dependentDialects = [
+    "mlir::affine::AffineDialect",
+    "mlir::memref::MemRefDialect"
+  ];
+}
+
 def ResolveTensorDimsPass
     : Pass<"hip-resolve-tensor-dims", "mlir::func::FuncOp"> {
   let summary = "Fold `tensor.dim` queries via upstream reify patterns "
--- a/lib/Conversion/OnnxToHip/CMakeLists.txt
+++ b/lib/Conversion/OnnxToHip/CMakeLists.txt
@@ -23,6 +23,7 @@ add_library(OnnxToHip STATIC
   MultiHeadAttentionConversion.cpp
   GatherConversion.cpp
   GatherShapeFold.cpp
+  ReshapeShapeFold.cpp
   ShapeConversion.cpp
   ReshapeConversion.cpp
   CausalConvWithStateConversion.cpp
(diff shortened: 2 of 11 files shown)


include/hip/Dialect/Transforms/Passes.td
```diff
@@ -335,6 +335,51 @@ def InferShapesPass : Pass<"hip-infer-shapes", "mlir::ModuleOp"> {
   ];
 }
 
+def RelaxMultiDynExpandShapePass
```
fhanuman: Hi @AMDmoore, We don't need most of the changes in this PR. All this is addressed as part of the following PRs already. Please check the following PRs and remove what's not needed.

https://github.com/ROCm/onnx-hipdnn-ep/pull/286
https://github.com/ROCm/onnx-hipdnn-ep/pull/293
https://github.com/ROCm/onnx-hipdnn-ep/pull/287
https://github.com/ROCm/onnx-hipdnn-ep/pull/288

fhanuman (changes requested):

lib/Dialect/Transforms/PoolAllocs.cpp
```diff
@@ -34,8 +34,11 @@
 
 #include "mlir/Dialect/Arith/IR/Arith.h"
```
fhanuman, line 51: Please revert changes to this file. A lot of these fixes here are hacks and I have fixed them properly as part of the following PRs.

PR #286
PR #293
PR #287
PR #288

fhanuman (changes requested):

lib/Dialect/Transforms/PoolAllocs.cpp
AMDmoore: ok, will revert changes in this file.

include/hip/Dialect/Transforms/Passes.td
AMDmoore: ok, will revert these changes and the related caller implementation.

fhanuman (dismissed):

fhanuman:
Hi @AMDmoore, Upstream `memref-expand-strided-metadata` in the LLVM we currently pin (22.1.x) can't lower a `memref.expand_shape` whose reassociation has more than one `dynamic dim` in a single group — it asserts `There must be at most one dynamic size per group` instead of consulting the op's `output_shape` operands. Our IR hits this legitimately (canonical case: ONNX Range -> Reshape([bs, ss]) for 2-D position_ids). The pass `RelaxMultiDynExpandShape` we had is a targeted workaround that rewrites only that case to `memref.reinterpret_cast`; everything else is left to upstream. So, I am putting this pass back till we do LLVM upgrade.

This was fixed upstream (expand-strided-metadata now reads output_shape), so once we bump to LLVM 23, the pass becomes a no-op and can be reverted.

fhanuman (approved):

lib/Dialect/Transforms/PoolAllocs.cpp
AMDmoore: changes are reverted.

PR #305, opened 2026-06-08
feat(nonzero,scatter_nd): thread GPU row-count via count_buf/valid_count to match #284 ABI

AMDmoore:
## Summary

Lowering-half of the #284 ABI train (**NonZero / ScatterND slice**, the GPU row-count dataflow). #284's runtime ships `wrap_nonzero` (9-arg, +`count_ptr`/`input_dims`) and `wrap_scatter_nd` (16-arg, +`count_ptr`), but the compiler half still emitted the old 7-/15-arg calls -- a latent ABI mismatch that only the VLM embedding `NonZero -> Transpose -> ScatterND` placeholder-substitution chain exercises. This brings the op defs, conversions and lowerings in line.

**NonZero** gains a second DPS output `count_buf` (`memref<1xi32>`) that receives the data-dependent non-zero count at runtime (kernel atomicAdd):
- `getDpsInitsMutable` now spans `{y, count_buf}` so the shared `HipDpsOp::reifyResultShapes` reifies both for free.
- `autoInfer=0` + a hand-written 2-result `inferReturnTypes` (the auto-generated body pushes only the single `outsAccessor` result).
- `NonZeroLowering` emits the 9-arg call (`count_ptr` + a stack `input_dims` array).

**ScatterND** gains a `valid_count` input operand + `has_valid_count` attr:
- `ScatterNDConversion` backtraces the `indices` operand (through Transpose/Reshape/Cast/... and their hip/tensor equivalents) to an upstream `hip.nonzero` and wires its `count_buf` as `valid_count` (`has_valid_count=true`); otherwise a dummy `1xi32` (`false`, never read). Defers when an unconverted `onnx.NonZero` is found so the greedy rewriter retries post-conversion.
- `ScatterNDLowering` emits the 16-arg call (`count_ptr` = the `valid_count` ptr, or null when `has_valid_count=false`). Operand-only change -> single-result auto-infer is untouched (low risk).

> This is the `count_buf`/reify-surgery half that PR-3 was split into. It is **independent of** and **parallel to** #304 (Gather slice) -- both branch directly off #284 and touch disjoint files, so they can merge in any order.

## Test plan

- [x] Full rebuild + install (tablegen regen) -- green
- [x] `ctest -R MorphizenMLIRLitTests` -- 100% pass
- [x] hip-to-llvm + onnx-to-hip `test_nonzero` / `test_scatter_nd` updated to the new assembly + sigs
- [x] Added `NonZero -> Transpose -> ScatterND` backtrace regression (asserts `has_valid_count = true` + 2-result `hip.nonzero`)
- [x] `hip-infer-shapes` nonzero case updated to the 2-out form
- [x] pre-commit (clang-format + license) -- pass

## Dependencies

Base: **#284** (the matching runtime). Parallel to #304; merge after #284.

### DIFF
--- a/lib/Conversion/OnnxToHip/ScatterNDConversion.cpp
+++ b/lib/Conversion/OnnxToHip/ScatterNDConversion.cpp
@@ -9,6 +9,65 @@ namespace mlir {
 namespace hip {
 namespace {
 
+/// Backtrace the SSA chain from `indices` through ops that preserve element
+/// identity (Transpose/Reshape/Squeeze/Unsqueeze/Cast/Gather and their HIP /
+/// tensor equivalents) to find a `hip.nonzero`. If found, return its count_buf
+/// (result[1]) — the runtime row count to bound ScatterND's writes.
+///
+/// Returns {count_value, should_defer}:
+///   count_value : non-null if a converted `hip.nonzero` is found upstream.
+///   should_defer: true if an UNCONVERTED `onnx.NonZero` is found — the caller
+///                 returns failure() so the greedy rewriter retries this op
+///                 after NonZero converts (and exposes its count_buf result).
+static std::pair<mlir::Value, bool> traceToNonZeroCount(mlir::Value indices) {
+  mlir::Value current = indices;
+  // Bounded walk: the NonZero->ScatterND index chain in real exports is short
+  // (Transpose + a reshape or two); 16 hops is generous slack.
+  for (int depth = 0; depth < 16; ++depth) {
+    auto *defOp = current.getDefiningOp();
+    if (!defOp)
+      return {nullptr, false};
+
+    if (auto nonzeroOp = llvm::dyn_cast<mlir::hip::NonZeroOp>(defOp)) {
+      if (nonzeroOp->getNumResults() >= 2)
+        return {nonzeroOp->getResult(1), false};
+      return {nullptr, false};
+    }
+
+    // Unconverted onnx.NonZero: defer so the rewriter retries post-conversion.
+    if (defOp->getName().getStringRef() == "onnx.NonZero")
+      return {nullptr, true};
+
+    llvm::StringRef opName = defOp->getName().getStringRef();
+    // ONNX shape/identity ops carry the index data through operand 0.
+    if (opName == "onnx.Transpose" || opName == "onnx.Reshape" ||
+        opName == "onnx.Squeeze" || opName == "onnx.Unsqueeze" ||
+        opName == "onnx.Cast" || opName == "onnx.Gather") {
+      if (defOp->getNumOperands() >= 1) {
+        current = defOp->getOperand(0);
+        continue;
+      }
+    }
+    // HIP DPS equivalents take ctx as operand 0; the data input is operand 1.
+    if (opName == "hip.transpose" || opName == "hip.cast") {
+      if (defOp->getNumOperands() >= 2) {
+        current = defOp->getOperand(1);
+        continue;
+      }
+    }
+    // tensor reshape/cast descriptor ops carry data through operand 0.
+    if (opName == "tensor.cast" || opName == "tensor.collapse_shape" ||
+        opName == "tensor.expand_shape") {
+      if (defOp->getNumOperands() >= 1) {
+        current = defOp->getOperand(0);
+        continue;
+      }
+    }
+    return {nullptr, false};
+  }
+  return {nullptr, false};
+}
+
 /// onnx.ScatterND -> hip.scatter_nd
 ///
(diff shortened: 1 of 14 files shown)


lib/Dialect/IR/HipDialect.cpp
```diff
@@ -1399,17 +1399,43 @@ void ScatterNDOp::getEffects(
...
 }
 
+// Hand-written because NonZero is the only HIP DPS op with two outputs, and the
+// auto-generated inferReturnTypes (Hip_DpsOp autoInfer=1) pushes only the
+// single outsAccessor result. Push both `y` and `count_buf` in DPS-init order
+// so the SSA result order (result[0]=y, result[1]=count_buf) matches what
+// NonZeroConversion builds. Only ranked-tensor outs become SSA results; in
+// memref mode (post-bufferize) both are memrefs and nothing is pushed.
+LogicalResult
+NonZeroOp::inferReturnTypes(MLIRContext *, std::optional<Location>,
```
fhanuman, line 1462: This is going to grow at some point when we specialize the `inferReturnTypes` for other operators, so please keep these in a separate utils file for `inferReturnTypes` like we have done  here:

https://github.com/ROCm/onnx-hipdnn-ep/blob/main/lib/Dialect/IR/HipShapeUtils.cpp
https://github.com/ROCm/onnx-hipdnn-ep/blob/main/lib/Dialect/IR/HipShapeUtils.cpp

lib/Conversion/OnnxToHip/ScatterNDConversion.cpp
```diff
@@ -9,6 +9,65 @@ namespace mlir {
...
+      return {nullptr, false};
+    }
+
+    // Unconverted onnx.NonZero: defer so the rewriter retries post-conversion.
+    if (defOp->getName().getStringRef() == "onnx.NonZero")
+      return {nullptr, true};
+
+    llvm::StringRef opName = defOp->getName().getStringRef();
+    // ONNX shape/identity ops carry the index data through operand 0.
+    if (opName == "onnx.Transpose" || opName == "onnx.Reshape" ||
```
fhanuman, line 43: This looks hacky to me -- What happens if we have another operator which needs same kind of treatment?

lib/Conversion/OnnxToHip/ScatterNDConversion.cpp
```diff
@@ -9,6 +9,65 @@ namespace mlir {
...
+    }
+    // HIP DPS equivalents take ctx as operand 0; the data input is operand 1.
+    if (opName == "hip.transpose" || opName == "hip.cast") {
+      if (defOp->getNumOperands() >= 2) {
+        current = defOp->getOperand(1);
+        continue;
+      }
+    }
+    // tensor reshape/cast descriptor ops carry data through operand 0.
+    if (opName == "tensor.cast" || opName == "tensor.collapse_shape" ||
```
fhanuman, line 59: Same comment as above.

fhanuman (changes requested):

lib/Conversion/OnnxToHip/ScatterNDConversion.cpp
AMDmoore: @fhanuman yes, this is a hacky solution in pr259, it allocates output buffer by upper bound and calculate the real count by kernel, then for ScatterND op, it should know the existence of real count. So, I think this is the logic to judge the existence of NonZero.
have you finished the correct implementation for NonZero or similar data dependent ops? if yes, I can remove the whole implementation for NonZero including the runtime part.

lib/Conversion/OnnxToHip/ScatterNDConversion.cpp
AMDmoore: same reply as above.

lib/Dialect/IR/HipDialect.cpp
AMDmoore: @fhanuman As you say, HipInferTypeUtils.cpp created.

lib/Conversion/OnnxToHip/ScatterNDConversion.cpp
fhanuman: Hi @AMDmoore, data dependent ops will be handled now see PR #320, once all the remaining `output buffer allocator` changes are merged -- PR #298, PR #301, and PR #306. 

So, please don't merge this hacky solution instead use the fixes from PR #320.

AMDmoore: ok, keep this pr open, if confirm unused after test, will close it.

PR #306, opened 2026-06-08
feat(ep): wire output allocator EP bridge + zero-copy (phase 5)

zpye:
## Summary

Phase 5 of the output-allocator feature: wires the MorphiZen EP to drive the Phase 4 allocator pipeline end-to-end, behind an explicit `use_output_allocator` provider option (default OFF, so the classic 3-arg path is byte-identical and unaffected).

Built on the attribute-driven Phase 4 (#301, now merged): the `hipdnn.output_allocator` module attribute is the source of truth, and the HIP-to-LLVM half reads it off the IR rather than a second flag.

- **Single source of truth = compile flag → module attribute + embedded metadata bit.** The `use_output_allocator` flag is set once on the OnnxToHip half in `CompilerDriver`, where the slot-4.5 pair `hip-use-output-allocator` (IR rewrite) + `hip-set-output-allocator-attr` (stamps `hipdnn.output_allocator`) runs. The HipToLLVM half (`convert-hip-to-llvm` + the single `generate-interface`) reads that module attribute to pick the 2-arg ABI — no separate flag, so the two halves can't disagree. The same flag is also embedded in the model `Metadata`. The EP reads its dispatch arity from the embedded metadata (not the live provider option), so the DLL ABI and EP dispatch can never disagree (correct even on a reused ORT EPContext).
- **2-arg dispatch + allocator callback.** `InferenceState` gains `compute_with_output_allocator` (2-arg `inference_compute(state, inputs)`) and `set_output_allocator` (resolved once, optional-export like `begin_compute`). `MlirCustomOp` installs a per-`Compute()`, `noexcept` callback that maps `out_idx` -> ORT output, asks ORT for the buffer at the DLL's **in-graph shape (used verbatim — no `present.*`/share-buffer override)**, and returns ORT's GPU pointer (zero-copy) or an EP-owned grow-on-demand GPU scratch for host outputs (blocking D2H after the DLL's stream-sync; real-build-only). An output-completeness guard `LOG(FATAL)`s on unsupported passthrough/aliased outputs.
- **KV cache needs no override on the allocator path.** A `present.*` output's dynamic seq dim is materialized in-graph by `hip.alloc_output` from `memref.dim %past_key` (the past input buffer's actual extent), which under OGA `past_present_share_buffer` already **is** the `max_length` capacity buffer — so `GetOutput` returns the pre-bound shared `OrtValue` and `past == present` pointer identity is preserved with zero special-casing. Pinned by a new e2e LIT guard (`test_gqa_output_allocator_present_dim.mlir`) that asserts each dynamic `present.*` `hip.alloc_output` is fed by a `memref.dim` of its matching `past_key_values.*` input. The classic `marshal_output_tensors` path keeps its **own inline** `present.*` override (it pre-computes shapes from `DimSource`→`attention_mask`, the tight current length) and is intentionally left byte-identical to `main` — the allocator path is decoupled from that OGA-specific logic (addresses the layering concern raised in review).
- **Docs.** `output-allocator-design.md` Phase 5 marked implemented + a "Usage" section (how to enable the provider option in ORT/OGA, and the cache-key caveat); `CLAUDE.md` gotcha (single-source flag→attribute+metadata, EPContext not keyed on provider options, `noexcept` callback, real-build-only host D2H, **allocator path needs no override / do-not-re-add guard**, plus the enable syntax).

Deferred (out of scope): per-model auto-enable heuristic; data-dependent outputs (NonZero/Range/ConstantOfShape); passthrough/aliased outputs.

## Test plan

All green on gfx1151 under `HIPDNN_EP_STRICT=1` (so any allocator-mode compile failure hard-aborts instead of silently falling back to CPU):

- [x] **Unit (GPU-free), `OutputAllocatorUnitTest`** -- runtime-contract coverage for the allocator entry points: `(self, out_idx, shape, rank, elem_size)` forwarding + return-pointer pass-through, and the null guards (no allocator installed / cleared / null state). PASS.
- [x] **LIT** -- allocator pipeline + codegen gates pass, incl. `hip-use-output-allocator`, `hip-set-output-allocator-attr`, `output-allocator-dealloc`, `test_output_allocator_model`, `test_main_graph_param_mismatch`, and the new **`test_gqa_output_allocator_present_dim.mlir`** invariant guard.
- [x] **Python e2e, Llama-1B (allocator):**
  - `test_ort_output_allocator_dynamic` -- one allocator session, multiple shapes (prefill sq=128/sq=64 + decode), host-D2H path; decode cosines ~0.9999 vs CPU golden (true GPU, not CPU fallback). GPU zero-copy sub-check skips gracefully where the ORT wheel can't allocate AMD device memory from Python (covered by OGA + latency paths).
  - `test_ort_output_allocator_vs_classic` -- allocator vs classic byte-identity gate; every output (logits + present KV, 32 tensors) cosine==1.000000, max_abs=0.
- [x] **`python build.py`** -- clean incremental build after the rebase reconciliation.
- [ ] Numeric -- optional; backend toggle not wired (deferred).
- [ ] OGA suite -- not re-run in this pass; OGA uses classic mode and shares the validated `Compute()` path.


wcy123 (commented):
## Naming: `compute_allocator` → `compute_with_output_allocator`

**Issue**: `compute_allocator()` reads like "compute an allocator" rather than "execute inference using output-allocator mode".

**Suggestion**: Rename to `compute_with_output_allocator()` for clarity and consistency:

```cpp
// InferenceState.h
- int compute_allocator(span_t *inputs) const;
+ int compute_with_output_allocator(span_t *inputs) const;
```

**Rationale**:
- Matches `MlirCustomOp::compute_with_output_allocator()` (consistency across layers)
- Aligns with `use_output_allocator` flag and `set_output_allocator()` method
- Call site becomes self-documenting: `inference_state_->compute_with_output_allocator(&inputs.span)`

wcy123 (changes requested):
## Layering violation: OGA-specific logic in general EP

**Issue**: `apply_present_share_buffer_override()` hard-codes OGA's `past_present_share_buffer` convention into the EP.

**Evidence**:
- Function name uses OGA terminology: "present", "share_buffer"
- Mapping `present_to_past_input_idx` encodes OGA's KV cache naming convention
- Comment (line 16) explicitly references "OGA binds the SAME OrtValue to past_key_values.N and present.N"

**Why this is wrong**:
- The EP is a **general compiler backend**, not an OGA-specific runtime
- Model-specific conventions should not be hard-coded in EP infrastructure
- This couples the EP to OGA's evolving design choices

---

**Root cause hypothesis**: ORT's `GetOutput(shape)` API doesn't distinguish between:
1. Logical shape (actual data extent): `[1, 8, 7, 64]` - what the DLL computed
2. Capacity shape (buffer allocation): `[1, 8, 128, 64]` - what ORT pre-allocated via IO binding

The override works around this by requesting capacity instead of logical shape to retrieve pre-allocated buffers.

**Before merging**: We need to explore ORT's `GetOutput()` implementation to understand:
- Is this distinction (logical vs capacity) a real ORT API limitation?
- How does `GetOutput(idx, shape)` behave when an OrtValue is pre-allocated via IO binding?
- Why does requesting the DLL's computed shape fail to return the pre-allocated buffer?
- Is there existing ORT API we're missing that handles this correctly?

The single source of truth should be `ctx->GetOutput(ort_idx, shape)`. If we're computing shape from the DLL then "fixing" it with model-specific overrides, we're papering over the real problem.

The output allocator design doc (`tech/notes/2026-06-06-output-allocator-design.md`) doesn't explain why this override is necessary or how it relates to the allocator design.

backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
```diff
@@ -78,6 +78,32 @@ struct span_t {
...
+// graph-output buffer at the point its shape is known. Re-declared here (not
+// shared) for the same decoupling reason as tensor_t.
+//
+// ABI / forward-compat: `struct_size` MUST stay first; the runtime setter
+// copies only min(caller_size, local_size) bytes, so the EP MUST set
+// `struct_size = sizeof(output_allocator_t)`. New callbacks are APPENDED after
+// `allocate`; existing fields never move. The runtime header explicitly
+// requires this EP-side copy to carry the same asserts.
+struct output_allocator_t {
+  size_t struct_size; // ABI size guard; MUST be first (offset 0)
```
wcy123: `size_t struct_size; // ABI size guard; MUST be first (offset 0)`
this might over-designed.

anyway, it is nice to have.

wcy123 (commented):
## Code structure: Group parallel vectors into struct

**Current design** uses parallel vectors for host output scratch:
```cpp
mutable std::vector<void *> host_out_scratch_ptr_;
mutable std::vector<size_t> host_out_scratch_cap_;
```

**Suggestion**: Group into a single struct:
```cpp
struct HostOutputScratch {
  void *ptr = nullptr;
  size_t capacity = 0;
};
mutable std::vector<HostOutputScratch> host_out_scratch_;
```

**Rationale**:
- Eliminates synchronization burden (no manual `resize` of two vectors)
- Better cache locality (ptr and capacity adjacent in memory)
- Cleaner function signatures: `ensure_host_out_slot(std::vector<HostOutputScratch>&, ...)`
- Less error-prone (cannot accidentally resize one vector but not the other)

backend-mlir-compiler/custom-op-mlir/src/MlirCustomOp.cpp
```diff
@@ -585,6 +575,152 @@ struct PerfSummaryPrinter {
...
+        past_idx < static_cast<int>(octx->input_index_map->size())) {
+      int ort_past_idx = (*octx->input_index_map)[past_idx];
+      auto past_tensor = octx->ctx->GetInput(ort_past_idx);
+      auto past_shape = past_tensor.GetTensorTypeAndShapeInfo().GetShape();
+      apply_present_share_buffer_override(output_meta.shape().data(),
+                                          past_shape.data(), out_shape.data(),
+                                          out_shape.size(), past_shape.size());
+    }
+
+    int ort_idx = (*octx->output_index_map)[static_cast<int>(out_idx)];
```
fhanuman, line 676: I am just curious: How would the `ORT` know what the output buffer-size should be for dynamic shapes? Does memory get allocated when we call `ctx->GetOutput(ort_idx, out_shape)` on the fly?

```
  int ort_idx = (*octx->output_index_map)[static_cast<int>(out_idx)];
    auto out_tensor = octx->ctx->GetOutput(ort_idx, out_shape);
    int mem_type =
        static_cast<int>(out_tensor.GetTensorMemoryInfo().GetDeviceType());
    void *ort_ptr = out_tensor.GetTensorMutableRawData();

    octx->allocated[static_cast<size_t>(out_idx)] = true;

    if (mem_type == TENSOR_MEMORY_GPU) {
      // Zero-copy: ORT buffer is GPU-accessible (EP hipHostMalloc allocator or
      // OGA device memory). The DLL writes into it directly.
      return ort_ptr;
    }
```

Also, is there way we can hash the output-size and use that hash to determine whether to call `ctx->GetOutput(ort_idx, out_shape)` or not if the output-shape is same for the second-invocation?


This is not blocker, so please think about it and let's get changes merged in soon. And address feedbacks in a follow-up if that requires more time.

fhanuman (dismissed):

backend-mlir-compiler/custom-op-mlir/src/custom_op_mlir.hpp
zpye: removed in this PR

zpye:
> ## Naming: `compute_allocator` → `compute_with_output_allocator`
> **Issue**: `compute_allocator()` reads like "compute an allocator" rather than "execute inference using output-allocator mode".
> 
> **Suggestion**: Rename to `compute_with_output_allocator()` for clarity and consistency:
> 
> ```c++
> // InferenceState.h
> - int compute_allocator(span_t *inputs) const;
> + int compute_with_output_allocator(span_t *inputs) const;
> ```
> 
> **Rationale**:
> 
> * Matches `MlirCustomOp::compute_with_output_allocator()` (consistency across layers)
> * Aligns with `use_output_allocator` flag and `set_output_allocator()` method
> * Call site becomes self-documenting: `inference_state_->compute_with_output_allocator(&inputs.span)`

fixed

zpye:
> ## Code structure: Group parallel vectors into struct
> **Current design** uses parallel vectors for host output scratch:
> 
> ```c++
> mutable std::vector<void *> host_out_scratch_ptr_;
> mutable std::vector<size_t> host_out_scratch_cap_;
> ```
> 
> **Suggestion**: Group into a single struct:
> 
> ```c++
> struct HostOutputScratch {
>   void *ptr = nullptr;
>   size_t capacity = 0;
> };
> mutable std::vector<HostOutputScratch> host_out_scratch_;
> ```
> 
> **Rationale**:
> 
> * Eliminates synchronization burden (no manual `resize` of two vectors)
> * Better cache locality (ptr and capacity adjacent in memory)
> * Cleaner function signatures: `ensure_host_out_slot(std::vector<HostOutputScratch>&, ...)`
> * Less error-prone (cannot accidentally resize one vector but not the other)

fixed

zpye:
> ## Layering violation: OGA-specific logic in general EP
> **Issue**: `apply_present_share_buffer_override()` hard-codes OGA's `past_present_share_buffer` convention into the EP.
> 
> **Evidence**:
> 
> * Function name uses OGA terminology: "present", "share_buffer"
> * Mapping `present_to_past_input_idx` encodes OGA's KV cache naming convention
> * Comment (line 16) explicitly references "OGA binds the SAME OrtValue to past_key_values.N and present.N"
> 
> **Why this is wrong**:
> 
> * The EP is a **general compiler backend**, not an OGA-specific runtime
> * Model-specific conventions should not be hard-coded in EP infrastructure
> * This couples the EP to OGA's evolving design choices
> 
> **Root cause hypothesis**: ORT's `GetOutput(shape)` API doesn't distinguish between:
> 
> 1. Logical shape (actual data extent): `[1, 8, 7, 64]` - what the DLL computed
> 2. Capacity shape (buffer allocation): `[1, 8, 128, 64]` - what ORT pre-allocated via IO binding
> 
> The override works around this by requesting capacity instead of logical shape to retrieve pre-allocated buffers.
> 
> **Before merging**: We need to explore ORT's `GetOutput()` implementation to understand:
> 
> * Is this distinction (logical vs capacity) a real ORT API limitation?
> * How does `GetOutput(idx, shape)` behave when an OrtValue is pre-allocated via IO binding?
> * Why does requesting the DLL's computed shape fail to return the pre-allocated buffer?
> * Is there existing ORT API we're missing that handles this correctly?
> 
> The single source of truth should be `ctx->GetOutput(ort_idx, shape)`. If we're computing shape from the DLL then "fixing" it with model-specific overrides, we're papering over the real problem.
> 
> The output allocator design doc (`tech/notes/2026-06-06-output-allocator-design.md`) doesn't explain why this override is necessary or how it relates to the allocator design.

fixed

wcy123 (approved):
All requested changes addressed:

1. ✅ Renamed `compute_allocator` → `compute_with_output_allocator` (consistency across layers)
2. ✅ Grouped parallel vectors into `HostOutputScratch` struct (eliminates sync burden)
3. ✅ Removed `apply_present_share_buffer_override()` (layering violation fixed)

LGTM.

backend-mlir-compiler/custom-op-mlir/src/MlirCustomOp.cpp
zpye: There is currently no such implementation, we can discuss it later.

PR #308, opened 2026-06-08
feat(onnx-to-hip): vision/projector compute ops + Conv/Matmul ABI (PR-9, ex #259)

AMDmoore:
## Summary (PR-9 of the #259 split — train tail, base #284)

Adds the vision/projector **compute-op** slice of #259 and finishes the Conv/Matmul **lowering ABI** whose runtime half already shipped in #284.

- **ProjectorOpsRewrites** (net-new): patch-embed Conv-ND -> Reshape/Gemm/Reshape, AveragePool(kernel==stride) -> Reshape/Transpose/ReduceMean, Pow(x,c) -> Mul chain, broadcasting Div -> Mul(Reciprocal).
- **FastGeluFusion**: exact-Gelu (Erf) + inlined-vision Gelu variants; peek-through Reciprocal/Div constants.
- **ConvLowering**: emit data_type (derived from output elem type) -> matches wrap_miopenConvolutionForward in #284.
- **MatmulLowering**: emit _batch_stride (compile-time-folded leading-dim product; rank-2 / leading-one B -> 0) -> matches wrap_hipblasLtMatmul in #284.
- **OnnxToHip pre-lowering**: FastGelu+Projector now run as a fixed-point quiescence loop (cap 4) so projector-emitted ops are re-visited next round. The upstream multi-round driver's `inferOnnxShapes` call is dropped — that ONNX-level forward inference was superseded on main by the HIP-dialect `--hip-infer-shapes` pass. Newly-emitted ops are typed **in-place at emission** (result types built explicitly from the dims the rewriter already knows); the `--hip-infer-shapes` pass (pipeline tail, post-conversion) resolves any residual dynamic dims.

> **Update — addressed review comment:** an earlier revision added an `OnnxResultTypeInference` helper (pure result-type rules for onnx.* ops) to type newly-emitted ops at emission. As noted in review, this is unnecessary now that shape inference + propagation (`--hip-infer-shapes`) covers it. The helper (`OnnxResultTypeInference.cpp/.h`) has been **removed** and its call sites inlined — each rewrite constructs its emitted ops' result types directly from the dims it already knows.

## Test plan

- [x] Full reconfigure + build + install clean (all new TUs compile).
- [x] LIT 208/208 pass — new `test_projector_rewrites.mlir`; `test_matmul.mlir` updated to 10-param + rank-3 leading-one B case.
- [ ] (post-merge of train) e2e vision-encoder accuracy on the dynamic-shape PR-1..#284 stack.

### DIFF
--- a/lib/Conversion/HipToLLVM/ConvLowering.cpp
+++ b/lib/Conversion/HipToLLVM/ConvLowering.cpp
@@ -17,7 +17,27 @@ namespace {
 //                                     output, output_h, output_w, kernel_h,
 //                                     kernel_w, stride_h, stride_w, pad_top,
 //                                     pad_left, pad_bottom, pad_right,
-//                                     dilation_h, dilation_w, group)
+//                                     dilation_h, dilation_w, group, data_type)
+//
+// Before:
+//   %out = hip.conv(%ctx) ins(%in, %w, %b :
+//                              memref<1x3x896x896xf16, 1>,
+//                              memref<1152x3x14x14xf16, 1>,
+//                              memref<1152xf16, 1>)
+//                          outs(%o : memref<1x1152x64x64xf16, 1>)
+//                          {kernel_shape=[14,14], strides=[14,14], ...}
+// After:
+//   llvm.call @wrap_miopenConvolutionForward(%ctx, %in, 1, 3, 896, 896,
+//                                              %w, 1152, %b, %o, 64, 64,
+//                                              14, 14, 14, 14, 0, 0, 0, 0,
+//                                              1, 1, 1,
+//                                              /*data_type=*/1 /* f16 */)
+//
+// The `data_type` value is derived from the OUTPUT memref's element type and
+// applied uniformly to all three MIOpen tensor descriptors. MIOpen requires
+// input / weights / output to share the same dtype; this is enforced by the
+// host-side typing rule in OnnxToHip (`init` tensor allocated with
+// `resultType.getElementType()`).
 struct ConvOpLowering : public ConvertOpToLLVMPattern<ConvOp> {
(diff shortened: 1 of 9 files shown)


lib/Conversion/OnnxToHip/OnnxResultTypeInference.cpp
```diff
@@ -0,0 +1,453 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
```
fhanuman: We probably don't need this after `shape inference + shape propagation` changes?

lib/Conversion/OnnxToHip/OnnxResultTypeInference.h
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+//===- OnnxResultTypeInference.h - Pure result-type rules for onnx.* ops -===//
```
fhanuman: Same comment as above we probably don't need this -- `Shape inference  + shape propagation` should handle this?

fhanuman (changes requested):

lib/Conversion/OnnxToHip/OnnxResultTypeInference.cpp
AMDmoore: I see, you are right. I will remove these two files.

qianglin-amd (dismissed):

qianglin-amd (dismissed):

qianglin-amd (approved):

fhanuman (approved):

PR #310, opened 2026-06-09
feat(conversion): onnx.LpNormalization decomposition (#259 split 4/N)

AMDmoore:
﻿Part 4 of the #259 mega-PR split. **Base = `pr259-split/09-vision-projector-compute` (PR #308)** — this PR stacks on PR-9 and should be reviewed/merged after it.

## What
Adds `onnx.LpNormalization` (opset 22) support by decomposing it at the ONNX level into a chain of ops that already have HIP converters. No new HIP dialect op or runtime kernel.

- `p == 2`: `x_sq = Mul(x,x)` -> `ReduceSum(x_sq, axis, keepdims=1)` -> `Sqrt` -> `Div(x, norm)`
- `p == 1`: `x_sq = Mul(x,x)` -> `Sqrt` (= `|x|`) -> `ReduceSum(.., axis, keepdims=1)` -> `Div(x, norm)`

## Why (rationale carried from #259)
- **No 1:1 primitive.** There is no MIOpen / hipBLASLt op that maps directly to LpNormalization, so we rewrite rather than add a bespoke kernel.
- **Why the pre-lowering loop, not `convertComputeOps`.** The pattern is registered next to `FastGeluFusion` / `ProjectorOpsRewrites` in the `convert-onnx-to-hip` `ExistingOps` fixed-point round loop. Freshly-emitted `onnx.*` primitives become "existing" by the next round and reach the `convertComputeOps` driver as ordinary ops. A pattern inside `convertComputeOps` (single-shot) would leave the new primitives unconverted and trip the survival check.
- **Why this depends on PR-9.** The final step emits a **broadcasting** `onnx.Div` (`norm` has shape `[...,1,...]`). `hip.div`'s flat element-wise kernel does **not** broadcast, so the broadcasting Div is intentionally left to `BroadcastDivToMulReciprocal` (introduced in PR-9) which rewrites it into `Mul(x, Reciprocal(norm))` before HIP lowering.
- **Spec edge case.** ONNX 22 defines zero-norm output as zero, but the accompanying example divides directly (NaN/Inf on zero norm). We match ORT CPU behaviour and the example with the plain division; strict compliance would cost an `Equal + Where` pair around the final Div. Revisit if a downstream model trips the all-zero edge case.

## Tests
- `test/lit/Conversion/onnx-to-hip/test_lpnormalization.mlir` (p=1, p=2, axis variants)
- `test/lit/e2e/test_lpnormalization_model.mlir`
- Full lit suite: **210/210 pass** (208 baseline + 2 new).

### DIFF
--- a/lib/Conversion/OnnxToHip/LpNormalizationConversion.cpp
+++ b/lib/Conversion/OnnxToHip/LpNormalizationConversion.cpp
@@ -1,0 +1,225 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+//===- LpNormalizationConversion.cpp - Decompose onnx.LpNormalization ----===//
+//
+// `onnx.LpNormalization` (opset 22) computes
+//
+//     output = input / Lp_norm(input, axis)
+//
+// where `Lp_norm` reduces along a single user-specified `axis` and `p` is
+// either 1 or 2. There is no MIOpen / hipBLASLt primitive that maps 1:1 to
+// this op; instead we rewrite it at the ONNX level as a small chain of ops
+// that already have HIP converters:
+//
+//   p == 2:
+//     %x_sq    = onnx.Mul(%x, %x)                        // same shape as %x
+//     %sum_sq  = onnx.ReduceSum(%x_sq, [axis], keepdims=1)
+//     %norm    = onnx.Sqrt(%sum_sq)
+//     %result  = onnx.Div(%x, %norm)                     // broadcasts on axis
+//
+//   p == 1:
+//     %x_sq    = onnx.Mul(%x, %x)
+//     %abs_x   = onnx.Sqrt(%x_sq)                        // |x| via sqrt(x^2)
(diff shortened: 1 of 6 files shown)


lib/Conversion/OnnxToHip/LpNormalizationConversion.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+namespace mlir {
+namespace hip {
+
+namespace {
+
+/// Decomposes `onnx.LpNormalization` into a chain of already-supported ONNX
+/// primitives. See file header for the rewrite shape and the rationale for
+/// living in the pre-lowering loop instead of `convertComputeOps`.
+struct LpNormalizationDecompose : public mlir::RewritePattern {
```
fhanuman, line 86: Since this is a plain conversion pass, can someone from Beijing team help review this?

AMDmoore: @amd-genmingz please help to review this pr.

fhanuman (approved):
I am going to approve this PR, so we can unblock the QWEN integration test. This pass itself could use cleanup, but lets get the qwen working we will cleanup the codebase in parallel.

PR #311, opened 2026-06-09
fix(qwen-vision): PrecisionFreeCast (PFC) eliminates fp16 saturation NaN cascade (#259 split)

AMDmoore:
Split out of #259. PFC is an **ONNX-graph-level** precision fix, independent of the runtime/ABI splits — so it bases directly on `main`.

## Problem
Qwen 3.5 9B vision encoder (and any transformer block whose fp16 residual / GEMM saturates at one row position) silently produces **100% NaN `image_features` on the EP** but finite output on ORT CPU — even though the fp16 inputs to the saturating op are **bit-identical** between the two. Flickr30k VLM PPL: 2 of 50 prompts returned NaN logits pre-fix.

Strict first-divergence point (per-op probe of vision encoder block 21→22):

| op | CPU | GPU |
|---|---|---|
| `add_3461` (residual Add) | `[728,1151] = -Inf` | `[728,1151] = -Inf` |
| `Cast(fp16→fp32)` of add_3461 | finite + Inf | finite + Inf |
| `LayerNormalization_43` (block 21 norm2) | **0 NaN**, range `[-304, 9]` | **1152 NaN** at row 728 |
| `linear_88` (QKV proj) | 0 NaN | `3*norm_size` NaN |
| block 22 inputs | finite | 100% NaN |

The "one NaN row at LN → `3*norm_size` NaN after the QKV Gemm → 100% NaN 1–3 blocks downstream" signature is the diagnostic fingerprint.

## Root cause
fp16 saturates at the producer of the residual `Add` (and the GEMM feeding it): one row overflows fp16 max (~65504) → ±Inf. `Cast(fp16,fp32)` preserves the Inf; `LayerNormalization` then computes `var = E[x²] - E[x]² = Inf - Inf = NaN`, poisoning the whole row, which fans out through attention.

ORT CPU does **not** diverge because MLAS keeps the dot product in fp32 **and** ORT's `InsertCastTransformer` / "PrecisionFreeCast" optimization fuses `fp16 → Cast(fp32) → fp32` chains so the fp16 storage is never materialized — fp32 has the headroom so the same sum does not saturate. Verified by 3 minimal control models (`Cast`, `Add→Cast→LN`, `Gemm→Cast→LN`) under `ORT_DISABLE_ALL`: CPU produces 0 NaN even with Inf-injected inputs; our per-op kernels faithfully reproduce the saturation.

## Fix
New `simplify-onnx` sub-pass `simplifyPrecisionFreeCast` (PFC) at the ONNX-dialect level, mirroring ORT CPU PrecisionFreeCast. Rewrites

```
fp16-saturating-producer -> onnx.Cast(fp16->fp32) -> fp32-pure-consumer
```
into
```
fp32 producer -> fp32 consumer
```

with per-input `fp16→fp32` casts and a `fp32→fp16` bridge cast for any remaining fp16 users of the original producer.

- **Promotable producers:** `onnx.Add`, `onnx.Sum`, `onnx.Gemm`, `onnx.MatMul`.
- **Fp32-pure consumers:** `onnx.LayerNormalization` OR any op whose results are all fp32.
- All producer attributes preserved (`Gemm` alpha/beta/transA/transB; node names; etc.).
- The pass **loops until no new anchors appear** (cap 64; converges in 2–3 rounds because Add promotion exposes Gemm anchors and vice versa); the input scan **peeks through** `Cast(fp32→fp16)` bridges so a multi-block residual chain promotes in one walk.

Strictly compiler-level — no kernel band-aid. (The LayerNorm `isfinite()` filter alternative was rejected: it only masks the symptom, leaves per-row divergence permanent, and catches saturation in no other op.)

**Coverage gap (documented in CLAUDE.md):** PFC anchors only on `{Add, Sum, Gemm, MatMul}` today. A future model saturating fp16 in `Mul`/`Pow`/`Conv`/`MatMulNBits`/`ReduceSum`/`Softmax` → Cast → fp32 needs that op added to `isPromotableProducer`.

## Verification
- **9 LIT cases** in `test_simplify_onnx_precision_free_cast.mlir` (Add / Gemm / MatMul anchors, cascading residual, bridge insertion, negative + fp32-pure-non-LN positive, already-fp32 no-op, chained Gemm+Add iteration). All pass.
- Qwen 3.5 9B vision hardhats: `image_features` NaN 100% → 0; `cos(GPU, CPU) = 1.000001`.
- Qwen 3.5 9B Flickr30k VLM PPL (50 samples, `max_length=2048`): pre-fix 48/50 finite (samples 5,7 NaN); post-fix **50/50 finite, EP PPL 91.490 vs CPU 91.496 (delta −0.0074%)**.

## Files
- `lib/Conversion/OnnxToHip/SimplifyOnnx.cpp` — `simplifyPrecisionFreeCast` helper + dispatcher integration.
- `include/hip/Dialect/Transforms/Passes.td` — pass description extended (anchor/consumer set + iteration).
- `test/lit/Conversion/onnx-to-hip/test_simplify_onnx_precision_free_cast.mlir` — 9-case LIT coverage.
- `CLAUDE.md` — symptom / root cause / fix / coverage gap.


fhanuman (commented):

AMDmoore:

@fhanuman sure, but this pr is for accuracy, not for function.
we can hold it until accuracy fixing.

fhanuman:
> 
> @fhanuman sure, but this pr is for accuracy, not for function. we can hold it until accuracy fixing.

Hi @AMDmoore, Good idea. Let's keep some of these PRs open till we verify numeric/accuracy.

PR #313, opened 2026-06-09
fix(qwen-vision): hip-fix-loop-accumulator-offset for Concat-grown hip.loop accumulators (#259 split)

AMDmoore:
﻿Split out of #259 (commits `4c4c33cd` "dynamic-shape Qwen vision support" which added the pass, and `1c438063` "three accuracy fixes" which refined its offset to `cu_seqlens_k[iter]`).

**Self-contained MLIR pass with no dependency on the other splits** — it operates on bufferized memref IR and references no symbol from PR-9 / the runtime split, so it bases directly on `main`. (Earlier draft based it on PR-9 for "memrefCopy pairing"; that was unnecessary — see Scope below.)

## Problem
Outlined `hip.loop` body funcs are emitted by OneShotBufferize + BufferResultsToOutParams as **in-place writers**: the `v_in` (loop-carried in) and `v_out` (`bufferize.result`) arg pairs alias the SAME memref descriptor (`LoopLowering.cpp::createOrGetTrampoline`, the `// v_out_i == v_in_i (same buffer, single-pass kernel safety)` line).

The body's typical **Concat-grow accumulator** computes the chunk-append byte offset as `arith.muli %v_in.dim(N), %stride`. But MLIR memref descriptors are **immutable by-value SSA**, so `v_in.dim(N)` is FROZEN at `v_init`'s compile-time value — the trampoline reloads the same descriptor every iteration. Every iter therefore writes its chunk to the **same byte offset**, and only the **last** iteration's contribution survives.

Canonical break: a transformer vision encoder's windowed-attention `onnx.Loop`, where cosine vs the CPU baseline collapses to ~`1/max_trip` (≈ 0.48 for `max_trip=2`).

## Fix
New post-bufferize pass `--hip-fix-loop-accumulator-offset`. It finds every `memref.subview` whose offset for dim N is `memref.dim %v_in_arg, %cN` and replaces that offset with an **iter-driven** value:

- **Preferred:** the TRUE cumulative start `cu_seqlens_k[iter]`, loaded (or synthesized) from the body's captured seqlens-style `hip.gather` output — accurate even when chunks are unequal (the `1c438063` refinement; avoids the `iter*sizes[i]` OOB write that occurs when `SliceConversion` over-allocates the slice output to `data.dim(axis)`).
- **Fallback:** `iter * chunk_size` (`chunk_size` = the subview's own size for dim N), for fixed-stride loops.

Uses in **size** positions are intentionally left alone, so the body's self-copy subview (`sizes=[..]`, `offsets=all-0`) is unaffected. **No-op** on any body func without the frozen-dim-in-offset pattern.

**Before** (loop body, post-BufferResultsToOutParams):
```mlir
%dim_9 = memref.dim %arg3, %c1            // FROZEN at v_init's dim(1)
%off   = arith.muli %dim_9, %strides#1
%dst   = memref.reinterpret_cast %base[%off] sizes=[1, %chunk, 1152]
memref.copy %chunk_src, %dst
```
**After:**
```mlir
%iter_i64 = memref.load %arg1[]
%iter_idx = arith.index_cast %iter_i64 : i64 to index
%off_e    = arith.muli %iter_idx, %chunk
%off      = arith.muli %off_e, %strides#1
%dst      = memref.reinterpret_cast %base[%off] sizes=[1, %chunk, 1152]
memref.copy %chunk_src, %dst
```

**Pipeline placement** (`Pipelines.cpp` step 5b): AFTER `BufferResultsToOutParams` (the in-place writer pattern only exists post-conversion) and the deallocation pipeline, BEFORE `OptimizeMemRefs` / `PoolAllocs` (so the new iter-driven arith chain is visible to pool-hoist analysis). A cross-reference comment is added at the `LoopLowering.cpp` aliasing line; the pass file header lists architecturally-cleaner alternatives (separate `v_out` buffer, side channel) for a future retirement of the patch.

## Scope / verification
This is the **compiler half** of a multi-part vision-accuracy fix. The full end-to-end `cos=1.0` result also needs:
- the `memrefCopy` strided-copy (D2D) fix (`lib/Runtime/real/hip.cpp`) — already in the runtime/projector slice of #259, and
- the dynamic-`Slice` empty-extent path (`SliceConversion.cpp`) — ships with dynamic-shape vision support.

So on `main` this pass is a **correct-but-dormant** transform until those land; that is also why it has no build dependency on any other split.

- Builds on `main`; registers as `--hip-fix-loop-accumulator-offset`.
- Full LIT suite green (the pass is a no-op on all existing in-tree IR; no model in-tree yet exercises the dynamic windowed-attention Loop).
- End-to-end regression when the full vision stack is present: `test/python/test_qwen3_5_9b.py::...::test_vision_grid_small_matches_cpu_cosine`.

## Files
- `lib/Dialect/Transforms/FixLoopAccumulatorOffset.cpp` — new pass.
- `include/hip/Dialect/Transforms/Passes.td` — `FixLoopAccumulatorOffsetPass` def (with Before/After IR snippet).
- `lib/Dialect/Transforms/Pipelines.cpp` — step 5b wiring.
- `lib/Dialect/Transforms/CMakeLists.txt` — new source.
- `lib/Conversion/HipToLLVM/LoopLowering.cpp` — cross-reference comment at the aliasing line.
- `CLAUDE.md` — gotcha (accurately scoped to the compiler half + cross-refs).


include/hip/Dialect/Transforms/Passes.td
```diff
@@ -386,6 +254,65 @@ def MaterializeHostScalarsPass
   ];
 }
 
+def FixLoopAccumulatorOffsetPass
```
fhanuman, line 257: I understand you didn't write this code. But, I don't understand why we need this pass in the first place. Let's start testing `qwen` without this change and see if we run into any issues. I will help fix if we run into any issues.

AMDmoore: @fhanuman sure. so, shall we keep this pr open or close it?

fhanuman (commented):
Hi @AMDmoore, As I mentioned in the Teams channel, please hold off on merging this change. I didn’t need this change to get QWEN model compiled -- we probably don't need this change. We will know more about it while debugging segmentation fault issue that I mentioned.

PR #316, opened 2026-06-09
feat(onnx-to-hip): fix dynamic-shape sizing for Expand/Tile/Range

AMDmoore:
## Summary

Split from #259. Three independent dynamic-shape **sizing** correctness fixes
for the shape ops, ported onto main's `InferTypeOpInterface` `create()` (no
explicit result-type churn). Each fix only affects how the DPS `tensor.empty`
init buffer for a dynamic output dim is sized.

| Op | Bug (from #259) | Fix |
|----|-----------------|-----|
| Tile | dynamic dim sized as `input.dim[i]` (via `createEmptyTensor`), under-sizing the output by the tile factor | size as `input.dim[i] * repeats[i]` |
| Expand | (1) `shape_value` used directly, shrinking a `>1` input dim under an identity-broadcast shape `[1,1,...]`; (2) `tensor.extract` on a shared `Shape` vector breaks pool-allocs dominance | (1) `max(input_dim, shape_value)` via `arith.maxsi`; (2) trace shape entries back to `tensor.dim`/const |
| Range | (1) only rank-0 scalars accepted; (2) `tensor.extract` on rank-1 size-1 operand breaks pool-allocs dominance | (1) accept rank-1 size-1; (2) trace scalar producers to pure SSA |

### Tile -- tile-factor under-size
The init alloc for a dynamic output dim must be `input.dim[i] * repeats[i]`,
not `input.dim[i]` alone. Under-sizing leaves the tiled axis at its pre-tile
extent; a downstream broadcast `Mul` then fails the MIOpen rank-broadcast
check (`BTensor dim != 1 && BTensor dim != CTensor dim`) and the EP propagates
partial data, collapsing per-model cosine. Fix sizes each dynamic dim as
`input.dim[i] * extract(repeats, i)`.

### Expand -- broadcast max + pool-allocs traceback
1. **Shrink bug.** ONNX `Expand` is broadcast, not `numpy.broadcast_to`:
   `output_dim = max(input_dim, shape_value)`. Using `shape_value` directly
   collapsed a `>1` input dim when the shape carries `1` at that axis
   (identity-broadcast `Expand(x, [1, 1, ...])`). Fix emits
   `arith.maxsi(tensor.dim(input, k), shape_value)` for opaque shape operands,
   with compile-time fast paths when the input dim is statically `1` or absent.
2. **pool-allocs dominance.** `tensor.extract` on a `Shape` vector shared by
   two `Expand`s bufferizes to `memref.load` from one scratch buffer, which
   `hip-pool-allocs` cannot hoist -> `operand does not dominate this use`.
   Fix traces each shape entry back to `tensor.dim(source, k)` or a constant
   (handles both `tensor.from_elements` and `onnx.Shape` producers), keeping
   the size chain pure SSA.

`ShapeConversion` benefit is bumped `1 -> 2` so `onnx.Shape` lowers before
`onnx.Expand`, giving Expand a `tensor.from_elements` vector to trace instead
of racing the rewrite into the `tensor.extract` fallback.

### Range -- rank-1 scalars + pool-allocs traceback
1. Accept rank-1 size-1 scalar operands (`tensor<1xi64>`), not just rank-0 --
   both are valid ONNX scalar encodings.
2. `tensor.extract` on a rank-1 size-1 operand bufferizes to `memref.load`,
   which `hip-pool-allocs` only hoists for pure SSA shape arithmetic ->
   dominance failure on the dynamic-size Range output alloc. Fix traces scalar
   producers (inline `onnx.Constant`, `tensor.from_elements`, rank-0 tensors,
   through reshape views) to a pure-SSA scalar before falling back to
   `tensor.extract`.

Carries two shared helpers into `OnnxToHipUtils.h`:
`getInlineScalarFromOnnxConstant` and `materializeScalarFromDenseAttr`.

## Test plan

- [x] New `test_expand_shape_trace.mlir` (traceback + opaque fallback).
- [x] `test_tile.mlir` -- assert `arith.muli` sizing on the dynamic case.
- [x] `test_expand.mlir` -- assert `arith.maxsi` on the opaque-shape case.
- [x] `test_range.mlir` -- rank-1 size-1 scalar + `from_elements` traceback.
- [x] Full `hip-mlir-opt` lit suite: 212/212 pass.
- [x] Full reconfigure + build + install clean.

### DIFF
--- a/lib/Conversion/OnnxToHip/ExpandConversion.cpp
+++ b/lib/Conversion/OnnxToHip/ExpandConversion.cpp
@@ -5,16 +5,137 @@
 
 #include "OnnxToHipUtils.h"
 
+// Lowers `onnx.Expand` to `tensor.empty` + `hip.expand`.  For dynamic result
+// dims that come from the 1-D `shape` operand, we avoid `tensor.extract` on a
+// `tensor.from_elements` shape vector when each element traces to
+// `tensor.dim(source, k)` or a compile-time constant.  That keeps the
+// bufferized size chain in the `memref.dim` / `arith.*` form `hip-pool-allocs`
+// can hoist, instead of `memref.load` from a shared shape scratch buffer
+// (dominance failure when two Expands reuse one Shape result).
+//
+// Symmetric to `GatherShapeFold.cpp`, which collapses Gather(Shape, const) ->
+// tensor.dim.  Here the idiom is Shape(x) -> from_elements -> Expand, and we
+// read dims directly from x for `tensor.empty` sizing.
+//
+// ONNX Expand broadcast semantics (v8/v13 spec):
+//   per-axis output_dim[k] = max(input_dim[k'], shape_value[s']).
+// Critically: ONNX Expand is NOT numpy.broadcast_to.  When shape[s] == 1 but
+// input has input_dim > 1 at the corresponding axis, the spec says the output
+// dim is input_dim (identity broadcast), NOT 1.  This is rare in LLM-style
+// exports (which build shape from `Shape(x)` so SSA-traced values already
+// equal input_dim), but some HF vision-encoder exports emit literal
+// `Expand(x, [1, 1, ...])` with x rank > 0 -- a pure no-op at runtime, but
+// only if we honour max(input_dim, shape_value).  When the shape operand is an
+// opaque constant we cannot peek into, we emit
+// `arith.maxsi(tensor.dim(input, k), index_cast(extract(shape, k)))` instead
+// of just `index_cast(extract(shape, k))`.
+//
+// Before (broken on `Expand([3136,1152]xf16, dense<[1,1]>) -> [?, 1152]`):
+//   %v = tensor.extract %shape[%c0] : tensor<2xi64>      // = 1
+//   %d = arith.index_cast %v : i64 to index              // = 1
+//   %t = tensor.empty(%d) : tensor<?x1152xf16>           // = tensor<1x1152>
+//
+// After:
+//   %sv = tensor.extract %shape[%c0] : tensor<2xi64>     // = 1
+//   %sd = arith.index_cast %sv : i64 to index            // = 1
+//   %id = tensor.dim %input, %c0 : tensor<3136x1152xf16> // = 3136
+//   %d  = arith.maxsi %id, %sd : index                   // = 3136
+//   %t  = tensor.empty(%d) : tensor<?x1152xf16>          // = tensor<3136x1152>
+//
+// Compile-time fast paths (skip the `tensor.dim` + `maxsi` op pair):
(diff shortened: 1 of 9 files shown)


lib/Conversion/OnnxToHip/ExpandConversion.cpp
```diff
@@ -5,16 +5,137 @@
...
+//                                shape_value.  Preserves IR shape for the
+//                                canonical attention-mask Expand idiom (input
+//                                `[1, 1, S]`, shape via `from_elements`).
+
 namespace mlir {
 namespace hip {
 namespace {
 
-/// onnx.Expand -> hip.expand
+/// If \p value is `arith.index_cast` of an index-typed SSA value, return the
```
AMDmoore, line 57: @fhanuman These changes seem to get the dim source, with the shape infer PRs, they are not needed any more, right? revert changes in this file?

lib/Conversion/OnnxToHip/RangeConversion.cpp
```diff
@@ -187,24 +187,114 @@ struct RangeToHip : public RewritePattern {
...
     if (failed(verifyConstantDeltaNonZero(op, op->getOperand(2), elemTy)))
       return failure();
 
-    Value startE = tensor::ExtractOp::create(rewriter, loc, op->getOperand(0),
-                                             ValueRange{});
-    Value limitE = tensor::ExtractOp::create(rewriter, loc, op->getOperand(1),
-                                             ValueRange{});
-    Value deltaE = tensor::ExtractOp::create(rewriter, loc, op->getOperand(2),
-                                             ValueRange{});
+    // Extract scalar values for Range start/limit/delta operands.
```
AMDmoore, line 209: @fhanuman as you say, memory pooling and output allocator PRs can support data dependent OPs, so changes for Range can be reverted, right?

lib/Conversion/OnnxToHip/ShapeConversion.cpp
```diff
@@ -69,9 +69,14 @@ namespace {
...
 // surfacing a "shouldn't happen" error.
+//
+// Benefit 2 (onnx.Expand is 1): onnx.Shape must lower before onnx.Expand so
+// ExpandConversion sees a `tensor.from_elements` shape vector it can trace
+// back to `tensor.dim` (pure SSA sizing), rather than racing the Expand
+// rewrite and falling back to `tensor.extract`.
 struct ShapeToTensorDims : public mlir::RewritePattern {
   ShapeToTensorDims(mlir::MLIRContext *ctx)
-      : RewritePattern("onnx.Shape", /*benefit=*/1, ctx) {}
+      : RewritePattern("onnx.Shape", /*benefit=*/2, ctx) {}
```
AMDmoore, line 79: change for expand op, so the same process way as expand changes.

fhanuman (commented):

PR #317, opened 2026-06-09
feat(slice): size dynamic Slice output to exact logical extent (#259 split)

AMDmoore:
## Summary

Part of the #259 split (op-by-op). This PR ports the **dynamic `Slice`
exact-extent output sizing** from #259 onto current `main`, scoped to the
self-contained, non-regressing subset.

`SliceToHip` previously sized every dynamic output dim to the **data-dim upper
bound** (`tensor.dim`). When a sliced axis has a **dynamic input dim with
constant indices**, the `hip.slice` kernel writes only the logical prefix and
zero-pads the over-allocated tail. Downstream reductions / matmuls then fold
that zero-padded tail into the result, and the output cosine collapses on
dynamic-shape vision encoders.

## What this PR does

`SliceToHip` now sizes each dynamic output dim to the **exact logical extent**
whenever the params fold:

- **(A) touched axis + static input dim** → compile-time `clampSlice`
  `IntegerAttr`.
- **(B) touched axis + dynamic input dim + constant starts/ends/steps** →
  host-side `index` arith on `tensor.dim` (`emitRuntimeExtent`: ONNX neg-index
  + clamp + ceildiv). Descriptor-only, **no GPU read**. This is the headline
  fix.
- **(C) untouched axis / unfoldable params** → data-dim upper bound (the legacy
  contract; Slice can never widen an axis).

**Empty-slice (extent == 0)** is guarded back to the upper bound
(`guardWithUpper`) rather than sized to 0, for two self-contained reasons: ONNX
Slice can never widen an axis (so `data.dim(axis)` is always a sound
over-allocation and the runtime honours the real extent via `output_shape[i]`),
and a true dim-0 buffer is rejected by MIOpen broadcast descriptors.

## Scope decisions (vs #259)

- **Kept `main`'s `SliceDecompose`** (benefit=2): const-index + all-static-input
  slices still lower to a zero-copy `tensor.extract_slice` — the LLM
  static-prefix fast path. No regression; existing lit Tests 1–3/6 still assert
  `extract_slice`.
- **Deferred** #259's **runtime-operand case** (per-token attention slice with
  RUNTIME `starts`/`ends` inside an outlined `hip.loop` body). It needs
  `followCaptureToParent` + a `hipdnn.fold_value` constant sidecar that is **not
  present on `main`**; it lands with the dynamic-vision loop infra.

## Runtime gating

The empty-input / `logical_extent` partial-fill behaviour in
`lib/Runtime/real/slice.cpp` ships in **PR-9/#308**. On bare `main` this
conversion compiles and lit-passes (pure ONNX→HIP rewrite); the empty /
partial-fill **e2e** is exercised once #308's runtime lands.

## Test plan

- [x] `test_slice.mlir` **Test 7** updated (upper-bound → exact-extent arith);
      new **Test 8** (multi-dim case B).
- [x] Conversion + Dialect lit suite **211/211** (`ctest -R Lit`).
- [x] Full reconfigure + build + install clean.
- [x] `pre-commit` clean (clang-format / lintrunner / license).
- [ ] e2e empty/partial-fill: gated on PR-9/#308 runtime (see above).

### DIFF
--- a/lib/Conversion/OnnxToHip/SliceConversion.cpp
+++ b/lib/Conversion/OnnxToHip/SliceConversion.cpp
@@ -10,6 +10,9 @@
 
 #include "llvm/ADT/APInt.h"
 
+#include <algorithm>
+#include <limits>
+
 namespace mlir {
 namespace hip {
 namespace {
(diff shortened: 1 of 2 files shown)


fhanuman (commented):

AMDmoore:

@fhanuman for accuracy bug fix as well.

fhanuman:
> 
> @fhanuman for accuracy bug fix as well.

Hi @AMDmoore, sounds good. Let's first test without this change and see what happens from the main branch if we have not already done?

PR #318, opened 2026-06-09
feat(onnx-to-hip): fold ConstantOfShape->Where to rank-0 linalg.fill scalar

AMDmoore:
﻿## Summary

Splits the **ConstantOfShape -> Where scalar fold** out of the #259 mega-PR as a self-contained change. Adds a third, highest-benefit lowering path for `onnx.ConstantOfShape` and wires the minimal linalg/scf lowering it requires.

### Problem (from #259)
On a multimodal prompt, the Qwen3.5 `embedding.onnx` mask-fill pattern

```
Shape -> ConstantOfShape(value=V) -> Where(mask, _, input_ids)
```

faulted on-device with **HIP 719 (unspecified launch failure)**. The dynamic `ConstantOfShape` path materialized a full-size fill buffer, and the natural rank-0 shortcut `arith.constant dense<v> : tensor<>` bufferizes to a **host** `memref.global "private" constant` (DLL `.data` section) -- a host VA the GPU where-kernel cannot dereference. The text-only path masked the bug (mask all-false -> kernel never reads `x[0]`); the moment any mask bit is true, `x[0]` is read on-device and the launch faults.

### Solution
`ConstantOfShapeAsScalar` (benefit=3): when **every** consumer is `onnx.Where` (which already broadcasts a scalar operand at stride 0), replace the op with a rank-0 fill-value buffer via `tensor.empty + linalg.fill`. This survives canonicalization (unlike `tensor.splat(const)`, which folds to the host global) and bufferizes to `memref.alloc + linalg.fill`. For rank-0 the fill is a single store (no loops); `hip-materialize-host-scalars` then redirects the small static alloc to host-mapped scratch (host-writable **and** GPU-readable at the same VA on UMA), so the where kernel reads a valid device address. The full-size buffer and the `splat -> linalg.map -> scf.for` chain are elided on every inference.

### Pipeline wiring (required prerequisite)
`linalg.fill` is the **first** linalg op this compiler emits, so this PR also wires its lowering:
- `convert-linalg-to-loops` at tail **step 5a** -- placed **before** `materialize-host-scalars` so the rank-0 fill becomes a `memref.store` that the host-scalar scan recognises. Running it *after* would leave a `linalg.fill` user, and the rank-0 buffer would land in the GPU pool with a host store on it -> device fault.
- `scf-to-cf` + `reconcile-unrealized-casts` after `lower-affine` in the HipToLLVM pipeline.

Both are no-ops on graphs that emit no linalg ops. Adds `MLIRLinalgDialect/Transforms` + `MLIRSCFToControlFlow` link deps and the `LinalgDialect` dependent-dialect on `ConvertOnnxToHipPass`.

### Not included
The dead `buildRank0ScalarAttr` helper from the #259 branch (the AsScalar pattern uses `buildScalarFillValue` + `linalg.fill`, never the dense-attr path) is intentionally dropped.

## Test plan
- [x] `test_constant_of_shape.mlir` Test 6: `ConstantOfShape` consumed only by `Where` -> `linalg.fill` (no `splat`/`dense`); Tests 4/5 (non-Where -> `tensor.splat`) and the static fold path unchanged.
- [x] Full **211-test** lit suite passes (`ctest -R MorphizenMLIRLitTests` + `llvm-lit test/lit`).
- [x] e2e `test_constant_of_shape_model.mlir` (static const, `onnx.Return` consumer) still folds to `arith.constant`.
- [x] pre-commit (lintrunner + clang-format + license) clean.
- [ ] **GPU e2e (gated):** the HIP-719 fix itself is only observable on a Qwen3.5 multimodal mask-fill run on a real device -- validated as part of the dynamic-vision stack, not reproducible from lit alone.


fhanuman (commented):

AMDmoore:

@fhanuman if your prs can make this op work, we can close this one.

fhanuman:
> 
> @fhanuman if your prs can make this op work, we can close this one.

Hi @AMDmoore, let's not close this PR yet, until we verify end-to-end for qwen and make sure numeric/accuracy is fine

PR #319, opened 2026-06-09
perf(gemm): autotune hipBLASLt algo + fused-bias epilogue (vision GEMM TTFT ~-40%)

BoarQing:
## Summary
- `wrap_gemm` took the hipBLASLt heuristic's top algo (`heurs[0]`), which is ~33x off optimal for the small-N vision-encoder GEMM shapes, and applied a broadcast `[N]` bias via a separate MIOpen op + `beta=1` in-place accumulate (`C==output`) that blocks fast split-K algorithms.
- Now: enumerate the full solution set (`hipblaslt_ext::getAllAlgos` + `matmulIsAlgoSupported`), benchmark the top candidates, cache the measured-fastest algo per `(shape, transpose, dtype, bias-epilogue)`.
- For a per-output-feature `[N]`/`[1,N]` bias with `beta==1`, use `HIPBLASLT_EPILOGUE_BIAS` (bias = per-row-of-D vector) with `beta=0`, dropping the separate broadcast op and the `C==output` in-place read.
- Falls back to the existing heuristic + broadcast path for unsupported dtypes / `[M,N]` C / `beta != 1`.

## Perf (controlled A/B, Qwen3.5-9B vision, dog image, 3 runs each, same machine)
| metric | baseline | this PR | delta |
|---|---|---|---|
| TTFT | 28.3 s | 16.4 s | ~-40% |
| Wall-clock throughput | 55.6 tps | 82.5 tps | +48% |
| Decode | 15.06 tps | 15.40 tps | flat (vision GEMM is prefill-only) |

TTFT distributions do not overlap (baseline 26.6-29.6 s vs this PR 16.1-16.6 s).

## Correctness
Generated text is bit-identical to baseline across runs.

## Test plan
- [ ] Build EP (gfx1151), clear `%TEMP%/morphizen_mlir_*`.
- [ ] `oga_mm_smoketest.py --image_path dog.jpg -g128 -m2048 -r5 -w2`; confirm TTFT drop + identical text.
- [ ] Spot-check a non-bias / `[M,N]`-C GEMM path still uses the fallback.

### DIFF
--- a/lib/Runtime/real/gemm.cpp
+++ b/lib/Runtime/real/gemm.cpp
@@ -8,9 +8,12 @@
 #include "error_check_macros.h"
 #include "runtime_types.h"
 
+#include <hipblaslt/hipblaslt-ext.hpp>
+
 #include <cstdio>
 #include <cstring>
 #include <functional>
+#include <mutex>
 #include <unordered_map>
 #include <vector>
 
@@ -58,9 +61,11 @@ static bool resolveGemmTypes(int64_t typeCode, hipDataType &dataType,
 
 struct GemmCacheKey {
   int64_t M, N, K, transA, transB, typeCode;
+  bool bias_epilogue; // distinct algo for the fused-bias-epilogue problem
   bool operator==(const GemmCacheKey &o) const {
     return M == o.M && N == o.N && K == o.K && transA == o.transA &&
-           transB == o.transB && typeCode == o.typeCode;
+           transB == o.transB && typeCode == o.typeCode &&
+           bias_epilogue == o.bias_epilogue;
   }
 };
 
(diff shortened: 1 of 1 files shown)


wcy123 (dismissed):
**LGTM** - Strong performance results (TTFT -42%, throughput +48%, text bit-identical).

**Readability issue:**

`wrap_gemm()` is now ~400 lines with multiple responsibilities and deep nesting:

```cpp
if (!have_cached) {
    if (getAllAlgos success) {
        for (each algo) {
            if (returned < max) {
                if (isAlgoSupported) {
                    // add to candidates
                }
            }
        }
    }
    if (!ok) {
        // try 4 different heuristics
    }
    if (ok) {
        if (returned > 1) {
            // benchmark loop (another 60 lines)
            for (int i = 0; i < returned; ++i) {
                if (workspace check) {
                    if (warmup success) {
                        if (timing success) {
                            // ...
                        }
                    }
                }
            }
        }
    }
}
```

Suggest extracting helpers in follow-up:
```cpp
static bool enumerateGemmAlgos(...);
static int benchmarkGemmAlgos(...);
static GemmCacheEntry selectGemmAlgo(...);
```

**Cache design note:**

Global `g_gemm_algo_cache` is semantically correct (algo selection is process-wide, not session-specific) but has limitations:
- Unbounded growth (no LRU/eviction)
- Global mutex (all sessions contend)
- Never cleaned up (lives for process lifetime)

This PR doesn't worsen these issues. Future work: consider per-session cache with bounded lifetime.

amd-mingw (approved):

PR #330, opened 2026-06-10
perf(memrefCopy): parallel strided-copy kernel for strided memref.copy

BoarQing:
## Summary
Standalone, main-targeted version of the strided `memref.copy` perf fix (originally #300, which is stacked on `pr259_with_perf`). This rebases just the feature onto `main`.

`memrefCopy` (the runtime helper for MLIR `memref.copy`, emitted by Concat/Slice/Tile/insert_slice bufferization) handled the non-contiguous, multi-outer-dim case with a host loop issuing one `hipMemcpyAsync` per row on the default stream -- ~2.07M tiny 72-byte launches per Qwen3.5 vision prefill, dominating GPU prefill time.

- Add `hip_strided_copy`: a single parallel strided D2D copy kernel (arbitrary rank/strides, element sizes 1/2/4/8 bytes). Host per-row loop kept as a correctness fallback.
- Route `memrefCopy` off the default stream onto the session stream (published per-Compute via a thread-local), removing default-stream serialization.

General fix (not model-specific); no new env vars.

## Perf (vision on MorphiZenEP)
- tower TTFT 18942 -> 6042 ms (-68%; vision prefill 15087 -> 2217 ms)
- dog   TTFT 106735 -> 39143 ms (-63%; vision 95448 -> 28542 ms)
- Generated text bit-identical (greedy decode unchanged).

## Tests
Adds `TestConcat` + `TestStridedCopyTier3` to `test/numeric/tests/test_shape_ops.py`, driving the kernel through last-axis Concat on rank >= 3 across 2/4/8-byte widths (incl. large outer extents exercising the grid clamp + grid-stride loop). All 14 pass EP-vs-CPU on gfx1151.

## Note
Same change as #300 but targeted at `main` directly. #300 additionally carries the `pr259_with_perf` base stack and base CI-cleanup that do not apply on `main`.

### DIFF
--- a/3rd-party/custom_kernels/CMakeLists.txt
+++ b/3rd-party/custom_kernels/CMakeLists.txt
@@ -58,6 +58,7 @@ hip_add_library(hip_custom_kernels STATIC
     hip/scatter_nd_kernel.hip
     hip/softmax_kernel.hip
     hip/nonzero_kernel.hip
+    hip/strided_copy_kernel.hip
     INCLUDE_DIRECTORIES
         ${CMAKE_CURRENT_SOURCE_DIR}/include
 )
--- a/3rd-party/custom_kernels/hip/strided_copy_kernel.hip
+++ b/3rd-party/custom_kernels/hip/strided_copy_kernel.hip
@@ -1,0 +1,126 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+// Parallel strided device-to-device copy for MLIR `memref.copy` (memrefCopy)
+// when neither side is fully contiguous and the copy spans >1 outer dim.
+//
+// The runtime's memrefCopy previously walked the outer index space on the host
+// and issued one hipMemcpyAsync PER contiguous row. For the Qwen3.5 vision
+// encoder that is ~2 million 72-byte launches per prefill (Reshape/Concat/Tile
+// chains over ~19k rows x ~108 copies) and dominated GPU prefill time. This
+// kernel does the whole strided copy in ONE launch: one element per thread,
+// with the contiguous inner suffix mapped to consecutive threads (coalesced),
+// and the outer multi-index decomposed per thread to apply src/dst strides.
+
+#include <hip/hip_runtime.h>
+#include <cstdint>
+#include <cstdio>
+
+#include "hip_custom_kernels.h"
+
+namespace {
+constexpr int SC_MAX_RANK = HIPDNN_MAX_MEMREF_RANK;
+
+struct StridedCopyParams {
+  int64_t outerSizes[SC_MAX_RANK];      // sizes of the outer (non-contiguous) dims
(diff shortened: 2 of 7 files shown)


wcy123 (approved):
**LGTM** - Strong performance results (Tower TTFT -68%, Dog TTFT -63%, text bit-identical).

**Why `memrefCopy` exists:**

MLIR's standard `memref.copy` lowering uses a two-tier pattern system:
- **Benefit 10 (high priority):** Our `MemRefCopyOpLowering` handles statically analyzable cases
  - Contiguous → `wrap_hipMemcpyAsync`
  - 2D pitched → `wrap_hipMemcpy2DAsync`
- **Benefit 1 (fallback):** MLIR's default lowering handles dynamic/complex strides
  - Generates call to `memrefCopy` runtime helper
  - Generated by: Concat/Slice/Tile bufferization (strided `memref.subview` copies)

**Architecture (no compiler changes):**
```
memref.copy op
    ↓ (MemRefCopyOpLowering - benefit 10, static cases)
    ├→ Static contiguous → wrap_hipMemcpyAsync
    ├→ Static 2D → wrap_hipMemcpy2DAsync
    └→ Fallback → MLIR default lowering (benefit 1)
                    ↓
                 memrefCopy runtime helper (THIS PR optimizes)
                    ├→ Tier 1: contiguous → hipMemcpyAsync
                    ├→ Tier 2: 2D → hipMemcpy2DAsync
                    ├→ Tier 3 (NEW): strided → hip_strided_copy kernel
                    └→ Tier 4: host per-row loop (fallback)
```

**Changes:**
1. **Parallel strided copy kernel** - template on elem_size (1/2/4/8), grid-stride loop for arbitrary size
2. **Session stream routing** - thread-local `g_hipdnn_ep_current_stream` to avoid default-stream serialization

**Architectural note:**

Thread-local global works for this PR, but there's a cleaner solution requiring minimal code change (~50 lines):

**Override MLIR's default lowering** with our own fallback pattern (benefit 1) that passes RuntimeState:

```cpp
struct MemRefCopyFallbackLowering : public ConvertOpToLLVMPattern<memref::CopyOp> {
  MemRefCopyFallbackLowering(const LLVMTypeConverter &converter)
      : ConvertOpToLLVMPattern(converter, PatternBenefit(1)) {}

  LogicalResult matchAndRewrite(...) {
    auto llvmFn = op->getParentOfType<LLVM::LLVMFuncOp>();
    Value statePtr = llvmFn.getArgument(0);  // ✅ All LLVM functions have RuntimeState* as arg 0
    call @memrefCopy(ptr %statePtr, i64 %elemSize, ptr %src, ptr %dst)
  }
};
```

Then update `memrefCopy` signature to accept `RuntimeState*` as first parameter and read stream directly from state instead of thread-local.

**Benefits:**
- Eliminates thread-local global
- Explicit state passing (no hidden dependencies)
- Consistent with existing `MemRefCopyOpLowering` (benefit 10) which already does this

**Recommendation:**
- **Accept PR as-is** (fixes real perf problem, thread-local is workable)
- **File follow-up issue** to add fallback lowering pattern with state parameter

PR #337, opened 2026-06-10
feat(hip): add hip-infer-loop-body-shapes pass to rank outlined loop bodies

fhanuman:
The MorphiZen importer emits unranked `tensor<*xT>` for values it can't infer, most often an outlined `hip.loop` body's loop-carried `onnx.Concat` output (e.g. a Concat accumulating along the sequence axis). `convert-onnx-to-hip`'s converters require ranked result types (`ConcatConversion` bails on unranked), so the op stays unconverted, the body `func.return` operand becomes type-incompatible with the func signature, and conversion/bufferization fail downstream. This blocks compiling models with data-dependent loops (hit on Qwen3.5-9B).

**Minimal example that demonstrates the issue (one outlined body func):**

```mlir
// `onnx.Concat` result is unranked, so ConcatConversion bails, the op stays
// unconverted, and the return operand no longer matches the func signature.
func.func private @loop_body(%ctx, %i, %c, %acc: tensor<1x?x1152xf16>, ...)
    -> tensor<*xf16> {
  %a = hip.multi_head_attention ... : tensor<?x?x?xf16>
  %r = "onnx.Concat"(%acc, %a) {axis = 1} :
         (tensor<1x?x1152xf16>, tensor<?x?x?xf16>) -> tensor<*xf16>  // unranked
  return %r : tensor<*xf16>
}
```

**The fix:**

A new pre-conversion pass `--hip-infer-loop-body-shapes`, run after `onnx-loop-outline` and before `convert-onnx-to-hip`. For every outlined `hip.loop` body func it:

* Seeds the loop-carried block args from the loop op's `$v_init` operand types (ranked post-outline).
* Forward-infers unranked `onnx.*` results from operand types (currently `onnx.Concat`), iterated to a fixed point.
* Loop-contract backstop: any loop-carried body output still unranked is set to the matching `$v_init` type (the ONNX Loop spec guarantees body-output type == loop-carried-input type).
* Reconciles the body func's declared result types from the now-ranked terminator operands.

Post-conversion `--hip-infer-shapes` is unchanged and still handles all `?`-dim narrowing on HIP-dialect ops; this pass only establishes *rank* so conversion can proceed.

### DIFF
--- a/docs/design/hip-shape-inference.md
+++ b/docs/design/hip-shape-inference.md
@@ -432,6 +432,43 @@ HIP DPS ops (GQA, RmsNorm, SkipRmsNorm, Attention, RotaryEmbedding,
 …) gain reify impls, the wiring lets each new op participate without
 any change to `Pipelines.cpp`.
 
+### Pre-conversion sibling: `--hip-infer-loop-body-shapes`
+
+`--hip-infer-shapes` only *narrows* `?` dims within a known rank on
+HIP-dialect ops; it cannot *establish* rank on a `tensor<*xT>`, and it
+runs after conversion. That leaves a gap for outlined `hip.loop` body
+functions: the importer emits `tensor<*xT>` for values whose shape it
+cannot infer (canonically a loop-carried `onnx.Concat` accumulating
+along the sequence axis), and `convert-onnx-to-hip`'s converters require
+ranked results (`ConcatConversion` bails on unranked). An unranked body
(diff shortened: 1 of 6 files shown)


fhanuman:
This workaround is needed for `QWEN3.5-9B`, so I will go ahead and merge the changes. A proper fix needs dependency on `onnx-mlir`. We need to come back to this, maybe after the ROCm EP 0.2 release

PR #338, opened 2026-06-10
fix(hip): expand arith.ceildivsi/floordivsi in HIP->LLVM lowering

fhanuman:
`ConvertHipToLLVM` hand-rolls its conversion pattern set and called only `arith::populateArithToLLVMConversionPatterns`, which has no direct lowering for `arith.ceildivsi` / `floordivsi` / `ceildivui`. A stray `ceildivsi` (e.g. from dynamic-shape index arithmetic in an outlined loop body) then survives `applyPartialConversion` and aborts MLIR→LLVM translation with `missing LLVMTranslationDialectInterface ... for op: arith.ceildivsi` — which silently drops the graph to CPU. Hit on Qwen3.5-9B once the loop body started converting (stacked on #337).

**Minimal example that demonstrates the issue:**

```mlir
// Without the ceil/floor-div expand patterns, this arith.ceildivsi has no
// ToLLVM lowering, survives the partial conversion, and aborts translation:
//   "missing LLVMTranslationDialectInterface ... for op: arith.ceildivsi"
func.func @ceildiv(%a: i64, %b: i64) -> i64 {
  %0 = arith.ceildivsi %a, %b : i64
  return %0 : i64
}
```

**The fix:**

Upstream's `-convert-arith-to-llvm` pass populates `arith::populateCeilFloorDivExpandOpsPatterns` ahead of the ToLLVM patterns — the two are a pair. Because we hand-roll the pattern set, we replicate that pairing:

* Add `arith::populateCeilFloorDivExpandOpsPatterns(patterns)` before `arith::populateArithToLLVMConversionPatterns(...)`, so `ceil/floor div` expand into primitive arith ops (cmp/sub/add/div/select) that lower in the same `applyPartialConversion`.
* Link `MLIRArithTransforms`.
* New LIT test asserts no `arith.ceildivsi` / `floordivsi` survives `--convert-hip-to-llvm`.

P.S. Also bumps `3rd-party/morphizen` to pick up #228 ("preserve unranked tensors at the ORT boundary"), part of the same Qwen3.5-9B unranked-tensor enablement. Stacked on #337; base will retarget to `main` once that merges.

### DIFF
--- a/lib/Conversion/HipToLLVM/CMakeLists.txt
+++ b/lib/Conversion/HipToLLVM/CMakeLists.txt
@@ -59,4 +59,5 @@ target_link_libraries(HipToLLVM PUBLIC
   MLIRLLVMCommonConversion
   MLIRMemRefDialect
   MLIRTransforms
+  MLIRArithTransforms
 )
--- a/lib/Conversion/HipToLLVM/HipToLLVM.cpp
+++ b/lib/Conversion/HipToLLVM/HipToLLVM.cpp
@@ -344,6 +344,15 @@ void ConvertHipToLLVMPass::runOnOperation() {
   // stages would require a reconcile-unrealized-casts cleanup pass.
   populateFuncToLLVMConversionPatterns(typeConverter, patterns);
   populateFinalizeMemRefToLLVMConversionPatterns(typeConverter, patterns);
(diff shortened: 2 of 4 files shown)


fhanuman:
Will go ahead and merge the changes to unblock QWEN integration, and also the changes are trivial.

PR #339, opened 2026-06-10
fix(hip): loop-body out-param ABI and duplicate elementwise_mul link symbol

fhanuman:
I hit two compile/link blockers on Qwen3.5-9B after #337/#338: outlined loop bodies stayed on a return-based memref ABI that `convert-hip-to-llvm` LoopLowering rejects, and `hip_elementwise_mul` was defined in two custom-kernel translation units.

**Loop body ABI mismatch:**

```mlir
// After one-shot-bufferize, private *_loop_body_* helpers still return
// loop-carried memrefs. LoopLowering expects v_in + v_out out-params.
func.func private @main_loop_body_n0(..., %v_in: memref<16xf32>) -> memref<16xf32> {
  %out = memref.alloc() : memref<16xf32>
  return %out : memref<16xf32>
}
// → convert-hip-to-llvm aborts: "body arg count mismatch: expected N, got M"
```

Module-level `buffer-results-to-out-params` promotes `@main_graph` in the classic pipeline but skips private helpers; the allocator pipeline skips module-level out-params on `@main_graph` entirely. Neither path promotes outlined loop bodies.

**The fix (loop bodies):**

* Add `--hip-loop-body-to-out-params`: run MLIR `promoteBufferResultsToOutParams` with a `*_loop_body_*` name filter after one-shot-bufferize and before buffer-deallocation (classic and allocator pipelines).
* LIT: [`test/lit/Pipeline/loop-body-out-params.mlir`](https://github.com/ROCm/onnx-hipdnn-ep/blob/fix/dup-mul-loop-body-outparams/test/lit/Pipeline/loop-body-out-params.mlir).

**Duplicate `hip_elementwise_mul` symbol:**

`elementwise_mul_kernel.hip` and `elementwise_binary_kernel.hip` both exported `hip_elementwise_mul`, causing a duplicate symbol at link time.

**The fix (custom kernels):**

* Remove the standalone `elementwise_mul_kernel.hip`; keep the implementation in `elementwise_binary_kernel.hip` only.
* Consolidate the public header docs for all binary elementwise ops under one section.


fhanuman:
Will go ahead and merge the changes to unblock qwen integration

PR #342, opened 2026-06-11
refactor(ep): remove classic DimSource path

zpye:
## Summary

Following review (@wcy123), this PR now **deletes the classic DimSource path entirely** and **removes the `use_output_allocator` provider option** — the output-allocator ABI is the only mode at the EP front-end. This is the "Update this PR to delete Classic path" outcome the review requested, rather than the original dual-path-with-TODO approach.

### What changed

- **DimSource mechanism removed.** The `DimSource` proto message and `Output.dim_sources` field are retired (field number 5 reserved in `metadata.proto`). `build_metadata_json` no longer emits per-output-dim resolution and no longer parses `dim_params_map`; it emits each output shape verbatim (static extent or `-1`).
- **EP consumer simplified.** `marshal_output_tensors` is now static-output-only: no DimSource resolution and no `present.*` shared-buffer override. The `present_to_past_input_idx_` plumbing + `build_present_to_past_input_idx` helper are gone.
- **`use_output_allocator` provider option removed.** `pass_main.cpp::load_config` hardwires `config.useOutputAllocator = true`; there is no string provider option. The classic out-param pipeline still exists in `lib/` (so `hip-compiler` CLI / LIT tests can exercise both ABIs) but is unreachable from a live EP session.
- The metadata `use_output_allocator` bit + the EP's dispatch-arity read are **retained** to keep the EP's dispatch in agreement with a loaded DLL's actual ABI (EPContext reuse / older-DLL guard). A follow-up can remove these once the classic codegen is dropped from `lib/`.

## Answers to the review's questions

> **1. Does the Classic approach work for any existing models?**

For dynamic-output models (every LLM / KV-cache graph): the classic path resolved `present.*` dynamic dims via DimSource→`attention_mask` plus a shape override. It "works" only for the 1:1-copy symbolic case; it **cannot** handle data-dependent outputs (VLM `NonZero`), and the symbolic-transformation cases (Concat `M1+M2`, Split, Reshape) were never implemented. So classic is strictly a subset.

> **2. If Classic works, does the output allocator ALSO work for those same models?**

Yes. The allocator path sizes every dynamic output in-graph from its producer operands. For `present.*` outputs the seq dim is `memref.dim %past_key` (the pre-bound buffer's actual extent = `max_length` under `past_present_share_buffer`), so `GetOutput` returns the shared `OrtValue` and `past == present` pointer identity is preserved **with no EP-side override** — the exact invariant the classic override existed to maintain. Pinned by `test/lit/e2e/test_gqa_output_allocator_present_dim.mlir`.

> **If the output allocator works for everything → delete Classic path, delete `use_output_allocator` parameter, one approach, no configuration.**

Done — that is this PR.

## Test plan

- [x] `pytest test/python/test_output_allocator_unresolved_dim.py` — passes (data-dependent output symbol on no input compiles cleanly in the allocator ABI).
- [x] EP rebuilt cleanly (proto regenerated); no compile fallout.
- [x] `pre-commit run --files ...` — all hooks pass.

Docs synced: `docs/design/output-allocator-design.md`, `docs/design/morphizen-ep-integration.md` (retired the `dim_params_map` Contract section), and `CLAUDE.md`.

## Notes

- The full classic-codegen removal in `lib/` (and the now-vestigial metadata bit) is intentionally deferred to a follow-up so this PR stays focused on the EP-front-end ABI collapse and the DimSource deletion.


wcy123 (changes requested):
**REQUEST CHANGES**

**Problem:**

This PR adds conditional path while keeping both approaches:
```cpp
if (useOutputAllocator) {
  // New approach (handles data-dependent outputs)
} else {
  // TODO: delete later  
  // Old approach (cannot handle VLMs)
}
```

**Comparison:**

| Model Type | Classic (Pre-allocation) | Output Allocator |
|-----------|-------------------------|------------------|
| **Static shapes only** (no symbolic dims) | ✅ Works | ✅ Works? |
| **Symbolic dims (1:1 copy)** | ✅ Works (DimSource maps input→output) | ✅ Works? |
| **Symbolic dims (transformations)** | ❌ Requires symbolic calculus (Concat: M1+M2, Split, Reshape, etc.) | ✅ Works? |
| **Dynamic shapes, symbolic dims missing** | ❌ Cannot resolve | ✅ Works? |
| **Data-dependent outputs** | ❌ Cannot work (CHECK fails) | ✅ Works |

**Key questions:**

1. **Does Classic approach work for any existing models?**
   - If NO → Update this PR to delete Classic path

2. **If Classic works, does output allocator ALSO work for those same models?**
   - If YES → Update this PR to delete Classic path (no reason to keep both)
   - If NO → Need migration plan (what's broken in output allocator for those models)

**If output allocator works for everything:**
- Delete Classic path entirely
- Delete `use_output_allocator` parameter (no option needed)
- One approach, no configuration

**Current PR has TODO with no plan:**

Line 271: `// TODO(output-allocator): delete this else branch once allocator mode is default-on`

Need concrete answers before merging dual-path maintenance.

**Suggestion:**

Update this PR to delete Classic path. Let validation results determine if we need it back.

wcy123 (commented):
let's discuss and decide later

wcy123:
Clarification: The questions above are for discussion, not blocking merge. Author should decide based on validation results.

zpye:
Thanks @wcy123 — took the suggestion: this PR now **deletes the Classic path entirely** instead of keeping the dual-path + TODO. Pushed as a rebase on latest `main`.

Concretely:
- Removed the `DimSource` proto message + `Output.dim_sources` field and all DimSource emission/resolution (producer in `build_metadata_json`, consumer in `marshal_output_tensors`, plus the `present.*` share-buffer override and its `present_to_past` plumbing).
- Removed the `use_output_allocator` provider option — `load_config` hardwires the allocator ABI. There is no configuration knob anymore; one approach.
- The classic out-param codegen still lives in `lib/` for the `hip-compiler` CLI / LIT tests, but it's unreachable from a live EP session.

Answers to your two key questions (also folded into the PR description):

1. **Does Classic work for any existing model?** Only the static + 1:1-copy symbolic cases. It can't do data-dependent outputs (VLM `NonZero`) and the symbolic-transformation cases (Concat `M1+M2`, Split, Reshape) were never implemented. Strict subset.
2. **Does the allocator also work for those same models?** Yes — it sizes every dynamic output in-graph; for `present.*` the seq dim comes from `memref.dim %past_key`, so `GetOutput` returns the shared buffer and `past == present` identity holds **with no EP-side override** (the exact thing the classic override existed for). Pinned by `test/lit/e2e/test_gqa_output_allocator_present_dim.mlir`.

So: allocator covers everything Classic did and more → Classic deleted, option deleted.

One scoping note: I deferred ripping the classic *codegen* out of `lib/` (and the now-vestigial metadata `use_output_allocator` bit + EP dispatch read) to a follow-up, to keep this PR focused on the EP-front-end ABI collapse + DimSource deletion. The metadata bit is retained for now because it guards EPContext reuse / older-DLL ABI matching. Happy to fold that in here instead if you'd prefer a single sweep.

wcy123:
## ✅ LGTM - Classic Path Deleted, Allocator Mode Default

**Clean migration:**
- Classic pre-allocation deleted ✅
- `use_output_allocator` hardcoded to `true` ✅  
- Single code path, no conditional logic ✅

**Architectural significance:**

This completes the transition to a fundamentally more capable output sizing approach:
- **Shape-derived outputs** (MatMul, Transpose): ✅ Works  
- **Data-dependent outputs** (NonZero, VLM features): ✅ Works (Classic could not handle)  
- **Symbolic calculus** (Concat, Split, Reshape): Future-ready (Classic required complex DimSource logic)

**Minor observation (non-blocking):**

`output_proto->add_shape` (line 267-270) is now dead code in allocator mode:
- EP provider never reads `metadata_.outputs().shape()` in allocator mode  
- `marshal_output_tensors()` (the only consumer) is unreachable after allocator early-return  
- Shapes computed in-graph, passed to EP via allocator callback

No action needed now (harmless metadata bloat), but could be cleaned up in follow-up if desired.

**Test coverage:** Realistic data-dependent graph (NonZero → Transpose) validates compile-time fix.

Thanks for the clean deletion of the Classic path - much better than maintaining dual approaches.

wcy123 (dismissed):
## ✅ LGTM - Output Allocator Hardcoded, DimSource Deleted

**Clean migration:**
- DimSource emission deleted ✅
- Output allocator now hardcoded throughout codebase ✅  
- Fixes data-dependent shape compilation (NonZero, VLM features) ✅

**Observation: Extensive dead code remains (non-blocking follow-up):**

Now that output allocator is the only path, large portions of the metadata system and runtime dispatch are dead code:

**1. Output metadata fields (proto):**
```protobuf
message Output {
  string name = 1;           // ✅ USED (output name mapping)
  int32 rank = 2;            // ❌ DEAD (allocator gets rank from DLL callback)
  int32 elem_type = 3;       // ❌ DEAD (allocator gets type from DLL callback)
  repeated int64 shape = 4;  // ❌ DEAD (allocator gets shape from DLL callback)
  // Field 5 retired (was dim_sources)
}
```
- Allocator mode: only `.name()` used (line 676 in MlirCustomOp.cpp)
- Everything else emitted but never consumed

**2. Compiler-side emission (pass_main.cpp):**
```cpp
// Lines 267-282: emit rank, elem_type, shape
output_proto->set_rank(...);      // Dead
output_proto->set_elem_type(...); // Dead
output_proto->add_shape(...);     // Dead (your observation)
```

**3. Compile flags (MlirCompiler.h, pass_main.cpp):**
```cpp
bool useOutputAllocator = true;  // Hardcoded, no variability
config.useOutputAllocator = true;  // Line 50, unconditional
metadata.set_use_output_allocator(true);  // Line 161, unconditional
```
- Could delete flag entirely, simplify to single path

**4. Runtime dispatch (MlirCustomOp.cpp):**
```cpp
bool use_output_allocator_ = false;  // Always set to true (line 611)
if (use_output_allocator_) {  // Line 713, always taken
  compute_with_output_allocator(context);
  return;
}
// Lines 721-800: Unreachable (classic path)
marshal_output_tensors(...);  // Unreachable function
```

**Recommendation for follow-up PR:**

**Delete:**
- `Output.rank`, `Output.elem_type`, `Output.shape` fields
- `Metadata.use_output_allocator` flag
- `CompilationConfig::useOutputAllocator` flag
- `MlirCustomOp::use_output_allocator_` member
- Classic dispatch path (lines 721-800 in Compute)
- `marshal_output_tensors()` function

**Keep:**
- `Output.name` (needed for index mapping)
- `Input` message (used for input marshaling)

**Result:** Metadata reduces to output names only. Code simplifies from dual-path dispatch to single allocator implementation.

---

**This PR is ready to merge** - completes the user-facing fix. Dead code cleanup is separate refactoring.

fhanuman (approved):

PR #345, opened 2026-06-11
fix(hip): register arith BufferDeallocationOpInterface external models

mingyueliuh:
## Summary
- Register the `arith` `BufferDeallocationOpInterface` external model in `registerAllDialects` (and in `hip-mlir-opt` for consistency).
- The ownership-based buffer-deallocation pass walks `arith` ops (e.g. `arith.select` on buffers) present in hybrid / MoE graphs (e.g. Qwen3.5-35B-A3B). Without this external model the pass aborts with `LLVM ERROR: ... interface BufferDeallocationOpInterface ... promised by dialect 'arith' but never implemented`.
- Simple graphs (e.g. single-op Gather) never exercised this path, so the gap was latent until a graph with such `arith` ops was compiled.

## Test plan
- [ ] `check-hip-mlir-lit` passes (194/194)
- [ ] `ctest -R ^E2E_` passes (56/56)
- [ ] Compiling a graph that contains `arith` ops requiring buffer deallocation no longer aborts in `OwnershipBasedBufferDeallocationPass`.

### DIFF
--- a/include/hip/InitAllPasses.h
+++ b/include/hip/InitAllPasses.h
@@ -11,6 +11,7 @@
 #include "hip/Dialect/Transforms/Passes.h"
 
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Arith/Transforms/BufferDeallocationOpInterfaceImpl.h"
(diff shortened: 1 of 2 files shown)


mingyueliuh:
Superseded: re-opening from the origin (ROCm) branch instead of a fork.

PR #348, opened 2026-06-11
feat(hip): slice NonZero output to its true extent via hip.readback_dim

fhanuman:
NonZero's output column count `N` (the number of non-zero elements) is data-dependent, so `convert-onnx-to-hip` over-allocates the indices buffer to the worst-case capacity `numel(X)`. The problem: that *capacity* — not the true count — was flowing downstream as the dynamic dim, so consumers and the ORT-reported output shape saw `numel(X)` columns and read undefined data past the real count `N`.

**Minimal example that demonstrates the issue:**

```mlir
// X: int64[A, B]  (A, B dynamic)
%nz = "onnx.NonZero"(%X)   : (tensor<?x?xi64>) -> tensor<2x?xi64>  // 2nd dim = #nonzero, data-dependent
%y  = "onnx.Sub"(%nz, %nz) : (tensor<2x?xi64>, tensor<2x?xi64>) -> tensor<2x?xi64>  // consumes the ? dim
```

The `?` dim of `%nz` is only known after the kernel scans `X`. With the indices buffer sized to `numel(X)`, the `Sub` (and the graph output) inherited `numel(X)` as its extent instead of the true count, exposing the over-allocated tail.

**The fix:**

* `hip.nonzero` gains a second DPS destination `count` (a device `i32` scalar) that the kernel writes with the true non-zero count `N`.
* A new generic `hip.readback_dim` op synchronizes the stream and reads that device scalar back to a host `index`.
* The conversion slices the over-allocated indices buffer to `[R, N]` (`tensor.extract_slice`), so the true extent — not the capacity — flows to consumers and the output allocator.
* `hip.readback_dim` is non-speculatable (declares a `Read` effect on the scalar): buffer-deallocation sees a *known* (non-allocating) effect, and `hip-hoist-alloc-size-arith` cannot lift the extent-sized allocation above the producing kernel.

`hip.readback_dim` is deliberately operator-agnostic: any op that computes a runtime extent into a device scalar can size a dynamic shape through it. This change wires it up for NonZero, but it is not coupled to NonZero's lowering or kernel.

**IR walkthrough (`convert-onnx-to-hip`):**

<details>
<summary><b>Before</b> — ONNX input; the <code>?</code> dim of the NonZero result is data-dependent</summary>

```mlir
func.func @main_graph(%arg0: !hip.context, %arg1: tensor<?x?xi64> {onnx.name = "X"}) -> (tensor<2x?xi64> {onnx.name = "y1"}) {
[...]

PR #349, opened 2026-06-11
fix(hip): pass scatter_nd count_ptr and materialize host scalars reached through views

fhanuman:
Two pre-existing bugs that only the VLM `NonZero -> ScatterND` embedding path exercises, so no prior model surfaced them. #348 added that path (data-dependent NonZero extent), which is what first drove `hip.scatter_nd` and the merged host-scalar shape buffer end-to-end. Both reproduce as a hard SIGSEGV on a text-only prompt (zero image tokens, so the NonZero count `N` is 0).

**Bug 1 - `hip.scatter_nd` lowers to the wrong runtime ABI.**

`wrap_scatter_nd` is a 16-parameter C function whose 6th parameter is `count_ptr` (a nullable device pointer to a dynamic valid-slice count). The HIP->LLVM lowering built the call with only 15 arguments, omitting `count_ptr`:

```mlir
// hip.scatter_nd(%ctx) ins(%data, %indices, %updates) outs(%out) {reduction = "none"}
//   lowered call (BEFORE):  state, data, indices, updates, out,
//                           data_shape, data_rank, idx_shape, idx_rank, ...   // 15 args
//   callee expects:         state, data, indices, updates, out,
//                           count_ptr, data_shape, data_rank, idx_shape, ...  // 16 params
```

Every argument after `out` is shifted by one slot, so the callee reads the `data_shape` *pointer* as `count_ptr`, the `data_rank` *integer* as the `data_shape` pointer, and then dereferences a small integer as a pointer -> SIGSEGV.

**The fix:**

* Pass the missing `count_ptr` so the 16-arg call matches the C declaration in `hipdnn_ep_runtime.h`.
* The op carries no runtime valid-count operand and its `indices` buffer is already truncated to the valid extent upstream (`readback_dim -> extract_slice -> transpose`), so we pass `null`; the kernel then processes all `num_slices` rows, which is correct because the row count already equals the count.
* The hip-to-llvm LIT test encoded the buggy 15-arg signature; it now asserts the 16-arg ABI.

**Bug 2 - `MaterializeHostScalars` misses host scalars reached through a view op.**

The pass redirects tiny host-fed scalar `memref.alloc`s out of the GPU pool into host-mapped scratch (otherwise a host `memref.store` lands in real device memory and SEGVs). Its candidate filter bailed at the first user it could not classify - and after bufferization + CSE fuse two `tensor.from_elements` shape buffers into one alloc and `reinterpret_cast` it for the second use, the host store reaches its hip consumer *through* the cast:

```mlir
%alloc = memref.alloc() : memref<3xi64>          // candidate
memref.store %d0, %alloc[%c0] : memref<3xi64>    // host store (direct)
hip.expand ins(%e, %alloc) ...                   // hip consumer (direct)
%rc = memref.reinterpret_cast %alloc to offset: [0], sizes: [1], strides: [1]
        : memref<3xi64> to memref<1xi64>
memref.store %n, %rc[%c0] : memref<1xi64>        // host store (via view) -- filter bailed here
hip.slice ins(..., %rc) ...                      // hip consumer (via view)
```

The lone `reinterpret_cast` user made the filter reject `%alloc`, so it stayed in the GPU pool and the host store SEGV'd.

**The fix:**

* The user classifier recurses through pure descriptor/view ops (`reinterpret_cast`, `expand_shape`, `collapse_shape`, `subview`, `view`, `cast`) - they touch no memory, so the backing buffer is still a host-staged scalar - and accepts the alloc when every transitive leaf user is host I/O, metadata, or a hip consumer.
* A view chain that escapes to any other consumer still rejects the alloc (conservative, unchanged behaviour for the GPU-pool case).
* New LIT cases cover both the `reinterpret_cast` redirect and the view-escape rejection.

With both fixes the VLM OGA benchmark (text-only prompt, `--prompt 128 --gen 64`) runs to completion and produces coherent output instead of SIGSEGV. Full build + LIT + lint clean.

P.S. `MaterializeHostScalars` is a post-bufferization heuristic (size + element type + host-I/O users). The principled, upstream-aligned design is to carry a host memory space on these buffers as a typed `memref` attribute (assigned at bufferization via `defaultMemorySpaceFn`) and have `hip-pool-allocs` skip non-device spaces (IREE preserves `getMemorySpace()` when packing). That is a cross-cutting change (conversion + bufferization callback + pool filter + lowering + verifiers) and is intentionally out of scope here; I'll track it separately.

### DIFF
--- a/lib/Conversion/HipToLLVM/ScatterNDLowering.cpp
+++ b/lib/Conversion/HipToLLVM/ScatterNDLowering.cpp
@@ -11,6 +11,7 @@ namespace {
 
 // hip.scatter_nd(ctx, data, indices, updates, output) {reduction}
 //   -> wrap_scatter_nd(state, data_ptr, indices_ptr, updates_ptr, out_ptr,
+//                      count_ptr,
 //                      data_shape_ptr, data_rank,
 //                      indices_shape_ptr, indices_rank,
 //                      updates_shape_ptr, updates_rank,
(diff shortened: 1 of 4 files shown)


fhanuman:
Will go ahead and merge the changes to unblock Qwen integration.

PR #357, opened 2026-06-12
fix(onnx-to-hip): synchronized GPU-to-host scalar readback for Range/Reshape/Expand/Loop

zpye:
﻿## Summary

This PR makes ONNX shape/scalar operators that read GPU-computed values back to the host (Range, Reshape, Expand, Loop) correct on AMD GPUs by replacing unsafe ONNX-level rewrites with safe, GPU-to-host **synchronized** scalar readback via a post-bufferize MLIR path (`hip.readback_scalar`).

### Changes
- **Synchronized host readback** for Range/Reshape scalars and Expand/Reshape shape entries.
- **New `hip.readback_scalar` op** (`HipOps.td`) with HIP->LLVM lowering (`ReadbackScalarLowering.cpp`) and runtime support (`memory.cpp`, mock stub, `hipdnn_ep_runtime.h`).
- **Loop accumulator offset fix** (`FixLoopAccumulatorOffset.cpp` pass + `Passes.td` + pipeline wiring): iteration-driven offsets for growing `Concat` accumulators inside `Loop` bodies, with LIT coverage.
- **Refactors:** consolidated synchronized-readback helpers (`ReadbackScalar.h`, `OnnxToHipUtils.h`), collapsed two readback bufferization models into one templated model (`HipBufferize.h`), removed the duplicate registration in `hip-mlir-opt.cpp`.
- **Submodule bump:** `3rd-party/morphizen` -> merged `main` (`4d2da1c6`, ROCm/MorphiZen#233 squash-merged), which corrects control-flow fusion output calculation by lifting region-capture / subgraph-yield consumers to the enclosing control-flow node.

## Test plan
- [x] LIT tests pass (`ctest -R MorphizenMLIRLitTests`)
- [x] Numeric EP-vs-CPU on affected ops (`test_shape_ops.py`: 55/55)
- [x] Qwen3.5-9B vision / embedding / text: EP-vs-CPU all PASS (worst cosine 0.9942 on text)
- [x] GPU dispatch confirmed (fp16-GPU-like 0.99x cosines, not exact 1.0)

Validated on the merged morphizen `main` (`4d2da1c6`).


wcy123 (commented):
# PR #357 Review: Request Changes

**PR:** https://github.com/ROCm/onnx-hipdnn-ep/pull/357
**Author:** Zupei (zpye)
**Title:** fix(onnx-to-hip): synchronized GPU-to-host scalar readback for Range/Reshape/Expand/Loop
**Changes:** +854/-303, 26 files
**Date:** 2026-06-13

---

## Summary

This PR adds `hip.readback_scalar` operation with synchronized D2H copy (`hipMemcpyAsync` + `hipStreamSynchronize`) to fix race conditions when host reads GPU-computed shape values.

**Recommendation: Request changes.** This PR papers over five architectural failures. Fix the architecture instead.

---

## Five Architectural Failures

### 1. Missing Shape Inference (MOST CRITICAL)

**The fundamental issue:** Compiler doesn't infer output shapes for data-independent operators.

**Example: onnx.Expand**
```python
input = tensor<1x1x256xf32>
target_shape = [2, 128, 256]
output = Expand(input, target_shape)
# Output shape = [2, 128, 256] - known from target_shape operand
```

**Current broken behavior:**
```mlir
%output = onnx.Expand(%input, %target) -> tensor<?x?x256xf32>
//                                         ↑ DYNAMIC - compiler doesn't infer
// Bufferization must read target_shape values → D2H readback
```

**Correct behavior with shape inference:**
```mlir
%output = onnx.Expand(%input, %target) -> tensor<2x128x256xf32>
//                                         ↑ STATIC - inferred from operand
// Bufferization allocates directly, no D2H needed
```

**All operators in this PR have data-independent shapes:**

| Operator | Output Shape | Inference Rule |
|----------|--------------|----------------|
| Expand | `target_shape` | Copy target_shape operand |
| Reshape | `target_shape` | Copy target_shape operand |
| Range | `[(limit-start)/delta]` | Arithmetic on operands |
| ReduceMax | Input shape with axes removed | Remove dimensions at axes |

**None require GPU execution or host reads to know output shape.** Compiler has all information.

**Data-dependent vs data-independent:**
- **Data-independent** (Expand, Reshape, Range, ReduceMax): Output shape from input shapes + attributes. Should infer statically.
- **Data-dependent** (NonZero, Unique): Output shape depends on actual data values. Cannot infer at compile time.

**Fix:** Implement shape inference pass. Produces static shapes for data-independent operators. Eliminates all D2H readback for these operators.

**Result:** This PR becomes unnecessary.

---

### 2. Pipeline Stalls from Synchronization

**The performance problem:** Every `hip.readback_scalar` calls `hipStreamSynchronize`.

**From memory.cpp:**
```cpp
void hipdnn_ep_readback_scalar(...) {
  hipMemcpyAsync(host_dst, device_src, nbytes, D2H, stream);
  hipStreamSynchronize(stream);  // BLOCKS ENTIRE PIPELINE
}
```

**What synchronization does:**
- Blocks host until ALL queued work on stream completes
- Destroys pipeline overlap:
  - Kernel execution with memory transfers
  - Multiple kernel executions
  - Host work with GPU work

**Performance impact:**

Without sync (ideal):
```
[Kernel1][Kernel2][Kernel3]
[  H2D  ][  H2D  ]
[   D2H   ][   D2H   ]
↑ Overlapped - maximum throughput
```

With sync after each kernel (this PR):
```
[Kernel1]--SYNC--[Kernel2]--SYNC--[Kernel3]--SYNC--
[  H2D  ]        [  H2D  ]
[   D2H   ]      [   D2H   ]
↑ Pipeline bubbles - serialized execution
```

**Concrete overhead:**
- Range operator: 3 sync calls (start, limit, delta)
- Every Reshape with dynamic shape: 1+ sync calls
- Every Expand reading shape: 1+ sync calls
- Model with 50 dynamic shapes: 50+ sync points

**Cost:**
- `hipStreamSynchronize`: ~1-10μs per call
- Pipeline serialization: GPU idle while waiting
- Cumulative: Severe degradation at scale

**PR #357 test results:** Reports numerical accuracy (cosine 0.9942), **does not report performance impact**. Likely significant slowdown not measured.

**GPU programming principle:** Minimize synchronization, maximize overlap. This PR does the opposite.

---

### 3. Shape Computation on GPU Instead of Host

**The question:** Why are shape values on GPU?

**Example from Qwen3.5:**
```python
grid_h = ReduceMax(image_grid_thw[1])  # GPU kernel → 16
grid_w = ReduceMax(image_grid_thw[2])  # GPU kernel → 16
num_patches = grid_h * grid_w          # GPU kernel → 256
# Now: num_patches in GPU memory, host needs it for allocation
# PR #357: hip.readback_scalar + sync to read 256
```

**Why this is wrong:**
- Shapes are small scalars (2-3 values)
- CPU is faster for small operations (no kernel launch overhead ~10-50μs)
- Host needs values for allocation anyway
- GPU kernel overhead >> CPU arithmetic cost

**Correct architecture:**
```python
grid_h = cpu_reduce_max(image_grid_thw[1])  # CPU reads → 16
num_patches = grid_h * grid_w               # CPU arithmetic → 256
# Host has value immediately, no D2H copy
```

**Why shapes belong on CPU:**
- Metadata, not bulk data
- Computed where used (on host)
- No kernel launch overhead
- No D2H transfer overhead

**Compiler should distinguish:**
- **Data tensors** (bulk computation → GPU): `image_pixels [batch, 3, 224, 224]`
- **Shape tensors** (small scalars → CPU): `grid_dims [height, width]`

**Fix:** Keep shape computation on host. CPU reads shape tensors, performs scalar arithmetic, provides values for allocation. Zero GPU kernels, zero D2H overhead.

---

### 4. Compile-Time Constants Treated as Runtime Values

**Same issue as PR #361 Pad fix.**

**Example: Range with constant bounds:**
```mlir
%start = onnx.Constant {value = dense<0> : tensor<i64>}
%limit = onnx.Constant {value = dense<256> : tensor<i64>}
%delta = onnx.Constant {value = dense<1> : tensor<i64>}

// Current broken:
// Constant externalization strips inline values
// RangeConversion can't fold, emits runtime readback
%range = onnx.Range(%start, %limit, %delta) -> tensor<?xi64>
// → hip.readback_scalar(start) + sync
// → hip.readback_scalar(limit) + sync
// → hip.readback_scalar(delta) + sync

// Correct:
// Compiler folds (256-0)/1 = 256
%range = onnx.Range(%start, %limit, %delta) -> tensor<256xi64>
// → Zero runtime overhead
```

**Pipeline ordering bug:**
1. Constant externalization: strips inline value, stores in blob
2. RangeConversion tries to fold: inline value gone, fold fails
3. Falls back to runtime: emits `hip.readback_scalar`

**Fix:** Converters read externalized constant blobs at compile time. Fold values during compilation.

**Note:** Issue #1 (shape inference) also solves this - if output shape inferred statically, no need to read operand values.

---

### 5. Address Space Confusion

**Same issue as PR #361.**

**The problem:** Type system doesn't enforce host vs device memory.

**Current broken architecture:**
```mlir
// GPU kernel computes shape:
%target = some.gpu.op : tensor<3xi64>

// Bufferizes to:
%target = memref.alloc() : memref<3xi64>  ← address space 0 (default)
                                          ← but allocates GPU memory (hipMalloc)

// Host reads:
%dim = memref.load %target[0]  ← HOST reads GPU memory (RACE!)
```

**Proper architecture:**
```mlir
// GPU produces device memory:
%target_device = memref.alloc() : memref<3xi64, 1>  ← address space 1 = GPU

// Explicit D2H transition:
%target_host = memref.copy %target_device : memref<3xi64, 1> to memref<3xi64, 0>
                                          ↑ type system enforces

// Host reads safely:
%dim = memref.load %target_host[0] : memref<3xi64, 0>  ← verified safe
```

**Benefits:**
- Static verification: type system catches host reading device memory
- Explicit transitions: IR shows where D2H happens
- Correct allocation: address space 0 → host memory, address space 1 → GPU memory

**Fix:** Use explicit address spaces. Fix `MemRefAllocOpLowering` to respect address spaces (file: `lib/Conversion/HipToLLVM/MemoryLowering.cpp:407-441`).

---

## Concrete Example: The Race Condition

**Scenario: onnx.Expand reading GPU-computed shape**

```python
data = tensor<1x1x256xf32>
target_shape = ReduceMax(...)  # GPU computes [2, 128, 256]
result = Expand(data, target_shape)
```

**Timeline (old buggy code):**
```
T0: GPU kernel writes [2, 128, 256] to device memory
T1: Host reads shape (no sync!)
    %dim = memref.load %target_device[0]  ← RACE
T2: Host gets stale values [0, 0, 0]  ← uninitialized memory
T3: Allocates result with shape [0, 0, 256]  ← WRONG
T4: GPU kernel completes (too late)
T5: Execution fails (0 elements instead of 65,536)
```

**Timeline (this PR - correct but slow):**
```
T0: GPU kernel writes [2, 128, 256]
T1: hip.readback_scalar called
T2: hipMemcpyAsync queued (D2H after kernel)
T3: hipStreamSynchronize BLOCKS host
T4: GPU kernel completes, D2H copy completes
T5: Host reads fresh values [2, 128, 256] ✓
T6: Allocates correctly
```

**Timeline (with shape inference - correct and fast):**
```
T0: Compiler infers output shape = [2, 128, 256]
T1: Bufferization allocates tensor<2x128x256xf32>
    No GPU kernel, no D2H, no sync
```

**On UMA systems:** Race is silent (host-accessible GPU memory), reads stale values, wrong results masked.

**On discrete GPU:** Immediate crash `0xC0000005` access violation.

---

## Why This Fix Is Wrong

The PR treats symptoms, not root causes:

✅ Correctly identifies race condition
✅ Correctly adds synchronization
❌ Does not fix shape inference (Issue #1)
❌ Does not fix shape computation placement (Issue #3)
❌ Does not fix constant folding (Issue #4)
❌ Does not fix address space confusion (Issue #5)
❌ Adds severe performance overhead (Issue #2)
❌ Every GPU→host read requires manual insertion
❌ Sets bad precedent for other code

**After this PR:**
- Address space confusion remains
- Compile-time constants still treated as runtime values
- Shape computation still on GPU
- Synchronizes per scalar read - destroys pipeline
- Potentially dozens of sync points per forward pass
- No performance measurement (only numerical accuracy tested)
- Violates GPU programming principle: minimize sync, maximize overlap

---

## Path Forward

**Fix the architecture properly (priority order):**

**1. Implement shape inference (MOST CRITICAL):**
- Propagate shapes through graph
- Infer output shapes from input shapes + attributes
- Produce static shapes for data-independent operators
- Result: Zero D2H, zero sync, this PR unnecessary

**2. Keep shape computation on host:**
- Distinguish data vs metadata tensors
- CPU reads shape tensors, performs scalar arithmetic
- Compiler analysis: "value needed for allocation → compute on host"
- Result: Zero GPU kernels for shapes

**3. Fix constant folding:**
- Read externalized constant blobs at compile time
- Fold during compilation
- Result: Zero runtime overhead for constants

**4. Fix address space handling:**
- Use explicit address spaces: `memref<T, 0>` (host) vs `memref<T, 1>` (GPU)
- Fix `MemRefAllocOpLowering` to respect address spaces
- Type system enforces GPU↔host boundaries

**5. If sync truly needed (rare after 1-4):**
- Batch multiple reads into one operation
- Sync at coarse granularity (graph boundaries)
- Measure and report performance impact

**Then this PR becomes unnecessary:**
- Shapes inferred statically or computed on host
- No D2H readback
- No sync overhead
- GPU pipeline overlap maintained

---

## Implementation Quality (Technical Correctness)

The implementation is technically correct as a workaround:

**`hip.readback_scalar` op design:**
- Proper traits: `DeclareOpInterfaceMethods<MemoryEffectsOpInterface>` (non-pure, has side effects)
- Type handling: preserves element type (i64/f32/f16), handles Index specially (i64 in ABI)
- Documentation: explains race condition rationale

**HIP→LLVM lowering:**
- Allocates host stack slot with proper alignment
- Uses `extractMemRefDataPtr` (offset-aware) for packed scalars
- Calls runtime helper: state, host_dst, device_src, nbytes

**Bufferization models:**
- `HipDstBufferizableModel<OpTy>`: templated DPS model
- `HipReadbackBufferizableModel<OpTy>`: correct semantics (bufferizesToMemoryRead=true, Write=false, no aliasing)

**Loop accumulator fix:**
- Surgical rewrite: only outlined loop bodies
- Finds gather producing `memref<1xi32>` from capture arg
- Fallback to `iter * chunk_size` if no gather

**Test coverage:**
- LIT tests: `test_reshape.mlir`, `hip-fix-loop-accumulator-offset.mlir`
- Numeric EP-vs-CPU: `test_shape_ops.py` (55/55)
- End-to-end: Qwen3.5-9B vision/embedding/text (cosine 0.9942)

**But it's solving the wrong problem.** Fix the type system and shape inference instead.

---

## MorphiZen#233 Dependency

PR depends on squash-merged MorphiZen#233 (submodule bump to 4d2da1c6). Verify this is merged to `main` before considering this PR.

fhanuman:
Hi @wcy123 ,
 
Thanks for helping review these PRs — much appreciated. One request: please don't post AI-generated reviews directly on PRs — see the No automated reviews section in our [contribution guidelines](https://github.com/ROCm/onnx-hipdnn-ep/blob/main/CONTRIBUTING.md). Using AI to understand the code is totally fine, but a posted AI review doesn't add much beyond what the author has likely already run themselves, and it tends to crowd out the thing we actually want from review — your own independent read on the architecture and the code. Independent thoughts and views are how we, as a group, improve the codebase.
 
If you spot something that looks under-reviewed, I would rather you nudge the author to do that pass than post an AI review in its place. And if any of the points in that review are ones you personally find compelling, I would be glad to hear them in your own words — that's exactly the kind of architectural discussion we want here.
 
Thanks again for pitching in here.

amd-mingw:
#318 merged, contain this PR context, close this PR.

PR #359, opened 2026-06-13
feat(ci): report OGA model_benchmark peak GPU memory

zz002:
﻿## Summary

Add a peak GPU memory column to the OGA model_benchmark CI results, sampled externally via Windows perf counters (no EP/inference changes).

## Why

CI had no GPU memory figure for the OGA benchmark, and TheRock's Windows distribution ships no `rocm-smi`/`amd-smi`. An earlier revision sampled inside the EP (DXGI in `MlirCustomOp`); per review that mixed diagnostics into the inference path and was fully reverted. Measurement now lives in CI: poll the `model_benchmark` PID's `\GPU Process Memory` counters. Host peak is already printed by `model_benchmark`, so only GPU is added; scope is the OGA suite only.

## What

1. **Sampling** in `Run OGA benchmark` of [.github/workflows/windows-build.yml](.github/workflows/windows-build.yml): start `model_benchmark.exe` via `Start-Process -PassThru`, poll `\GPU Process Memory(pid_*)\Dedicated Usage` + `\Shared Usage` for the peak, append as `[GPUMEM]` lines. `$proc.Handle` is cached so `ExitCode` is readable (else `Start-Process` may leave it `$null`, and `$null -ne 0` is `$true`, failing the step).
2. **Publish**: one `GPU Mem (GB)` column = dedicated + shared. The WDDM split drifts run-to-run on the UMA APU; the sum is stable, so report the total.
3. **No changes** to the EP, perf_test / EPContext / L2 tables, or dependencies (`Get-Counter` is built in).

## Test plan

- [x] Poller validated locally: captures PID, returns sane peaks.
- [x] OGA TTFT/TPS unchanged vs main within noise (Llama TPS 41.2 vs 41.6); external sampling does not perturb the benchmark.
- [x] Total GPU stable across runs (~13.8 GB gpt-oss, ~22.4 GB Llama) despite split drift.
- [ ] gpu-test on StrixHalo: OGA step green, `GPU Mem (GB)` populated.

## Notes for reviewers

PDH polled every ~1.5 s, peak only. PID polling requires `Start-Process` with separate stdout/stderr, so the EP's stderr lines group after stdout instead of interleaving (`2>&1`); metrics and parsing are unaffected.

- [ ] **Incremental** -- 1 file, ~50 LOC.
- [ ] **AI policy** -- Cursor-assisted; disclosed via commit `Co-authored-by` trailer.
- [ ] **Reviews resolved**
- [ ] **CODEOWNERS routing**

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -1024,11 +1024,28 @@ jobs:
           $sha = $rawSha.Substring(0, 7)
           $prNumber = "${{ github.event.pull_request.number }}"
 
+          # Peak GPU memory (MB) from the EP's [PEAKMEM] lines (emitted in bytes;
+          # one set per process run, multiple if a step appends several runs into
+          # one file). Returns Dedicated / Shared (max of each segment).
+          function Get-GpuMem($content) {
+            $d = [regex]::Matches($content, '\[PEAKMEM\]\s*Dedicated GPU memory:\s+(\d+)\s+bytes')
+            $s = [regex]::Matches($content, '\[PEAKMEM\]\s*Shared GPU memory:\s+(\d+)\s+bytes')
+            if ($d.Count -eq 0 -and $s.Count -eq 0) {
+              return [pscustomobject]@{ Dedicated = "-"; Shared = "-" }
+            }
+            $dMax = if ($d.Count) { ($d | ForEach-Object { [long]$_.Groups[1].Value } | Measure-Object -Maximum).Maximum } else { 0 }
+            $sMax = if ($s.Count) { ($s | ForEach-Object { [long]$_.Groups[1].Value } | Measure-Object -Maximum).Maximum } else { 0 }
+            return [pscustomobject]@{
+              Dedicated = [long]($dMax / 1MB)
+              Shared    = [long]($sMax / 1MB)
+            }
+          }
+
           # Helper: build table rows from a results dir.
           # Each file: extract Number of inferences per second / Session
-          # creation / 1st infer / Avg CPU / Peak WS for the standard
-          # 6-column format (Model | QPS | Session (s) | 1st Infer (ms) |
-          # CPU% | Mem (MB)).
+          # creation / 1st infer / Avg CPU / Peak WS / peak GPU mem (dedicated,
+          # shared) for the 8-column format (Model | QPS | Session (s) |
+          # 1st Infer (ms) | CPU% | Mem (MB) | GPU Ded (MB) | GPU Shared (MB)).
           function Get-PerfRows($dir) {
             $rows = ""
             foreach ($f in (Get-ChildItem "$dir/*.txt" -ErrorAction SilentlyContinue)) {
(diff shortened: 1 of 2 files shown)


wcy123 (commented):
# PR #359 Review: Request Changes

**PR:** https://github.com/ROCm/onnx-hipdnn-ep/pull/359
**Author:** Chi Zhang (zz002)
**Title:** feat(ep): report peak host + GPU memory per process

## Request Changes

This PR embeds diagnostic measurement code into the production inference path. This is wrong design.

### Problem 1: Separation of Concerns

Production inference code should do inference. Measurement and profiling is a separate concern that should be handled by external tooling.

This PR adds:
- 160 LOC of Windows-specific diagnostic code
- COM dependencies (dxgi.lib, psapi.lib)
- Global state (atomics, DXGI adapter cache, instance counting)
- DXGI API sampling on every `Compute()` call (unconditional, no disable)

**None of this is inference logic.** All of it exists solely to produce CI metrics.

### Problem 2: Maintenance Burden

The EP team now owns:
- Windows DXGI COM initialization and lifecycle management
- DXGI adapter enumeration and caching
- Driver compatibility across Windows versions and AMD driver versions
- Debugging instance counting, atomic operations, and memory ordering
- Testing diagnostic code across configurations

**This is not the EP's job.** This code delivers zero value to production users.

### Problem 3: Risk to Main Functionality

Even though this is "just diagnostics" with "silent degradation":

- **Global state on hot path**: Atomics updated on every `Compute()` call
- **COM lifecycle**: DXGI adapter initialization and release interleaved with inference
- **Complexity**: Windows-specific measurement infrastructure mixed with inference logic
- **Failure modes**: "Silent degradation" means bugs hide until metrics are wrong

The risk/benefit ratio is unacceptable. The benefit is CI metrics. The risk is production inference code carrying COM initialization, global state, and diagnostic sampling on the hot path.

### Problem 4: Wrong Architecture

The problem statement is: "CI needs GPU memory metrics and TheRock's Windows distribution ships no rocm-smi/amd-smi."

This is a CI infrastructure problem. The solution should be CI infrastructure (external measurement tooling), not modifying production inference code to measure itself.

**The PR description reveals the misguided thinking:**
> "Sampling lives in the EP itself (rather than a CI-side poller) so one change covers every suite"

This is presented as an advantage. It is wrong architecture.

### Recommendation

**Request changes.** Diagnostic requirements should not drive production code architecture.

GPU memory measurement should be done by external tooling (PowerShell scripts, Windows Performance Counters, or other measurement tools) that does not modify the production inference path.

wcy123 (approved):
## Approve
# PR #359 Follow-up Comment

## Minor Observation

**Sampling interval:** 1500ms polling may miss short GPU memory spikes in model loading/initialization phases. For production profiling, consider shorter intervals (100-500ms) or event-driven sampling.

**For CI metrics:** 1500ms is acceptable - captures steady-state memory usage during inference, which is the primary concern for regression detection.

PR #361, opened 2026-06-13
fix(vision): host-safe Pad externalized-constant reads + shape-staging scalars

AMDmoore:
## Summary

Stacked on top of PR #357 (`fix/range-reshape-readback-scalar`). Targets that branch as base; rebase onto `main` once #357 merges.

## Problem
The dynamic-shape vision encoder (Qwen3.5 rope-grid path) crashed with an access violation in `inference_compute` on true-device-memory targets (silently masked on UMA targets, where device memory happens to be host-accessible). Two distinct sites faulted the same way: the host directly read/wrote a buffer that physically lives in device memory.

## Root cause
1. **Pad externalized-constant size reads.** `onnx.Pad` with a dynamic output shape: `buildPadOutputInit` computes each padded dim as `data_dim + pads[i] + pads[i+N]`. It first tries to fold the `pads` vector from a `memref.global`'s inline value, but constant externalization strips the inline value from small shape constants, so the fold fails and the runtime path runs `tensor.extract %pads[idx]`. That bufferizes to an **unsynchronized host `memref.load`** of the device-resident constants blob -> access violation.
2. **Shape-staging scalar buffers.** `MaterializeHostScalars` rejected `memref.copy` and `memref.reshape` as users of a tiny host-staged scalar buffer. The bufferized Concat/insert_slice that stages grid-dim scalars into a shape vector uses `memref.copy`, and dynamic Reshape uses the scalar buffer as its shape operand. Rejected -> those scalar allocs stayed in the GPU pool, and the synchronous host `memref.store` into them SEGV'd.

## Solution
1. `PadConversion`: `extractAsIndex` / `buildPadOutputInit` now thread the `!hip.context` and read each pad entry via `readbackShapeEntryToHost` -- folds the value when the constant is still inline, otherwise emits `extract_slice -> collapse_shape -> hip.readback_scalar` (D2H + stream sync) so the host observes the real device bytes.
2. `MaterializeHostScalars`: accept `memref.copy` (terminal, host-mapping-safe stream op -- the runtime `memrefCopy` issues `hipMemcpyAsync`, which accepts host-mapped operands) and `memref.reshape` (recurse into the result when our value is the source; terminal accept when it is the shape operand). The shape-staging scalar buffers now stay out of the GPU pool, so the host store no longer faults.

## Both fixes are load-bearing (A/B isolation, identical valid inputs)
Verified on `Qwen3.5-9B-fp16-ve-...-dml/vision.onnx` via `hip-onnx-runner` (`image_grid_thw=[1,16,16]`, `num_patches=256`). The Pad node (`node_pad`, topo idx 406) has dynamic output `u9+1` and an externalized `[2]` const `pads`, so it hits the buggy path:

| Build (identical inputs) | Result |
|---|---|
| Both fixes | exit 0; `wrap_pad`/`hip_pad` dispatched; `image_features` produced |
| **Pad fix reverted** (host-scalar fix kept) | **`0xC0000005` in `inference_compute`** |
| (baseline) PR357-only | `0xC0000005` upstream at the ReduceMax/Range chain (host-scalar fix territory) |

Note: random inputs (the default) cannot exercise this -- a random `image_grid_thw` makes the upstream ReduceMax/Range shape arithmetic blow up before execution reaches the Pad at idx 406. Valid grid inputs are required to reach and prove the Pad path.

## Changes
- `lib/Conversion/OnnxToHip/PadConversion.cpp`
- `lib/Dialect/Transforms/MaterializeHostScalars.cpp`
- LIT: `test/lit/Conversion/onnx-to-hip/test_pad.mlir` (readback_scalar path), `test/lit/Dialect/hip-materialize-host-scalars.mlir` (`memref.copy` + `memref.reshape` cases)
- `CLAUDE.md`

## Test plan
- [x] `hip-mlir-opt` builds Release
- [x] LIT `test_pad.mlir` + `hip-materialize-host-scalars.mlir` pass
- [x] Vision encoder end-to-end via `hip-onnx-runner` with valid grid inputs: both fixes -> exit 0 + `image_features`; reverting the Pad fix -> `0xC0000005` at `inference_compute` (proves the Pad fix is required, not incidental)


wcy123 (changes requested):
# PR #361 Review: Host-Safe Pad Externalized-Constant Reads + Shape-Staging Scalars

**PR:** https://github.com/ROCm/onnx-hipdnn-ep/pull/361
**Author:** Shoucai (AMDmoore)
**Title:** fix(vision): host-safe Pad externalized-constant reads + shape-staging scalars
**Stacked on:** PR #357

## Summary

Fixes two load-bearing bugs in dynamic-shape vision encoder (Qwen3.5 rope-grid arithmetic):

1. **Pad externalized-constant size reads**: Unsynchronized host `memref.load` of device-resident constants → `0xC0000005` access violation
2. **Shape-staging scalar buffers**: Rejected `memref.copy` and `memref.reshape` uses → tiny scalars stay in GPU pool → SEGV on host store

Both fixes are correctly diagnosed, properly implemented, and well-tested.

## Concrete Examples

### Bug 1: Pad Externalized-Constant Crash

**ONNX operation in Qwen3.5 vision encoder:**
```python
output = Pad(input, pads=[0, 0, 1, 1, 0, 0, 1, 1])
#                          ↑ compile-time constant in ONNX model
```

**Root cause: Compiler pipeline ordering problem**

The constant `pads=[0,0,1,1,0,0,1,1]` is **compile-time known** and SHOULD be inlined. But the compiler pipeline runs in the wrong order:

1. **Constant externalization pass runs FIRST**:
   ```mlir
   // Before: inline constant (value present in IR)
   %pads = onnx.Constant {value = dense<[0,0,1,1,0,0,1,1]> : tensor<8xi64>}

   // After externalization: value stripped, only reference remains
   memref.global "private" constant @pads_blob : memref<8xi64>
   //                                            ↑ values moved to binary blob
   //                                            ↑ NO inline value in IR
   ```

2. **PadConversion tries to fold (too late)**:
   ```cpp
   // buildPadOutputInit tries to read pads[i] at compile time
   // But the inline value is gone → fold fails
   // Falls back to runtime path
   ```

2. **Host needs to compute output shape**:
   ```cpp
   // Running on CPU (host code):
   for (int i = 0; i < rank; i++) {
     output_shape[i] = input_shape[i] + pads[i] + pads[i + rank];
     //                                  ↑ NEEDS TO READ FROM GPU
   }
   ```

3. **Old path crashes**:
   ```mlir
   // Generates unsynchronized host read:
   %pad_left = tensor.extract %pads[%i] : tensor<8xi64>
   // ↓ bufferizes to:
   %pad_left = memref.load %pads_device_ptr[%i] : memref<8xi64>
   //           ↑ HOST CPU READS GPU MEMORY DIRECTLY → 0xC0000005
   ```

**Why:** On true-device-memory targets (non-UMA), host CPU cannot directly access GPU device memory → access violation.

**Fix applied: Runtime GPU→Host readback (WRONG ARCHITECTURE)**
```mlir
%pad_slice = tensor.extract_slice %pads[%i][1]
%pad_scalar = tensor.collapse_shape %pad_slice
%pad_left = hip.readback_scalar %ctx, %pad_scalar
//          ↑ Runtime GPU-to-host copy with stream synchronization
//          ↑ FOR A COMPILE-TIME CONSTANT
```

**This is wrong.** `pads=[0,0,1,1,0,0,1,1]` is a **compile-time constant** in the ONNX model. The compiler should fold it at compile time, not use runtime GPU-to-host readback.

**The correct fix:** PadConversion should read the constant from the externalized blob **at compile time** and fold it into the output shape computation. The constant values are right there in the blob - the compiler has all the information it needs.

**Current approach:** Treats compile-time constant as runtime value, pays synchronization overhead, adds complexity. This is a compiler bug being papered over with a workaround instead of fixing the compiler pipeline properly.

**Why this matters:** Every Pad with constant padding now pays runtime GPU-to-host copy + stream synchronization overhead for values that should be folded at compile time.

### Bug 2: Shape-Staging Scalar Buffer Crash

**ONNX operations:**
```python
# Compute grid dimensions on GPU:
grid_h = ReduceMax(image_grid_thw[1])  # GPU scalar: 16
grid_w = ReduceMax(image_grid_thw[2])  # GPU scalar: 16

# Build shape vector for Reshape (needs to run on HOST):
num_patches = grid_h * grid_w  # 256
shape = [batch, num_patches, embed_dim]  # HOST needs this array
```

**Root cause: memref address spaces not used properly**

The fundamental architectural issue is **address space confusion**:

1. **Hip ops use address space 1** (GPU global memory):
   ```mlir
   hip.alloc : memref<?x128xf32, 1>  // address space 1 = GPU
   ```

2. **Standard memref.alloc uses address space 0** (default, conventionally host):
   ```mlir
   %shape = memref.alloc() : memref<3xindex>  // address space 0
   %grid_h = memref.alloc() : memref<1xi32>   // address space 0
   //        ↑ BOTH use address space 0, but...
   ```

3. **Runtime allocation ignores address space**:
   - ALL `memref.alloc` ops lower to GPU device memory (`hipMalloc`)
   - Address space 0 does NOT mean host memory in this codebase
   - `MemRefAllocOpLowering` has a bug: calls `hipMalloc`, doesn't cast to memref's declared address space

4. **MaterializeHostScalars is a WORKAROUND**:
   - Uses runtime heuristic to detect "tiny scalar buffers with host I/O"
   - Routes them to `hip.get_host_scratch` (host-mapped memory)
   - This is patching the symptom, not using the type system properly

**What goes wrong in this specific case:**

1. `%shape = memref.alloc() : memref<3xindex>` → allocates in GPU pool (not host)
2. Host code tries: `memref.store %batch, %shape[0]` → writes to GPU memory → SEGV
3. MaterializeHostScalars should detect this needs host memory, but rejects it because:
   - `memref.copy` user wasn't recognized as "safe terminal use"
   - So `%shape` stays in GPU pool

**Fix applied: Patch the heuristic (still a workaround)**
```cpp
// Accept memref.copy as safe terminal use:
if (auto copy = dyn_cast<memref::CopyOp>(user)) {
  return true;  // Runtime hipMemcpyAsync handles host-mapped memory
}
```

**Result:** `%shape` routed to host-mapped memory → host store succeeds.

**This is STILL a workaround.** The proper fix would be:
- Use explicit address spaces: `memref<3xindex, 0>` for host, `memref<3xindex, 1>` for GPU
- Fix `MemRefAllocOpLowering` to respect address spaces (cast pointers properly)
- Let the type system statically verify host vs device access
- No runtime heuristics needed

**Why this matters:** Every new host-access pattern requires updating the MaterializeHostScalars heuristic. The type system should enforce this statically via address spaces.

### Test Evidence (from PR description)

| Build | Result |
|-------|--------|
| Both fixes | ✅ exit 0, `image_features` produced |
| Pad fix reverted | ❌ `0xC0000005` in `inference_compute` |
| Both reverted | ❌ `0xC0000005` upstream at ReduceMax chain |

**Proof:** Each fix addresses distinct crash. Both are load-bearing.

## Assessment: Request Changes - Split PR, Fix Pad Properly

### 1. Root Cause Diagnosis: Correct

**Pad externalized-constant issue:**
- **Diagnosis:** Precise and correct. Constant externalization strips inline `initial_value` from small shape constants. When `buildPadOutputInit` fold fails, the old path emits `tensor.extract %pads[idx]` which bufferizes to unsynchronized host `memref.load` of device-resident buffer → access violation on true-device-memory targets.
- **Evidence:** A/B isolation test proves necessity: reverting Pad fix causes `0xC0000005` in `inference_compute` on valid grid inputs.
- **Why it was masked:** On UMA with host-accessible pool, unsynchronized read accidentally works. On true-device-memory targets (gfx1150+), it reads stale/garbage.

**Shape-staging scalar buffer issue:**
- **Diagnosis:** Correct. `MaterializeHostScalars` previously rejected `memref.copy` and `memref.reshape`, causing tiny host-staged shape buffers to stay in GPU pool. The synchronous host `memref.store` then SEGVs.
- **Evidence:** Happens in vision encoder's `ReduceMax→Reshape→Gather→Range` chain where bufferized `Concat/insert_slice` emits `memref.copy` of tiny scalars.
- **Why the fix is correct:** `memref.copy` is terminal and host-mapping-safe (runtime `memrefCopy` → `hipMemcpyAsync` accepts both UMA and host-mapped). `memref.reshape` with buffer as shape operand is safe (reads shape, doesn't alias data).

### 2. Fixes Applied: Both Are Workarounds

**Fix 1: PadConversion - Runtime readback for compile-time constant (WRONG)**

**Change:** `extractAsIndex` and `buildPadOutputInit` now use `readbackShapeEntryToHost` (from PR #357's `ReadbackScalar.h`).

**Implementation:**
- Folds compile-time constants when inline (no sync needed)
- When constant is externalized: emits `extract_slice → collapse_shape → hip.readback_scalar` (runtime D2H + stream sync)

**Assessment:** This is **wrong architecture**. The constant values exist in the compiler at compile time (in the externalized blob). PadConversion should read them from the blob and fold them, not emit runtime GPU-to-host readback code. This workaround treats compile-time constants as runtime values, paying unnecessary synchronization overhead.

**Correct fix:** PadConversion should read externalized constant blobs at compile time and fold the values into the output shape computation.

**Fix 2: MaterializeHostScalars - Accept safe uses (WORKAROUND)**

**Changes:** `classifyHostScalarUsers` now accepts:
- `memref.copy`: Terminal, host-mapping-safe (runtime → `hipMemcpyAsync`)
- `memref.reshape`:
  - Buffer is source: recurse into aliasing result
  - Buffer is shape operand: terminal accept (reads shape, doesn't alias data)

**Assessment:** This is a **workaround for the address space problem**. It patches the runtime heuristic instead of fixing the type system.

**Why it's a workaround:**
- MaterializeHostScalars exists because address spaces aren't used properly
- Should be: `memref<3xindex, 0>` (host) vs `memref<3xindex, 1>` (GPU) - type system enforces
- Actually is: All `memref.alloc()` use address space 0 but allocate GPU memory
- Workaround: Runtime heuristic detects "tiny scalars with host I/O" and routes to host-mapped memory
- This fix: Updates heuristic to recognize `memref.copy` pattern

**The fundamental issue:**
- `MemRefAllocOpLowering` has a bug: calls `hipMalloc` (GPU memory) but doesn't cast to memref's declared address space
- Result: Address space 0 does NOT mean host memory in this codebase
- Every new host-access pattern requires updating the heuristic
- Type system should statically enforce this via address spaces


## Recommendation

**Request changes.** Do not approve. Both fixes are workarounds for deeper architectural issues.

### Why Not Approve

**Both fixes patch symptoms, not root causes:**

**1. Fix 1 (Pad) - Wrong architecture:**
- ❌ Treats compile-time constant as runtime value
- ❌ Pays runtime GPU-to-host copy + stream synchronization for compile-time constant
- ❌ Compiler has constant blob at compile time - should fold, not emit runtime readback
- ❌ Sets bad precedent - other operators will copy this pattern
- ❌ Technical debt stays forever once merged

**2. Fix 2 (MaterializeHostScalars) - Patches heuristic, not type system:**
- ❌ Workaround for address space confusion
- ❌ All `memref.alloc()` use address space 0 but allocate GPU memory
- ❌ Runtime heuristic detects host-access patterns instead of using type system
- ❌ Every new host-access pattern requires updating the heuristic
- ❌ Address spaces should statically enforce host vs device, not runtime detection

**3. Fundamental architectural issues:**

**Address space confusion:**
- Hip ops use address space 1: `hip.alloc : memref<?xf32, 1>` (GPU)
- Standard memref.alloc uses 0: `memref.alloc() : memref<?xf32>` (should be host, but allocates GPU)
- `MemRefAllocOpLowering` bug: calls `hipMalloc`, doesn't cast to memref's declared address space
- MaterializeHostScalars exists as workaround for this type system gap

**Compiler pipeline ordering:**
- Constant externalization strips inline values before PadConversion can fold
- PadConversion falls back to runtime readback instead of reading blob at compile time

### Path Forward

**Do not merge as-is. Fix the architecture properly:**

**1. Fix address space handling:**
- Use explicit address spaces consistently: `memref<T, 0>` (host) vs `memref<T, 1>` (GPU)
- Fix `MemRefAllocOpLowering` to respect address spaces (cast pointers properly, like `AllocOpLowering` does)
- Eliminate MaterializeHostScalars heuristic - let type system enforce statically
- File: `lib/Conversion/HipToLLVM/MemoryLowering.cpp:407-441` (bug location)
- Reference: `lib/Conversion/HipToLLVM/MemoryLowering.cpp:48-62` (correct pattern with address space cast)

**2. Fix constant folding:**
- Make PadConversion read externalized constant blobs at compile time
- Fold values during compilation, not at runtime
- No runtime GPU-to-host readback for compile-time constants

**3. Only then consider merging:**
- With proper address space enforcement
- With compile-time constant folding
- Zero runtime overhead, correct architectural precedents

### Acknowledgment

The author did excellent root-cause analysis (A/B isolation testing, understanding constant externalization, clear documentation). The diagnosis is correct. But the Pad fix takes the wrong approach - it works around a compiler bug instead of fixing it properly.

Fix it right, not fast.

AMDmoore:
> * nstant externalization strips inline values before PadConversion can fold

@wcy123 Reworked the pads handling to fold the constant at compile time instead of reading it back at runtime. The original PadConversion runs after lowerOnnxConstants, so by then the inline pads value has been externalized. To fix this properly, I added a pre-lowering pattern (PadShapeFold, alongside GatherShapeFold/ReshapeShapeFold) that runs before externalization while pads/axes are still inline constants, captures their values, and stamps them onto the onnx.Pad op as attributes (hipdnn.pad_amounts/hipdnn.pad_axes). PadConversion then folds the output shape directly from those attributes. Is this solution ok?

lib/Dialect/Transforms/MaterializeHostScalars.cpp
```diff
@@ -159,6 +159,31 @@ static bool classifyHostScalarUsers(Value memrefVal, bool &sawHostIO) {
...
+    // to/from another buffer is host-mapping-safe: the runtime memrefCopy
+    // lowering issues hipMemcpyAsync, which accepts host-mapped (UMA) operands
+    // just as it accepts pool ones. The canonical site is the bufferized
+    // Concat/insert_slice that stages grid-dim scalars into a shape vector
+    // (vision rope `ReduceMax -> Reshape -> Gather -> Range` arithmetic): each
+    // scalar slot is filled via `memref.copy <scalar-view>, <shape-subview>`.
+    // It does NOT by itself flag host I/O (memref.copy is a stream op, not a
+    // host store), so a buffer used ONLY by hip ops + memref.copy stays in the
+    // GPU pool where it belongs.
+    if (isa<memref::CopyOp>(user))
```
fhanuman, line 187: I have fixed this with a proper fix in PR#370, so we can revert this change.

amd-mingw:
#318 merged, contain this PR context, close this PR.

PR #363, opened 2026-06-13
perf(ProcyonV2): elementwise broadcast kernel + thin-row strided copy

amd-genmingz:
## Summary

Two low-risk MorphiZen-EP optimizations for the ProcyonV2 vision models:

1. **Elementwise custom broadcast kernel** (`c589cea1`) — route fp16/fp32
   Add/Mul/Min/Max away from `miopenOpTensor` (pathologically slow on gfx1151 for the
   common vision broadcast shapes) into a dedicated 4D broadcast kernel. This is the
   dominant win.
2. **`memref.copy` thin-row strided copy** (`931e2362`) — route thin-row pitched 2D copies
   to an element-indexed `wrap_strided_copy` kernel instead of many micro `hipMemcpy2D`
   transfers.

> Scope note: this PR was slimmed down in response to review. The **conv MIOpen Find cache**
> and the **slice/pad constant-index readback cache** were removed and will be handled in
> separate PRs (conv cache → rebased onto the PR #366 op-state-slot pattern; slice/pad → a
> compile-time constant fold in the converters, not a runtime cache).

## Performance (Release, `onnxruntime_perf_test -r 20`, gfx1151)

Latency in ms (lower is better). `main` vs this branch, measured back-to-back.

| Model        |    main | this PR | speedup |
|--------------|--------:|--------:|:-------:|
| blip_decoder |   23.74 |   21.68 |  1.09x  |
| blip_encoder |   73.51 |   36.15 | **2.03x** |
| convnext     |  137.83 |  ~flat  |  ~1.0x (noise) |
| detr         |  167.10 |   58.90 | **2.84x** |
| esrgan       |  270.18 |  264.96 |  1.02x  |
| sam_decoder  |   28.85 |   25.48 |  1.13x  |
| sam_encoder  |  105.26 |  ~flat  |  ~1.0x (noise) |

Headline wins (blip_encoder 2.03x, detr 2.84x) come from the elementwise kernel. convnext /
sam_encoder are within run-to-run noise (re-measured 4x each: branch and main overlap, and a
repeat branch run landed faster than main) — no regression.

## Accuracy

Combined cosine vs CPU fp32 reference; detr / esrgan / blip_decoder / sam_decoder use real
pre-recorded inputs, the rest random seed 42.

| Model | cosine | Model | cosine |
|---|--:|---|--:|
| blip_decoder | 0.99999903 | esrgan | 0.99999701 |
| blip_encoder | 0.99999936 | sam_decoder | 0.99999962 |
| convnext | 0.99999967 | sam_encoder | 0.99999975 |
| detr | 0.99999328 | | |

All 7 models cosine ≥ 0.99999 — no accuracy regression. (The absolute-L2 threshold "FAIL"
some models print is expected for large-magnitude outputs; cosine is the scale-invariant
signal.)

## Test plan

- [x] Build (Release) + install MorphiZen EP
- [x] `onnxruntime_perf_test` across all 7 ProcyonV2 models
- [x] Accuracy (cosine vs CPU fp32) on all 7 models — real inputs for the image models
- [x] `lintrunner -a`


mingyueliuh:
Dynamic Shape Acc CI Pass
    http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/729/

amd-genmingz:
hipMemcpy2DAsync(dst, dst_pitch=4B,
                 src, src_pitch=2B,
                 width=2B, height=40000,
                 hipMemcpyDeviceToDevice, stream);

we observed that when the underlying transfer path involves copying data from device to host, the execution time can become extremely high. It appears that the implementation performs some fixed-cost operation approximately height times, resulting in significant overhead for cases with a very small width and a very large height.

To optimize this scenario, we switch these degenerate copies to use strided_copy when the row width is small but the number of rows is large. Benchmark results show substantial performance improvements on mechine with small dedicated GPU memory

amd-mingw:
File three separate architectural issues:
1. Address space type safety - affects all memory operations
2. Compile-time constant folding - PadConversion/SliceConversion should read externalized blobs at compile time, not emit runtime readback
    Eliminates need for const_readback_cache (200 LOC runtime complexity)
    Zero first-inference overhead (constants folded during compilation)
   Correct layer: compiler fix, not runtime workaround
   See PR #361 review for detailed analysis
3.Process-global cache elimination - Conv Find cache uses pattern PR #366 eliminates
   Current: static unordered_map + static mutex, never freed, process-shared
   Correct: per-session op-state slot with WeakStore for process-shared autotune
   Memory leak + defeats per-session isolation

amd-mingw (approved):
split the above 1/2/3 related code from this PR, and will with follow-up PRs to address them.

PR #366, opened 2026-06-15
Op-state slots: per-instance compile-time slot assignment, generic teardown, and static-cache migration

BoarQing:
## Summary

Replaces per-operator `RuntimeState` fields and process-lifetime file-scope static caches with a unified **op-state slot** mechanism: each stateful operator *instance* is assigned a dense slot at compile time, constructed at `inference_init`, and torn down generically.

**Compiler**
- `OpStateOpInterface` + `--hip-assign-op-state-slots`: each stateful op instance gets its own dense slot (`hip.op_state_slot` per op; slot count recorded on the module). Per-instance, not per-class, so two ops of the same kind never collide.
- `--hip-generate-op-state-init` weaves each op's self-emitted `hipdnn_ep_op_state_construct_<class>(...)` call into `inference_init`; teardown walks `op_states` and calls each state's `deletor`, so order and types are irrelevant at cleanup.
- HIP-to-LLVM lowerings thread the slot index as the trailing `wrap_*` argument.

**Runtime**
- New `op_state.h`: `OpState` base, `op_state<T>(state, slot)` accessor, `make_op_state<T>()`, growable device/pinned buffers, and `WeakStore` for opt-in cross-session sharing. `RuntimeState` gains an `op_states` array.

**State now on slots**
- Per-session (one cache per session instance): `matmul_nbits` (incl. its zero-point unpack cache), `gqa`, `multi_head_attention`, `causal_conv_with_state`.
- Process-shared via device-keyed `WeakStore` (one table per device, shared across sessions, freed when the last session drops): `matmul` and `gemm` autotuned-algo tables, plus the MIOpen descriptor caches migrated off process-lifetime statics — `activation` (`sigmoid`/`tanh`/`softplus`), op-tensor (`mul`/`add`/`min`/`max`), `rms_norm`, and `skip_rms_norm`. This removes the never-freed `static` descriptor leak and the previously unlocked-map data race (the table's mutex now guards find/insert).

**Deliberately left as-is**
- `power`'s MIOpen cache is unreachable dead code (`neg` lowers to `wrap_neg`; reciprocal/sqrt bypass to HIP kernels), so it keeps no slot.
- `conv` and `qmoe` keep their transient scratch in per-session `RuntimeState` fields (`conv_scratch`, `qmoe_scratch`/`qmoe_host_scratch`): the correct sharing scope is per-session, and one buffer grown on demand uses far less VRAM than one per instance.

**Mock runtime**
- `mock/op_state_stubs.cpp` defines the `hipdnn_ep_op_state_construct_<class>` symbols the compiler emits, so stateful-op models link against the mock runtime.

**Docs**
- `docs/design/op-state-slots-design.md` updated with the current slot-using op list and the device-keyed `WeakStore` sharing framing.

## Test plan
- [x] LIT suite green (277 tests): `AssignOpStateSlots/assign_slots.mlir` (+ `no_stateful_ops.mlir`), `GenerateInterface/test_basic_interface.mlir`, and per-op `GenerateOpStateInit/gen_init_*.mlir` (matmul_nbits, activation, elementwise, rms_norm, skip_rms_norm, gemm) proving each op emits its construct call into `inference_init`.
- [x] Numeric correctness on real GPU (102 tests): activation (sigmoid/tanh/softplus), binary + unary elementwise (mul/add/min/max/...), layer_norm (rms_norm + skip), matmul, multi_head_attention (gemm path).
- [x] `test/numeric/tests/test_concurrent_sessions.py` (5 tests): two concurrent sessions prove per-session cache isolation (activation, distinct shapes) and shared-autotune thread-safety (matmul).
- Note: some GQA dynamic-shape configs fail at session-init shape inference; this is pre-existing in the unchanged EP DLL and unrelated to this change.


wcy123:
# PR #366 Review Comment

**Date:** 2026-06-15

---

## Review Comments

**Summary:** This PR introduces compile-time slot assignment for per-session operator state, eliminating RuntimeState field bloat, static cache leaks, and C-ABI destroy shims. The implementation is correct and well-tested.

### Strengths

- Zero runtime overhead (compile-time slot assignment)
- Generic teardown via function pointers (eliminates per-class destroy shims)
- Per-session isolation validated by concurrent sessions test
- Implementation follows established patterns (mirrors pool allocation planning)
- Good design documentation

### Architectural Limitations

The current design has two fundamental limitations:

#### 1. Cannot Support Per-Instance State

All operators of a class share one state object per session:

```mlir
%h1 = hip.rnn(%x, ...) {hip.op_state_slot = 0}   // Layer 1
%h2 = hip.rnn(%h1, ...) {hip.op_state_slot = 0}  // Layer 2 - SAME slot!
%h3 = hip.rnn(%h2, ...) {hip.op_state_slot = 0}  // Layer 3 - SAME slot!
```

All three RNN layers access the same `RnnState` object, corrupting each other's hidden state. No clean workaround exists.

#### 2. Cannot Support Parameterized Construction

Constructor table enforces zero-parameter signature:
```cpp
OpState *hipdnn_ep_op_state_construct_<class>(RuntimeState *state);
```

This prevents:
- Multi-GPU device selection
- Buffer size hints
- Per-slot configuration
- Any construction-time parameters

Workarounds (RuntimeState pollution, environment variables, hardcoded values, process globals) all defeat the design goals.

### Design Suggestion: Code Generation Method

**Current design has unnecessary indirection:**
```
Op → getOpStateClass() → "activation" → hardcoded transform → constructor table → zero-param constructor
```

**Suggested redesign:**

Replace `getOpStateClass()` and the constructor table with a **code generation method** on the interface:

```cpp
def OpStateOpInterface {
  let methods = [
    InterfaceMethod<"Generate state initialization",
                    "::mlir::Value",
                    "generateOpStateInit",
                    (ins "::mlir::OpBuilder &":$builder,
                         "::mlir::Location":$loc,
                         "::mlir::Value":$statePtr,
                         "int32_t":$slot)>
  ];
}
```

**Op implementation (with parameters):**

```cpp
def GemmOp {
  let arguments = (ins ..., OptionalAttr<I32Attr>:$device_id);

  let extraClassDeclaration = [{
    Value generateOpStateInit(OpBuilder &builder, Location loc,
                              Value statePtr, int32_t slot) {
      // Read op attribute
      int32_t device = getDeviceId().value_or(0);

      // Declare parameterized constructor
      auto ctorFunc = declareFunc("hipdnn_ep_op_state_construct_gemm",
                                  ptrType, {ptrType, i32Type});

      // Generate call with parameter
      Value deviceVal = builder.create<LLVM::ConstantOp>(loc, i32Type, device);
      return builder.create<LLVM::CallOp>(loc, ctorFunc,
                                          ValueRange{statePtr, deviceVal})
                    .getResult();
    }
  }];
}
```

**Runtime constructor accepts parameters:**
```cpp
OpState *hipdnn_ep_op_state_construct_gemm(RuntimeState *state, int32_t device_id) {
  // Use parameters for construction
}
```

**Benefits:**
- ✅ Solves per-instance state limitation (each op instance has its own attributes)
- ✅ Solves constructor parameter limitation (generated code can pass any parameters)
- ✅ Removes unnecessary "class" abstraction
- ✅ Eliminates hardcoded naming convention
- ✅ Grouping still works (ops that generate identical code share slots)
- ✅ Full flexibility for each op

**This addresses both fundamental limitations.**

### Questions

1. **Are stateful ops planned?** (RNN, stateful convolution) If yes, current design cannot support them.

2. **Is multi-GPU support planned?** (per-slot device selection) If yes, parameterized construction is needed.

3. **Should the design be revised now** to support these use cases, or document the limitations for future consideration?

### Minor Suggestions

1. **Simplify optional op-state check** - `AssignOpStateSlots` is mandatory in pipeline, attribute always exists
2. **Add debug logging for construct failures** - helps diagnose OOM at session init
3. **Document limitations in design doc** - add "Architectural Limitations" section

lib/Dialect/Transforms/AssignOpStateSlots.cpp
```diff
@@ -0,0 +1,82 @@
...
+  void runOnOperation() override {
+    ModuleOp module = getOperation();
+    MLIRContext *ctx = &getContext();
+
+    // class string -> slot index, kept in first-appearance order.
+    llvm::MapVector<StringRef, int32_t> classToSlot;
+
+    module.walk([&](OpStateOpInterface op) {
+      StringRef cls = op.getOpStateClass();
+      auto it = classToSlot.find(cls);
```
zpye: Consider creating an op state for each op instance instead of creating one for each op class. For ops of the same class, the shared parts can share the same op class state across multiple op states.

BoarQing:
This PR does not consider multi-threading run() on a single session. This would lead to race condition for sure!

amd-mingw:
http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/854/

amd-mingw (dismissed):

amd-mingw:
http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/864/

amd-mingw (approved):

BoarQing:
http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/871/

PR #370, opened 2026-06-15
feat(hip): infer static shapes at the dialect level to run dynamic-shape vision encoders

fhanuman:
Vision-language models with dynamic input shapes (the image projector and the embedding-norm chain after it) couldn't run on GPU. The ONNX exporter drops the static sizes of tensors whose shapes are actually fixed, so `ReduceMean` / `AveragePool` reached our lowering looking fully dynamic and failed to compile.

My first fix did this at the ONNX level (`RefineReshapeResultType`), re-deriving the sizes by hand — fragile, since we don't depend on onnx-mlir so it matched ops by name and reimplemented onnx-mlir's reshape math. Recovering static shapes is the dialect layer's job, so this PR removes that pass and does the work there, plus the runtime fixes the path needs. Related readback work: #348.

**What changed:**

* Recover the static sizes in the HIP dialect: ops report their result shape through the standard MLIR reify interface, and `--hip-infer-shapes` + canonicalize/CSE rediscover the sizes that can be proven.
* New first-class `hip.reduce_mean` (was a sum + divide). This is the op that carries the inference above; doing the divide in-kernel also means it no longer needs a static reduce axis — that tolerance is what let me delete the ONNX pass.
* Pooling no longer bails on a dynamic spatial output. Different mechanism — not static inference, it computes the size at runtime from ONNX's output-size formula (the window params are constants and fold away). Honors `ceil_mode`. (The projector's own average-pool rides the `ReduceMean` path instead.)
* Runtime fixes the path needs: the i32 scalar readback uses `hipMemcpyDefault` (hard-coded device-to-host failed `invalid argument` on host-visible memory and zeroed the size); each `inference_compute` re-publishes its own GPU stream on entry (sessions sharing one model.dll otherwise race on a shared thread-local stream — garbled output, only when several are live); a no-op transpose is treated as a no-op; the pool-grow path is guarded against a zero-size request.
* `MaterializeHostScalars` detects buffer views via MLIR's `ViewLikeOpInterface` instead of a hand-maintained list of op types.

<details>
<summary><b>Before/after IR — dynamic <code>ReduceMean</code> and <code>AveragePool</code></b></summary>

```mlir
// ReduceMean (older opset, axis as an attribute)
//
// Before — axis is an attribute and the result size is left dynamic:
%y = "onnx.ReduceMean"(%x) {axes = [1], keepdims = 1}
       : (tensor<?x4096xf16>) -> tensor<?x1xf16>

// After — axis is now an operand, and because the divide is done in the
// kernel there is no need for a static reduce size. --hip-infer-shapes +
// canonicalize then carry %x's leading size through to the result.
%axes = arith.constant dense<[1]> : tensor<1xi64>
%init = tensor.empty(%n) : tensor<?x1xf16>
%y    = hip.reduce_mean(%ctx) ins(%x, %axes) outs(%init) {keepdims = 1}
          : tensor<?x1xf16>

// AveragePool (kernel = stride = 4, no padding)
//
// Before — fully dynamic spatial size; PoolToHip used to bail out and drop
// the graph to CPU:
%y = "onnx.AveragePool"(%x) {kernel_shape = [4,4], strides = [4,4]}
       : (tensor<1x1152x?x?xf16>) -> tensor<1x1152x?x?xf16>

// After — output spatial size computed from the input size at runtime;
// kernel/stride/padding already folded to constants:
%h    = tensor.dim %x, %c2 : tensor<1x1152x?x?xf16>
%t    = arith.subi %h, %c3 : index            // h - ((K-1)*d + 1)
%out  = arith.floordivsi %t, %c4 : index      // floor(t / stride) + 1
%init = tensor.empty(%out, %out) : tensor<1x1152x?x?xf16>
%y    = hip.pool(%ctx) ins(%x) outs(%init) {pool_type = 0 : i64, ...}
```

</details>

Tests: new lowering tests for `reduce_mean`, a shape-inference test, a dynamic `AveragePool` test, and a numeric fp16 `ReduceMean` test. All 257 LIT tests pass.

P.S. The general-purpose `hip.readback_scalar` op (the wider dynamic-reshape readback path) is a separate change and isn't on this branch; this PR only fixes the copy direction in the existing i32 readback.


fhanuman:
Will merge the changes to unlock VLM model integration so we can meet the ROCm EP 0.2 release time. I have a planned codebase cleanup, will use that opportunity get code reviewed then

PR #374, opened 2026-06-17
feat: add GreaterOrEqual, LessOrEqual, and Tanh ops

amd-genmingz:
## Summary
Adds three ONNX ops to the ONNX→HIP backend so models that use them stop
falling back to (the disabled) CPU EP:

- **GreaterOrEqual** and **LessOrEqual** (comparison)
- **Tanh** (activation)

### GreaterOrEqual / LessOrEqual
Implemented as rewrite-based decompositions on top of the existing `hip.less`
and `hip.not` ops — no new `hip.*` op, lowering, or runtime kernel needed:

- `GreaterOrEqual(A, B)` → `hip.not(hip.less(A, B))`   (`A >= B  <=>  !(A < B)`)
- `LessOrEqual(A, B)`    → `hip.not(hip.less(B, A))`   (`A <= B  <=>  !(B < A)`)

The decomposition builds the `hip.less` / `hip.not` ops **directly** rather than
emitting `onnx.Less` / `onnx.Not`. `convert-onnx-to-hip` runs its greedy rewrite
with `GreedyRewriteStrictness::ExistingOps`, so freshly created `onnx.*` ops are
never revisited by their own patterns — emitting them would leave the graph
un-bufferizable (`"op was not bufferized"`) and silently drop it to CPU.

### Tanh
New `hip.tanh` DPS op (mirrors `hip.sigmoid`), lowered to
`wrap_miopenActivationForward` with `miopenActivationTANH`. This unblocks any
model with a standalone `Tanh` outside the Gelu(tanh) fusion — e.g. Gemma-2
lm_head logit soft-capping (`cap * tanh(logits / cap)`), which previously
survived `convert-onnx-to-hip` unconverted and aborted one-shot-bufferize.

- `HipOps.td`: new `hip.tanh` op
- `HipDialect.cpp` / `HipBufferize.h`: DPS inits, effects, bufferization iface
- `ActivationConversion.cpp`: `onnx.Tanh` → `hip.tanh`
- `ActivationLowering.cpp`: `hip.tanh` → `wrap_miopenActivationForward(TANH)`
- Runtime `activation.cpp`: set `alpha = beta = 1` for `miopenActivationTANH`.
  MIOpen TANH computes `alpha * tanh(beta * x)`, so the previous `0,0,0` default
  degenerated to `0 * tanh(0) = 0` (LOGISTIC/SOFTRELU ignore alpha/beta). The
  descriptor cache is keyed by `activation_mode`, so the mode-derived alpha/beta
  never alias across modes.

The Gelu(tanh) fusion runs in the pre-lowering phase before `convertComputeOps`,
so the standalone `Tanh` pattern only claims the leftover `Tanh` and does not
interfere with Gelu fusion.

## Known limitations

**NaN semantics of GreaterOrEqual / LessOrEqual.** The identities
`A >= B  <=>  !(A < B)` and `A <= B  <=>  !(B < A)` hold only for a total order.
With NaN operands the decomposition deviates from ONNX/IEEE-754: ONNX requires a
comparison involving NaN to be `false`, but `hip.less` (C++ `<`) returns `false`
for any NaN comparison, so `!(A < B)` yields `true`. This is **exact for integer
inputs** (the dominant index/mask use) and for **NaN-free float models**; float
models that can produce NaN are not supported by this decomposition. This is a
deliberate trade-off (reuse `hip.less` + `hip.not` vs. a dedicated NaN-correct
`>=`/`<=` kernel) and is documented in the conversion sources.

## Tests
- LIT (conversion + hip-to-llvm + e2e), each with static **and dynamic-shape**
  cases for Tanh and GreaterOrEqual/LessOrEqual; Tanh covers fp16 and fp32.
- EP-vs-CPU accuracy validated exact (cosine = 1.0) for fp32/fp16/int32,
  equal-shape and dynamic-shape inputs.

> Note: a separate, unrelated compile-time crash in the existing reduce
> conversions (unranked result) and a test-harness input-range fix are tracked
> in their own PR, not here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ This project demonstrates the integration of HIP (Heterogeneous-compute Interfac
 ## Features
 
 - **MLIR Compiler Pipeline**: ONNX dialect → HIP dialect → LLVM IR → native DLL
-- **MIOpen Integration**: Conv, ConvTranspose, Softmax, RMS Norm, Mul, Sigmoid, Softplus, CausalConvWithState via MIOpen library
+- **MIOpen Integration**: Conv, ConvTranspose, Softmax, RMS Norm, Mul, Sigmoid, Tanh, Softplus, CausalConvWithState via MIOpen library
 - **hipBLASLt MatMul**: High-performance matrix multiplication
 - **Custom HIP Kernels**: GQA, RoPE, Cast, Sub, Gather, ReduceSum, Reciprocal, Sqrt, GELU, Range, LinearAttention
 - **Memory Pool Optimization**: `hip-pool-allocs` pass packs allocations into a single grow-on-demand buffer
@@ -33,6 +33,7 @@ This project demonstrates the integration of HIP (Heterogeneous-compute Interfac
 | Mul | MIOpen |
 | Add | MIOpen |
 | Sigmoid | MIOpen |
+| Tanh | MIOpen |
 | Softplus | MIOpen |
 | Gelu | Custom HIP kernel |
 | Reciprocal | Custom HIP kernel |
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ This project demonstrates the integration of HIP (Heterogeneous-compute Interfac
 | Mod | Custom HIP Kernel |
 | Sign | Custom HIP Kernel |
 | Less | Custom HIP Kernel |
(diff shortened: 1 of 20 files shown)


amd-mingw (approved):
LGTM

PR #375, opened 2026-06-17
feat: convert 1x1 conv to gemm

amd-genmingz:
This is a test PR that converts a 1x1conv module to gemm+add. Currently, it's only used for package testing. If performance benefits are confirmed, the PR description will be revised.


amd-genmingz:
Testing showed no performance improvement.

PR #384, opened 2026-06-19
fix: Whisper encoder miscompile (CSE buffer aliasing) + CI memory & AMDGPU EP rename

chizamd:
## Summary

Several fixes to get the Whisper accuracy suite green on the 32 GB UMA CI
machines, verified end-to-end on the AMDGPU-umbrella build on the **default
bitcode** path (no NATIVE).

## 1. Pipeline miscompile (encoder cosine 0.606)

**Root cause** (bisected to #370): a pre-bufferize CSE merged every HIP DPS op's
`tensor.empty` destination into one, so after bufferize multiple ops shared one
output buffer and clobbered each other — whole-encoder corruption, cosine ~0.6
(identical fp16/fp32 = structural, not precision).
**Fix:** run `empty-tensor-to-alloc-tensor` before that CSE so each op keeps its
own destination. Regression test added.

## 2. Bitcode path under the Python harness

The default bitcode (in-process LLVM-IR JIT) path crashed under pytest /
`transcribe_whisper.py`. Resolved by:

- **Build against a Tier-1 `llvm-install` (`find_package`), not the from-source
  Tier-2 fallback.** A Tier-2 build's JIT crashes at session-create on Windows;
  CI already builds a standalone `llvm-install`, so matching it locally
  (`build.py --cmake_prefix_path …`) fixes it. Documented in
  `docs/whisper_quick_start.md` §1. (Local build-hygiene, not a repo change.)
- **#401** (now in this branch's base after rebase) binds the JIT'd runtime's C
  runtime — stdio + the malloc family — to `ucrtbase` via a search generator,
  fixing the heap-corruption crash. No JIT code change is needed in this PR.
- **Stop the tests forcing `HIPDNN_EP_DEBUG=1`** (process-latched; floods stderr).
  The compile-failure tripwire (`_assert_compiled_on_gpu`) works without it.

**(Optional) To bypass the in-process JIT entirely**, set `HIPEP_ARTIFACT_FORMAT=NATIVE` —
  each model is compiled to a per-model DLL (lld-link + LoadLibrary) instead of
  JIT'd in process. Opt-in escape hatch for local debugging; default stays
  bitcode (see §4).

## 3. Whisper CI memory hygiene (32 GB UMA)

Peak-memory reduction for the shared 32 GB CPU+GPU pool: a `gc`-between-tests
fixture (host RSS −30%: 10.27 → 7.22 GB) + run one precision per pytest process
(two caches survive session destruction, so a same-process `gc` can't free them)
→ caps GPU peak at one precision (~11.5 GB) instead of ~15 GB combined. CI runner
should split the selection into `… and fp16` / `… and fp32` calls.

## 4. AMDGPU EP rename adaptation

EP now reached via the `AMDGPUExecutionProvider` umbrella (`hipep` backend,
`profile=llm`). Updated the test harness (`amdgpu-ep.dll`, registration/OGA names,
genai_config), renamed the override env var `MORPHIZEN_EP_BIN` → `HIPEP_EP_BIN`
(old name kept as fallback), and added an opt-in `HIPEP_ARTIFACT_FORMAT` escape
hatch (unset → bitcode default).

## Verification (default bitcode, real GPU, umbrella amdgpu-ep → hipep-backend → hipep)

- Whisper suite **20/20 pass** (encoder/decoder/e2e greedy JFK quote/librispeech/WER/30s, fp16+fp32); encoder cosine **0.606 → 0.9999**.
- `transcribe_whisper.py` → verbatim JFK quote on GPU (~48 tok/s).
- Pipeline fix also: llama1b ORT 7/7, LIT 258/258.

## Test plan

- [x] Whisper 20/20 on default bitcode (GPU), llama1b ORT 7/7, LIT 258/258
- [x] Memory: −30% host RSS; per-precision split caps GPU peak
- [ ] CI green on 32 GB UMA machine (split invocation + AMDGPU umbrella)


test/python/whisper/test_whisper.py
```diff
@@ -235,7 +249,18 @@ def _morphizen_session(model_name, model_dir=_MODEL_DIR):
...
+    # Force the NATIVE artifact format (compile each model to a per-model DLL +
+    # LoadLibrary) instead of the default in-process LLVM-IR ORC JIT. On Windows
+    # the JIT path crashes at session create (comgr LLVM-symbol collision; the
+    # version-script localization that fixes it is Linux-only). artifact_format
+    # is a hipep *backend* option, so it must be passed as a raw `ep.hipep.*`
+    # session-config entry (NOT a provider option — ORT validates those against
+    # the umbrella's declared options and rejects an unknown `artifact_format`);
+    # the AMDGPU umbrella forwards non-umbrella-prefixed config entries verbatim
+    # to the backend. Mirrors CI's `-C ep.hipep.artifact_format|NATIVE`.
+    so.add_session_config_entry("ep.hipep.artifact_format", "NATIVE")
```
amd-mingw: Native mode compile-phase output is model.dll, as discussed, maybe has security risk and dll sign issue, so we switched to bitcode mode by default.

amd-mingw (changes requested):

test/python/whisper/test_whisper.py
chizamd: **Previously JIT does not work on my end and I had to force to use NATIVE mode. After some investigations, the problems are resolved now.**

Root-caused and fixed — bitcode now works under the Python test harness (no NATIVE needed). Three independent issues, all addressed:

1. **Local build was Tier-2 (from-source LLVM) instead of Tier-1 (`llvm-install` + `find_package`)** → the JIT crashed at session-create. Fix: build against a standalone `llvm-install` like CI does (`build.py --cmake_prefix_path <llvm-install>`)

2. **The JIT pinned `new`/`delete` to the EP DLL's CRT but not `malloc`/`free`** → under `python.exe` the two split across different heaps and corrupted on free (random `0xc0000005` mid-inference). Fix: pin `malloc`/`free`/`calloc`/`realloc` to the EP DLL's CRT too. (5fcbd44b)

3. **The tests forced `HIPDNN_EP_DEBUG=1`**, whose bitcode logging path still crashes. Fix: removed it — the compile-failure tripwire works without it. (5fcbd44b)

Verified on default **bitcode** (no NATIVE): whisper suite **20/20 pass**, `transcribe_whisper.py` emits the verbatim JFK quote on GPU. `HIPEP_ARTIFACT_FORMAT=NATIVE` stays as an opt-in escape hatch but is no longer required.

backend-mlir-compiler/custom-op-mlir/src/LlvmIrJit.cpp
```diff
@@ -82,6 +83,27 @@ llvm::Error installPlatformSymbolShims(llvm::orc::LLJIT &jit) {
...
+  // this DLL's hid the split). Routing the whole malloc family through this
+  // DLL's CRT keeps a single consistent heap regardless of the host process.
+  add("malloc",
+      reinterpret_cast<void *>(static_cast<void *(*)(size_t)>(::malloc)));
+  add("free", reinterpret_cast<void *>(static_cast<void (*)(void *)>(::free)));
+  add("calloc", reinterpret_cast<void *>(
+                    static_cast<void *(*)(size_t, size_t)>(::calloc)));
+  add("realloc", reinterpret_cast<void *>(
+                     static_cast<void *(*)(void *, size_t)>(::realloc)));
+
```
amd-mingw: I think #401 solve it from the root cause and merged, you can sync with @BoarQing and double check it.

others LGTM.

chizamd:
@fhanuman — detailed explanation of §1 (pipeline numerics regression), for review:

**TL;DR:** the pre-bufferize `cse` introduced in #370 deduplicates the `tensor.empty` destinations of distinct HIP DPS ops into a single SSA value, so after bufferization those ops share one output buffer and overwrite one another's results. Bufferization still succeeds, so this surfaces only as incorrect numerics — the Whisper encoder cosine drops to ~0.6. The fix converts `tensor.empty` → `bufferization.alloc_tensor` before that CSE.

**Background.** #370 added `--hip-infer-shapes` followed by `canonicalize + cse` in `buildOnnxToHipPipelineTail`, to fold the dynamic-shape index arithmetic the dynamic-encoder conversions emit (so `--hip-pool-allocs` sees folded constant dims). The `canonicalize` does exactly that; the interaction below is specific to the `cse`.

**What happens.** Every HIP DPS op takes its own `tensor.empty` as the destination-passing `outs` init:

```mlir
%e0 = tensor.empty() : tensor<2x8xf16>
%a  = hip.gemm ... outs(%e0) : tensor<2x8xf16>
%e1 = tensor.empty() : tensor<2x8xf16>
%b  = hip.gemm ... outs(%e1) : tensor<2x8xf16>
```

`tensor.empty` is side-effect-free, so CSE coalesces all same-typed empties into one value:

```mlir
%e  = tensor.empty() : tensor<2x8xf16>
%a  = hip.gemm ... outs(%e)
%b  = hip.gemm ... outs(%e)   // same %e
```

After `one-shot-bufferize`, `%e` becomes a single `memref.alloc` and both ops bufferize in-place into it (the HIP DPS ops don't trigger copy-on-conflict), so `%b`'s write lands on `%a`'s buffer and a later read of `%a` observes `%b`'s data. On the Whisper encoder this coalesced **551 distinct `tensor.empty` → 8** and reduced **post-bufferize allocations 551 → 170**, which is enough to perturb the whole graph.

**Why it reads like a precision issue but isn't.** The cosine is **identical for fp16 and fp32 (~0.606)** — i.e. structural buffer aliasing, independent of element type, not numerical error. Because bufferization reports no diagnostic, an accuracy test against a CPU baseline simply sees a low cosine.

**Bisect:** encoder cosine is ~0.9999 at #369 and ~0.606 at #370.

**Fix.** Insert `bufferization::createEmptyTensorToAllocTensorPass()` immediately before the CSE. `bufferization.alloc_tensor` carries allocation semantics (it is not memory-effect-free), so CSE keeps each instance distinct — every DPS op retains its own destination — while still deduplicating the genuinely side-effect-free shape/index (`tensor.dim` / `arith`) recomputations the pass was added for. #370's intent is preserved; only the destination coalescing is prevented.

```mlir
%e0 = bufferization.alloc_tensor() : tensor<2x8xf16>   // distinct, CSE-safe
%a  = hip.gemm ... outs(%e0)
%e1 = bufferization.alloc_tensor() : tensor<2x8xf16>
%b  = hip.gemm ... outs(%e1)
```

**Verification.**
- Whisper suite 20/20 (encoder cosine **0.606 → 0.999984 fp16 / 1.0 fp32**; decoder / e2e / librispeech / WER / 30s).
- llama1b ORT 7/7 (incl. `dynamic_vs_fixed` bit-identical KV and `per_step_logits`); LIT 258/258.
- Post-bufferize allocation count restored to 551.
- Regression test `test/lit/Pipeline/cse-preserves-dps-dest.mlir`: a bare `--cse` coalesces the two same-typed empties into one; `--empty-tensor-to-alloc-tensor --cse` keeps two distinct `alloc_tensor` destinations.

fhanuman:
> @fhanuman — detailed explanation of §1 (pipeline numerics regression), for review:
> 
> **TL;DR:** the pre-bufferize `cse` introduced in #370 deduplicates the `tensor.empty` destinations of distinct HIP DPS ops into a single SSA value, so after bufferization those ops share one output buffer and overwrite one another's results. Bufferization still succeeds, so this surfaces only as incorrect numerics — the Whisper encoder cosine drops to ~0.6. The fix converts `tensor.empty` → `bufferization.alloc_tensor` before that CSE.
> 
> **Background.** #370 added `--hip-infer-shapes` followed by `canonicalize + cse` in `buildOnnxToHipPipelineTail`, to fold the dynamic-shape index arithmetic the dynamic-encoder conversions emit (so `--hip-pool-allocs` sees folded constant dims). The `canonicalize` does exactly that; the interaction below is specific to the `cse`.
> 
> **What happens.** Every HIP DPS op takes its own `tensor.empty` as the destination-passing `outs` init:
> 
> ```mlir
> %e0 = tensor.empty() : tensor<2x8xf16>
> %a  = hip.gemm ... outs(%e0) : tensor<2x8xf16>
> %e1 = tensor.empty() : tensor<2x8xf16>
> %b  = hip.gemm ... outs(%e1) : tensor<2x8xf16>
> ```
> 
> `tensor.empty` is side-effect-free, so CSE coalesces all same-typed empties into one value:
> 
> ```mlir
> %e  = tensor.empty() : tensor<2x8xf16>
> %a  = hip.gemm ... outs(%e)
> %b  = hip.gemm ... outs(%e)   // same %e
> ```
> 
> After `one-shot-bufferize`, `%e` becomes a single `memref.alloc` and both ops bufferize in-place into it (the HIP DPS ops don't trigger copy-on-conflict), so `%b`'s write lands on `%a`'s buffer and a later read of `%a` observes `%b`'s data. On the Whisper encoder this coalesced **551 distinct `tensor.empty` → 8** and reduced **post-bufferize allocations 551 → 170**, which is enough to perturb the whole graph.
> 
> **Why it reads like a precision issue but isn't.** The cosine is **identical for fp16 and fp32 (~0.606)** — i.e. structural buffer aliasing, independent of element type, not numerical error. Because bufferization reports no diagnostic, an accuracy test against a CPU baseline simply sees a low cosine.
> 
> **Bisect:** encoder cosine is ~0.9999 at #369 and ~0.606 at #370.
> 
> **Fix.** Insert `bufferization::createEmptyTensorToAllocTensorPass()` immediately before the CSE. `bufferization.alloc_tensor` carries allocation semantics (it is not memory-effect-free), so CSE keeps each instance distinct — every DPS op retains its own destination — while still deduplicating the genuinely side-effect-free shape/index (`tensor.dim` / `arith`) recomputations the pass was added for. #370's intent is preserved; only the destination coalescing is prevented.
> 
> ```mlir
> %e0 = bufferization.alloc_tensor() : tensor<2x8xf16>   // distinct, CSE-safe
> %a  = hip.gemm ... outs(%e0)
> %e1 = bufferization.alloc_tensor() : tensor<2x8xf16>
> %b  = hip.gemm ... outs(%e1)
> ```
> 
> **Verification.**
> 
> * Whisper suite 20/20 (encoder cosine **0.606 → 0.999984 fp16 / 1.0 fp32**; decoder / e2e / librispeech / WER / 30s).
> * llama1b ORT 7/7 (incl. `dynamic_vs_fixed` bit-identical KV and `per_step_logits`); LIT 258/258.
> * Post-bufferize allocation count restored to 551.
> * Regression test `test/lit/Pipeline/cse-preserves-dps-dest.mlir`: a bare `--cse` coalesces the two same-typed empties into one; `--empty-tensor-to-alloc-tensor --cse` keeps two distinct `alloc_tensor` destinations.

Thanks, @chizamd , for the clarification. I made the changes in PR #370. I will approve your PR to unblock the submission.
 
My concern is that the pass you are adding may prevent buffers optimization. Could you please run a quick experiment to confirm this?
 
Compare the buffer count for Qwen 3.5-9B and 35B with and without your changes, and the same for Gemma-int4-4B.
 
You can go ahead and merge the changes, and do this experiment separately.

fhanuman (approved):

backend-mlir-compiler/custom-op-mlir/src/LlvmIrJit.cpp
chizamd: rebased with latest main

amd-mingw (approved):

PR #386, opened 2026-06-22
refactor(op-state): simplify op-state API (create -> unique_ptr, plain-C++ set, get_op_state)

BoarQing:
## Summary

Implements the reviewer feedback on the op-state-slots API, tightening the boundary between the runtime bitcode and generated IR without changing the construct ABI or the emitted IR.

- `OpStateT::create()` now returns an owning `std::unique_ptr<T>` (with `deletor` already wired) instead of constructing-and-storing internally. Each `construct_<op>` hands the released pointer to the setter and returns `0`.
- `hipdnn_ep_op_state_set` is now plain C++ returning `void` (was `extern "C"` returning `bool`). It is only ever called from within the runtime bitcode by the `construct_<op>` functions, never by generated IR, so it needs no C ABI and no failure path; a debug `assert` documents the slot-validity contract.
- Renamed `OpStateT::get_slot` -> `get_op_state` for consistency with the setter/getter naming.
- Removed the redundant `hipdnn_ep_op_state_set` declaration in `hipdnn_ep_runtime.h` that conflicted with the `op_state.h` declaration (it was the source of the ambiguous-call build break).
- Updated all 10 real + 10 mock `construct_<op>` functions to the `set(state, slot, X::create().release())` form.
- Updated the MLIR-side comments (`HipOpStateInterface.td`/`.cpp`, `HipDialect.h`, `GenerateOpStateInit.cpp`), the design doc, and the `gen_init` lit test comments to reflect that the `construct_<op>` i8 result is now vestigial (always `0`) and that the store has no failure path.

No functional change: the `construct_<op>` ABI signature `(RuntimeState*, i32 slot, i64...) -> i8` and the generated IR are unchanged; the i8 is kept for a stable call signature and is not branched on.

## Test plan

- [x] `lintrunner` (clang-format / ruff): clean
- [x] Build under `vcvars64` (ninja, TheRock HIP SDK): clean
- [x] LIT suite: 277/277 pass
- [x] GPU numeric suite (EP DLL refreshed): 337 passed, 31 pre-existing failures unchanged vs `main` (no regressions)
- [ ] CI green on this PR

### DIFF
--- a/docs/design/op-state-slots-design.md
+++ b/docs/design/op-state-slots-design.md
@@ -98,7 +98,7 @@ ConvOp::generateOpStateInit(b, loc, statePtr, slot):
     ctor  = declare_extern("hipdnn_ep_op_state_construct_conv",
                            ret = i8, params = [ptr, i32, i64, i64])
     ok    = b.call(ctor, [ statePtr, const(slot), const(k), const(s) ])
-    return ok                                                # i8: store refused => 0
+    return ok                                                # i8: vestigial, always 0
 ```
 
 The runtime constructor takes those parameters, builds the state with `OpStateT::create` (which returns an owning `std::unique_ptr<T>` with `deletor` already wired), and hands the released pointer to `hipdnn_ep_op_state_set`, which stores it into the slot. `_set` is plain C++ internal to the runtime bitcode (never called by generated IR) and has no failure path: the compiler always supplies a valid slot into an already-allocated array, so the `RuntimeState` layout stays opaque without bounds-check ceremony:
--- a/docs/design/op-state-slots-design.md
+++ b/docs/design/op-state-slots-design.md
@@ -54,9 +54,9 @@ struct OpState {                 // base of every state struct
 
 template <class T> struct OpStateT : OpState {   // CRTP base each state derives from
   OpStateT() { deletor = [](OpState *p) { delete static_cast<T *>(p); }; }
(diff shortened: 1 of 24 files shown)


wcy123 (approved):
**LGTM** — Well done. This refactoring clarifies the API boundary with the right separation of concerns.

**Key improvements:**

1. **Standard C++ factory idiom:** `create()` returns an object (`unique_ptr<T>`). Construction is construction — the factory's job is to create and return the object, not to store it somewhere. This follows established C++ conventions.

2. **Ownership and placement separation:** Construction (`create()`) and placement (`set(state, slot, ...)`) are now distinct operations. The caller explicitly transfers ownership via `release()`. This makes the control flow clear and puts each responsibility where it belongs.

3. **Simplified setter contract:** `hipdnn_ep_op_state_set` should not and will not fail. It's an internal runtime call (never invoked from generated IR), so the `extern "C"` ABI and bool failure path were unnecessary complexity. Making it plain C++ `void` with a debug `assert` documents the actual contract.

4. **Naming clarity:** `get_slot` → `get_op_state` aligns with what the function actually does and matches the setter naming.

**Minor observation:** The `i8` return type in the construct function signature `(RuntimeState*, i32, i64...) -> i8` is non-idiomatic — standard C API design uses `int` for status codes, or `void` when there's no failure path. The return value is now always `0` and unused, but keeping it maintains ABI stability without requiring IR regeneration. Not critical to change now, but worth noting for future API design.

PR #389, opened 2026-06-22
fix(runtime): route all JIT'd runtime logging through CRT-free hipdnn_ep_log_emit

AMDmoore:
## Issue

Enabling `HIPDNN_EP_DEBUG=1` or `HIPDNN_EP_PERF=1` crashed the process with `0xC0000409` (STACK_BUFFER_OVERRUN) shortly after the first log line. The crash was **delayed** (it did not fault at the log call site) and reproduced only on the JIT/bitcode EP path (the default) — never on the mock/native build.

## Root cause

In bitcode mode the EP does not link a self-contained model DLL; `LlvmIrJit` (ORC LLJIT, `backend-mlir-compiler/custom-op-mlir/src/LlvmIrJit.cpp`) JIT-loads `runtime.bc` in-process and resolves external symbols **by name against the whole process image** (`DynamicLibrarySearchGenerator::GetForCurrentProcess`).

That resolver splits a single translation unit across **two CRTs**:

- `fputs` / `fwrite` / `fputc` / `fflush` (and `getenv` / `malloc`) bind to **msvcrt.dll**
- the `stdout` / `stderr` `FILE*` (`__acrt_iob_func`) come from **ucrtbase.dll**

Handing a ucrtbase `FILE*` to an msvcrt stdio function makes msvcrt interpret a foreign, incompatibly-laid-out `FILE` struct and write through bogus buffer pointers — corrupting memory that surfaces later as `0xC0000409`.

- `printf` / `fprintf` only *survived* because clang **inlines** them into JIT-private memory (self-consistent single CRT).
- `fflush` / `putchar` survived by UB luck.
- It is **NOT** stdout-vs-stderr (stderr's `FILE*` is also ucrtbase), **NOT** buffering, and **NOT** un-run global constructors (`jit->initialize()` does run `@llvm.global_ctors`).

Proven with a per-mode experiment (`HIPDNN_EP_DIAG_STDOUT`, since removed) plus an address/binding probe (`HIPDNN_EP_DIAG_CRT=1`, retained): `fputs`/`fwrite` fault 3/3, `printf`/`fprintf`/`fflush`/`putchar` do not.

## Solution

Replace **every** CRT-stdio output call (`fprintf`/`fputs`/`fwrite`/`fputc`/`fflush`) in the bitcode runtime with `hipdnn_ep_log_emit()`, which:

1. formats with `vsnprintf` into a stack buffer (no `FILE*`, no stream state), then
2. writes the bytes to fd 2 via Win32 `WriteFile(GetStdHandle(STD_ERROR_HANDLE), ...)` / POSIX `write(2, ...)`.

`op_profile.cpp` separator `fputc` loops become preformatted strings; redundant `fflush(stderr)` calls are dropped (`WriteFile` is unbuffered). Same class of fix as `GetEnvironmentVariableA` vs `getenv`.

**Intentionally excluded:** `std::fopen`+`std::fread` of the constants file (both bind to the same CRT, so the `FILE*` is self-consistent — not a cross-CRT handoff) and `lib/Runtime/mock/mock_gpu.cpp` (CPU-only mock, never JIT-loaded).

Also adds the `hipdnn_ep_diag_crt_binding` probe (`HIPDNN_EP_DIAG_CRT=1`) for future diagnosis and updates the `CLAUDE.md` gotcha to the proven cross-CRT mechanism.

## Test plan

- [x] `HIPDNN_EP_DEBUG=1` runs Llama end-to-end with no crash; `[REAL] wrap_*` dispatch traces print.
- [x] `HIPDNN_EP_PERF=1` runs Llama end-to-end with no crash; the `[PERF]` table prints.
- [x] `pre-commit run` (lintrunner clang-format + licenseheaders) passes on all staged files.


AMDmoore:
root cause was found here, but refer to #401 as the better solution.

PR #398, opened 2026-06-24
feat(runtime): Unified Memory Manager Phase 1 — HAL + MemoryManager foundation

tejus26:
## Summary

Phase 1 of a 5-phase Unified Memory Manager (UMM). Introduces `MemoryManager` (`lib/Runtime/mm/`) to replace six independent `ensure_*` memory patterns in `RuntimeState` with a single typed API.

- **`HalAllocator` interface** with two fully-specified backends:
  - `ApuHalAllocator` — APU/UMA: `hipHostMalloc(Mapped|NonCoherent)`, eviction/restore are zero-copy fence ops
  - `DiscreteHalAllocator` — dGPU: `hipMalloc` VRAM + pre-allocated pinned host; eviction is async D2H, restore is async H2D. Both backing stores allocated at `alloc()` time to avoid latency spikes at eviction time.
- **`MemoryManager`** wraps pool domains, workspace, host-scalar scratch, and qmoe host scratch with uniform **1.5× amortized growth** (fixes the pool-domain exact-growth bug)
- `seqlens_k` cross-layer GQA cache moved into MM; `begin_compute()` is the single invalidation point
- **Bug fix**: `hipHostMallocCoherent` → `hipHostMallocNonCoherent` in `HipGpuAllocator` (+1.6–4.4% decode TPS on gfx1151)
- **38 GPU-free unit tests** (`test/runtime/mm/`) — all pass

## No behavior change in Phase 1

All existing extern C functions (`hipdnn_ep_get_pool_base`, `hipdnn_ep_get_host_scratch_base`, `hipdnn_ep_state_ensure_workspace`, etc.) delegate to `mm->*` with identical semantics. Legacy fields kept in `RuntimeState` during transition.

## Future phases

Phase 2 (block pool + bump-pointer scratch), Phase 3 (pool domains via block pool), Phase 4 (EP-owned KV cache), Phase 5 (Tier-1 CPU offload).

## Test plan

- [x] 38 GPU-free unit tests: `test-memory-manager.exe` — **all passed**
- [x] Runtime bitcode compiled cleanly via clang (4 new `.bc` files + `runtime.bc` linked)
- [ ] CI: Windows build
- [ ] CI: Linux build
- [ ] CI: GPU tests

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/lib/Runtime/CMakeLists.txt
+++ b/lib/Runtime/CMakeLists.txt
@@ -270,7 +270,7 @@ endmacro()
 # Compiled to bitcode so the MM is merged into each per-model DLL alongside
 # the rest of the runtime. The HAL factory is included here; it selects the
 # correct backend at session-create time via hipDeviceGetAttribute.
-compile_to_bitcode(mm/hal_apu.cpp runtime_mm_hal_apu.bc)
+compile_to_bitcode(mm/hal_igpu.cpp runtime_mm_hal_igpu.bc)
 compile_to_bitcode(mm/hal_discrete.cpp runtime_mm_hal_discrete.bc)
 compile_to_bitcode(mm/hal_factory.cpp runtime_mm_hal_factory.bc)
 compile_to_bitcode(mm/memory_manager.cpp runtime_mm_memory_manager.bc)
@@ -343,7 +343,7 @@ if(NOT BUILD_MOCK_RUNTIME)
     compile_to_bitcode(real/global_pool.cpp runtime_global_pool.bc)
 
     set(RUNTIME_BC_MODULES
-        ${CMAKE_CURRENT_BINARY_DIR}/runtime_mm_hal_apu.bc
+        ${CMAKE_CURRENT_BINARY_DIR}/runtime_mm_hal_igpu.bc
         ${CMAKE_CURRENT_BINARY_DIR}/runtime_mm_hal_discrete.bc
         ${CMAKE_CURRENT_BINARY_DIR}/runtime_mm_hal_factory.bc
         ${CMAKE_CURRENT_BINARY_DIR}/runtime_mm_memory_manager.bc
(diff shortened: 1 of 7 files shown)


amd-mingw:
Hi @tejus26 , could you help to provide a brief view about this memory manager module? such as why we need it(solve which issues)? what benefits we can get from it? what's the difference between it and existed memory-pool mechanism? and so on.  Thanks!  cc @amd-genmingz

tejus26:
> Hi @tejus26 , could you help to provide a brief view about this memory manager module? such as why we need it(solve which issues)? what benefits we can get from it? what's the difference between it and existed memory-pool mechanism? and so on. Thanks! cc @amd-genmingz

Hi @amd-mingw , @amd-genmingz , 
I think we discussed this a while back. But here are the pointers to the spec: https://amdcloud-my.sharepoint.com/:w:/r/personal/tsiddaga_amd_com/Documents/Microsoft%20Teams%20Chat%20Files/LLM%20Memory%20management.docx?d=wd0f5ab1ffe1541f6b717753b2da6bb3a&csf=1&web=1&e=NXVHbA 

The main goal of this abstraction is to be able to implement various KV cache compression and management techniques without relying on OGA / ORT. This will help us quickly implement various algorithms and support latest features.

amd-mingw:
Hi @tejus26, I reviewed the document again, and I still has some confuse about it:
1. about the constants, there is already gpu_constants_blob deal with it, and seems it not the key target of this MM.
2. about the intermediate buffers, there is already memory-pool deal with it, and seems it not the key target of this MM.
3. about the IO, there is already EP binding(prepare_input) and hip.alloc_output(EP allocator), and KV cache(seems it's the focus point of this MM) is part of IO, not the whole IO, and only LLM/VLM has this type IO, other models not have this type IO, and I believe OGA has KV cache management, so I'm not sure the right location of this MM should be OGA or EP? in my opinion, EP should not care KV cache management, to EP, KV cache is same to other IO, and OGA care about KV cache management, it's the OGA application level, not the EP level. or this KV cache management can scalable to general IO management and apply to all the EP IO?
Thanks!

fhanuman (commented):
I skimmed through the changes, and I get the idea of what we are doing here. I
need some more time to think it through, but here are a few high-level comments
for now.

This is a good cleanup, and I like the abstraction for how we handle iGPU and
dGPU.

A couple of things I would like us to think about:

This memory manager seems to rely on the `hip-pool-allocs` pass I added and how we
`allocate pool domains` today. So I want to make sure we have thought about what
happens if we change the way we pool memory in the future.

I am also not sure if the memory manager is the right place for operator-level
details like qmoe, seqlens_k, etc.

One more thing. I think we should consider doing a quick POC to get some performance numbers
before we build out all the later phases. Phase 1 on its own doesn't give us a clear picture of how this is all going to help support larger context length. I just don't want us to spend all this time on the infrastructure and only then find out we don't get the improvement and larger context support we were after.

PR #403, opened 2026-06-24
perf(LinearAttention): chunked-parallel gated-delta prefill kernel (+ decode self-copy skip, NaN fix)

AMDmoore:
## Summary

Speeds up the gated-delta **LinearAttention** path (Qwen3.5 / Qwen3.5-MoE family) by replacing the per-token prefill launch loop with a single chunked-parallel kernel, plus a few supporting runtime/profiling changes. Includes a correctness fix for an underflow-induced NaN found during VLM validation.

Commits in this PR (on top of `main` @ 2c6b93a9):

- **`feat(runtime)`: chunked-parallel gated-delta LinearAttention prefill kernel** — one launch processes the whole sequence in `LA_CHUNK=32` token chunks (one block per `(batch, kv_head)`), keeping the recurrent state `S` in **fp32** across the chunk and doing dense per-chunk matmuls (coupling matrices `T`/`P` in LDS) instead of per-token rank-1 updates. This is strictly *more* accurate than the per-token kernel (which rounds `S` to fp16 every step) — cosine ~0.9999999 in fp16 vs the naive recurrence / ORT CPU. Only `gated_delta` with scalar log-decay is implemented; the launcher declines other configs so the runtime falls back to the per-token loop.
- **`perf(runtime)`: skip recurrent-state self-copy on aliased buffers** — when `past_state` and `present_state` alias the same GPU buffer, the redundant device-to-device self-copy is skipped (decode hot path).
- **`perf(kernels)`: register-block S reads + hoist phase-6 scale** — each `S[i,j]` is read from global once and folded into all `Ceff` partial sums (≈`LA_CHUNK`× fewer global S loads in phases 2 and 5); the per-`s` state-update scale is precomputed once instead of inside the `dk*dv` loop.
- **`fix(kernels)`: avoid 0/0 NaN in the phase-6 state update** — see below.
- **`chore(profile)`: op-profile rows for memcpy/readback paths** — these copies/readbacks include GPU sync cost; profiling them attributes that time correctly in the timeline (gated on `HIPDNN_EP_PERF`).

## The NaN fix

The phase-6 state-update scale was computed as the ratio `a_last / Asm[s] = exp(clog_last) / exp(clog_s)`. Under strong intra-chunk decay (log-decay `g << 0`), `clog_s` becomes very negative, so **both** `exp(clog_s)` and `exp(clog_last)` underflow to `0` in fp32 and the ratio becomes `0/0 = NaN`. That NaN is folded into the recurrent state `S` and propagates to every subsequent chunk → all-garbage output (repeated `!` tokens) on long sequences.

Fix: compute the scale directly as `exp(clog_last - clog_s)`. The difference is bounded (`<= 0` for log-decay `g <= 0`), so `exp()` is always in `(0, 1]` and finite. This also matches the `exp(clog_t - clog_s)` form already used for the `P`/`T` coupling ratios in the same kernel. Math is unchanged in the well-conditioned case.

## Performance (gfx1151, Qwen3.5 VLM, 1842-token image prompt)

vs the pre-branch baseline (`main` @ 2c6b93a9), `vlm_benchmark.py`, `HIPDNN_EP_PERF=0`:

| Model | TTFT (ms) | TTFT speedup | Decode (tok/s) | Decode speedup |
|---|---|---|---|---|
| Qwen3.5-9B (dense) | 8646 → 5043 | **1.71×** | 29.5 → 34.0 | **1.15×** |
| Qwen3.5-35B-A3B (MoE) | 9886 → 5541 | **1.78×** | 35.0 → 46.1 | **1.32×** |

(TTFT lower is better; Decode higher is better. Both `speedup` columns are expressed so that >1× = faster.)

Notes:
- Prefill throughput is `prompt_tokens / TTFT`, so it mirrors the TTFT speedup (same 1.7–1.8×) — omitted here to avoid double-counting the same number.
- The decode gain comes from the recurrent-state self-copy skip; the decode kernel itself is unchanged by this PR. MoE (35B) decode tok/s is sensitive to WMMA-autotune cache warmth and varies run-to-run, so treat its multiplier as approximate.

## Correctness

- Qwen3.5-9B: greedy output is **token-identical** to the pre-optimization baseline (dense; the kernel's fp32-internal accumulation matches closely enough that argmax never flips).
- Qwen3.5-35B-A3B: greedy output is **semantically correct** (coherent, on-topic). It diverges token-by-token from baseline because the chunked kernel is numerically different by construction and MoE expert routing amplifies any legal fp delta — this is expected, not a regression.
- The garbage-output (`!!!!`) symptom that motivated the NaN fix is gone on both models.

## Test plan

- [ ] `ctest` LIT/unit suites green
- [ ] Qwen3.5-9B VLM end-to-end (image prompt) — coherent output, no `!` garbage
- [ ] Qwen3.5-35B-A3B VLM end-to-end (image prompt) — coherent output
- [ ] Confirm GPU dispatch (`HIPDNN_EP_DEBUG=1` → `[custom_kernels] hip_linear_attention_prefill_chunked` / decode rows)

## Note for reviewers

Branch is based on `main` @ 2c6b93a9; `main` has since refactored op-state handling (incl. `lib/Runtime/real/linear_attention.cpp`). A rebase onto current `main` may be needed before merge — `linear_attention.cpp` is the likely conflict point.

### DIFF
--- a/3rd-party/custom_kernels/hip/linear_attention_kernel.hip
+++ b/3rd-party/custom_kernels/hip/linear_attention_kernel.hip
@@ -586,13 +586,18 @@ __global__ void linear_attention_prefill_chunked_kernel(
         }
         __syncthreads();
         const float a_last = Asm[Ceff - 1];
-        // Precompute the per-s state-update scale a_last/A_s ONCE (Ceff divisions
+        // Precompute the per-s state-update scale (A_{C-1}/A_s) ONCE (Ceff exps
         // total) instead of recomputing it inside the dk*dv phase-6 loop below.
-        // Kept as a division (not a_last * 1/A_s): a_last <= A_s so the quotient
-        // is <=1 and overflow-safe, whereas 1/A_s alone can overflow when A_s
-        // underflows under strong intra-chunk decay.
+        // Compute it as exp(clog_last - clog_s), NOT as the ratio a_last/A_s:
+        // under strong intra-chunk decay clog_s can be very negative, so both
+        // A_s = exp(clog_s) AND a_last = exp(clog_last) underflow to 0 in fp32
+        // and the ratio becomes 0/0 = NaN (which then poisons S and every
+        // subsequent chunk -> all-garbage output). The difference clog_last -
+        // clog_s is bounded (<= 0 for log-decay g <= 0), so exp() of it is in
+        // (0, 1] and always finite. This also matches the exp(clog_t - clog_s)
+        // form already used for the P/T ratios above (internal consistency).
         for (int s = tid; s < Ceff; s += nthreads)
-            rlast[s] = a_last / Asm[s];
+            rlast[s] = expf(clog[Ceff - 1] - clog[s]);
         __syncthreads();
 
         // B[t][j] = beta_t v_t[j] - beta_t A_t (S0^T k_t)[j]  -> Usm
--- a/3rd-party/custom_kernels/hip/linear_attention_kernel.hip
+++ b/3rd-party/custom_kernels/hip/linear_attention_kernel.hip
@@ -555,6 +555,7 @@ __global__ void linear_attention_prefill_chunked_kernel(
     float* Asm = Pm + LA_CHUNK * LA_CHUNK;          // [LA_CHUNK]
     float* clog = Asm + LA_CHUNK;                   // [LA_CHUNK]
     float* bsm = clog + LA_CHUNK;                   // [LA_CHUNK]
+    float* rlast = bsm + LA_CHUNK;                  // [LA_CHUNK] a_last/A_s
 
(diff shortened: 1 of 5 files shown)


AMDmoore:
From the CI verification, for qwen3.5-9B and 35B, TTFT uplift up to **41%~55%**, TPS uplift up to **10%~31%**, with no peak memory change.
optimization perf task: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/1000/
baseline perf task: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/1001/

image

amd-mingw:
image

amd-mingw (approved):

PR #412, opened 2026-06-25
refactor(hip): relocate custom GPU kernels from 3rd-party/ to lib/Runtime/Kernels

fhanuman:
We keep the custom HIP kernels under `3rd-party/`, which mislabels them — they are first-party AMD code, not vendored third-party dependencies. They are also the GPU-architecture-dependent half of the runtime, so they belong alongside the runtime that consumes them rather than being flagged as external. This moves them into the core tree at `lib/Runtime/Kernels`.

**The fix:**

* `git mv 3rd-party/custom_kernels lib/Runtime/Kernels` — recorded as renames, so history is preserved and the kernel sources are unchanged.
* Update the build wiring: the two top-level `add_subdirectory(...)` calls and the `-I` include flag + bitcode `DEPENDS` header in `lib/Runtime/CMakeLists.txt`. The `custom_kernels_<arch>` target/artifact names are kept intentionally — the EP's JIT arch-probing (`LlvmIrJit`) and the CI co-location checks key on that filename, so the output names are decoupled from the directory name.
* Carry the existing formatter / license-header exemption forward to `lib/Runtime/Kernels/**` in `.lintrunner.toml` and `.pre-commit-config.yaml`, so this stays a pure relocation rather than a whole-tree clang-format + license-header churn. Dropping those entries later brings the kernels under linting once their permanent home is settled.
* Sweep stale `3rd-party/custom_kernels` path references in comments and docs (`CLAUDE.md`, `docs/design/*`, `README.md`).

The kernels build, install (`libcustom_kernels_<arch>.so`), and all 278 LIT tests pass from the new location.

P.S. This is the location agreed for this release; the permanent home (MLSS / ROCm) is still under discussion.


amd-mingw (approved):
LGTM

PR #421, opened 2026-06-26
fix(hip): de-alias DPS init operands that CSE merged onto one tensor.empty

fhanuman:
A model that lowers to a multi-result HIP DPS op whose kernel reads back its own outputs can silently miscompile. CSE deduplicates `tensor.empty` (it is `Pure`); when it collapses the seeds of two init (DPS-out) operands of the *same* op onto one value, bufferization aliases both tied results into one buffer, and the kernel then reads one output as the other.

**Minimal example (post-CSE):**

```mlir
%e = tensor.empty() : tensor<AxBxCxDxf16>
// present_key and present_value share %e -> after bufferize they share one
// buffer -> hip.gqa reads V (present_value) as K (present_key) -> wrong output
hip.gqa(...) outs(%out, %e, %e : ...)
```

This is fine for single-use scratch, but wrong here. One-Shot Bufferize only separates two aliased writing inits when it can find a read-after-write conflict on a tied result, i.e. when the results are live. When they are dead (e.g. a present-KV output nothing downstream consumes), it keeps them merged and the kernel-internal read consumes the wrong bytes.

**The fix:**

New pass `--hip-split-duplicate-dps-inits`, run after CSE and before one-shot-bufferize:

* For each DPS op, re-point every init operand after the first occurrence of a shared `tensor.empty` at its own fresh `bufferization.alloc_tensor` (CSE-immune, documented to not alias any other buffer).
* Only `tensor.empty` seeds are rewritten -- their contents are undefined, so the substitution is value-preserving. A non-empty init (which may carry contents the op reads) is left untouched.
* Single-use scratch empties are unaffected, so benign CSE merges -- and the downstream pool packing that depends on them -- are preserved.

This is the surgical counterpart to running `empty-tensor-to-alloc-tensor` over every empty (the upstream blanket approach), which also blocks the benign merges and inflates the buffer / pool-domain count on activation-heavy graphs.

### DIFF
--- a/include/hip/Dialect/Transforms/Passes.td
+++ b/include/hip/Dialect/Transforms/Passes.td
@@ -765,4 +765,49 @@ def ResolveExternConstantsPass
   ];
 }
 
+def SplitDuplicateDpsInitsPass
+    : Pass<"hip-split-duplicate-dps-inits", "mlir::func::FuncOp"> {
+  let summary = "De-alias DPS init operands of an op that CSE merged onto one "
+                "tensor.empty";
+  let description = [{
+    `tensor.empty` is `Pure`, so CSE deduplicates same-typed empties. That is
+    harmless for single-use scratch, but WRONG when one merged empty feeds two
+    or more *init* (DPS-out) operands of the SAME op: bufferization then aliases
+    those results into one buffer, and an op whose kernel reads back its own
+    outputs -- e.g. `hip.gqa` reading K from `present_key` and V from
+    `present_value` -- consumes the same bytes for both and silently
+    miscompiles. One-Shot Bufferize separates such inits only when the tied
+    results are live; for dead scratch results it keeps them merged.
+
+    For each `DestinationStyleOpInterface` op, this pass re-points every init
+    operand after the first occurrence of a shared `tensor.empty` at its own
+    fresh `bufferization.alloc_tensor` (CSE-immune, documented to not alias any
+    other buffer). Only `tensor.empty` seeds are rewritten -- their contents are
+    undefined, so the substitution is value-preserving; non-empty inits, which
+    may carry contents the op reads, are left untouched. Single-use scratch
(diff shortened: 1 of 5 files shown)


amd-yanjunz:

- PR: ROCm/onnx-hipdnn-ep#421
- Source comment: [models=all](https://github.com/ROCm/onnx-hipdnn-ep/pull/421#issuecomment-4813310770)
- TEST_MODE: both
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/1095/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hipdnn_ep_oga&id=6a3efe555251324dc8440fa9

amd-yanjunz:

- PR: ROCm/onnx-hipdnn-ep#421
- Source comment: [modelzoo=LLM_ROCm_Whisper models=all](https://github.com/ROCm/onnx-hipdnn-ep/pull/421#issuecomment-4813354429)
- TEST_MODE: both
- MODELZOO: LLM_ROCm_Whisper
- TEST_MODELS: (all)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/1096/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hipdnn_ep_oga&id=6a3eff0223a4069c1377943d

amd-mingw:
http://xsjvitisaijenkins:8080/job/daily_hipdnn_ep_oga/1100/

http://xsjnts.xilinx.com/ROCm/compare?base_url=&base_job=daily_hipdnn_ep_oga&base=1091&curr_url=&curr_job=daily_hipdnn_ep_oga&curr=1100

amd-mingw (approved):

PR #427, opened 2026-06-30
remove unicode encoding for simpler ci generated report

BoarQing:
removed the unicode as it may produced gibberish for certain terminal


BoarQing:
@fhanuman Can u took a look at this? It would reduce the time for our performance optimization

amd-mingw (approved):
LGTM

PR #428, opened 2026-06-30
fix(ci): apply OGA PR patches with git am + ship libamdgpu-ep.so

zz002:
## Background

`microsoft/onnxruntime-genai` PR #2194 first **adds** `src/morphizen_ep/*`, then **modifies** it, and finally **renames** it to `src/amdgpu/*` (renaming the EP library to `libamdgpu-ep.so` and the provider to `amdgpu`) — all within its own commit series.

CI currently applies it with `git apply pull/2194.patch`. `git apply` pre-checks every hunk in the mbox against the base tree at once, so modifying/renaming a file that the same series only just added self-conflicts. The `Apply OGA PR patches` step fails and the whole build dies. This is also why the current main package's OGA is still the pre-rename `MorphiZenEP` build.

## Changes

- **`git apply` → `git am`**: `pull/<n>.patch` is a multi-commit mbox; `git am` replays each commit in order, so any add → modify → rename sequence applies cleanly. Set a committer identity (a bare runner has no git identity, which `git am` requires).
- **Bump the OGA cache key (`-am`)**: invalidate the stale git-apply artifact so it is not served from cache.
- **Copy `libhipgpu.so` → `libamdgpu-ep.so` in the bundle step**: the renamed AMDGPU integration `dlopen`s `libamdgpu-ep.so` at runtime, while this project still builds the EP as `libhipgpu.so`. The copy is then picked up by the dependency-closure staging and packaging that follow.

## Validation

Reproduced locally the same way CI runs: `git am --whitespace=nowarn pull/2194.patch` applies all 5 commits (including `rename → AMDGPU`) cleanly on top of a clean `v0.14.0`, producing `src/amdgpu/`.

## Note

The genai_config provider name is uppercase `AMDGPU`, while the source uses `ep_name_ = "amdgpu"` (lowercase) — worth confirming whether the dispatch is case-sensitive.

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -257,7 +257,7 @@ jobs:
         uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: ${{ runner.workspace }}/oga-artifacts
-          key: linux-oga-${{ env.OGA_VERSION }}-pr${{ env.OGA_PR_PATCHES }}-ort${{ env.ONNXRUNTIME_VERSION }}-ortpr${{ env.ONNXRUNTIME_PR_PATCHES }}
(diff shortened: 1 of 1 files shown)


amd-mingw (approved):
LGTM

PR #439, opened 2026-07-01
refactor(ep): make output-allocator ABI the only mode

zpye:
## Summary

Makes the **output-allocator ABI the only execution mode** and deletes the classic 3-arg out-param path entirely. Every model now compiles to a single 2-arg `inference_compute(state, inputs)`, with graph outputs allocated in-graph via `hip.alloc_output` → the EP's output-allocator callback.

Net **−677 lines** (45 files changed, 667 insertions / 1344 deletions).

### What was removed
- The dual pipeline / classic slot-3 `buffer-results-to-out-params` for the main graph (the internal `onnx.Loop` body out-param ABI is unaffected).
- The `use_output_allocator` compile flag, the `hipdnn.use_output_allocator` module attribute, and the proto field — no provider option, no mode switch anywhere.
- The `DimSource` output-dim resolution and the `prepare_output` / `finalize_output` runtime staging helpers.
- Classic EP dispatch; `InferenceState::compute` is now 2-arg only.

### What changed
- **Pipelines:** `hip-use-output-allocator` runs unconditionally.
- **Codegen:** `HipToLLVM` / `GenerateInterface` emit only the 2-arg ABI.
- **Proto:** metadata fields 4 (`use_output_allocator`) and 5 (`dim_sources`) retired and kept `reserved`.
- **`UseOutputAllocator` — new Step 2:** back passthrough / view / aliased outputs (and outputs behind a shape-adjusting `memref.cast`) with their own `hip.alloc_output` + `memref.copy`, so the allocator callback fires exactly once per output index. (Classic mode previously handled these via out-param copies.)
- **`MemoryLowering` fix:** `alloc_output` element size for sub-byte types (`i1`) now uses ceil-div bytes (was truncating to 0 → legalization failure).
- **`hip-test`:** installs a host-buffer output-allocator callback + 2-arg compute.
- **Docs/comments:** rewritten to single-mode and simplified.

## Test plan
- [x] `ctest` — E2E compile + execute, 278 LIT, OutputAllocator unit test
- [x] `pre-commit run` clean (lintrunner / clang-format / license headers)
- [ ] CI green on the PR

> Note: the `lpnormalization` e2e **execute** test remains a pre-existing failure on `main` (the `reduce_sum` kernel lacks an f32 path, so the sum stays 0 → `1/0 = inf`). That gap is ABI-independent and out of scope here; it will be fixed in a separate PR.


zpye:
Superseded by a 3-PR stack for easier review: #440 (compiler) -> #441 (ep/runtime) -> #442 (docs). No content dropped - the stack sums byte-identical to this branch.

PR #440, opened 2026-07-02
refactor(compiler): emit only the output-allocator 2-arg ABI [Phase 1/2]

zpye:
## Summary
**PR 1 of 2** in the output-allocator-only migration stack. Makes the output-allocator ABI the only compiled path.

- Pipeline runs `hip-use-output-allocator` unconditionally; `convert-hip-to-llvm` / `generate-interface` emit only the 2-arg `inference_compute(state, inputs)`. Classic 3-arg out-param codegen removed.
- Removes the compile-time `use_output_allocator` flag consumer wire (flatbuffer field, JSON writer, `CompilerDriver` option, pipeline option). The metadata flag + runtime dispatch are retired in #441.
- `UseOutputAllocator` change here is removal-only: it drops the now-dead `hipdnn.use_output_allocator` module-attribute stamp (the codegen halves no longer read any allocator attribute; they are unconditional). The returned-alloc -> `hip.alloc_output` rewrite is unchanged from `main`.
- `MemoryLowering`: `alloc_output` element size for sub-byte types (`i1`) now uses ceil-div bytes.
- `hip-test` migrated to a host-buffer allocator callback; LIT/e2e tests updated to the 2-arg ABI.

## Docs (co-located so this PR is self-consistent)
- `docs/technical-pipeline-walkthrough.md` and `docs/design/pool-allocs-memory-planning.md` updated to the allocator-only pipeline.

## Stack
1. **this PR (#440)**   compiler
2. #441   ep/runtime (delete dead classic dispatch, retire proto fields)

## Test plan
- [ ] CI green
- [x] `pre-commit run` clean

### DIFF
--- a/backend-mlir-compiler/level-1-pass/src/MlirCompiler.cpp
+++ b/backend-mlir-compiler/level-1-pass/src/MlirCompiler.cpp
@@ -48,8 +48,6 @@ std::string build_compiler_options_json(const CompilationConfig &config) {
   json << "\"opt_level\": " << config.optLevel;
   json << ", \"skip_constant_data\": "
        << (config.skipConstantData ? "true" : "false");
-  json << ", \"use_output_allocator\": "
-       << (config.useOutputAllocator ? "true" : "false");
   // output_mode maps to the flatbuffers OutputMode enum
   // (schemas/compilation_options.fbs). The flatbuffers JSON parser accepts
   // the enum value name.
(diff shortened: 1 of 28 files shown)


wcy123 (approved):
Clean refactor removing ~600 lines of dual-mode infrastructure. Pipeline ordering and sub-byte type fix verified. Combined with #441, this eliminates ~1,360 lines of complexity.

PR #441, opened 2026-07-02
refactor(ep/runtime): delete classic out-param dispatch, retire proto fields [Phase 2/2]

zpye:
## Summary
**PR 2 of 2** in the output-allocator-only migration stack. Deletes the now-dead classic 3-arg out-param path.

- `MlirCustomOp` / `InferenceState` dispatch only the 2-arg `inference_compute`.
- Removes the metadata `use_output_allocator` flag + its setter/reader, the `DimSource` output-dim resolution (and its `Output.dim_sources` field), and `CompilationConfig::useOutputAllocator`.
- The freed proto field numbers are NOT reserved - the schema is renumbered so field numbers stay continuous (`Metadata.artifact_format` is now field 4; `Output` is fields 1-4). Safe here because the metadata proto has a single producer (level-1 pass) and a single consumer (EP) that are always rebuilt together, with no persisted wire data across versions.
- Deletes the `prepare_output` / `finalize_output` runtime staging helpers; outputs are created in-graph via the allocator callback.

## Docs (co-located so this PR is self-consistent)
- `docs/design/output-allocator-design.md`, `CLAUDE.md`, `docs/design/morphizen-ep-integration.md`, and `backend-mlir-compiler/doc/INTEGRATION-DESIGN.md` updated to describe the single allocator ABI.

## Stack
1. #440 - compiler
2. **this PR (#441)** - ep/runtime

## Dependency
Stacked on #440 (base branch `feat/allocator-only-1-compiler`).

## Test plan
- [ ] CI green (after #440 merges / retargets to main)
- [x] `pre-commit run` clean


wcy123 (dismissed):
Clean runtime-side refactor removing ~750 lines of dual-dispatch code. Proto field renumbering safe (single producer/consumer, always rebuilt together). Completes the output-allocator-only migration started in #440.

wcy123 (approved):
LGTM

PR #444, opened 2026-07-02
Plugin interface: static-linkage core

fhanuman:
> Base of the plugin-interface stack: #444 → #445 → #446 → #443 → #447. Review/merge in order.

## Problem

A downstream needs a stable boundary to extend the compiler without editing in-tree files. There is no mechanism for it to register passes, a custom dialect, runtime bitcode, or link libraries and have the host pick them up.

## Change

This PR adds the static-linkage core of the plugin interface: the plugin entry-point macro and registry (`PluginAPI.h` / `PluginRegistry`), the CMake machinery that links the selected plugin static libraries into the host, the generated static registrar, and the three non-pipeline contribution points the host consumes — dialect registrations, runtime bitcode, and link libraries. It also runs plugin registration in `hip-mlir-opt` and adds the `HIPDNN_EP_PIPELINE` textual pipeline override. With no plugin selected every hook is a no-op and the default build and compile path are unchanged. Pipeline-slot consumption and the custom-op lowering hook follow in the next PR.

## Design and usage

[docs/design/plugin-interface.md](https://github.com/ROCm/onnx-hipdnn-ep/blob/plugin-split/01-design-doc/docs/design/plugin-interface.md)


Vandana9628 (approved):
Looks good, thanks.

PR #446, opened 2026-07-02
Plugin interface: sample plugin and tests

fhanuman:
> Stacked on #445 (plugin-interface stack: #444 → #445 → #446 → #443 → #447). Review/merge in order.

## Problem

The plugin interface has no in-tree exercise, so nothing keeps the ABI working in CI before any downstream plugin exists.

## Change

This PR adds a complete in-tree sample plugin that exercises every registry contribution — a pass at a slot, runtime bitcode, and a link library — plus a GPU-free unit test that asserts the static registrar records them, and two LIT tests. It wires `test/plugin` into the build and adds the `hip_static_plugins` lit feature so the Plugin tests run when the sample is selected (`-DHIPDNN_EP_COMPILER_PLUGINS=sample`) and are skipped otherwise. This keeps the plugin ABI exercised in CI.

## Design and usage

[docs/design/plugin-interface.md](https://github.com/ROCm/onnx-hipdnn-ep/blob/plugin-split/01-design-doc/docs/design/plugin-interface.md)


Vandana9628 (approved):
@fhanuman the sample plugin never calls the addDialectRegistration and has no CI coverage for it. So any refactoring here could break the `loadAllDialects` and `hasPromisedInterface`. Can we add a minimal dialect + op so both paths run in our own CI. It's not a blocker and okay to add a follow up to address this.

fhanuman:
> @fhanuman the sample plugin never calls the addDialectRegistration and has no CI coverage for it. So any refactoring here could break the `loadAllDialects` and `hasPromisedInterface`. Can we add a minimal dialect + op so both paths run in our own CI. It's not a blocker and okay to add a follow up to address this.

Thanks for the review. Makes sense. Let me address this in a follow up.

fhanuman:
> > @fhanuman the sample plugin never calls the addDialectRegistration and has no CI coverage for it. So any refactoring here could break the `loadAllDialects` and `hasPromisedInterface`. Can we add a minimal dialect + op so both paths run in our own CI. It's not a blocker and okay to add a follow up to address this.
> 
> Thanks for the review. Makes sense. Let me address this in a follow up.

Addressing this as part of below PR:

https://github.com/ROCm/hip-ep/pull/456

PR #449, opened 2026-07-07
refactor(compiler): trace returned buffers via alias analysis in hip-use-output-allocator

zpye:
## Summary

- Rework `hip-use-output-allocator` to decide whether each `memref.alloc` is a graph output using `BufferViewFlowAnalysis` instead of the previous cast-only `OpRewritePattern`. `resolve()` returns the alloc plus every value derived from it through **any** chain of view ops (`memref.cast` / `collapse_shape` / `expand_shape` / `subview`), so an alloc that is reshaped on the way to `func.return` — e.g. a rank-4 conv result collapsed to rank-2 — is now recognized and rewritten to `hip.alloc_output` with no per-op special-casing. The graph-output index comes from `getOperandNumber()` on the `func.return` use of any alias.
- Structure the pass as **analyze-then-transform**: phase 1 classifies allocs with the alias analysis (no IR mutation), phase 2 rewrites (no analysis queries), so the analysis (which caches `Value` handles) is never read after the IR it describes has been mutated.
- Declare `getDependentDialects` (`HipDialect` for the created `hip.alloc_output`, `MemRefDialect` for the view ops' `ViewLikeOpInterface`).
- Existing behavior is preserved: public-only functions, `!hip.context` arg-0 requirement, dealloc removal, first (lowest) `out_idx` for aliased multi-outputs, and program-order rewrite so `out_idx` values print in order.
- Extend the LIT test with `collapse_shape` / `expand_shape` / multi-op-chain / `subview` return cases; sync `Passes.td` and `docs/design/output-allocator-design.md`.

## Test plan

- [x] Incremental build via local build script (only `UseOutputAllocator.cpp` recompiled + relinked; clean).
- [x] `MorphizenMLIRLitTests` passes — covers direct / cast / collapse / expand / chain / subview / aliased / passthrough / no-context / private-helper cases plus the `output-allocator-dealloc.mlir` pipeline test.
- [x] `pre-commit run` clean (lintrunner/clang-format, license headers, whitespace).

### DIFF
--- a/docs/design/output-allocator-design.md
+++ b/docs/design/output-allocator-design.md
@@ -95,7 +95,7 @@ Introduce `hip.alloc_output` — an operation that allocates output buffers via 
 
 **Placement in pipeline:** *after* `buffer-deallocation`, *before* `hip-pool-allocs` (so pooling skips the EP-owned buffer). `hip-use-output-allocator` is FuncOp-scoped: it rewrites returned allocs (leaving each signature + `return` intact). The after-dealloc order is load-bearing — see [§2](#the-pipeline).
 
-The `hip-use-output-allocator` pass replaces the returned `memref.alloc` for each graph output with `hip.alloc_output`, reusing the alloc's dynamic-size operands.
+The `hip-use-output-allocator` pass replaces the returned `memref.alloc` for each graph output with `hip.alloc_output`, reusing the alloc's dynamic-size operands. An alloc counts as a graph output when `func.return` hands it back — either directly, or after it passes through view ops (`memref.cast`, `memref.collapse_shape`, `memref.expand_shape`, `memref.subview`, any number of them). The pass uses `BufferViewFlowAnalysis` to check this: an alloc is an output if any value made from it via those view ops is returned. This handles reshaped / cast / subview'd outputs without special-casing each op, and the view ops between the alloc and the return are left in place.
 
 **Example: Add → MatMul → Sigmoid**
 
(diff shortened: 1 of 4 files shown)


wcy123 (commented):
Excellent fix for the reshape bug using BufferViewFlowAnalysis! The backward alias analysis approach is exactly right. Two suggestions to make this more MLIR-idiomatic:

## 1. Test Coverage: Longer Chains

The longest test chain is 2 ops (`@chain_output`: collapse → cast). Since the bug analysis emphasizes "arbitrary-length chains," add a 3+ op test to demonstrate this works for longer chains:

```mlir
// CHECK-LABEL: func.func @long_chain_output
// CHECK-NOT:     memref.alloc
// CHECK:         %[[OUT:.*]] = hip.alloc_output(%{{.*}}) {out_idx = 0 : i64}
// CHECK:         %[[COL:.*]] = memref.collapse_shape %[[OUT]]
// CHECK:         %[[CST:.*]] = memref.cast %[[COL]]
// CHECK:         %[[EXP:.*]] = memref.expand_shape %[[CST]]
// CHECK:         return %[[EXP]]
func.func @long_chain_output(%ctx: !hip.context) -> memref<?x?x?xf32> {
  %x = memref.alloc() : memref<2x4x8xf32>
  %col = memref.collapse_shape %x [[0, 1, 2]] : memref<2x4x8xf32> into memref<64xf32>
  %cast = memref.cast %col : memref<64xf32> to memref<?xf32>
  %exp = memref.expand_shape %cast [[0, 1, 2]] output_shape [?, ?, ?]
       : memref<?xf32> into memref<?x?x?xf32>
  return %exp : memref<?x?x?xf32>
}
```

With these changes, the implementation is both simpler and has stronger test coverage.

wcy123 (dismissed):

wcy123 (approved):
LGTM

PR #450, opened 2026-07-07
build(runtime): add HIPDNN_EP_DISABLE_VENDOR_BLAS flag to build without MIOpen/hipBLASLt

Vandana9628:
### Summary:

Adds an opt-in build flag, `HIPDNN_EP_DISABLE_VENDOR_BLAS` (default OFF).

When it's OFF, nothing changes — every existing build behaves exactly as before.

When it's ON, the build stops requiring MIOpen and hipBLASLt: it skips looking for
those libraries, doesn't link them into the model, and doesn't create their
handles at startup. This is for GPU targets that don't ship a working
MIOpen/hipBLASLt.

This PR only sets up the flag and these outer changes. It builds fine, but a model
that actually uses a MIOpen/hipBLASLt op won't link yet — the pieces that make
those ops link come in the next PR. Splitting it this way keeps each PR small:
first this setup, then the op stubs, then some unrelated arch fixes.


fhanuman (approved):
LGTM

PR #451, opened 2026-07-07
Extra guards for FFM upstream.

Vandana9628:
### Summary

Three small, independent fixes so the compiler builds and runs on GPU targets
that differ from the RDNA3/gfx1151 assumptions in some kernels and tools. None
hard-codes a GPU name.

- mock_types.h: add a missing include for size_t, which a bitcode-only build
  doesn't get automatically.
- WMMA kernels (gemm_wmma, gqa, matmul_nbits): the fast matrix-multiply builtin
  only exists on RDNA3. On a GPU without it, the file now still compiles and, if
  that path is ever run, it stops with a clear crash instead of returning wrong
  numbers.
- hip-test: hide its own LLVM symbols at link time so they can't clash with a
  different LLVM the HIP runtime loads.

### DIFF
--- a/lib/Runtime/Kernels/hip/gemm_wmma_kernel.hip
+++ b/lib/Runtime/Kernels/hip/gemm_wmma_kernel.hip
@@ -11,6 +11,24 @@
 typedef _Float16 half16 __attribute__((ext_vector_type(16)));
 typedef float float8 __attribute__((ext_vector_type(8)));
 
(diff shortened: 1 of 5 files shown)


fhanuman (approved):
The failures you are seeing are due to a change in the repo name — Elliott renamed the repo from onnx-hipdnn-ep → hip-ep. I pushed fixes for these issues yesterday. Could you please rebase and check?

PR #454, opened 2026-07-07
Remove blas stubs

Vandana9628:
### Summary

Completes `HIPDNN_EP_DISABLE_VENDOR_BLAS` so a model can build without MIOpen/hipBLASLt. Those libraries provide functions the runtime calls; with them unlinked, those calls become missing symbols and the model won't link. This adds error-returning stubs for every such function, so the model links anywhere and an op that truly needs a vendor library just fails at runtime instead. All stubs are behind the flag, so a normal (flag-off) build is unchanged.

### DIFF
--- a/lib/Runtime/CMakeLists.txt
+++ b/lib/Runtime/CMakeLists.txt
@@ -180,6 +180,13 @@ macro(compile_to_bitcode SOURCE_FILE OUTPUT_BC)
         list(APPEND COMPILE_FLAGS "-fPIC")
     endif()
 
+    # Propagate the vendor-BLAS opt-out into the bitcode compile so the
+    # MIOpen/hipBLASLt-backed wrappers (real/{miopen,hipblas,elementwise,gemm}.cpp)
+    # compile their stub branch instead of the vendor-backed one.
+    if(HIPDNN_EP_DISABLE_VENDOR_BLAS)
+        list(APPEND COMPILE_FLAGS "-DHIPDNN_EP_DISABLE_VENDOR_BLAS")
+    endif()
+
     # Always add root directory for public headers
     list(APPEND COMPILE_FLAGS "-I" "${CMAKE_CURRENT_SOURCE_DIR}")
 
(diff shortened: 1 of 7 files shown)


lib/Runtime/real/elementwise.cpp
```diff
@@ -2,6 +2,76 @@
...
+  (void)out_h;
+  (void)out_w;
+  (void)data_type;
+  (void)tensor_op;
+  fprintf(stderr, "wrap_miopenOpTensor: MIOpen disabled at build time "
+                  "(HIPDNN_EP_DISABLE_VENDOR_BLAS); op is unavailable\n");
+  return -1;
+}
+
+int wrap_elementwise_sub(RuntimeState *state, void *lhs, void *rhs,
```
fhanuman: These #ifdef HIPDNN_EP_DISABLE_VENDOR_BLAS guards look redundant. These wrappers route vendor calls through *_CHECK_GOTO, which sets result = -1; goto cleanup; (returns, no abort) — and with the flag on those calls link against the error-returning stubs in vendor_blas_stub.cpp, so the wrapper already fails cleanly with -1; the guard only changes the log message. gemm/hipblas/elementwise call only functions the stub already defines, so their guards can go as-is. Can we remove the guards and rely on vendor_blas_stub.cpp?

lib/Runtime/CMakeLists.txt
```diff
@@ -280,6 +287,11 @@ if(NOT BUILD_MOCK_RUNTIME)
     compile_to_bitcode(real/hip.cpp runtime_hip.bc)
     compile_to_bitcode(real/miopen.cpp runtime_miopen.bc)
     compile_to_bitcode(real/hipblas.cpp runtime_hipblas.bc)
+    # Vendor-symbol stubs so a model links without libMIOpen / libhipblaslt (see
+    # real/vendor_blas_stub.cpp).
+    if(HIPDNN_EP_DISABLE_VENDOR_BLAS)
```
fhanuman, line 292: Is there a way we can add a test to exercise this flag? Currently, none of the test points seem to exercise this flag

fhanuman (changes requested):
The failures you are seeing are due to a change in the repo name — Elliott renamed the repo from `onnx-hipdnn-ep` → `hip-ep`. I pushed fixes for these issues yesterday. Could you please rebase and check?

lib/Runtime/real/elementwise.cpp
Vandana9628: Done. Removed the #ifdef guards from all four wrappers and rely on vendor_blas_stub.cpp. 
One addition: miopen.cpp's `wrap_miopenConvolutionTranspose` calls `miopenSetTransposeConvOutputPadding`, which wasn't in the stub yet — so dropping its guard would have left an undefined symbol at link. I added that one function to vendor_blas_stub.cpp, and now all four guards are gone.

lib/Runtime/CMakeLists.txt
Vandana9628: @fhanuman Added a lightweight CI check on the Linux build that exercises the flag. It configures with `-DHIPDNN_EP_DISABLE_VENDOR_BLAS=ON` and asserts the vendor backends are dropped.

fhanuman (approved):

PR #458, opened 2026-07-08
perf(profile): per-op GPU/CPU profiler + Chrome-trace export

BoarQing:
## What

Per-operator profiling for the EP, gated on `HIPDNN_EP_PERF=1`:

- **Per-op timing** — one fenceless GPU marker per op, differenced against a per-inference epoch (low distortion), plus host-side CPU wrapper time. Printed as a `[PERF]` table (calls / gpu ms / cpu ms / gpu%, with per-shape breakdown).
- **`[outer] Compute` total** — the whole `Compute()` CPU time, recorded separately and excluded from `TOTAL`, so `outer − Σ(per-op)` is the bubble (host/launch overhead not attributed to any op).
- **Chrome-trace export** — set `HIPDNN_EP_TRACE_FILE=<path>` to also emit a `chrome://tracing` / Perfetto JSON (GPU track + CPU track + H2D/Compute/D2H phase tracks). One file per session on the shared time axis.

## Tooling

- `tools/perf-report/merge_chrome_traces.py` — stitch multiple session traces into one (argparse CLI, `--zero` for cross-clock inputs).

## Notes

- Single GPU-timing path (no classic event-pair mode / `HIPDNN_EP_PERF_ISOLATE`).
- Includes minor dead-code cleanup in the runtime.

## Verified

`test_matmul_2d` and a Qwen3.5-9B VLM run (NATIVE); trace call-counts and GPU/CPU ms sums cross-checked line-for-line against the printed `[PERF]` tables.


lib/Runtime/debug_log.h
```diff
@@ -57,29 +58,76 @@ inline bool hipdnn_ep_perf_isolate_enabled() {
...
+// chrome://tracing JSON in addition to the aggregated [PERF] table.
+inline const std::string &hipdnn_ep_trace_path() {
+  static const std::string path = [] {
+#ifdef _WIN32
+    char buf[1024];
+    unsigned long n =
+        GetEnvironmentVariableA("HIPDNN_EP_TRACE_FILE", buf, sizeof(buf));
+    return (n > 0 && n < sizeof(buf)) ? std::string(buf, n) : std::string();
+#else
+    const char *v = std::getenv("HIPDNN_EP_TRACE_FILE");
```
BoarQing: i think HIPDNN_EP_PERF_TIMELINE and HIPDNN_EP_TRACE_FILE can be merged into one right? it does not make sense one if only one of them is ON. and the macro can just check if the string == ""

lib/Runtime/op_profile.cpp
```diff
@@ -103,12 +176,84 @@ hipEvent_t op_profile_get_stop_event(OpProfileState *ps, int index) {
...
+
+double op_profile_roofline_gbps() {
+  static double gbps = [] {
+    std::string e = hipdnn_ep_get_env("HIPDNN_EP_ROOFLINE_GBPS");
+    if (!e.empty()) {
+      double v = atof(e.c_str());
+      if (v > 0)
+        return v;
+    }
+    return 256.0; // Strix Halo LPDDR5X default
```
BoarQing: ideally this should be read directly from device. but i do perfer u just use a hardcode number here. and add a TODO

lib/Runtime/op_profile.cpp
```diff
@@ -200,21 +411,43 @@ void op_profile_resolve_and_print(OpProfileState *ps) {
...
+                          char *pkBuf, size_t n) {
+    if (bytes > 0 && ms > 0) {
+      double gbps = (double)bytes / 1e9 / (ms / 1000.0);
+      snprintf(mbBuf, n, "%.1f", (double)bytes / 1e6);
+      snprintf(gbBuf, n, "%.0f", gbps);
+      snprintf(pkBuf, n, "%.0f%%", 100.0 * gbps / peakGbps);
+    } else {
+      snprintf(mbBuf, n, "%s", "-");
+      snprintf(gbBuf, n, "%s", "-");
+      snprintf(pkBuf, n, "%s", "-");
```
BoarQing: i dont get why u change the prints. i think everything we change are used for chrome_traces?

amd-mingw (approved):

PR #461, opened 2026-07-09
AMDMLSS gfx1151: GQA is the only kernel candidate, and it gives zero porting value for decode

BoarQing:
## 1. Candidates from MLSS (gfx1151)
- **GQA** (CK WMMA fp16) — the only real candidate; every target model uses it.
- **MHA** — dormant (our models export GroupQueryAttention, not MHA).
- MVN / Conv1x1 / Conv-dilated — vision ops, absent from transformer graphs.
- RMSNorm / SiLU / GEMMGEMM / ConvMxN / GEMM / QGEMM — not on gfx1151 (RDNA4-only or scaffold).

## 2. Where it beats onnx-hipdnn-ep
Only in the WMMA comfort zone: **prefill, head_dim ≤ 80** (~12–14 TFLOP/s), plus a marginal ~1.2–1.3× MHA decode microbench.

## 3. Why it's still zero porting value
Our models are **GQA decode, head_dim 128/256** — outside that zone:
- **Decode starves WMMA:** always a 16-row query tile, so q_seq=1 uses 1/16 (q_seq=1 ≈ q_seq=16 ≈ 100µs). ~17% of peak BW, 6× over the memory floor → **regresses** vs our decode path.
- **head_dim 128 misses the tuned tiles** (tuned only ≤80) → fallback at 4–8% of peak → prefill is a no-op too.

**Net: decode regresses, prefill no-op → zero value today.** Correctness isn't the blocker — GQA ABI is decoded and validates at ~6e-5 (see `docs/amdmlss_gfx1151_assessment.md`). Revisit only if AMD ships a decode-tuned or head_dim-128 blob for gfx1151.

### DIFF
--- a/docs/amdmlss_gfx1151_assessment.md
+++ b/docs/amdmlss_gfx1151_assessment.md
@@ -1,0 +1,107 @@
+# AMDMLSS gfx1151 (Strix Halo) kernel port assessment
+
+Status: **assessed, not merged.** This document records which AMDMLSS kernels are
+portable to `onnx-hipdnn-ep` on gfx1151, what was actually validated, and why none
+of them is worth enabling for the target workload (LLM **decode**) right now.
+
+## TL;DR
+
+- The complete set of gfx1151-eligible AMDMLSS kernels is small: **MHA, GQA, MVN,
+  Conv1x1, Conv-dilated**. Everything else (RMSNorm, SiLU, GEMMGEMM, ConvMxN,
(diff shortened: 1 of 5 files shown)


BoarQing:
Closing as assessed / not-merged: no gfx1151 AMDMLSS kernel is worth enabling for the LLM decode workload. Decode regresses (WMMA 16-row query tile wastes 15/16 at q_seq=1); prefill only helps head_dim<=80 while all target models are head_dim 128/256 (untuned fallback, 4-8% of peak). Kept the decoded GQA ABI + assessment doc for future revival.

PR #473, opened 2026-07-13
feat(dialect): scaffold hipsr op pipeline

mingyueliuh:
Scaffold the op-generation pipeline for the `hipsr` dialect. No ops are defined yet — this only wires up the base class and TableGen op generation so later PRs can add ops one at a time.

- Add `HipsrOps.td` with the `Hipsr_Op` base class.
- Add op decls/defs generation (`-gen-op-decls` / `-gen-op-defs`) to `HipsrDialectIncGen`.
- Add `HipsrOps.h` / `HipsrOps.cpp`, link into `HipsrDialectIR`, and register the (empty) op list in `initialize()`.


wcy123 (approved):
LGTM

PR #474, opened 2026-07-13
feat(dialect): add hipsr shape_yield op and ShapeRegionInterface

mingyueliuh:
Add the shape-region terminator and the interface that will be the single source of truth for output-shape computation on hipsr ops.

- Add `hipsr.shape_yield`: a `Pure`, `Terminator` op yielding the computed output dimensions as `index` values (zero operands = rank-0 result). Operands are always `index`, so the assembly omits the redundant types.
- Add `ShapeRegionInterface` with `populateShapeRegion` (declared), `populateCapacityShapeRegion` (default no-op), and `getShapeRegion` / `getCapacityShapeRegion` accessors. Its verifier runs `verifyShapeRegionStructure` (defensive: single block ending in `hipsr.shape_yield`) then `verifyShapeRegionScoping` (selective isolation — may reference the op's operands but no other outer values). Implementing ops must also add the `SingleBlockImplicitTerminator<"ShapeYieldOp">` trait (documented in the interface).
- Wire op-interface TableGen and link the interface into `HipsrDialectIR`. Split the TableGen by concern (MLIR upstream style): `HipsrInfrastructureOps.*` (structural ops), `HipsrOps.*` (compute ops), and per-interface `HipsrShapeRegionInterface.*`.
- Add a round-trip lit test for `shape_yield`.

## Deferred to a follow-up PR

The design (pool-allocs shape-region architecture, Appendix C) has `ShapeRegionInterface` extend `ReifyRankedShapedTypeOpInterface` and provide a default `reifyResultShapes`. Both are intentionally deferred here:

- `baseInterfaces = [ReifyRankedShapedTypeOpInterface]`
- default `reifyResultShapes` that reads the shape region's `hipsr.shape_yield` terminator and returns its `index` operands as the reified dims (the Phase-1 form; the doc's Appendix C reads a `hip.empty_yield` tensor, which is the later Stage-3 form).

Reason: no op implements `ShapeRegionInterface` yet, so a `reifyResultShapes` default cannot be exercised by a lit test in this PR. It also couples the interface to the concrete `ShapeYieldOp`. Both will be added when the first shaped ops attach a shape region, where the reify path is testable end-to-end.


wcy123 (changes requested):
three documentation/organization suggestions.

Clean shape-region infrastructure. Scoping verification correctly enforces selective isolation (shape regions access op's operands but not arbitrary outer values).

**Suggestions:**

**1. Document `populateCapacityShapeRegion` contract in `HipsrInterfaces.td`:**

```tablegen
InterfaceMethod<
  [{Populate the capacity shape region for EndBarrier operations.
    
    Only EndBarrier operations override this method. Regular and StartBarrier
    operations use the default no-op implementation.
    
    PRECONDITION: Caller must verify the operation implements EndBarrierInterface
    and that capacityShapeRegion (region 1) exists before calling this method.
    The Region& parameter (reference, not pointer) assumes the region is valid.
  }],
  "void", "populateCapacityShapeRegion",
  (ins "::mlir::OpBuilder &":$builder, "::mlir::Region &":$capacityShapeRegion),
  ...
```

**2. Document `verifyShapeRegionScoping` design rationale in `HipsrInterfaces.h`:**

```cpp
/// Design note: We use custom scoping verification instead of IsolatedFromAbove
/// trait because shape regions need selective isolation:
/// - IsolatedFromAbove: Region cannot access ANY outer values (complete isolation)
/// - verifyShapeRegionScoping: Region CAN access the op's operands, but nothing else
///
/// Shape regions need to read the op's inputs to compute output shapes,
/// e.g., `tensor.dim %input, %c0` where %input is the op's operand.
/// With IsolatedFromAbove, this would require verbose region arguments.
/// Current design allows direct access to op operands for ergonomics while
/// preventing accidental capture of arbitrary outer scope values.
::mlir::LogicalResult verifyShapeRegionScoping(::mlir::Operation *op);
```

**3. TableGen organization for scale:** With 70+ operations planned, consider splitting: `HipsrShapeOps.td` (shape_yield + shape infrastructure), `HipsrOps.td` (compute ops). Follows MLIR upstream patterns (Linalg, GPU dialects).

```

include/hip/Dialect/Hipsr/IR/
├── HipsrOps.td                    # Data compute operations (matmul, add, conv, ...)
├── HipsrInfrastructureOps.td      # Terminators, control flow (shape_yield, pool_domain, ...)
└── HipsrInterfaces.td             # Interfaces (already separate)
Or simpler:

├── HipsrOps.td                    # All compute operations
├── HipsrShapeOps.td               # Shape region infrastructure (shape_yield, etc.)
└── HipsrInterfaces.td
```
Two-region support (capacity + real) for EndBarrier ops is well-designed. Deferring `reifyResultShapes()` until implementing ops land is right incremental approach.

wcy123 (commented):
Adding one more suggestion:

**4. Simplify `shape_yield` assembly format:**

```tablegen
// Current (redundant):
let assemblyFormat = "attr-dict ($dims^ `:` type($dims))?"
// Prints: hipsr.shape_yield %d0, %d1 : index, index

// Simplified:
let assemblyFormat = "attr-dict ($dims^)?"
// Prints: hipsr.shape_yield %d0, %d1
```

Types are redundant - always `index` from `Variadic<Index>`. Follows MLIR practice (scf.yield, func.return).

wcy123 (changes requested):
Two file header comments reference old file names:

**1. `HipsrOps.td` line 5-6:**
```tablegen
// Compute ops for the hipsr dialect (matmul, add, ...). They are added
// incrementally. Shape-region infrastructure lives in HipsrShapeOps.td; the
// shared op base class lives in HipsrBase.td.
```

Should be:
```tablegen
// Compute ops for the hipsr dialect (matmul, add, ...). They are added
// incrementally. Infrastructure ops live in HipsrInfrastructureOps.td; the
// shared op base class lives in HipsrDialect.td.
```

**2. `HipsrInfrastructureOps.td` line 5-7:**
```tablegen
// Infrastructure ops for the hipsr dialect: shape-region terminators and
// other structural ops that are not compute kernels. Compute ops live in
// HipsrOps.td; the shared op base class lives in HipsrBase.td.
```

Should be:
```tablegen
// Infrastructure ops for the hipsr dialect: shape-region terminators and
// other structural ops that are not compute kernels. Compute ops live in
// HipsrOps.td; the shared op base class lives in HipsrDialect.td.
```

Both reference `HipsrBase.td` (removed) and `HipsrOps.td` references `HipsrShapeOps.td` (never existed).

wcy123 (changes requested):
Add defensive terminator verification and clarify trait requirement.

Currently, if a developer forgets to add the `SingleBlockImplicitTerminator<"ShapeYieldOp">` trait, wrong terminators (e.g., `func.return`) pass verification and cause crashes downstream when code assumes `shape_yield`.

**Changes:**

1. **Document trait requirement** in interface description
2. **Split verification into two functions:**
   - `verifyShapeRegionStructure`: Defensive structural check (single block + correct terminator)
   - `verifyShapeRegionScoping`: Semantic scoping rules (existing logic)

This follows MLIR pattern: interfaces can't auto-attach traits, so we document the requirement and add defensive verification for better error messages.

**Suggested changes:**

---

**include/hip/Dialect/Hipsr/IR/HipsrShapeRegionInterface.td:**

```tablegen
def ShapeRegionInterface : OpInterface<"ShapeRegionInterface"> {
  let description = [{
    Base interface for shape computation.
    
    REQUIREMENT: Operations implementing this interface MUST also add the
    SingleBlockImplicitTerminator<"ShapeYieldOp"> trait. Example:
    
      def Hipsr_MatMulOp : Hipsr_Op<"matmul", [
          ShapeRegionInterface,
          SingleBlockImplicitTerminator<"ShapeYieldOp">
      ]> { ... }
    
    The trait provides automatic terminator insertion and structural verification.
    The interface verifier will defensively check the terminator type and enforce
    scoping rules.
    
    Shape regions compute output shapes for the operation. The interface provides:
    - populateShapeRegion(): Abstract method defining shape computation logic
    - populateCapacityShapeRegion(): Optional override for EndBarrier operations
  }];

  let cppNamespace = "::mlir::hipsr";

  let methods = [
    // ... existing methods ...
  ];

  let verify = [{
    if (failed(::mlir::hipsr::verifyShapeRegionStructure($_op)))
      return ::mlir::failure();
    return ::mlir::hipsr::verifyShapeRegionScoping($_op);
  }];
}
```

---

**include/hip/Dialect/Hipsr/IR/HipsrInterfaces.h:**

```cpp
namespace mlir {
namespace hipsr {

/// Verifies that the shape region has exactly one block terminating with
/// hipsr.shape_yield. This is a defensive check in case the developer forgets
/// to add the SingleBlockImplicitTerminator<"ShapeYieldOp"> trait.
///
/// Called by ShapeRegionInterface's verify block before scoping verification.
::mlir::LogicalResult verifyShapeRegionStructure(::mlir::Operation *op);

/// Verifies that every value used inside an op's shape region is either
/// defined within the region, a block argument of the region, or one of the
/// op's own operands. Referenced by ShapeRegionInterface's verifier.
///
/// Design note: We use custom scoping verification instead of IsolatedFromAbove
/// trait because shape regions need selective isolation:
/// - IsolatedFromAbove: Region cannot access ANY outer values (complete isolation)
/// - verifyShapeRegionScoping: Region CAN access the op's operands, but nothing else
///
/// Shape regions need to read the op's inputs to compute output shapes,
/// e.g., `tensor.dim %input, %c0` where %input is the op's operand.
/// With IsolatedFromAbove, this would require verbose region arguments.
/// Current design allows direct access to op operands for ergonomics while
/// preventing accidental capture of arbitrary outer scope values.
::mlir::LogicalResult verifyShapeRegionScoping(::mlir::Operation *op);

} // namespace hipsr
} // namespace mlir
```

---

**lib/Dialect/Hipsr/IR/HipsrInterfaces.cpp:**

```cpp
#include "hip/Dialect/Hipsr/IR/HipsrInterfaces.h"
#include "hip/Dialect/Hipsr/IR/HipsrInfrastructureOps.h"
#include "mlir/IR/Operation.h"

using namespace mlir;
using namespace mlir::hipsr;

LogicalResult mlir::hipsr::verifyShapeRegionStructure(Operation *op) {
  auto shapeRegionOp = cast<ShapeRegionInterface>(op);
  Region *shapeRegion = shapeRegionOp.getShapeRegion();
  
  // Verify single block
  if (shapeRegion->empty() || !shapeRegion->hasOneBlock())
    return op->emitOpError("shape region must have exactly one block");
  
  Block &block = shapeRegion->front();
  Operation *terminator = block.getTerminator();
  
  // Verify terminator exists and is the correct type
  if (!terminator)
    return op->emitOpError("shape region block must have a terminator");
  
  if (!isa<ShapeYieldOp>(terminator))
    return op->emitOpError("shape region must terminate with hipsr.shape_yield, got '")
           << terminator->getName() << "'. Did you forget to add "
           << "SingleBlockImplicitTerminator<\"ShapeYieldOp\"> trait?";
  
  return success();
}

LogicalResult mlir::hipsr::verifyShapeRegionScoping(Operation *op) {
  auto shapeRegionOp = cast<ShapeRegionInterface>(op);
  Region *shapeRegion = shapeRegionOp.getShapeRegion();
  
  // Structure already verified by verifyShapeRegionStructure
  Block &block = shapeRegion->front();
  
  // Verify scoping: walk all operations and check each operand
  for (Operation &nestedOp : block) {
    for (Value operand : nestedOp.getOperands()) {
      Operation *definingOp = operand.getDefiningOp();
      
      // Case 1: Block argument - allowed
      if (!definingOp)
        continue;
      
      // Case 2: Defined within shape region - allowed
      if (shapeRegion->isAncestor(definingOp->getParentRegion()))
        continue;
      
      // Case 3: Op's operand - allowed
      bool isOpOperand = false;
      for (Value opOperand : op->getOperands()) {
        if (operand == opOperand) {
          isOpOperand = true;
          break;
        }
      }
      if (isOpOperand)
        continue;
      
      // Case 4: Disallowed outer value
      return op->emitOpError("shape region references disallowed outer value");
    }
  }
  
  return success();
}

#include "hip/Dialect/Hipsr/IR/HipsrInterfaces.cpp.inc"
```

---

This defensive verification catches the "forgot trait" mistake immediately with a clear error message instead of crashing downstream.

wcy123:
Minor: Update the example in `shape_yield` description to match the simplified assembly format.

**In `HipsrInfrastructureOps.td`:**

```tablegen
Example:
```mlir
hipsr.shape_yield %d0, %d1 : index, index  // ❌ Old format with types
```
```

Should be:

```tablegen
Example:
```mlir
hipsr.shape_yield %d0, %d1  // ✅ Types omitted (per assemblyFormat change)
```
```

The assembly format was correctly updated to omit redundant types, but the example still shows the old format.

**Comment by wcy123** [comment]
**Design Rationale: Custom Verification vs IsolatedFromAbove**

The custom scoping verifier is the right choice for shape regions. Here's the detailed rationale:

**Why Not IsolatedFromAbove?**

Custom verification enables concise syntax:

```mlir
hip.matmul ins(%input, %weight) outs(%init)
  shape_region {
    %d0 = tensor.dim %input, %c0  // ← Direct operand access
    hip.shape_yield %d0, %c1024
  }
```

IsolatedFromAbove would require mirroring all operands as block arguments:

```mlir
hip.matmul ins(%input, %weight) outs(%init)
  shape_region(%input_arg: tensor<?x?xf16>,     // ← Boilerplate
               %weight_arg: tensor<?x1024xf16>,
               %init_arg: tensor<?x1024xf16>) {
    %d0 = tensor.dim %input_arg, %c0
    hip.shape_yield %d0, %c1024
  }
```

**Cost comparison:**
- Custom verifier: ~70 lines (one-time)
- IsolatedFromAbove: ~5 lines/op × 70+ ops = 350+ lines boilerplate

**MLIR Upstream Pattern:**

Operations WITHOUT IsolatedFromAbove (allow implicit capture):
- `scf.for` - loop regions access outer values directly
- `scf.if` - conditional regions access outer values directly
- `scf.while` - loop regions access outer values directly

Example from MLIR upstream tests:

```mlir
func.func @std_if(%arg0: i1, %arg1: f32) {
  scf.if %arg0 {
    %0 = arith.addf %arg1, %arg1 : f32  // ← Implicit capture
  }
}
```

Operations WITH IsolatedFromAbove (require explicit arguments):
- `func.func` - function boundaries require explicit interface

**Pattern:** Control-flow constructs use implicit capture; function-like constructs use explicit arguments.

**Shape regions align with control flow** - they are local, tightly coupled computations, not reusable functions.

**Verification Strategy:**

Two-phase verification:

1. **Structure** (`verifyShapeRegionStructure`) - Defensive check
   - Single block
   - Terminates with `hipsr.shape_yield`
   - Clear error: "Did you forget SingleBlockImplicitTerminator trait?"

2. **Scoping** (`verifyShapeRegionScoping`) - Selective isolation
   - Allow: values from within region
   - Allow: block arguments (for EndBarrier ops)
   - Allow: operation's operands (direct capture)
   - Forbid: arbitrary outer values

Implementation handles three allowed cases:

```cpp
// Case 1: Block argument (EndBarrier) - allowed
if (isa<BlockArgument>(operand))
  continue;

// Case 2: Nested region (scf.for, scf.if) - allowed
Region *definingRegion = operand.getParentRegion();
if (definingRegion && shapeRegion.isAncestor(definingRegion))
  continue;

// Case 3: Operation's operand - allowed
if (allowedValues.contains(operand))
  continue;
```

This provides ergonomic operand access while preventing accidental capture of arbitrary outer values, following MLIR's control-flow pattern.

**Suggestion: Add Negative Test Cases**

Current tests verify positive cases (round-trip). Add negative tests to prove verifiers catch errors:

```mlir
// test/lit/Dialect/Hipsr/shape_region_verify.mlir
// RUN: hip-mlir-opt %s -verify-diagnostics

// ===== verifyShapeRegionStructure Tests =====

// Test 1: Wrong terminator (catches forgotten trait)
func.func @wrong_terminator() {
  // expected-error@+1 {{shape region must terminate with hipsr.shape_yield, got 'func.return'}}
  "test.shape_region_holder"() ({
    func.return %c0 : index
  }) : () -> ()
}

// Test 2: Missing terminator
func.func @missing_terminator() {
  // expected-error@+1 {{shape region block must have a terminator}}
  "test.shape_region_holder"() ({
  }) : () -> ()
}

// Test 3: Multiple blocks
func.func @multiple_blocks() {
  // expected-error@+1 {{shape region must have exactly one block}}
  "test.shape_region_holder"() ({
    cf.br ^bb1
  ^bb1:
    hipsr.shape_yield
  }) : () -> ()
}

// Test 4: Empty region
func.func @empty_region() {
  // expected-error@+1 {{shape region must have exactly one block}}
  "test.shape_region_holder"() ({
  }) : () -> ()
}

// ===== verifyShapeRegionScoping Tests =====

// Test 5: Disallowed outer value (not operand/block-arg/nested)
func.func @disallowed_outer(%input: tensor<?xf32>) {
  %outer = arith.constant 42 : index
  // expected-error@+1 {{shape region references disallowed outer value}}
  "test.shape_region_holder"() ({
    hipsr.shape_yield %outer
  }) {operands = []} : () -> ()
}

// Test 6: Sibling region value (not ancestor)
func.func @sibling_region(%input: tensor<?xf32>) {
  %result = scf.if %cond -> index {
    scf.yield %c0 : index
  } else {
    scf.yield %c1 : index
  }
  // expected-error@+1 {{shape region references disallowed outer value}}
  "test.shape_region_holder"() ({
    hipsr.shape_yield %result
  }) {operands = []} : () -> ()
}
```

**Positive tests (verify allowed patterns):**

```mlir
// RUN: hip-mlir-opt %s
// (No -verify-diagnostics: should parse successfully)

// Test 7: Block arguments allowed (EndBarrier ops)
func.func @block_args_allowed() {
  "test.shape_region_holder"() ({
  ^bb0(%result: tensor<?xi64>):
    %d = tensor.dim %result, %c0 : tensor<?xi64>
    hipsr.shape_yield %d
  }) : () -> ()
}

// Test 8: Nested region values allowed (scf.for/scf.if)
func.func @nested_region_allowed(%input: tensor<?xf32>) {
  "test.shape_region_holder"() ({
    %max = scf.if %cond -> index {
      %d = tensor.dim %input, %c0 : tensor<?xf32>
      scf.yield %d : index
    } else {
      scf.yield %c1 : index
    }
    hipsr.shape_yield %max
  }) {operands = [%input]} : () -> ()
}

// Test 9: Operation operands allowed
func.func @operands_allowed(%input: tensor<?x?xf32>) {
  "test.shape_region_holder"() ({
    %d0 = tensor.dim %input, %c0 : tensor<?x?xf32>
    %d1 = tensor.dim %input, %c1 : tensor<?x?xf32>
    hipsr.shape_yield %d0, %d1
  }) {operands = [%input]} : () -> ()
}
```

wcy123 (approved):
LGTM

PR #475, opened 2026-07-13
feat(dialect): add hipsr memory-space attribute

zpye:
## Summary

Adds a self-contained `#hipsr.mem<...>` memory-space attribute to the `hipsr` dialect:

- A `MemorySpaceKind` enum with four kinds — `host`, `device`, `pinned`, `managed` (numeric values matching the AMDGPU address spaces).
- A `MemorySpaceAttr` (`#hipsr.mem<device>`) usable as a `memref` memory space, e.g. `memref<4xf32, #hipsr.mem<device>>`.

This is **definitions only** — no type constraints, ops, lowering, or runtime. Those are deferred to later stacked PRs to keep this one small and easy to review. Follows the same wiring pattern as the existing `hip` dialect (enum + attrdef tablegen, `useDefaultAttributePrinterParser`, `addAttributes` in `initialize()`).

## Design note

Every hipsr `memref` is intended to carry an explicit memory space; a memref with no memory space is not allowed. Unlike the `hip` dialect, hipsr has no lenient "accept unspecified space" toggle. This PR documents that invariant at the definition site and demonstrates it in the test (all memrefs carry a space); the checks that mechanically enforce it land with the future type-constraints PR.

## Test plan

- [x] `HipsrAttributes.td` builds (enum + attrdef `.inc` generated, `HipsrDialect.cpp` compiles and links).
- [x] New LIT test `test/lit/Dialect/Hipsr/memory-space-attr.mlir` passes:
  - all four kinds round-trip as a memref memory space,
  - the attribute round-trips as a plain attribute on an op,
  - an unknown kind fails to parse (diagnostic asserted).
- [x] Full `MorphizenMLIRLitTests` suite passes (100%).

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/CMakeLists.txt
+++ b/include/hip/Dialect/Hipsr/IR/CMakeLists.txt
@@ -12,3 +12,10 @@ mlir_tablegen(HipsrOps.h.inc -gen-op-decls)
 mlir_tablegen(HipsrOps.cpp.inc -gen-op-defs)
 
 add_public_tablegen_target(HipsrDialectIncGen)
+
+set(LLVM_TARGET_DEFINITIONS HipsrAttributes.td)
+mlir_tablegen(HipsrEnums.h.inc -gen-enum-decls)
+mlir_tablegen(HipsrEnums.cpp.inc -gen-enum-defs)
+mlir_tablegen(HipsrAttributes.h.inc -gen-attrdef-decls -attrdefs-dialect=hipsr)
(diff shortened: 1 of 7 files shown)


wcy123 (approved):
Approve.

Numeric values match AMDGPU address spaces (0=flat, 1=global, 2=pinned, 3=managed) - clean lowering path. `genSpecializedAttr = 0` correctly forces `#hipsr.mem<device>` syntax over raw integers. Tests cover all four kinds plus error diagnostic.

PR #476, opened 2026-07-13
feat(dialect): add hipsr strict memory-space memref type constraints

zpye:
## Summary

Add strict per-memory-space `memref` type constraints to the `hipsr` dialect. A `memref` matches only when it carries an explicit `#hipsr.mem<...>` memory space — one with no space (or a space set by some other attribute) does not match.

New constraints in `HipsrTypes.td`, one per space:

- `Hipsr_HostMemRef`, `Hipsr_DeviceMemRef`, `Hipsr_PinnedMemRef`, `Hipsr_ManagedMemRef`
- `Hipsr_TensorOr*MemRef` variants (ranked tensor OR the named-space memref) for ops that run both before and after bufferization.

Each constraint is backed by a C++ predicate in `HipsrTypes.cpp` that checks for a `MemRefType` whose space is a `#hipsr.mem` attribute of the expected kind.

Also splits the `MemorySpaceKind` enum out into `HipsrEnums.td`.

This is foundation-only: the constraints land here but no op adopts them yet, and op-verifier tests are deferred to a follow-up.

## Test plan

- [x] Full build clean.
- [x] LIT suite passes; `hipsr` memory-space round-trip and bad-kind diagnostic verified.
- [x] `pre-commit` passes on all changed files.


wcy123 (approved):
Clean foundation for strict memory-space type checking. The new `Hipsr_*MemRef` constraints enforce that every memref carries an explicit `#hipsr.mem<kind>` attribute - memrefs with no space or non-hipsr spaces correctly rejected by the predicates.

File split (HipsrAttributes.td → HipsrEnums.td + HipsrAttrs.td) is mostly mechanical; real contribution is the type constraint infrastructure in HipsrTypes.td/h/cpp.

LGTM for foundation work.

PR #477, opened 2026-07-13
perf(linear_attention): parallel dk-split decode kernel (default on)

BoarQing:
## Summary
Parallelize the per-token `linear_attention_decode_kernel` (the largest decode kernel on gfx1151) and make it the default. ~2x kernel speedup, ~10-14% faster decode.

## Problem (verified by SQTT + code)
`linear_attention_decode_kernel` was ~**18.6% of decode GPU-busy** (~108 us/call, ~30 calls/token in the 40-layer hybrid model) and under-parallelized:
- **50% idle threads:** work loops are `for (j=tid; j<dv; j+=nthreads)` but `block=256` while `dv=128` -> half the threads idle.
- **latency stall:** `grid = B*Hkv` (~16 blocks) with too few wavefronts to hide the global recurrent-state (`S`) reads.

## Fix
`linear_attention_decode_kernel_par`: `G` threads per output column split the `dk` reduction (`block = G*dv`), combining partial `sk` and output sums in fp32 via LDS before the single fp16 store.
- Each `S[i,j]` is updated exactly once (disjoint `i`-slices) -> **math matches** the original fp32-internal readout; the recurrence is across tokens, so columns/partitions are independent.
- **Default `G=4` (on).** `HIPDNN_EP_LA_PARALLEL` overrides: `0/1` -> original per-column kernel, `2/8` -> other split. Auto-falls back to the original when the shape can't satisfy the split.

## Results (gfx1151 / Radeon 8050S, `vlm_benchmark.py`, interleaved, unpinned)
- **Kernel:** 108.6 -> **53.1 us/call (~2x)**, SQTT-confirmed (share 18.6% -> 6.7%).
- **Decode P50:** ~**10-14% faster**, repeatable across interleaved rounds (par ran on the warmer GPU each round, so if anything understated).
- **Correctness:** output within the model's run-to-run noise floor (isolated `hip-onnx-runner` L2: par-vs-base logits ~= base-vs-base / par-vs-par; KV cache identical).

## Test plan
- [ ] CI perf run (`vlm_benchmark.py`, `--max_tokens 128 --max_length 8192 -n 5 -w 1`, dog image) on a clean build: confirm decode P50 improves and TPS is stable.
- [ ] Token-identity / accuracy check on a clean build (my local venv has an unrelated prefill GQA bug that garbled end-to-end text, so I validated numerics via the isolated runner instead).
- [ ] Sanity: `HIPDNN_EP_LA_PARALLEL=1` reproduces the original kernel exactly.
- [ ] (Follow-up) apply the same fix to `linear_attention_prefill_chunked`.

### DIFF
--- a/lib/Runtime/Kernels/hip/linear_attention_kernel.hip
+++ b/lib/Runtime/Kernels/hip/linear_attention_kernel.hip
@@ -9,6 +9,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <cmath>
+#include <cstdlib>
 
 #include "hip_custom_kernels.h"
 #include "debug_log.h"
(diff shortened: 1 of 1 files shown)


BoarQing:
CI Perf: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1301/ 
Accuracy: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1317/

**Note: Only VLM models have linear attention. The performance diff for other models are noise**

amd-mingw (approved):
LGTM

PR #480, opened 2026-07-14
feat(dialect): add Hipsr_DpsOp shaped-op base class

mingyueliuh:
Add `Hipsr_DpsOp`, the base class for shaped compute ops (MatMul, Add, ...). Each op built on it:

- carries a shape region (region 0) terminated by `hipsr.shape_yield`, via the `SingleBlockImplicitTerminator<"ShapeYieldOp">` trait;
- implements `ShapeRegionInterface` (single source of truth for shape computation; each op provides `populateShapeRegion()`);
- uses destination-passing style (`DestinationStyleOpInterface`, `getDpsInitsMutable` op-defined);
- yields `Variadic<AnyRankedTensor>` results.

Regular and StartBarrier ops use exactly this one shape region; EndBarrier ops (which also need a capacity shape region) will extend it separately.

No concrete op uses the base yet, so it generates no code — this is pure scaffolding. The first shaped op (`hipsr.matmul`) will exercise the shape-region / reify / DPS path end-to-end (and add lit coverage) in a follow-up.

## Test plan

- Build passes; TableGen resolves the base class's interface/trait/type references.
- LIT suite green (0 failures).

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrOps.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrOps.td
@@ -7,8 +7,10 @@
 #define HIPSR_OPS
 
 include "mlir/IR/OpBase.td"
+include "mlir/Interfaces/DestinationStyleOpInterface.td"
 include "hip/Dialect/Hipsr/IR/HipsrDialect.td"
 include "hip/Dialect/Hipsr/IR/HipsrTypes.td"
+include "hip/Dialect/Hipsr/IR/HipsrShapeRegionInterface.td"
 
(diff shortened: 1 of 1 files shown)


wcy123 (approved):
LGTM. Clean base class design:

- Shape region structure enforced via `SingleBlockImplicitTerminator<"ShapeYieldOp">`
- Interface contracts clear (populateShapeRegion + getDpsInitsMutable)
- Proper trait composition with `!listconcat`
- DPS pattern correctly applied

Ready for first concrete op (hipsr.matmul) to exercise the full path.

PR #483, opened 2026-07-14
feat(dialect): add hipsr.cast op and onnx.Cast -> hipsr.cast conversion

mingyueliuh:
﻿Adds the `hipsr.cast` op and the `onnx.Cast -> hipsr.cast` conversion. 

## What

- `hipsr.cast`: elementwise dtype cast, destination-passing style, implements `ShapeRegionInterface`.
- `onnx.Cast -> hipsr.cast` conversion, registered in `convert-onnx-to-hipsr`.

## Shape region

Per the design, shape regions are filled by a dedicated later pass, not at conversion time. The conversion therefore emits `hipsr.cast` with an **empty** shape region (printed via an optional group, so it round-trips as empty). `populateShapeRegion` is the single source of truth for the output shape and is invoked by that later pass (hipsr-populate-shape-region).

## Tests

- `test/lit/Conversion/onnx-to-hipsr/cast.mlir`: `onnx.Cast -> hipsr.cast` with an empty shape region.
- `test/lit/Dialect/Hipsr/shape_region_verify.mlir`: negative structural + scoping tests (empty block, multiple blocks, missing terminator, wrong terminator, illegal outer capture).

## Not covered

This PR has no way to exercise `populateShapeRegion`: the conversion leaves the shape region empty and no pass invokes `populateShapeRegion` yet. It will be covered by the follow-up PR that adds the dedicated shape-region population pass, so no `populateShapeRegion` test case is added here.


wcy123 (changes requested):
Good foundation work, but found a type constraint inconsistency:

**Issue:** `HipsrCastOp.td` signature accepts `Hipsr_TensorOrDeviceMemRef` (both tensor and memref), but `HipsrCastOp.cpp:27` uses `tensor::DimOp` which only accepts tensors. If a memref is passed, the operation will fail.

**Fix:** Use shape dialect for unified tensor/memref handling (aligns with pool-allocs design):

```cpp
void CastOp::populateShapeRegion(OpBuilder &builder, Region &shapeRegion) {
  auto shapedTy = cast<ShapedType>(getInput().getType());
  SmallVector<Value> dims;
  
  // Shape dialect: works for BOTH tensor and memref
  Value shape = builder.create<shape::ShapeOfOp>(loc, getInput());
  
  for (int64_t i = 0; i < shapedTy.getRank(); ++i) {
    Value dim = builder.create<shape::GetExtentOp>(
      loc, builder.getIndexType(), shape, builder.getIndexAttr(i));
    dims.push_back(dim);
  }
  builder.create<ShapeYieldOp>(loc, dims);
}
```

Benefits: unified handling (no type dispatch), static dimensions fold automatically, returns index directly.

Everything else looks good - proper base class usage, scoping verification works correctly, zero-operand builder enables auto-terminator insertion.

wcy123:
**Additional issue: `populateShapeRegion()` never tested**

The tests are round-trip tests (parse IR → print IR), which do NOT exercise `populateShapeRegion()`. The shape region content in test files is manually written, not generated by the implementation.

Example from `cast.mlir`:

```mlir
%0 = hipsr.cast ins(%input : tensor<?x8xf32>) outs(%init : tensor<?x8xf16>)
  shape_region {
    %d0 = tensor.dim %input, %c0 : tensor<?x8xf32>
    hipsr.shape_yield %d0, %c8
    // ^ Manually written - populateShapeRegion() never runs
  }
```

The "single source of truth" functionality (the main point of the design) is completely untested. The implementation could be wrong and tests would still pass.

**To properly test:** Need a test that programmatically creates the op (which calls `populateShapeRegion()`), then verifies the generated shape region content. Options:
1. Conversion test (e.g., `onnx.Cast` → `hipsr.cast`)
2. Test pass that explicitly invokes `populateShapeRegion()` and verifies output
3. C++ unit test

wcy123 (commented):
LGTM.

**Note:** `populateShapeRegion` is not tested end-to-end in this PR. Depends on https://github.com/wcy123/onnx-hipdnn-ep/issues/106 - should be tested properly in next PR that adds the shape region population pass.

wcy123 (approved):
LGTM

PR #484, opened 2026-07-14
feat(conversion): add convert-onnx-to-hipsr pass framework

mingyueliuh:
Scaffold the ONNX -> hipsr conversion pass. Framework only: the pass is registered and runs a (currently empty) greedy pattern set. Per-op conversion patterns are added by follow-up layers (starting with Cast) that introduce both the hipsr op and its pattern.

This is the base of the stack `main -> hipsr-onnx-to-hipsr -> hipsr-cast`, so all Cast-related changes (the op and its conversion pattern + tests) live in the Cast PR on top.

## What

- Add the pass under `include/hip/Conversion/OnnxToHipsr/` + `lib/Conversion/OnnxToHipsr/`, matching the repo's (and upstream MLIR's) convention for cross-dialect conversions (alongside `OnnxToHip` / `OnnxToHipDNN` / `HipToLLVM`), not a dialect's `Transforms/` dir.
- Follow the upstream MLIR aggregate layout: pass `def`s live in one `include/hip/Conversion/Passes.td` (tablegen target `HipsrConversionPassIncGen`); the per-conversion header `OnnxToHipsr/OnnxToHipsr.h` pulls just its declaration via `GEN_PASS_DECL_CONVERTONNXTOHIPSRPASS` (no hand-written factory); `Conversion/Passes.h` does the `GEN_PASS_REGISTRATION`.
- Register through `registerConversionPasses()` (`lib/Conversion/Passes.cpp`) like the other conversions, via the generated umbrella `registerHipsrConversionPasses()`. `InitAllPasses.h` is unchanged vs main.
- Greedy driver config mirrors `convert-onnx-to-hip` (`ExistingOps`). This stage does not populate shape regions -- shape regions are filled uniformly by a dedicated later pass -- so it emits no in-region constants and needs no constant-CSE workaround.

## Review feedback addressed

- Directory structure: moved from `Dialect/Hipsr/Transforms/` to `Conversion/OnnxToHipsr/` (+ aggregate `Conversion/Passes.td`).
- Redundant factory declaration: removed; now uses `GEN_PASS_DECL_CONVERTONNXTOHIPSRPASS`.
- `enableConstantCSE(false)`: removed. Rather than work around constant hoisting breaking shape-region self-containment, the design moves shape-region population out of this stage into a later dedicated pass, so the problem does not arise here.

## Tests

No lit test: with no patterns the pass is a no-op, so a test would only assert "registered and doesn't crash". The first real conversion (Cast, next PR) provides the end-to-end test.

## Test plan

- Build + install pass; `convert-onnx-to-hipsr` is registered.
- LIT suite green (0 failures).


wcy123 (changes requested):
Review findings - recommend following modern MLIR best practices:

**1. Directory structure** - Conversion pass headers should be in `Conversion/`, not `Transforms/`:

Current:
```
include/hip/Dialect/Hipsr/Transforms/Passes.{h,td}  ← Transforms = same dialect
lib/Conversion/OnnxToHipsr/                         ← Conversion = ONNX → hipsr
```

MLIR convention: conversions (dialect A → B) in `Conversion/`, transforms (same dialect) in `Transforms/`. Compare with `OnnxToHip` which correctly uses `include/hip/Conversion/OnnxToHip/`.

Suggest: Move headers to `include/hip/Conversion/OnnxToHipsr/` to match MLIR conventions and existing `OnnxToHip` pattern.

**2. Redundant factory declaration** - `Passes.h` manually declares `createConvertOnnxToHipsrPass()` which TableGen auto-generates:

Modern MLIR pattern ([mlir/Transforms/Passes.h](https://mlir.llvm.org/doxygen/Transforms_2Passes_8h_source.html), [D131839](https://reviews.llvm.org/D131839)):
```cpp
#define GEN_PASS_DECL_CONVERTONNXTOHIPSRPASS  // Auto-generates declaration
#define GEN_PASS_REGISTRATION
#include "Passes.h.inc"
// No manual createPass() unless custom constructor needed (e.g., with options)
```

Your `.cpp` correctly uses `GEN_PASS_DEF_CONVERTONNXTOHIPSRPASS` which auto-generates the factory definition. The manual declaration in `.h` is unnecessary since this pass has no custom constructor.

Suggest: Remove manual declaration, add `GEN_PASS_DECL_CONVERTONNXTOHIPSRPASS` instead.

**3. `config.enableConstantCSE(false)` - Necessary?**

This doesn't exist in `OnnxToHip` (origin/main) Is this:
- Preventative for when Cast PR adds shape regions?
- Based on evidence that CSE hoists constants from regions?
- Workaround for MLIR CSE not recognizing custom `IsolatedFromAboveButAllowOperands` trait?

Worth clarifying the specific motivation since the pass is currently a no-op.

**Otherwise solid framework** - implementation correctly uses modern `GEN_PASS_DEF` pattern, build system complete, clean separation from Cast PR.

**References:**
- [MLIR Pass Infrastructure](https://mlir.llvm.org/docs/PassManagement/)
- [D131839: Split autogenerated pass declarations](https://reviews.llvm.org/D131839)
- [MLIR Transforms/Passes.h example](https://mlir.llvm.org/doxygen/Transforms_2Passes_8h_source.html)

wcy123 (dismissed):
LGTM! All review concerns addressed.

**Changes made (since initial review):**

1. ✅ **Directory structure fixed** - Moved to `include/hip/Conversion/OnnxToHipsr/` following MLIR conventions and matching `OnnxToHip` pattern

2. ✅ **Modern MLIR pass pattern** - Removed redundant manual declaration, now uses:
   ```cpp
   #define GEN_PASS_DECL_CONVERTONNXTOHIPSRPASS  // Auto-generates declaration
   #include "hip/Conversion/Passes.h.inc"
   ```

3. ✅ **CSE rationale documented** - Excellent 20+ line comment explaining:
   - ShapeRegionInterface scoping requirement (self-contained regions)
   - Concrete example showing CSE violation
   - Clear justification (correctness, not defensive)

**Architecture improvements:**

- **Better than OnnxToHip** - Uses aggregate `Conversion/Passes.td` mirroring upstream `mlir/Conversion/Passes.td`
- **Modern patterns** - Full `GEN_PASS_DECL`/`GEN_PASS_DEF` split (post-2022 TableGen infrastructure)
- **Well documented** - Design choices clearly explained in code

Ready to merge!

wcy123 (approved):
LGTM

PR #485, opened 2026-07-14
feat(dialect): add hipsr IsolatedFromAboveButAllowOperands trait

zpye:
## Summary

Adds a native op trait `IsolatedFromAboveButAllowOperands` to the `hipsr` dialect and applies it to `Hipsr_DpsOp`.

It works like MLIR's `IsolatedFromAbove` (a region can't reference values from the enclosing scope), but also lets region ops reference **the op's own operands**. Shape regions need this: they read the op's inputs to compute output shapes (e.g. `tensor.dim %input, %c0`), so full isolation would force threading every input through region arguments.

## Notes for reviewers

- The verifier follows upstream `verifyIsIsolatedFromAbove` closely, adding just one rule: the op's operands are allowed.
- This replaces the ad-hoc `verifyShapeRegionScoping` that previously lived in the `ShapeRegionInterface` verifier. The interface now only requires the trait to be present and lets the trait do the scoping check.
- No concrete op carries the trait yet (`Hipsr_DpsOp` is a base class), so scoping is exercised by later PRs that add concrete ops.

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrOps.h
+++ b/include/hip/Dialect/Hipsr/IR/HipsrOps.h
@@ -14,6 +14,9 @@
 #include "mlir/IR/OpImplementation.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
 
+// Must precede the .inc: generated op classes use the trait template.
+#include "hip/Dialect/Hipsr/IR/HipsrTraits.h"
+
 #define GET_OP_CLASSES
 #include "hip/Dialect/Hipsr/IR/HipsrOps.h.inc"
 
--- a/include/hip/Dialect/Hipsr/IR/HipsrOps.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrOps.td
@@ -10,6 +10,7 @@ include "mlir/IR/OpBase.td"
 include "mlir/Interfaces/DestinationStyleOpInterface.td"
 include "hip/Dialect/Hipsr/IR/HipsrDialect.td"
 include "hip/Dialect/Hipsr/IR/HipsrTypes.td"
+include "hip/Dialect/Hipsr/IR/HipsrTraits.td"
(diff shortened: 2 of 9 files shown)


wcy123 (approved):
LGTM - clean refactor that fixes a scoping bug and follows MLIR patterns.

**Bug fix:** Old `verifyShapeRegionScoping` skipped ALL block arguments with `if (isa<BlockArgument>(operand)) continue;`. This incorrectly allowed outer block arguments:

```mlir
func.func @outer(%arg0: index) {
  %0 = hipsr.cast ... shape_region {
    hipsr.shape_yield %arg0  // BUG: %arg0 is outer scope, should be rejected!
  }                           // Old: skipped (block arg), New: caught
}
```

New implementation fixes this by using `operand.getParentRegion()` which works for BOTH:
- **Operation results:** returns the defining op's region
- **Block arguments:** returns the owning block's region

So `region.isAncestor(operandRegion)` correctly:
- Allows block arguments from shape region itself (shape region CAN have block args - EndBarrier ops use this)
- Allows block arguments from nested regions (e.g., scf.if inside shape region)  
- Rejects block arguments from outer scopes (the bug - old code skipped ALL block args)

**Suggestion:** Add code comment explaining the key insight:

```cpp
// operand.getParentRegion() works for both op results and block arguments.
// For block arguments, returns owning block's region. This allows shape
// region's own block args while correctly rejecting outer block args
// (which the old blanket skip incorrectly allowed).
if (!region.isAncestor(operandRegion) && !allowedOperands.contains(operand)) {
```

**Other improvements:**

- Follows MLIR patterns: mirrors `IsolatedFromAbove`, uses `verifyRegionTrait` hook (runs after nested ops verified)
- Respects nested isolation: skips ops with their own isolation traits
- Simplifies interface: removes 42 lines of ad-hoc code, delegates to reusable trait
- Handles all regions: not hardcoded to region 0

PR #490, opened 2026-07-15
feat(dialect): redesign hipsr.shape_yield with VariadicOfVariadic operands

zpye:
## What

Redesign `hipsr.shape_yield` to describe each result's shape as its own group of dims, and add matching accessors to `ShapeRegionInterface`.

## Changes

- **Op:** `hipsr.shape_yield` now takes `VariadicOfVariadic<Index, "ranks">` shapes, a `DenseI32ArrayAttr` rank-segment attr, and a `TypeArrayAttr` of per-result element types. Assembly prints dim groups in parens and element types in brackets, e.g. `hipsr.shape_yield (%m, %n) : [f16]`.
- **Verifier:** only checks the element-type count (ODS already verifies the rank segmentation).
- **Interface accessors:** `getShapeRegionResultShapes` (per-result dim values) and `getShapeRegionResultTypes` (per-result `RankedTensorType`, all extents dynamic), plus capacity-region siblings. Logic lives in free functions in the `.cpp`; the interface methods just delegate, so the interface header stays op-agnostic.

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrShapeRegionInterface.h
+++ b/include/hip/Dialect/Hipsr/IR/HipsrShapeRegionInterface.h
@@ -6,8 +6,12 @@
 #ifndef HIPSR_SHAPE_REGION_INTERFACE_H
 #define HIPSR_SHAPE_REGION_INTERFACE_H
 
+#include "mlir/IR/BuiltinTypes.h"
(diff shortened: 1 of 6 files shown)


wcy123 (approved):
LGTM

wcy123 (approved):
LGTM

PR #493, opened 2026-07-16
feat(dialect): add hipsr.pool_domain op with pool_domain_yield terminator

zpye:
Add the `hipsr.pool_domain` region op and its `hipsr.pool_domain_yield` terminator, defined in separate TableGen files.

- `pool_domain` scopes a group of allocations that share a lifetime.
- `pool_domain_yield` returns the domain results and is inserted implicitly as the region terminator.
- `pool_domain` is `IsolatedFromAboveButAllowOperands` with a single-block region and a verifier that checks the yield operands match the domain results in count and type.


wcy123 (approved):
LGTM

PR #494, opened 2026-07-16
feat(dialect): allow an empty (unpopulated) shape region on Hipsr_DpsOp

mingyueliuh:
﻿## Why

`hipsr` shaped ops (those built on `Hipsr_DpsOp`) are created **before** their output-shape computation exists. In the ONNX -> hipsr flow, the ops are emitted first, and a **separate, dedicated pass later populates every op's shape region uniformly** (keeping shape computation as a single source of truth, with consistent timing). For that to work, an op must be able to legally exist with **no shape computation yet** -- i.e. the shape region has to be **optional**.

The previous base (`SizedRegion<1>` + `SingleBlockImplicitTerminator<"ShapeYieldOp">`) forced the opposite:
- `SizedRegion<1>` requires **exactly one block**, so an unpopulated (empty) region is illegal.
- `SingleBlockImplicitTerminator` **auto-inserts** a `shape_yield` on parse and **elides** it on print. So a region left empty silently becomes one holding a bare `shape_yield`, and "absent" vs "present-but-empty" are indistinguishable in text and do **not** round-trip stably.

## What

- `Hipsr_DpsOp`: `SizedRegion<1>` -> `MaxSizedRegion<1>` (0 or 1 block), and `SingleBlockImplicitTerminator<"ShapeYieldOp">` -> `SingleBlock`.
  - `MaxSizedRegion<1>` allows the 0-block (unpopulated) state.
  - `SingleBlock` keeps a non-empty region to a single block but neither auto-inserts nor elides the terminator, so an empty region **stays empty** (and round-trips faithfully) while a populated region prints its `shape_yield` explicitly. "Absent" becomes a first-class, stable, distinguishable state.
- `verifyShapeRegionStructure`: accept an empty region; a **non-empty** region must still be a single block ending with `hipsr.shape_yield`.
- Update `ShapeRegionInterface` docs to match.

## Notes

- Infra-only. No concrete `Hipsr_DpsOp` op exists on `main` yet, so there is no functional test here; the consumer + tests land in #483 (stacked on this branch).

## Test plan

- [x] `hip-mlir-opt` builds locally on latest `main` + this change.
- [ ] CI green.

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrOps.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrOps.td
@@ -27,9 +27,14 @@ class Hipsr_Op<string mnemonic, list<Trait> traits = []> :
 //===----------------------------------------------------------------------===//
 
 // Base for shaped compute ops (MatMul, Add, ...). Every such op:
-//   - carries a shape region (region 0) that computes its output shape and is
-//     terminated by `hipsr.shape_yield`. The SingleBlockImplicitTerminator
-//     trait auto-inserts that terminator and structurally checks it;
+//   - carries an optional shape region (region 0): it holds either zero blocks
+//     (no shape computation) or exactly one block that computes its output
+//     shape and is terminated by `hipsr.shape_yield`. MaxSizedRegion<1> allows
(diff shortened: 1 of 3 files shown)


wcy123 (approved):
LGTM

PR #496, opened 2026-07-16
feat(dialect): add StartBarrier / EndBarrier pool-domain op interfaces

zpye:
## Summary

Adds two op interfaces to the hipsr dialect for the pool-allocs shape-region work:

- `StartBarrierInterface` — marks an op that starts a new pool domain. Its output shape comes from its input data (e.g. Expand). Method `isStartBarrier()` defaults to `true`.
- `EndBarrierInterface` — marks an op that ends the current pool domain. Its output shape comes from its output data (e.g. NonZero). Method `isEndBarrier()` defaults to `true`.

Whether an op is a barrier can depend on its operands: an Expand with a constant shape operand is not a barrier (its output shape is static), while one with a runtime shape operand is. Each op can override the default per instance to reflect this.

Each interface lives in its own file (`HipsrStartBarrierInterface.{td,h,cpp}`, `HipsrEndBarrierInterface.{td,h,cpp}`), matching the sibling `ShapeRegionInterface` layout. No ops adopt them yet.

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/CMakeLists.txt
+++ b/include/hip/Dialect/Hipsr/IR/CMakeLists.txt
@@ -11,6 +11,14 @@ set(LLVM_TARGET_DEFINITIONS HipsrShapeRegionInterface.td)
 mlir_tablegen(HipsrShapeRegionInterface.h.inc -gen-op-interface-decls)
 mlir_tablegen(HipsrShapeRegionInterface.cpp.inc -gen-op-interface-defs)
 
+set(LLVM_TARGET_DEFINITIONS HipsrStartBarrierInterface.td)
+mlir_tablegen(HipsrStartBarrierInterface.h.inc -gen-op-interface-decls)
+mlir_tablegen(HipsrStartBarrierInterface.cpp.inc -gen-op-interface-defs)
+
+set(LLVM_TARGET_DEFINITIONS HipsrEndBarrierInterface.td)
+mlir_tablegen(HipsrEndBarrierInterface.h.inc -gen-op-interface-decls)
+mlir_tablegen(HipsrEndBarrierInterface.cpp.inc -gen-op-interface-defs)
+
 set(LLVM_TARGET_DEFINITIONS HipsrShapeYieldOp.td)
 mlir_tablegen(HipsrShapeYieldOp.h.inc -gen-op-decls)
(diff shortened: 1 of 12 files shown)


wcy123 (changes requested):
The pure marker trait approach cannot handle conditional barrier behavior. Operations like Expand and Reshape are barriers only when their shape operand is non-constant - this requires runtime inspection of the IR, not static trait attachment.

**Problem:**

```mlir
// Same operation, different barrier status based on operands
%c = arith.constant dense<[1,2,3]> : tensor<3xi64>
%expand1 = hipsr.expand(%data, %c)  // NOT a barrier (constant shape)

%shape = hipsr.compute(...)
%expand2 = hipsr.expand(%data, %shape)  // IS a barrier (runtime shape)
```

Traits are attached at definition time and cannot inspect operands. Need interfaces with `isStartBarrier()` / `isEndBarrier()` methods that check IR at partition time.

**Recommended approach:**

Replace traits with interfaces:
- `StartBarrierInterface` with `isStartBarrier()` method (default: `true`)
- `EndBarrierInterface` with `isEndBarrier()` method (default: `true`)

Each op implements custom logic:
- Expand: return `false` if shape operand is constant (opt-out), otherwise default `true`
- Reshape: return `false` if shape from `tensor.dim` (opt-out), otherwise default `true`
- NonZero: use default `true` (always barrier)

This enables per-instance conditional behavior while maintaining clean abstraction.

**Design docs updated:**

The following design docs now include conditional barrier interface specifications:

1. [Pool-allocs Refactor](https://amd.atlassian.net/wiki/spaces/~712020a865f68fb11a44a19832319935c49c26/pages/1765599352) - Appendix C: Interface Specifications
   - `isStartBarrier()` / `isEndBarrier()` methods with default `true`
   - Example implementations for Expand, Reshape, NonZero showing conditional logic
   - Updated partition algorithm calling interface methods

2. [Domain Partition Algorithm](https://amd.atlassian.net/wiki/spaces/~712020a865f68fb11a44a19832319935c49c26/pages/1789840183) - Barrier Detection via Interfaces
   - Explanation of conditional barrier checking
   - Interface method signatures and usage patterns
   - Cross-reference to pool-allocs design for complete specs

wcy123 (approved):
LGTM

PR #495, opened 2026-07-16
feat(dialect): add hipsr.constant op

zz002:
## Summary

Add the `hipsr.constant` op (foundation of the hipsr constant subsystem) plus `#hipsr.file_source` / `#hipsr.mem_source` attributes and a dialect-level memory-mapped file cache. Op and attributes only; the passes that produce and consume the op are follow-ups.

## Why

The hipsr migration needs a constant carrier that decouples the op form from file I/O (the current `hip` path couples conversion and serialization in one pass). `hipsr.constant` models the three data-source forms a constant can take (inline value / external file or memory source / externalized offset+size in the sidecar) so the later conversion, externalization, and lowering passes can transform between them without re-deriving shapes or touching private pass state. The result type is `AnyMemRef` rather than a tensor: a constant is fundamentally a read-only buffer, matching the existing pipeline's memref model; the memref-to-tensor bridge for tensor-level consumers is left to the conversion pass. Produced memrefs carry `#hipsr.mem<device>` to stay within the hipsr memory-space discipline.

## What

1. **`hipsr.constant` op** at `include/hip/Dialect/Hipsr/IR/HipsrConstantOp.td` (+ `.h` / `lib/.../HipsrConstantOp.cpp`). Five optional attrs (`value` / `source` / `offset` / `size` / `externalize`), result `AnyMemRef`. The verifier enforces exactly one data-source form `{value} | {source} | {offset+size}` with `offset` / `size` paired. Adds form helpers and a templated `getDataValues<T>()` returning a typed view across the inline / mem-source / file-source forms.
2. **Source attributes** in `include/hip/Dialect/Hipsr/IR/HipsrAttrs.td`: `FileSourceAttr{path, offset, size}` and `MemSourceAttr{address, size}`.
3. **Dialect file cache** on `HipsrDialect` (`getOrLoadFileMap`, an `llvm::StringMap` of `MemoryBuffer`s) backing the file-source path of `getDataValues<T>()`.
4. **Registration / build wiring**: op added to `HipsrDialect::initialize()`; tablegen and library entries in the two `CMakeLists.txt`.
5. **Intentional non-changes**: no onnx-to-hipsr conversion, externalization, or LLVM lowering — this PR is the op/attribute foundation only; those land in follow-up PRs.

## Test plan

- [x] `test/lit/Dialect/Hipsr/constant.mlir`: round-trips all four forms (inline / file_source / mem_source / externalized) and checks the verifier rejects an op carrying two forms.
[...]

PR #500, opened 2026-07-17
feat(hipsr): broaden constant result type and add onnx.Constant conversion

zz002:
﻿## Summary

Broadens `hipsr.constant`'s result type to ranked-tensor-or-device-memref and adds the `onnx.Constant -> hipsr.constant` conversion pattern, the first producer of the constant op that landed in #495. Also picks up the two minor review items from #495.

## Why

#495 landed `hipsr.constant` with an `AnyMemRef` result and a hand-written verifier check forcing `#hipsr.mem<device>`. hipsr compute ops are tensor-DPS, so a memref-only constant would force a `memref->tensor` bridge at every consumer in the conversion pass. Broadening the result type to `Hipsr_TensorOrDeviceMemRef` (already defined in `HipsrTypes.td`) lets the pass produce `hipsr.constant : tensor<...>` directly (pre-bufferization) while still allowing device memref post-bufferization; the ODS type constraint subsumes the verifier's device check, so that check is removed (it also mis-fired on tensor results).

The conversion pattern deliberately keys only on storage form, not size. The "keep small constants inline vs externalize" threshold policy is left to a later externalization pass, keeping this stage pure-IR with no file I/O and no policy, per the three-phase constant-subsystem split.

Alternative considered for rank-0: emitting a scalar `arith.constant` (bare `f32`). Rejected because it breaks the types of existing users that expect `tensor<f32>`; instead we keep the tensor-typed `arith.constant` (compile-time constant, type-preserving), matching the existing `OnnxToHip` behaviour.

## What

1. **Result type** at `include/hip/Dialect/Hipsr/IR/HipsrConstantOp.td`: `AnyMemRef` -> `Hipsr_TensorOrDeviceMemRef`; description updated.
2. **Verifier** at `lib/Dialect/Hipsr/IR/HipsrConstantOp.cpp`: drop the hand-written device memory-space check (now enforced by ODS); keep value-XOR-source and paired offset/size.
3. **Formatting** at `include/hip/Dialect/Hipsr/IR/HipsrConstantOp.h`: brace the two single-statement `if` guards in `getDataValues` (review request from #495).
4. **New conversion pattern** `lib/Conversion/OnnxToHipsr/ConstantConversion.cpp`: `ConstantOpLowering` (a `RewritePattern` matching `onnx.Constant` by name, no onnx-mlir headers). Four branches: rank-0 scalar -> `arith.constant`; rank>=1 inline value -> `hipsr.constant {value}`; `location == "*/_ORT_MEM_ADDR_/*"` -> `mem_source`; other location -> `file_source`.
5. **Registration**: `populateOnnxToHipsrConstantPatterns` declared in `OnnxToHipsr.h`, called from `OnnxToHipsr.cpp::runOnOperation`; `ConstantConversion.cpp` added to the CMake target.
6. **Intentional non-changes**: no size threshold / inline policy here (deferred to the externalization pass); no externalization or LLVM lowering.

## Test plan

- [x] LIT green: `check-hip-mlir-lit` -> 289 discovered, 287 passed, 0 failed, 2 unsupported (pre-existing plugin-gated).
- [x] Updated `test/lit/Dialect/Hipsr/constant.mlir`: non-device result now rejected by the ODS constraint (`result #0 must be ranked tensor or device memref`); added a tensor-result positive round-trip.
- [x] New `test/lit/Conversion/onnx-to-hipsr/test_constant.mlir` covers all four branches.
- [x] `pre-commit run --all-files` clean, no auto-fixes.

## Notes for reviewers

None.

## Checklist

- [x] **Incremental** -- 9 files, ~190 LOC.
- [x] **AI policy** -- pattern and tests drafted with Cursor; every design decision above is human-owned and answerable in review.
- [x] **Reviews resolved**
- [x] **CODEOWNERS routing**

### DIFF
--- a/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
+++ b/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
@@ -6,6 +6,7 @@
 #ifndef HIP_CONVERSION_ONNXTOHIPSR_ONNXTOHIPSR_H
 #define HIP_CONVERSION_ONNXTOHIPSR_ONNXTOHIPSR_H
 
+#include "mlir/IR/PatternMatch.h"
 #include "mlir/Pass/Pass.h"
 
 #include <memory>
@@ -20,6 +21,12 @@ namespace hipsr {
 // hip/Conversion/Passes.h.
 #define GEN_PASS_DECL_CONVERTONNXTOHIPSRPASS
 #include "hip/Conversion/Passes.h.inc"
+
+// Populates patterns that convert `onnx.Constant` into `hipsr.constant`
+// (rank>=1 inline / mem_source / file_source) or `arith.constant` (rank-0
+// scalar). Pure IR rewrite -- no file I/O and no size-threshold policy (that
+// is layered on in the externalization pass).
+void populateOnnxToHipsrConstantPatterns(::mlir::RewritePatternSet &patterns);
 
 } // namespace hipsr
 } // namespace mlir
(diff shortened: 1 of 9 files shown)


wcy123 (approved):
LGTM

PR #501, opened 2026-07-17
feat(hipsr): add hipsr-populate-shape-region pass

mingyueliuh:
## Summary
- **`-hipsr-populate-shape-region` pass** — walk every op implementing `ShapeRegionInterface` and fill each **empty** shape region via `populateShapeRegion()` (idempotent; already-populated regions are skipped). 
- New `Dialect/Hipsr/Transforms` module (mirrors the `hip` dialect's `Transforms` layout), registered via `registerHipsrPasses()` in `InitAllPasses`, linked into `hip-mlir-opt` and `LibHipCompiler`.

## Motivation

`convert-onnx-to-hipsr` leaves shape regions empty by design; a single dedicated pass fills them uniformly. This pass is also the first thing to run `populateShapeRegion()` end-to-end — the region is generated by the op and checked with FileCheck, not hand-written in tests.

## Test

`test/lit/Dialect/Hipsr/cast.mlir` covers: populate on the tensor form, idempotency, and the whole-function walk (two chained casts populated in one run).

### DIFF
--- a/include/hip/Dialect/Hipsr/CMakeLists.txt
+++ b/include/hip/Dialect/Hipsr/CMakeLists.txt
@@ -4,3 +4,4 @@
 ##
 
 add_subdirectory(IR)
+add_subdirectory(Transforms)
--- a/include/hip/Dialect/Hipsr/IR/HipsrCastOp.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrCastOp.td
@@ -47,13 +47,15 @@ def Hipsr_CastOp : Hipsr_DpsOp<"cast"> {
     Hipsr_TensorOrDeviceMemRef:$init
   );
 
-  // The shape region is optional: an empty (zero-block) region prints nothing
-  // and round-trips as empty. A non-empty region prints as
-  // `shape_region { ... }`.
+  // The result clause is optional so the op round-trips in both DPS forms
+  // (cf. Hip_DpsOp): tensor mode prints `-> tensor<...>`; the bufferized memref
+  // mode has no tensor result and omits it entirely. The shape region is
+  // likewise optional: an empty (zero-block) region prints nothing, a non-empty
+  // one prints as `shape_region { ... }`.
   let assemblyFormat = [{
     `ins` `(` $input `:` type($input) `)`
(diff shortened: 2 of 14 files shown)


wcy123 (changes requested):
Three design issues to address:

**1. Missing %ctx threading parameter** ([HipsrCastOp.td:54-58](https://github.com/ROCm/hip-ep/blob/bc7b373f175d50f2a16aa60e55b8df47f7ed97ca/include/hip/Dialect/Hipsr/IR/HipsrCastOp.td#L54-L58))

`hip.cast` signature: `hip.cast(%ctx) ins(...) outs(...)`
`hipsr.cast` signature: `hipsr.cast ins(...) outs(...)`

Should hipsr.cast thread %ctx like hip.cast does, or is context threading only added during lowering?

**2. Syntax inconsistency with hip dialect** ([HipsrCastOp.td:58](https://github.com/ROCm/hip-ep/blob/bc7b373f175d50f2a16aa60e55b8df47f7ed97ca/include/hip/Dialect/Hipsr/IR/HipsrCastOp.td#L58))

`hip.cast` uses: `(: type($result_tensors)^)?`
`hipsr.cast` uses: `(-> type($result_tensors)^)?`

hipsr should stay in line with hip dialect unless there's a compelling reason. Use `:` to match hip.cast or document why `->` is better.

**3. Comment explains "what" not "why"** ([HipsrCastOp.td:50-54](https://github.com/ROCm/hip-ep/blob/bc7b373f175d50f2a16aa60e55b8df47f7ed97ca/include/hip/Dialect/Hipsr/IR/HipsrCastOp.td#L50-L54))

Comment describes what the code does but not why the optional clause is needed. Should explain: bufferization removes results from DPS ops (memref-form has no result), and trying to print a non-existent result type would crash without the optional clause.

Recommend: **Request changes**

mingyueliuh:
**the `hipsr.cast` bufferized-form fix** has been split out of this PR (reverted here) and will land separately,

wcy123 (commented):
LGTM

wcy123 (approved):

PR #502, opened 2026-07-17
feat(hipsr): add -hipsr-externalize-constants pass

zz002:
﻿## Summary

Adds the `-hipsr-externalize-constants` pass (Phase 2 of the hipsr constant subsystem, aligned with the v6 constant design), the first consumer of the constant op's `getDataValues()` / `shouldExternalize()` API from #495. It walks every `hipsr.constant` in the module, stamps a 64-byte aligned cumulative `offset`/`size` on each opted-in one, and (when a FileSystem is injected) writes the bytes to a single shared sidecar. Also extracts the shared sidecar writer into its own library.

## Why

#495 landed `hipsr.constant` with an append-only externalization model (`offset`/`size` are added, `value`/`source` are kept so `getDataValues()` still works) and #500 added the ONNX producer. This pass assigns sidecar placement and writes it.

Design points:

1. **Module-scoped (`ModuleOp` pass).** Offsets accumulate into one sidecar shared by the whole module, so the pass must see every constant at once. A per-function pass would restart offsets at 0 per function (overlapping placements) and write the shared file once per function (races under MLIR's parallel function passes / last-writer-wins when serial).
2. **Reuse the hip.* sidecar writer.** Entries are built as `hip::ConstantEntry` and written through `writeConstantsBinToFileSystem` (`ConstantsIO`), the same layout/streaming logic the hip.* pipeline uses. file_source entries carry only their path/offset (`data = nullptr`) so the writer streams them on demand -- the multi-GB weight file is never mmap'd by this pass.
3. **FileSystem-gated write, no fallback.** Phase 1 (offset/size + entry collection) always runs, so the IR transform is deterministic. The write (Phase 2) happens only when a `morphizen::FileSystem` is injected on the `HipsrDialect` (new non-owning `set/getFileSystem`, mirroring `getOrLoadFileMap`). The sink is not synchronized -- it must be injected before passes are scheduled; the pass only reads it. Standalone runs (`hip-mlir-opt`) inject none and get a pure IR transform.

To avoid coupling the hipsr pass to the legacy `OnnxToHip` conversion lib (which shrinks as hipsr takes over), the shared `ConstantsIO` writer is extracted into its own `HipConstantsIO` library under `lib/Support`. Both `OnnxToHip` and the hipsr pass now depend on that small lib, not on each other.

Out of scope: no module-level descriptor arrays (runtime metadata generation is a separate follow-up); the hybrid / three-emit-mode differentiation is deferred until the hipsr pipeline runs end-to-end.

## What

1. **hipsr Transforms infrastructure** (new): `Passes.td` + `Passes.h` + `-name Hipsr` tablegen; `registerHipsrPasses()` wired into `InitAllPasses.h`. Lib linked into `hip-mlir-opt` and `LibHipCompiler`.
2. **Dialect FileSystem sink** at `HipsrDialect.td`/`.h`: non-owning `set/getFileSystem` + forward-declared `morphizen::FileSystem`.
3. **The pass** `lib/Dialect/Hipsr/Transforms/ExternalizeConstants.cpp`: `ModuleOp`-scoped. Phase 1 builds `ConstantEntry` (file_source -> path/offset, no read; inline/mem_source -> `getDataValues<uint8_t>()` byte view; `splat_elem_size = 0`) and stamps `offset`/`size`. Phase 2 calls `writeConstantsBinToFileSystem` once when a FileSystem is present. Option `constants-file`.
4. **ConstantsIO extraction**: moved `ConstantsIO.{h,cpp}` from `Conversion/OnnxToHip` to `Support` as a standalone `HipConstantsIO` target (namespace `mlir::hip` unchanged).
5. **Intentional non-changes**: no size threshold / inline policy; no module descriptor arrays; read side `getOrLoadFileMap` untouched.

## Test plan

- [x] **LIT** `test/lit/Dialect/Hipsr/test_externalize_constants.mlir` -- one case per distinct Phase 1 behavior: inline, file_source, cumulative 64-byte alignment, cross-function accumulation (two funcs in one module -> offsets 0/1024), opt-out, idempotent skip. CHECK directives grouped per case.
- [x] **Unit test** `test/unit/test_externalize_constants.cpp` (GPU-free, plain main) -- injects an in-memory FileSystem to cover Phase 2: exact bytes written for inline/mem/file sources, alignment + gap/trailing padding, the multi-function single-sidecar contract, and the write-failure path. 20/20 checks pass.
- [x] `check-hip-mlir-lit` -> 293 discovered, 291 passed, 0 failed, 2 unsupported (pre-existing plugin-gated).
- [x] Full build (`hip-compiler`, `hip-compiler.dll`, `hipgpu.dll`) links after the ConstantsIO extraction; OnnxToHip constant LIT still passes.

## Notes for reviewers

- Append-only keeps the (possibly large) inline `value` `DenseElementsAttr` in the IR after externalization -- required by the verifier (value XOR source is always present) and by `getDataValues()`. Known trade-off.
- Rebased onto latest main (picks up the new hipsr.cast / hipsr.placeholder ops); no conflicts.

## Checklist

- [x] **Incremental** -- the ConstantsIO move is a pure relocation (git tracks it as a rename); the rest is new hipsr scaffolding + tests.
- [x] **AI policy** -- pass and tests drafted with Cursor; every design decision above is human-owned and answerable in review.
- [x] **Reviews resolved**
- [x] **CODEOWNERS routing**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,6 +207,9 @@ if(BUILD_HIP_TOOLS)
     # Static compiler-plugin tests (sample plugin + unit test).
     # See docs/design/plugin-interface.md.
     add_subdirectory(test/plugin)
+
+    # MLIR dialect/pass unit tests (GTest; cover Phase 2 write paths LIT can't).
+    add_subdirectory(test/unit)
   endif()
 
   # Out-of-tree plugin dirs (HIPDNN_EP_COMPILER_PLUGIN_PATHS): add each so it can
--- a/include/hip/Dialect/Hipsr/IR/HipsrDialect.h
+++ b/include/hip/Dialect/Hipsr/IR/HipsrDialect.h
@@ -18,6 +18,13 @@
 #include <memory>
 #include <mutex>
 
+// The dialect holds a non-owning FileSystem* (constants sidecar sink). Forward
+// declaration only -- avoids pulling morphizen-foundation into every dialect
+// consumer.
+namespace morphizen {
+class FileSystem;
+} // namespace morphizen
+
 #include "hip/Dialect/Hipsr/IR/HipsrDialect.h.inc"
 
 // Enum header first: MemorySpaceAttr uses MemorySpaceKind.
--- a/include/hip/Dialect/Hipsr/IR/HipsrDialect.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrDialect.td
@@ -35,10 +35,6 @@ def Hipsr_Dialect : Dialect {
     /// Non-owning; the compile driver injects the same FileSystem it hands the
     /// hip.* pipeline. Null when nobody injected one (e.g. hip-mlir-opt), in
     /// which case the pass writes no sidecar (pure IR transform).
-    ///
-    /// Not synchronized: inject once before scheduling passes; passes only read
-    /// it. (Unlike the file-map cache above, this sink has no mutex, which is
-    /// safe only under that inject-before-run contract.)
     void setFileSystem(::morphizen::FileSystem *fs) { fileSystem = fs; }
     ::morphizen::FileSystem *getFileSystem() { return fileSystem; }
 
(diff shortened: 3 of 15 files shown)


wcy123 (dismissed):
LGTM.

**Follow-up:** Issue #122 tracks module metadata generation (Foundation-5) - next step after this PR.

wcy123 (approved):
LGTM

PR #505, opened 2026-07-17
feat(dialect): add hipsr.placeholder op for DPS init operands

zpye:
## Summary

- Add `hipsr.placeholder`, a typed stand-in that mirrors a hipsr DPS op's result type(s). It re-emits the given result type(s) as its own results, which are then passed as the DPS op's `outs` init operands. This lets a DPS op be built without computing its output shape first, and the init/result types always match so the DPS verifier is happy.
- The op is `Pure` and variadic (one result per DPS output). It has no bufferization interface, so it must be replaced before bufferization runs.
- Use it in the `onnx.Cast -> hipsr.cast` conversion: the init operand is now a placeholder mirroring the result type, dropping the manual `tensor.empty` + `tensor.dim` shape computation. The shape region stays the source of truth for output dimensions.
- Wire the op into tablegen, the build, and dialect registration, with LIT tests for placeholder round-trips, the updated cast conversion, and an omitted shape region.

## Why not just use `tensor.empty`

`tensor.empty` cannot express a dynamic dim from the type alone; it requires an `index` SSA value for every `?` in the result type. This is baked into the op:

- Its operands *are* the dynamic sizes: `let arguments = (ins Variadic<Index>:$dynamicSizes);` (`mlir/Dialect/Tensor/IR/TensorOps.td`).
- Its verifier rejects any mismatch between the operand count and the number of dynamic dims:

```cpp
// mlir/lib/Dialect/Tensor/IR/TensorOps.cpp
LogicalResult EmptyOp::verify() {
  return verifyDynamicDimensionCount(getOperation(), getType(),
                                     getDynamicSizes());
}

// mlir/lib/Dialect/Utils/VerificationUtils.cpp
LogicalResult mlir::verifyDynamicDimensionCount(Operation *op, ShapedType type,
                                                ValueRange dynamicSizes) {
  int64_t expectedCount = type.getNumDynamicDims();
  int64_t actualCount = dynamicSizes.size();
  if (expectedCount != actualCount)
    return op->emitOpError("incorrect number of dynamic sizes, has ")
           << actualCount << ", expected " << expectedCount;
  return success();
}
```

So building a `tensor.empty` for a `tensor<?x8xf16>` init forces the caller to first produce each dynamic extent as an `index` value (e.g. `tensor.dim %input, %c0`) and thread it in. That is the exact shape computation we want to defer.

## Shape inference is not needed at conversion time

The conversion already knows the result type from the ONNX op's output, so there is nothing to infer when building the op. Computing the dynamic extents (the `tensor.dim` calls) was only busywork to satisfy `tensor.empty`, not information the conversion actually needed.

`hipsr.placeholder` takes no size operands: it just re-emits the result type. The concrete extents are described once, later, by the DPS op's shape region, so the conversion no longer infers or computes any shape up front.

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrPlaceholderOp.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrPlaceholderOp.td
@@ -1,0 +1,55 @@
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// Licensed under the MIT License.
+//
+// The hipsr.placeholder op: a typed stand-in that mirrors a hipsr DPS op's
+// result type(s) so the DPS op can be built without first computing its output
+// shape. The shaped DPS op base class lives in HipsrOps.td.
+//
+
+#ifndef HIPSR_PLACEHOLDER_OP
+#define HIPSR_PLACEHOLDER_OP
+
+include "mlir/Interfaces/SideEffectInterfaces.td"
+include "hip/Dialect/Hipsr/IR/HipsrOps.td"
+
+//===----------------------------------------------------------------------===//
+// Hipsr_PlaceholderOp
+//===----------------------------------------------------------------------===//
+
+def Hipsr_PlaceholderOp : Hipsr_Op<"placeholder", [Pure]> {
+  let summary = "Typed placeholder for a hipsr DPS op's init operand(s)";
+  let description = [{
+    Stands in for the `outs` init operand(s) of a hipsr DPS op. It takes the DPS
+    op's result type(s) and re-emits the same type(s) as its own results, which
+    are then passed as the DPS op's init operands. This lets a DPS op be built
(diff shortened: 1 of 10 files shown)


include/hip/Dialect/Hipsr/IR/HipsrPlaceholderOp.td
```diff
@@ -0,0 +1,55 @@
...
+    `hipsr.placeholder` is transient: it has no bufferization interface, so it
+    must be replaced before bufferization runs.
+
+    Example:
+    ```mlir
+    %inits:2 = hipsr.placeholder : tensor<?x?xf16>, tensor<?xi64>
+    ```
+  }];
+
+  let results = (outs Variadic<AnyRankedTensor>:$results);
```
wcy123, line 43: it is used for multiple results, mimic hipsr.empty()

wcy123 (approved):
LGTM

PR #507, opened 2026-07-17
feat(dialect): add hipsr.matmul op and onnx.MatMul -> hipsr.matmul conversion

zpye:
## Summary

- Add a DPS `hipsr.matmul` op. Its shape region computes the result shape and asserts the runtime checks for ONNX/NumPy matmul: K dims must match, 1-D operands are promoted then stripped from the result, and batch dims broadcast (equal, or one is 1).
- Add the `convert-onnx-to-hipsr` pattern that turns `onnx.MatMul` into `hipsr.matmul` with a placeholder `init` and an empty shape region (filled in by a later pass).
- Add LIT tests for the dialect round-trip and the conversion.

### DIFF
--- a/test/lit/Conversion/onnx-to-hipsr/matmul.mlir
+++ b/test/lit/Conversion/onnx-to-hipsr/matmul.mlir
@@ -3,11 +3,10 @@
 //
 //===----------------------------------------------------------------------===//
 // Converts onnx.MatMul to hipsr.matmul. The conversion mirrors the result type
-// into a hipsr.placeholder init, threads the !hipsr.context taken from the onnx
-// op's own first operand (the ONNX phase prepends it) onto the op, and leaves
-// the shape region empty (a later pass fills it). It does not branch on
-// rank/shape, so two cases cover it: a plain 2-D matmul and a 1-D-operand
-// matmul (the rank-reducing ONNX case).
+// into a hipsr.placeholder init, threads the !hipsr.context from function arg 0
+// (added in the ONNX phase) onto the op, and leaves the shape region empty (a
+// later pass fills it). It does not branch on rank/shape, so two cases cover
+// it: a plain 2-D matmul and a 1-D-operand matmul (the rank-reducing ONNX case).
 //
 // Positive CHECKs were seeded with mlir/utils/generate-test-checks.py (piped
 // from the -convert-onnx-to-hipsr output) and then refined: captures renamed,
@@ -30,8 +29,7 @@
 // CHECK-NOT:     shape_region
 func.func @matmul_2d(%ctx: !hipsr.context, %a: tensor<?x4096xf16>,
                      %b: tensor<4096x1024xf16>) -> tensor<?x1024xf16> {
-  %0 = "onnx.MatMul"(%ctx, %a, %b)
-      : (!hipsr.context, tensor<?x4096xf16>, tensor<4096x1024xf16>)
+  %0 = "onnx.MatMul"(%a, %b) : (tensor<?x4096xf16>, tensor<4096x1024xf16>)
       -> tensor<?x1024xf16>
   return %0 : tensor<?x1024xf16>
 }
@@ -52,8 +50,7 @@ func.func @matmul_2d(%ctx: !hipsr.context, %a: tensor<?x4096xf16>,
 // CHECK-NOT:     shape_region
 func.func @matmul_1d_rhs(%ctx: !hipsr.context, %a: tensor<?x4096xf16>,
                          %b: tensor<4096xf16>) -> tensor<?xf16> {
-  %0 = "onnx.MatMul"(%ctx, %a, %b)
-      : (!hipsr.context, tensor<?x4096xf16>, tensor<4096xf16>)
+  %0 = "onnx.MatMul"(%a, %b) : (tensor<?x4096xf16>, tensor<4096xf16>)
       -> tensor<?xf16>
   return %0 : tensor<?xf16>
 }
--- a/test/lit/Conversion/onnx-to-hipsr/matmul.mlir
+++ b/test/lit/Conversion/onnx-to-hipsr/matmul.mlir
@@ -1,0 +1,106 @@
+// Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// Licensed under the MIT License.
+//
+//===----------------------------------------------------------------------===//
+// Converts onnx.MatMul to hipsr.matmul. The shape region is left empty (zero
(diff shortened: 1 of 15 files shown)


wcy123 (commented):
Three design issues to address:

**1. Missing %ctx threading parameter** ([HipsrMatMulOp.td:56-66](https://github.com/ROCm/hip-ep/blob/13d213eb6a5478aec269d555b0f6dc01e3f101e9/include/hip/Dialect/Hipsr/IR/HipsrMatMulOp.td#L56-L66))

`hip.cast` signature: `hip.cast(%ctx) ins(...) outs(...)`
`hipsr.matmul` signature: `hipsr.matmul ins(...) outs(...)`

Should hipsr.matmul thread %ctx like hip.cast does, or is context threading only added during lowering?

**2. Syntax inconsistency with hip dialect** ([HipsrMatMulOp.td:66](https://github.com/ROCm/hip-ep/blob/13d213eb6a5478aec269d555b0f6dc01e3f101e9/include/hip/Dialect/Hipsr/IR/HipsrMatMulOp.td#L66))

`hip.cast` uses: `attr-dict (: type($result_tensors)^)?`
`hipsr.matmul` uses: `attr-dict -> type($result_tensors) ...`

hipsr should stay in line with hip dialect unless there's a compelling reason. Use `:` to match hip.cast or document why `->` is better.

**3. Should use shape.cstr_eq instead of cf.assert** ([HipsrMatMulOp.cpp:90-97](https://github.com/ROCm/hip-ep/blob/13d213eb6a5478aec269d555b0f6dc01e3f101e9/lib/Dialect/Hipsr/IR/HipsrMatMulOp.cpp#L90-L97))

Shape constraints should use the shape dialect's `shape.cstr_eq` pattern instead of `arith.cmpi + cf.assert`. Current approach breaks the shape constraint abstraction and prevents shape-aware optimizations.

Shape dialect pattern:

```mlir
%witness = shape.cstr_eq %ka, %kb : index, index
shape.assuming %witness -> tensor<?x?xf16> {
  // computation
  shape.assuming_yield %result : tensor<?x?xf16>
}
```

Benefits: passes can merge witnesses, eliminate redundant checks, hoist constraints, defer checking to a single point. Current cf.assert forces runtime checks that can't be optimized even when shapes are statically known.

Recommend: **Request changes**

test/lit/Conversion/onnx-to-hipsr/matmul.mlir
```diff
@@ -0,0 +1,106 @@
...
+// 2-D x 2-D.
+// CHECK-LABEL: func.func @matmul_2d
+func.func @matmul_2d(%a: tensor<?x4096xf16>, %b: tensor<4096x1024xf16>)
+    -> tensor<?x1024xf16> {
+  // The DPS init is a hipsr.placeholder that mirrors the result type, so no
+  // output-shape computation (tensor.empty / tensor.dim) is emitted here.
+  // CHECK: %[[INIT:.+]] = hipsr.placeholder : tensor<?x1024xf16>
+  // The shape region is left empty (zero blocks) here; a later pass populates
+  // it. The optional region group prints nothing when the region is empty.
+  // CHECK: hipsr.matmul ins(%[[A:.+]], %[[B:.+]] : tensor<?x4096xf16>, tensor<4096x1024xf16>) outs(%[[INIT]] : tensor<?x1024xf16>) -> tensor<?x1024xf16>
```
wcy123: should check `%init = hipsr.placeholder`

wcy123:
**4. Hand-written shape region tests add no value**

`test/lit/Dialect/Hipsr/matmul.mlir` contains 347 lines testing hand-written shape regions. These tests don't verify any source code:

**What the hand-written regions DON'T test:**
- NOT testing `populateShapeRegion()` (that code isn't called - regions are hand-written)
- NOT testing conversion (already covered by `test/lit/Conversion/onnx-to-hipsr/matmul.mlir`)
- NOT testing assembly format (already covered by conversion tests)

**What actually has value:**
- Verifier tests (rank-0 operands, DPS violations) - ~50 lines
- Round-trip with empty shape region - ~20 lines

**The problem:** 347 lines just verify that hand-written MLIR parses correctly. Per [MLIR Testing Guide](https://mlir.llvm.org/getting_started/TestingGuide/), tests should verify actual functionality, not that arbitrary IR parses.

**Recommendation:**

1. **Conversion tests** (`test/lit/Conversion/onnx-to-hipsr/matmul.mlir`): ✅ Keep as-is (106 lines)
2. **Dialect tests** (`test/lit/Dialect/Hipsr/matmul.mlir`): ❌ Remove hand-written regions, keep only verifier tests (~50 lines)
3. **Populate pass tests** (after PR #501): Add new file testing `-hipsr-populate-shape-region` on `hipsr.matmul` - this will verify `populateShapeRegion()` actually works

Current 347-line file is test bloat - it doesn't exercise `populateShapeRegion()` and duplicates conversion coverage.

wcy123 (dismissed):
**LGTM** - all four review issues comprehensively addressed.

1. ✅ **%ctx threading** ([8130d3a1](https://github.com/ROCm/hip-ep/commit/8130d3a1)) - Correctly sourced from `onnx.MatMul` operand[0] with type guard, matching ONNX phase convention
2. ✅ **shape.cstr_eq** ([cc38f21f](https://github.com/ROCm/hip-ep/commit/cc38f21f)) - Proper shape dialect witness pattern, enables optimization passes to merge/eliminate/hoist constraints
3. ✅ **Test cleanup** ([ad4d43c7](https://github.com/ROCm/hip-ep/commit/ad4d43c7)) - Removed redundant coverage, kept only meaningful tests
4. ✅ **Assembly format** - Now uses `:` to match `hip.cast` convention

The shape constraint implementation is particularly well done: static matmuls fold the witness away (zero overhead), dynamic matmuls can have checks merged/hoisted by shape-aware passes.

**Follow-up:** After PR #501 (shape region populate pass) is merged, add test cases that verify `populateShapeRegion()` actually works - run `-hipsr-populate-shape-region` on `hipsr.matmul` and verify the generated shape region matches the expected ONNX/NumPy semantics. The current tests verify conversion (empty region) and verifier behavior, but don't exercise the shape region population code path.

**Minor note (non-blocking):** `MLIRControlFlowDialect` dependency appears in `lib/Dialect/Hipsr/IR/CMakeLists.txt` but isn't used after the `cf.assert` removal (commit 10892fe4). Commit message says it was removed, but the diff shows it being added. Likely just an unused link dependency - can be cleaned up in a follow-up if needed.

wcy123 (approved):
LGTM

PR #513, opened 2026-07-21
refactor(constants): rename sidecar -> constants file / MemSource

zz002:
## Summary

Mechanical naming cleanup for the constant-data subsystem: the legacy term `sidecar` is renamed to `constants file` everywhere it referred to the constant blob, and the FlatBuffer descriptor `SidecarSource`/`sidecar_offset` becomes `MemSource`/`mem_offset`. No behavior change. AI tool (Cursor) was used to perform the rename; every change is a rename I can account for in review.

## Why

`sidecar` was an imprecise legacy name: the runtime always reads the constant blob from `constants.bin` / `model.constants.bin`, and the descriptor formerly called `SidecarSource` specifically identifies a constant whose bytes were packed into that file from a memory address (mem-addr alias), as opposed to `SplatSource` / `FileRefSource`. `MemSource` names that origin directly, and `constants file` names the artifact by what it is. Doing this now, as a standalone rename, keeps the diff reviewable and unblocks upcoming constant-lowering work from inheriting the old vocabulary.

Alternative considered: leave the names as-is. Rejected because the `sidecar` term keeps generating confusion in reviews and the union member name (`MemSource`) is about to be referenced by new op-first metadata code.

## What

1. **FlatBuffer schema** (schemas/model_metadata.fbs): table `SidecarSource` -> `MemSource`, field `sidecar_offset` -> `mem_offset`. Union `ConstantSource` member order is unchanged (`SplatSource`, `FileRefSource`, `MemSource` = kinds 1/2/3), so previously serialized `__metadata_blob` payloads still parse. The generated accessor header is regenerated by flatc at build time (not checked in).
2. **Consumers updated** to the regenerated accessors: lib/Dialect/Transforms/GenerateInterface.cpp (`MemSourceT` / `mem_offset` / reads `hipdnn.constant_mem_offsets`), lib/Runtime/hipdnn_ep_runtime_state.cpp (`source_as_MemSource` / `mem_offset` / local reader helpers), lib/Conversion/OnnxToHip/OnnxToHip.cpp (stamps `hipdnn.constant_mem_offsets`).
3. **Module attribute rename**: `hipdnn.constant_sidecar_offsets` -> `hipdnn.constant_mem_offsets` (producer in OnnxToHip, consumer in GenerateInterface).
4. **Prose rename**: `sidecar` -> `constants file` across comments, docs (README, technical-pipeline-walkthrough), pass `.td` descriptions, and LIT test comments.
5. **Intentional non-changes**: real file names `constants.bin` / `model.constants.bin` / `.constants.bin` / `.constants.json` are untouched. Two Python spots keep the word `sidecar` because there it means an unrelated companion file, not the constant blob: `.data sidecars` in 	est/python/conftest.py (ONNX external-data files) and `_emit_genai_dml_sidecars` in 	ools/onnx-model-splitter/export_chunk_model.py (genai `*_dml.json` configs).

## Test plan

- `python build.py --install_dir D:/ROCm/local --build_dir D:/ROCm/build/hip-ep --cmake_generator Ninja`: build green, flatc regenerated the metadata accessors, LIT 294/296 passed (2 unsupported, unchanged from baseline).
- `ctest -R HipsrExternalizeConstantsUnit`: passed (this unit test contains the renamed test function `testMultiFunctionSharesOneConstantsFile`).
- `pre-commit run --all-files`: converged (clang-format reflowed a few now-longer comment lines; re-staged and re-ran to green).

## Notes for reviewers

Wire compatibility is the one load-bearing invariant: the union member order and the `ConstantInfo` field layout are unchanged, so this is a pure identifier/prose rename with no schema-format break. None otherwise.

## Checklist

- [x] **Incremental.** Standalone rename PR, self-contained, no functional change.
- [x] **AI policy.** Cursor performed the mechanical rename; disclosed here and via the commit trailer. Every change is a rename I can explain in review.
- [x] **No `good first issue` automation.**
- [x] **Reviews resolved.**
- [x] **CODEOWNERS routing.** Touches schemas / runtime / conversion / dialect under the existing ownership map.

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -860,9 +860,9 @@ jobs:
           exit /b %RC%
 
       # ----------------------------------------------------------------------
-      # EPContext export + import perf test (PR #132 sidecar-in-tar path).
+      # EPContext export + import perf test (PR #132 constants-file-in-tar path).
       # Limited to full_model_seq128 to bound runtime: each model triggers
-      # one export pass (~60 s + N GB sidecar write) plus one import perf
+      # one export pass (~60 s + N GB constants-file write) plus one import perf
       # pass (-t 30). The _ctx.onnx + _MORPHIZEN.bin are deleted right after
       # import to keep runner disk usage bounded.
       # ----------------------------------------------------------------------
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ This project demonstrates the integration of HIP (Heterogeneous-compute Interfac
 - **hipBLASLt MatMul**: High-performance matrix multiplication
 - **Custom HIP Kernels**: GQA, RoPE, Cast, Sub, Gather, ReduceSum, Reciprocal, Sqrt, GELU, Range, LinearAttention, Softmax, CausalConvWithState(fast path)
 - **Memory Pool Optimization**: `hip-pool-allocs` pass packs allocations into a single grow-on-demand buffer
-- **Constant Externalization**: Large model weights stored in sidecar `.constants.bin` files
+- **Constant Externalization**: Large model weights stored in `.constants.bin` constants files
 - **Mock Runtime**: GPU-free development and testing with `BUILD_MOCK_RUNTIME=ON`
 - **MorphiZen EP Integration**: Plugs into ONNX Runtime as the MorphiZen Execution Provider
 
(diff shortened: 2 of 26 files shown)


wcy123 (approved):
LGTM

PR #514, opened 2026-07-21
feat(hipsr): lower hipsr.constant to LLVM via ConvertToLLVMPatternInterface

zz002:
## Summary

Lower `hipsr.constant` to LLVM through HipsrDialect's `ConvertToLLVMPatternInterface` rather than a dedicated pass: an externalized `hipsr.constant` becomes an `hipdnn_ep_constant_get(ctx, index)` runtime call wrapped in a memref descriptor. Because HipsrDialect implements the interface, any ConvertToLLVM driver -- this repo's `convert-hip-to-llvm` and the upstream generic `-convert-to-llvm` (now registered in `hip-mlir-opt`) -- discovers and applies hipsr lowering alongside the standard dialects. Delivers the HipSR LLVM lowering scaffold (wcy123/onnx-hipdnn-ep#23) and the Phase 3 constant lowering (wcy123/onnx-hipdnn-ep#105). AI tool (Cursor) was used; each decision is accounted for below.

## Why

Using `ConvertToLLVMPatternInterface` means there is no `-convert-hipsr-to-llvm` pass to maintain: the driver already walks the module and invokes every dialect that implements the interface, so hipsr lowers uniformly with func/memref/arith/cf. The lowering is index-based (`hipdnn_ep_constant_get(ctx, index)` returns `gpu_constants[index]`), reusing the existing runtime entry that `hip.get_constant` already targets; `index` is the constant's ordinal, stamped by `hipsr-externalize-constants`, and `ctx` is the enclosing `llvm.func`'s arg 0 (the runtime-state pointer, per the EP ABI). The interface implementation and the per-op pattern live in the dialect IR library, so hipsr owns its whole LLVM lowering in one place with no separate Conversion sub-library.

Alternatives considered: (a) a dedicated `-convert-hipsr-to-llvm` pass -- rejected, the driver discovers interface dialects so the pass is redundant. (b) offset-based `wrap_get_global(ctx, offset, size)` -- dropped in favour of the existing index entry, so no new runtime symbol is added and the ABI matches `hip.get_constant`. (c) an `externalize` opt-out flag on the op -- removed, it was never set false in production, and a consumer wanting a compile-time value replaces the op with the concrete constant.

## What

1. **Interface + registration** in `lib/Dialect/Hipsr/IR/HipsrDialect.cpp`: `HipsrConvertToLLVMInterface` adds the `#hipsr.mem<kind>` -> integer address-space conversion, marks the hipsr dialect illegal, and aggregates the per-op patterns; `registerConvertHipsrToLLVMInterface(registry)` attaches it (declared in `HipsrDialect.h`, called from `registerAllDialects` and `hip-mlir-opt`).
2. **Constant pattern** in `lib/Dialect/Hipsr/IR/HipsrConstantOp.cpp` (co-located with the op): an externalized `hipsr.constant` lowers to `hipdnn_ep_constant_get(ctx, index)` + address-space cast + row-major memref descriptor; `ctx` = enclosing `llvm.func` arg 0; a constant reaching this before externalization diagnoses instead of mis-lowering.
3. **Op / pass changes**: `hipsr.constant` gains an `index` attribute (grouped with `offset`/`size` in the verifier); the `externalize`/`disableExternalize` opt-out is removed; `hipsr-externalize-constants` stamps `offset`/`size`/`index` and always re-stamps (no idempotency skip -- the single-run pipeline is unaffected and a full re-run is deterministic).
4. **Generic `-convert-to-llvm`** enabled for the hipsr(+standard-dialect) path: `hip-mlir-opt` registers the upstream func/memref/arith/cf/index `ConvertToLLVMPatternInterface`s and the `-convert-to-llvm` pass. The production hip pipeline still lowers through `convert-hip-to-llvm`.
5. **Runtime**: no new symbol -- reuses `hipdnn_ep_constant_get`; the earlier offset-based `wrap_get_global` / `gpu_constants_blob_size` were dropped.
6. **Intentional non-changes**: constant module metadata is not stamped here (moves to wcy123/onnx-hipdnn-ep#122); hip's index constant path (`hip.get_constant` / `gpu_constants[]`) is untouched (removal is a follow-up under epic wcy123/onnx-hipdnn-ep#21); `!hip.context` -> `!llvm.ptr` is left to the hip path, so the generic `-convert-to-llvm` targets IR whose ctx is already `!llvm.ptr`.

## Test plan

- `python build.py --install_dir D:/ROCm/local --build_dir D:/ROCm/build/hip-ep --cmake_generator Ninja`: build green, LIT green (2 unsupported), plugin + externalize unit tests pass.
- `test/lit/Conversion/hipsr-to-llvm/test_constant.mlir` (RUN: `hip-mlir-opt %s --convert-to-llvm`): externalized constants lower to `hipdnn_ep_constant_get(%ctx, %index)` + `addrspacecast ... to !llvm.ptr<1>`, ctx = arg 0.
- `test/lit/Dialect/Hipsr/test_externalize_constants.mlir`: `offset`/`size`/`index` stamped, module-scoped cumulative 64-byte-aligned offsets, and a pre-stamped constant re-stamped from scratch.
- `test/lit/Dialect/Hipsr/constant.mlir`: op round-trip + verifier (exactly one of value/source; offset/size/index grouped; device-memory result).
- `test/unit/test_externalize_constants.cpp`: constants-file bytes match the stamped offsets.
- `pre-commit run --all-files`: converged.

## Notes for reviewers

Not yet end-to-end: hipsr is not sequenced in a production pipeline (`Pipelines.cpp` still runs the onnx->hip path); the hipsr passes are exercised in isolation via LIT. Running a real model through hipsr additionally needs the constant metadata from wcy123/onnx-hipdnn-ep#122 (so the runtime can size/load the blob that `gpu_constants[index]` points into) and a `hipsr-externalize-constants -> ... -> convert-to-llvm` pipeline.

Follow-ups (separate issues): production hipsr pipeline wiring; op-first constant metadata as `hipdnn.metadata` (wcy123/onnx-hipdnn-ep#122); removal of hip's index constant path (epic wcy123/onnx-hipdnn-ep#21); `!hip.context` support in the generic `-convert-to-llvm`; per-source descriptors; inline device-memref lowering via `memref.global`. The negative LIT for the lowering's error paths was dropped; the `emitError` guards remain in the code.

## Checklist

- [x] **Incremental.** Interface + one stateless pattern in the dialect IR lib; hip path untouched.
- [x] **AI policy.** Cursor wrote the code; disclosed here and via commit trailers; each design decision is explained above and answerable in review.
- [x] **No `good first issue` automation.**
- [x] **Reviews resolved.**
- [x] **CODEOWNERS routing.** Touches `lib/Dialect/Hipsr`, `include/hip`, `lib/Runtime`, `tools` under the existing ownership map.


lib/Dialect/Hipsr/Transforms/ExternalizeConstants.cpp
```diff
@@ -60,8 +60,9 @@ struct HipsrExternalizeConstantsPass
     // is deterministic with or without a sink.
     std::vector<hip::ConstantEntry> entries;
     int64_t filePos = 0;
+    int64_t constantIndex = 0;
     module.walk([&](ConstantOp c) {
-      if (!c.shouldExternalize() || c.isExternalized()) {
+      if (c.isExternalized()) {
```
wcy123: add comment
// must run --externalize-constant before running `--convert-to-llvm`

wcy123 (dismissed):
LTGM

include/hip/Dialect/Hipsr/IR/HipsrConstantOp.h
```diff
@@ -80,6 +83,9 @@ template <typename T>::llvm::ArrayRef<T> ConstantOp::getDataValues() {
   llvm_unreachable("ConstantOp has neither value nor source");
 }
 
+void populateHipsrConstantLoweringPatterns(const LLVMTypeConverter &converter,
+                                           RewritePatternSet &patterns);
+
```
wcy123, line 88: this is very good

lib/Dialect/Hipsr/IR/HipsrConstantOp.cpp
```diff
@@ -13,23 +21,119 @@ LogicalResult ConstantOp::verify() {
...
+        getTypeConverter()->getMemRefAddressSpace(memRefType);
+    if (failed(addrSpace)) {
+      return failure();
+    }
+
+    Value dataPtr = callOp.getResult();
+    if (*addrSpace != 0) {
+      dataPtr = LLVM::AddrSpaceCastOp::create(
+          rewriter, loc, LLVM::LLVMPointerType::get(ctx, *addrSpace), dataPtr);
+    }
```
wcy123, line 107: add comment
// TODO: get_constant_wrapper should return !llvm.ptr<1>

wcy123 (approved):
LGTM

PR #516, opened 2026-07-21
fix(dialect): add a ctx operand to hipsr.cast and make its result optional

mingyueliuh:
Re-does the `hipsr.cast` bufferized-form fix dropped from #501, aligning `hipsr.cast` with the other hipsr DPS ops.

### Dialect fix
- Add the `!hipsr.context` operand (prints as a leading `(%ctx)`), threaded by `convert-onnx-to-hipsr` from function arg 0.
- Make the result type optional and switch `->` to `:` so the result-less (post-bufferization, memref) form round-trips.
- `populateShapeRegion` reads the output element type from `init` instead of `getResult(0)`, so it works on the no-result memref form.

Resolves the three #501 review comments (ctx threading, `:` vs `->`, why-comment).

### Test layout
Pass-level cases are split out of `cast.mlir` into their own file so per-op test files stay op-centric: each `ShapeRegionInterface` op checks only its own shape computation, and adding an op means adding a file rather than editing (and re-checking) a shared one. Idempotency and the whole-function walk are properties of the `-hipsr-populate-shape-region` pass, not of `hipsr.cast`, so keeping them in `cast.mlir` would misattribute them and force every future op file to duplicate the same cases.

- `cast.mlir` — op-centric: `cast_tensor`, `hipsr.cast`'s own tensor-form shape computation.
- `populate_shape_region.mlir` (new) — pass-level behavior of `-hipsr-populate-shape-region`: `already_populated` (idempotency) and `whole_function_walk` (every `ShapeRegionInterface` op filled in one run, across op kinds — a cast feeding a matmul).

LIT passes locally.

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrCastOp.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrCastOp.td
@@ -23,6 +23,10 @@ def Hipsr_CastOp : Hipsr_DpsOp<"cast"> {
     Destination-passing style: the result buffer is passed as the `init`
     operand.
 
+    The `ctx` operand (`!hipsr.context`) is the per-session execution context.
+    Every hipsr op carries it as its first operand; it is added in the ONNX
+    phase and threaded through unchanged.
+
     Regular op: carries an optional shape region (region 0). When populated it
     copies the input shape (via `populateShapeRegion`, which emits
     `shape.shape_of` + `shape.get_extent` so it works for both tensor and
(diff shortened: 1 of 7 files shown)


wcy123 (commented):
Changes look correct. The context threading, optional result handling, and test reorganization all follow good patterns. Two minor style suggestions:

**1. Comment style suggestion:**

Two comments explain historical mistakes rather than intent:

1. HipsrCastOp.td lines 55-57:
```cpp
// Result type is optional: Bufferization removes results from DPS ops (the
// memref form has no result), so trying to print a non-existent result type
// would crash without the optional clause.
```

2. HipsrCastOp.cpp lines 37-38:
```cpp
// Output element type from the `init` operand, not getResult(0): the
// bufferized (memref) form has no tensor result.
```

Both comments document what NOT to do and explain why a mistake would crash. Per LLVM/MLIR style, comments should explain intent, not historical mistakes. The code is self-documenting:

- `(: type($result_tensors)^)?` - assembly format shows the optional result
- `getInit().getType()` - code shows using init operand

**Recommendation:** Remove both comments. If explanation is needed, a brief "Supports both pre/post-bufferization forms" in the td file would suffice, but the optional syntax already documents this clearly.

**2. Curly braces for single-statement if:**

CastConversion.cpp lines 38-39:
```cpp
if (::mlir::failed(ctx))
  return ::mlir::failure();
```

Per [LLVM Coding Standards](https://llvm.org/docs/CodingStandards.html), single-statement bodies should use braces for early returns (guard clauses):

```cpp
if (::mlir::failed(ctx)) {
  return ::mlir::failure();
}
```

The braces improve readability for error-handling paths and make the guard clause immediately obvious.

wcy123 (approved):
LGTM

PR #519, opened 2026-07-21
refactor(hipsr): isolate shape regions and clean up ShapeRegionInterface

zpye:
## Summary

Follow-up cleanup to the hipsr `ShapeRegionInterface` from #474. A shaped op carries an isolated shape region (region 0) that computes its output shape; because the region can't read the op's operands, `hipsr-populate-shape-region` threads the DPS inputs in as entry-block args (arg `i` mirrors DPS input `i`).

## Key changes

- Use the built-in `IsolatedFromAbove` + `SingleBlock` traits instead of a custom trait. The verifier requires both traits and a single block ending in `hipsr.shape_yield` whose args match the DPS inputs one-for-one.
- Move the region/result accessors out of the interface into free functions defined once (taking a `ShapeRegionInterface` handle). Only `populateShapeRegion` stays per-op.
- Add a bounds-checked `ShapeRegionArgs` helper so each op names its block args (`getInput()`, `getA()/getB()`).
- Add `shape_region_isolation.mlir` (asserts `-cse`/`-canonicalize` don't break isolation) and update the cast/matmul/verify/populate tests.


wcy123 (approved):
LGTM

PR #455, opened 2026-07-08
feat: support operators in 0.3 release

amd-genmingz:
## New operator:

| # | Operator | Type | Summary |
|---|----------|------|---------|
| 1 | BiasGelu (com.microsoft) | New op | `hip.bias_gelu`, native kernel (`bias_gelu_kernel.hip`) |
| 2 | FastGelu (com.microsoft) | New op | `hip.fast_gelu`, native kernel (`fast_gelu_kernel.hip`) |
| 3 | Abs | New op | `hip.abs`, unary kernel (`elementwise_unary_kernel.hip`) |
| 4 | Ceil | New op | `hip.ceil`, unary kernel |
| 5 | Log | New op | `hip.log`, unary kernel |
| 6 | Or | New op | `hip.or`, boolean binary kernel (with broadcast) |
| 7 | ReduceMin | New op | `hip.reduce_min`, reuses ReduceMax axes/keepdims handling |
| 8 | Greater | New op | No new op; decompose → `hip.less(B, A)` |
| 9 | Compress | New op | `hip.compress`, native kernel |
| 10 | OneHot | New op | `hip.one_hot`, native kernel |
| 11 | GatherElements | New op | `hip.gather_elements`, native kernel |
| 12 | ScatterElements | New op | `hip.scatter_elements`, native kernel (includes reduction variant `scatter_reduction.inc.hip`) |
| 13 | TopK | New op | `hip.top_k`, native kernel |
| 14 | RMSNormalization (opset 23+) | New op | `hip.rms_norm`, **reuses** existing SimplifiedLayerNorm runtime; no new kernel |
| 15 | onnx.If | New op (control flow) | `hip.if` + branches outlined to func.func; new outline pass / lowering / runtime |
| 16 | onnx.Attention (ai.onnx opset 23/24) | New op (generic conversion) | Generic lowering to `hip.gqa`, reuses GQA runtime (`OnnxAttentionConversion.cpp`, 571 lines) |
| 17 | onnx.RotaryEmbedding (com.microsoft / ai.onnx) | New op (generic conversion) | `OnnxRotaryEmbeddingConversion.cpp` (167 lines) lowers to rope path |

## Extend the supported use cases of existing operators.

| # | Operator | Type | Summary |
|---|----------|------|---------|
| 1 | Conv (1D path) | Extension | Support `group != 1` (grouped/depthwise) + dynamic result dims (batch/channel/spatial) |
[...]

PR #521, opened 2026-07-21
perf(bufferize): opt-in copy-before-write to unblock huge models

fhanuman:
## Issue

On very large models (e.g. Nemotron-3-Super-120B — ~400k ops in a single function), the `one-shot-bufferize` pass hangs for ~1.5h. This is a known upstream MLIR issue: the bufferization analysis has O(n²) behavior in its aliasing read/write-set construction, and it gets much worse as the op count grows. Confirmed with gdb — the whole time is spent inside that analysis. See https://discourse.llvm.org/t/one-shot-bufferization-compilation-time-issue-with-large-irs/81439.

## Workaround

Add an opt-in env flag `HIPDNN_EP_BUFFERIZE_COPY_BEFORE_WRITE=1`. When set, one-shot-bufferize skips the analysis (copies before every write instead of proving in-place safety), which brings compile time down from ~1.5h to a few seconds.

It's **off by default**, so all other models keep the optimal in-place path and existing tests are unaffected. This unblocks end-to-end testing of large models now.

## Real fix

The copies cost some runtime performance, so this is a temporary unblock, not the final answer. I'm looking into the real fix (memoizing the aliasing-set construction so we can keep the in-place path even on large graphs).


fhanuman:
Superseded by #522 (this branch had PR #455 merged in, so its diff wrongly included #455). #522 is the clean version with only the copy-before-write commit.

PR #520, opened 2026-07-21
feat(tools): add identify_missing_ops for hip-ep compatibility reporting

amd-mounikk:
Add tools/common/identify_missing_ops.py to smoke-run ONNX models via onnxruntime_perf_test, or parse existing perf_test logs and MLIR IR dumps.

The script classifies graph ops as compile blockers, lowered, fused, or unsupported by discovering OnnxToHip converters and runtime kernels from the repo. It prints aligned gap tables for schema, converter, fusion, and kernel coverage to guide hip-ep enablement work.

PR #528, opened 2026-07-23
feat(cursor-skill): add export-onnx skill for Olive ONNX export

anilmartha:
## Summary
Adds a Cursor **agent skill** (export-onnx) that guides an agent through exporting a vision/ML model into an optimized fp16 ONNX model by authoring and running an Olive recipe under `olive-recipes/`.

## What's included
- `.cursor/skills/export-onnx/SKILL.md` — end-to-end workflow:
  - Inputs (model link / input shape / dtype) and exportability scope (CNNs & standard transformers; excludes multi-view 3D-detection stacks).
  - One-time setup (shallow-clone Olive repos, `hipdnn-ep` env deps, `onnxruntime-directml` caveat).
  - Recipe layout (`user_script.py` / `config.json` / `run.py`), incl. the torchvision shortcut and the required run-from-`olive/` gotcha.
  - `model_type` auto-detection for `OrtTransformersOptimization` (CNN vs transformer).
  - The fp16 duplicate-node post-fix.
  - Verification steps and reference recipes to mirror.

## Notes
- Skill only; no code/runtime changes. Perf-report tooling changes are intentionally excluded.
- Paths are genericized (`%USERPROFILE%\workspace`); no user-specific IDs.

### DIFF
--- a/.cursor/skills/export-onnx/SKILL.md
+++ b/.cursor/skills/export-onnx/SKILL.md
@@ -1,3 +1,7 @@
+<!--
+Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+Licensed under the MIT License.
+-->
 ---
 name: export-onnx
 description: >-


asparkhi:
We should add .cursor/commands wrappers for the skill to be available as a command to all.
This was missed out for find-missing-ops in #520. Can you please include one for that too?

First stage in the skill should be a dependency resolver. It should report out the missing dependencies in the environment/path. Let's add if its not already there.

PR #529, opened 2026-07-23
perf(pipeline): drop ownership-based buffer deallocation (superseded by pool-allocs)

fhanuman:
## Issue

On very large models (a single graph with hundreds of thousands of ops), compilation hangs in the buffer-deallocation pass. It builds one huge `bufferization.dealloc`, and the deallocation-simplification step then compares every buffer against every other to decide if they alias — an O(n²) scan that can take hours.

## Fix

Stop running the buffer-deallocation pipeline. This EP doesn't need it: instead of allocating and freeing each temporary individually, `hip-pool-allocs` packs all temporaries into session-owned pool(s) that live for the whole session and are freed once at the end, and graph outputs are owned by the runtime allocator. So there are no per-buffer allocations to free — the pass was pure overhead, and its output was already being discarded later in the pipeline.

Verified that the final compiled code is unchanged (same pooling, no leaks) on qwen3.5-35B, qwen3.5-9B, and gemma3; broader CI validation to follow.

### DIFF
--- a/docs/design/output-allocator-design.md
+++ b/docs/design/output-allocator-design.md
@@ -49,8 +49,7 @@ Dynamic output shapes are computed **inside** the graph (e.g. `M = memref.dim %i
 
 ```mermaid
 graph TD
-    A[one-shot-bufferize] --> B[buffer-deallocation]
-    B --> C[hip-use-output-allocator]
+    A[one-shot-bufferize] --> C[hip-use-output-allocator]
     C --> D[hip-pool-allocs]
     D --> E[convert-hip-to-llvm]
     E --> F[generate-interface]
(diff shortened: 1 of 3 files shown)


amd-yanjunz:

- PR: ROCm/hip-ep#529
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/529#issuecomment-5059856215)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- NODE: (OGA job default)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1458/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_oga&id=6a627450c72089b2f1f7b3be

PR #533, opened 2026-07-24
fix(hip-to-llvm): align alloc_output callback shape with func.return rank

amd-genmingz:
## Summary

- When a graph output is returned through collapse/expand views at a different rank than the internal alloc, `hipdnn_ep_alloc_output` now receives the ONNX/OGA ABI rank and dimensions.
- Callback shape is built at the alloc site by propagating sizes through the view chain, avoiding dominance-invalid `memref.dim` on the return value.
- Adds LIT coverage for Gemma3 vision and Conv flatten paths; compile-time verification against `hipdnn.output_shapes` rank.

## Test plan

- [ ] Gemma3 OGA vision: `image_features` shape binding + generation smoke test

## Checklist

- [ ] **Incremental.** This PR is either standalone, or a planned step
      in a documented series. If it exceeds ~500 LOC or ~10 files, the
      Summary / Why explains why it cannot be split, OR links the
      precursor PRs and the design issue. See:
      https://llvm.org/docs/DeveloperPolicy.html#incremental-development
- [ ] **AI policy.** If AI tools (Cursor, Claude, Copilot, etc.)
      generated substantial code or text, the Summary discloses it,
      the commit messages carry the trailers documented in
      [`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md#commit-message-trailers),
      and you can answer questions about each design decision in
      review without delegating back to the tool. See:
      https://llvm.org/docs/AIToolPolicy.html
- [ ] **No `good first issue` automation.** This PR is not an AI-tool
      fix to an issue tagged `good first issue` (those are reserved as
      learning opportunities for new contributors).
- [ ] **Reviews resolved.** All review-comment threads from prior
      rounds are resolved (enforced by branch protection on `main`).
- [ ] **CODEOWNERS routing.** Reviewers were assigned automatically by
      [`CODEOWNERS`](../blob/main/.github/CODEOWNERS). If the change
      touches a path outside the current ownership map, the PR
      description names the operational owner.


amd-mingw:
limitation of this solution:

1. View chain support is narrow (single collapse or expand, optional cast only)
stampAbiReshapeAttrs bails out on subview, reinterpret_cast, or any collapse+expand combo, and does not stamp ABI attrs.
BufferViewFlowAnalysis can still mark the alloc as a graph output, but the allocator callback may keep using the internal rank → ORT bind can still fail with a shape mismatch.
Not covered today: e.g. alloc → collapse_shape → subview → return. If production graphs use that pattern, a follow-up is needed.

2. Expand-path edge case (multiple dynamic external dims in one group)
If one internal dim expands into a group with more than one dynamic external dim, the usedInternal check rejects the attrs and the lowering falls back to the internal rank.
Current DETR/Gemma patterns should not hit this; more general logic would be needed if they do.

amd-mingw (approved):
this solution is not the better/general one as above comment, but in order to not block, plan to merge it, and will design a general solution with follow-up PR.

PR #542, opened 2026-07-24
perf(qmoe): single-launch grouped WMMA int4 MoE prefill (default-on)

BoarQing:
## What

Adds a single-launch **grouped Matrix-Core (WMMA) int4 GEMM** fast path for the quantized MoE (`qmoe`) prefill, replacing the host per-expert loop, and makes the associated scratch reservation conditional so it doesn't inflate the memory-pool high-water mark.

## Why

The previous prefill ran a **host per-expert loop** — roughly `5 x active_experts` dispatches plus a D2H copy / sync per MoE layer. That serializes launches and stalls on host round-trips, dominating prefill GPU time on high-activation models.

## Changes

- **Grouped WMMA int4 MoE prefill (default-on).** Each FC now runs one grouped int4 GEMM over all active experts (`blockIdx.z = expert`), reusing the tuned double-buffered int4 pipeline (~6 launches/layer instead of ~5x per-expert dispatches + D2H/sync). A fill-aware post-bucket gate keeps tiny prefills on the host-loop path via a minimum-token threshold. No env vars required.
- **Conditional grouped scratch reservation.** The grouped-only scratch (`pf_act` / `pf_accum` / `active_eids`) was previously reserved for *every* prefill (including sub-threshold and VLM image prefills that always take the host loop); since the per-state pool never shrinks, that unused reservation also leaked into the decode phase. The reservation is now gated on the pre-bucket portion of the grouped gate (`grouped_shape_ok && grouped_scale_ok`), known before `bucket_tokens` writes into the pool (realloc-safe). Below the token threshold / for non-WMMA shapes the grouped scratch is skipped, so the pool high-water matches the host-loop footprint. The same predicates are reused at the post-bucket dispatch so the reservation and the grouped decision can never disagree.

## Results / validation

- Grouped path still engages above the token threshold (>2048 tokens) on Qwen3.6-35B-A3B (dense_pct 91–93).
- Memory-pool high-water for sub-threshold / VLM image prefills and decode no longer carries the unused grouped scratch.

## Scope

int4 quantized MoE prefill only; decode and non-WMMA/sub-threshold shapes retain the existing host-loop path.

### DIFF
--- a/include/hip/env.h
+++ b/include/hip/env.h
@@ -71,4 +71,26 @@ inline std::string env_string(const char *name) {
   return (n > 0 && n < sizeof(buf)) ? std::string(buf, n) : std::string();
 }
 
+// Env var as an int; `dflt` when unset/blank. Manual parse (no std::stoi) to
+// keep the runtime bitcode free of MSVC vectorized helpers the JIT can't
+// resolve.
+inline int env_int(const char *name, int dflt) {
+  char buf[16];
+  unsigned long n = read_env(name, buf, sizeof(buf));
+  if (n == 0)
+    return dflt;
+  int sign = 1, i = 0;
+  if (buf[0] == '-') {
+    sign = -1;
+    i = 1;
+  }
+  long v = 0;
+  bool any = false;
+  for (; buf[i] >= '0' && buf[i] <= '9'; ++i) {
+    v = v * 10 + (buf[i] - '0');
+    any = true;
+  }
+  return any ? (int)(sign * v) : dflt;
+}
+
 } // namespace hipdnn_ep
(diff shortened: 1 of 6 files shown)


BoarQing:
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1476&base_job=daily_hip_ep_oga&base=1476&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1473&curr_job=daily_hip_ep_oga&curr=1484

PR #546, opened 2026-07-24
perf(gqa): occupancy-aware v8 prefill autotune (ND=2/MT=1/BKV=16)

BoarQing:
## Summary
- Folds the low-LDS `BKV=16` variant into the `gqa_flash_prefill_v8` (d=256) autotune candidate set and adds an **occupancy-aware tiebreak**: among candidates within 3% of the fastest per-call time, prefer the one with the lowest LDS footprint.
- The tuner now selects **ND=2, MT=1, BKV=16** automatically. This cuts LDS 25792 -> 10880 B, raising LDS-limited occupancy from 2 -> 6 workgroups/CU.
- No default behavior change beyond candidate selection; stays adaptive to shape/hardware.

## Measured impact (Qwen3.6-35B-A3B-fp16, gfx1151 / Radeon 8060S)
- **Per-call kernel latency: 4.13 -> 3.01 ms (~27% faster)** at sq=skv=1950, d=256.
- **End-to-end TTFT ~4.5%** (P128 -4.7%, P2048 -4.3%, same-session N=3). Smaller than the per-call gain because full-attention GQA is a sparse fraction of prefill (most LLM layers are linear-attention + conv).
- **Guardrails flat:** decode TPS 59.5 vs 59.6; peak memory +5 MB (negligible). Prefill kernels don't run in decode.
- Correctness: winner output is bit-identical to a baseline run; remaining diffs are within the baseline's own run-to-run GPU nondeterminism (flash online-softmax is tile-size invariant).

## Why this config (roofline)
Clock-scaling analysis shows the kernel is **compute/on-chip-throughput bound**, not memory-bandwidth or occupancy bound: pinning GPU clock 2900->700 MHz scaled runtime 3.98x (~0.96 of the 4.14x clock ratio). Higher occupancy is not the win - the 50%-occupancy variant (`4_1_16`) is 26% slower; `ND=2` wins by doing less redundant work. The tiebreak is a safeguard for future near-ties, not the primary selector here.

## Test plan
- [x] Same-session A/B sweep (N=3) P128/P2048, decode-TPS + peak-mem guardrails
- [x] Per-candidate `ms/call` via `HIPDNN_PREFILL_TUNE_DEBUG=1`
- [x] RGP occupancy confirmation (thermal-independent)
- [x] Output correctness vs baseline


BoarQing:
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1476&base_job=daily_hip_ep_oga&base=1476&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1475&curr_job=daily_hip_ep_oga&curr=1473

BoarQing:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1559/

PR #554, opened 2026-07-24
Streamline shape inference and pool planning design docs

fhanuman:
## Summary
- Replace stale shape-inference and PoolAllocs implementation histories with concise design contracts.
- Align pipeline ordering, unranked-tensor ownership, compiler/runtime metadata boundaries, and related source comments.


fhanuman:
Cleaned up stale information and details from the doc. No functional changes.

PR #557, opened 2026-07-24
fix(use-output-allocator): report graph-output rank for reshaped-view outputs

anilmartha:
## Summary

`UseOutputAllocator` rewrote a graph-output `memref.alloc` into `hip.alloc_output` using the **producer alloc's type**. When the returned value is a reshaping view of that alloc (e.g. a rank-2 Gemm result `expand_shape`'d to rank-3 with no trailing op — DETR's `class_labels_classifier` `logits` head), the EP registered the alloc's rank with ORT instead of the declared graph-output rank, so ORT rejected the output:

```
Invalid rank for output: logits Got: 2 Expected: 3
```

Outputs with a trailing op (e.g. `pred_boxes`' `sigmoid`, whose alloc is already at the output rank) were unaffected — which is why only `logits` failed.

## Fix

- Allocate `hip.alloc_output` at the `func.return` operand type (the graph-output shape).
- When it differs from the producer alloc's type, rebuild the **inverse** of the view chain (`cast` / `expand_shape` / `collapse_shape`, static output) so the producer keeps writing into the same EP-owned buffer viewed as its original type — no extra copy.
- Fast path (return type == alloc type) is unchanged; `subview` or dynamic-shaped outputs fall back to the legacy behavior, so no existing case regresses.

Applies to both the `LLVM_IR` and `NATIVE` artifact paths.

## Test plan

- [x] `facebook/detr-resnet-50` (fp16): `logits` now returns `{1,100,92}`; `onnxruntime_perf_test` runs clean.
- [ ] Re-run existing lit/numeric suites for regressions (esp. conv/reshape output models).
- [ ] Spot-check a model with a rank-4 conv output collapsed before return.

### DIFF
--- a/lib/Dialect/Transforms/UseOutputAllocator.cpp
+++ b/lib/Dialect/Transforms/UseOutputAllocator.cpp
@@ -14,9 +14,18 @@
 //
 // For each matching alloc the rewrite:
 //   - sets `out_idx` to its position in func.return (the graph output index),
-//   - reuses the alloc's dynamic-size operands unchanged,
-//   - deletes any `memref.dealloc` of it (the EP frees it, not us),
-//   - leaves the view ops in place -- only the alloc op itself changes.
+//   - allocates the EP-owned buffer at the GRAPH-OUTPUT type (the func.return
+//     operand type), not the producer alloc's type. These differ when the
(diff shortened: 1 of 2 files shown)


amd-mingw:
https://github.com/ROCm/hip-ep/pull/533 is fixing on it. cc @amd-genmingz

PR #558, opened 2026-07-24
perf(bufferize): opt-in copy-before-write to unblock huge models

fhanuman:
Port of closed PR [#522](https://github.com/ROCm/hip-ep/pull/522). The original description is copied below; its large-model timing refers to the original branch at commits `ab60d202` and `89a3dd1b`.

## Issue

On very large models (e.g. Nemotron-3-Super-120B — ~400k ops in a single function), the `one-shot-bufferize` pass hangs for ~1.5h. This is a known upstream MLIR issue: the bufferization analysis has O(n²) behavior in its aliasing read/write-set construction, and it gets much worse as the op count grows. Confirmed with gdb — the whole time is spent inside that analysis. See https://discourse.llvm.org/t/one-shot-bufferization-compilation-time-issue-with-large-irs/81439.

## Workaround

Add an opt-in env flag `HIPDNN_EP_BUFFERIZE_COPY_BEFORE_WRITE=1`. When set, one-shot-bufferize skips the analysis (copies before every write instead of proving in-place safety), which brings compile time down from ~1.5h to a few seconds.

It's **off by default**, so all other models keep the optimal in-place path and existing tests are unaffected. This unblocks end-to-end testing of large models now.

## Actual fix

Fix is in progress -- Spoke with Elliott and will upstream the fix soon.


amd-yanjunz:

- PR: ROCm/hip-ep#558
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/558#issuecomment-5072901906)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- NODE: (OGA job default)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1484/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_oga&id=6a63e9ba7f574705d7b95fd3

amd-mingw (dismissed):

PR #561, opened 2026-07-25
Apply plugin dialect registrations to hip-mlir-opt

Vandana9628:
`hip-mlir-opt` registers plugin passes (via `dispatchPluginRegistrationsOnce()`) but never registers plugin dialects into its DialectRegistry. As a result, a statically linked plugin's custom ops parse and bufferize, but convert-hip-to-llvm finds no lowering interface for them and they survive unlowered to the final LLVM output. This applies `pluginDialectRegistrations()` to the tool's registry, mirroring the EP's loadAllDialects, so the plugin op set never drifts between the tool and the EP. No-op when no plugins are selected.


tools/hip-mlir-opt/hip-mlir-opt.cpp
```diff
@@ -109,6 +109,15 @@ int main(int argc, char **argv) {
   // (HIPDNN_EP_COMPILER_PLUGINS empty). See cmake/HipEpPlugins.cmake.
   hip::compiler::dispatchPluginRegistrationsOnce();
 
+  // Companion to dispatchPluginRegistrationsOnce() above: that registers plugin
+  // passes, this registers plugin dialects (custom ops + their bufferization
+  // and HIP->LLVM-lowering interface models) into the tool's registry,
+  // mirroring hip::compiler::loadAllDialects so the op set never drifts between
+  // the tool and the EP. Without it convert-hip-to-llvm finds no lowering for
+  // plugin ops and they survive unlowered. No-op when no plugins are selected.
+  for (auto registerFn : hip::compiler::pluginDialectRegistrations())
```
fhanuman, line 124: Thanks for fixing this. Could we add a plugin-gated LIT test that runs a sample plugin dialect op through --`convert-hip-to-llvm` to catch any future issues?

fhanuman (changes requested):

fhanuman (approved):

PR #578, opened 2026-07-28
fix(hipsr): preserve device address spaces in LLVM lowering

mingyueliuh:
## Summary
- preserve the address space of pointers extracted from lowered Hipsr memref descriptors
- declare runtime state pointers in AS0 and device buffer pointers in AS1 for `hipsr.add` and `hipsr.constant`
- update focused LLVM lowering LIT checks for the corrected signatures

## Problem
Hipsr device memrefs lower to LLVM address space 1, but `extractContiguousMemRefPtr` converted non-AS0 pointers to AS0. The existing Add and Constant runtime declarations also described device buffers as AS0 pointers. This discarded the device-memory distinction and produced LLVM calls whose pointer types did not match the runtime ABI.

## Solution
Return the aligned memref pointer in its original address space and declare each runtime argument according to its role: host/runtime state remains AS0, while device buffers remain AS1. The generated call signatures now preserve the memory-space semantics without unnecessary `addrspacecast` operations.

## Test plan
- [x] Run the focused `hipsr.add` and `hipsr.constant` LLVM lowering LIT tests
- [x] Run the complete MorphizenMLIR LIT suite (353 passed, 2 unsupported)

### DIFF
--- a/lib/Dialect/Hipsr/IR/HipsrAddOp.cpp
+++ b/lib/Dialect/Hipsr/IR/HipsrAddOp.cpp
@@ -74,7 +74,9 @@ struct AddLowering : public ConvertOpToLLVMPattern<AddOp> {
                   ConversionPatternRewriter &rewriter) const override {
     Location loc = op.getLoc();
     ModuleOp module = op->getParentOfType<ModuleOp>();
-    Type ptrType = getPtrType();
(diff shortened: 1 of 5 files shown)


wcy123 (approved):
LGTM

PR #590, opened 2026-07-30
refactor(hipsr): use RuntimeFunc for add and constant lowerings

mingyueliuh:
## Summary
- extend `RuntimeFunc` with `callWithResult` for result-producing runtime calls
- refactor `hipsr.add` to declare its runtime ABI and convert device pointers, slot index, and constants through `RuntimeFunc`
- refactor `hipsr.constant` to consume its returned device pointer through `RuntimeFunc`

## Test plan
- [x] Build `hip-mlir-opt`
- [x] Run `MorphizenMLIRLitTests` (359 passed, 3 unsupported)

## AI assistance
Cursor assisted with the refactoring and test validation. The resulting code and generated LLVM behavior were reviewed, built, and validated with the full MLIR LIT suite.

### DIFF
--- a/include/hip/Conversion/HipsrToLLVM/HipsrToLLVM.h
+++ b/include/hip/Conversion/HipsrToLLVM/HipsrToLLVM.h
@@ -170,17 +170,25 @@ public:
   }
 
   template <typename... Args> LogicalResult call(Args &&...args) {
+    createCall(std::forward<Args>(args)...);
+    return success();
+  }
+
+  template <typename... Args> Value callWithResult(Args &&...args) {
+    return createCall(std::forward<Args>(args)...).getResult();
+  }
+
+private:
+  template <typename... Args> LLVM::CallOp createCall(Args &&...args) {
     static_assert(sizeof...(Params) == sizeof...(Args),
                   "argument count must match parameter count");
 
     llvm::SmallVector<Value, sizeof...(Params)> convertedArgs{
         detail::ArgConverter<Params, std::decay_t<Args>>::convert(
             rewriter, loc, std::forward<Args>(args))...};
-    LLVM::CallOp::create(rewriter, loc, funcOp, convertedArgs);
-    return success();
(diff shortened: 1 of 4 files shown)


wcy123 (commented):
**Issue 1: Unnecessary API split**

[`HipsrToLLVM.h:172-187`](https://github.com/ROCm/hip-ep/blob/40790fda7accd8db6cec9f13afa0ea82ae3f4c5e/include/hip/Conversion/HipsrToLLVM/HipsrToLLVM.h#L172-L187)

Why have both `call()` (void) and `callWithResult()` (Value)? Just return `FailureOr<Value>` from a single method - callers ignore the value when not needed.

**Issue 2: Unjustified infallibility assumption**

[`HipsrCastOp.cpp:102-106`](https://github.com/ROCm/hip-ep/blob/40790fda7accd8db6cec9f13afa0ea82ae3f4c5e/lib/Dialect/Hipsr/IR/HipsrCastOp.cpp#L102-L106)

The second commit removes this failure check:

```cpp
// Before:
if (failed(castFunc->call(...))) {
  return failure();
}

// After:
castFunc->call(...);
```

Commit message claims "argument conversion and call emission are infallible" without evidence. What if `ArgConverter::convert()` fails? What if `extractContiguousMemRefPtr()` encounters incompatible memref layout? Either prove these operations cannot fail, or restore `FailureOr<Value>` return type.

**Design note**

[`HipsrAddOp.cpp:100-113`](https://github.com/ROCm/hip-ep/blob/40790fda7accd8db6cec9f13afa0ea82ae3f4c5e/lib/Dialect/Hipsr/IR/HipsrAddOp.cpp#L100-L113)

The RuntimeFunc design is sound - template parameters declare the runtime function signature, `ArgConverter` overloads allow passing either runtime `Value` or compile-time `int64_t`, both convert to the declared type.

wcy123 (approved):
LGTM

PR #582, opened 2026-07-29
feat(hipsr): add hipsr-pool-alloc pass skeleton

zz002:
## Summary

Register `-hipsr-pool-alloc` as a no-op `ModuleOp` pass and add the lit file the rest of the series grows into. No IR changes yet.

## Related issue or design

Upstream [`wcy123/onnx-hipdnn-ep#108`](https://github.com/wcy123/onnx-hipdnn-ep/issues/108) (Phase 1, PR1 of `wcy123/onnx-hipdnn-ep#19`), which links the HIPSR Pool Allocation Pass design page.

## Why

Registration is zero extra code once the `def` exists — `Passes.h` pulls `GEN_PASS_REGISTRATION` and `InitAllPasses.h` already calls `registerHipsrPasses()`. Landing the skeleton on its own therefore gives each follow-up (liveness utility, single-group emit, multi-group offsets, opt-in report) a registered pass and a lit file to extend, so every one of them stays small and reviewable.

## What

- `Passes.td`: `def HipsrPoolAllocPass` over `mlir::ModuleOp`. The description states the target semantics and that this skeleton leaves the IR unchanged. `dependentDialects` already lists memref/arith so the follow-ups that emit views and size arithmetic do not have to revisit the def.
- `HipsrPoolAllocPass.cpp`: empty `runOnOperation()`.
- `test/lit/Dialect/Hipsr/pool_alloc.mlir`: `empty_domain` (a domain with only the implicit yield) and `noalloc_noop` (a tensor-mode domain). Both assert unchanged IR; the RUN line resolving `-hipsr-pool-alloc` is what proves the pass is registered.

## Test plan

- [x] `lit --filter pool_alloc` — `MorphizenMLIR :: Dialect/Hipsr/pool_alloc.mlir` PASS.
- [x] Full `check-hip-mlir-lit` — 351 passed, 0 failed.
- [x] `hip-mlir-opt --help` lists `--hipsr-pool-alloc`.
- [x] `pre-commit run --all-files` clean.

## Notes for reviewers

- The pass is not in any pipeline; it is reachable only through `hip-mlir-opt` until the rewrite lands.
- Follow-ups land in upstream order `wcy123/onnx-hipdnn-ep#109`-`#118`, each on top of the previous one.

## Checklist

- [x] The change is focused, or links a design/series explaining its scope.
- [x] Relevant tests were added or updated and the results are documented.
- [x] User-facing or design documentation was updated when needed.
[...]

PR #110, opened 2026-04-13
ci: disable scheduled trigger for GPU perf/accuracy test workflow

zz002:
## Summary
- Remove the daily `schedule` cron trigger (`0 16 * * *`) from `gpu-perf-accuracy-test.yml`
- Workflow is now triggered exclusively via `workflow_dispatch` (manual)

## Motivation
The daily automated run is no longer needed. Manual dispatch gives better control over when tests are executed.

## Test plan
- [x] Workflow file validates (no YAML errors)
- [ ] Confirm workflow no longer appears in scheduled runs

PR #591, opened 2026-07-30
feat(hipsr): add pool-alloc liveness analysis

zz002:
## Summary

`-hipsr-pool-alloc` now computes each pool domain's alloc live ranges. The pass still rewrites nothing; a new `emit-pool-report` option reports each range as a remark so lit can assert it.

## Related issue or design

Upstream [`wcy123/onnx-hipdnn-ep#109`](https://github.com/wcy123/onnx-hipdnn-ep/issues/109) (liveness analysis utility), a step of [`wcy123/onnx-hipdnn-ep#19`](https://github.com/wcy123/onnx-hipdnn-ep/issues/19), which links the HIPSR Pool Allocation Pass design page. Follows [#582](https://github.com/ROCm/hip-ep/pull/582) (`#108`, skeleton).

## Why

A live range starts at the first write, not at the alloc. After materialize-init-tensors the allocs sit together at the top of the domain, so ranges keyed on the alloc index all overlap and nothing can ever share pool space. Taking the first DPS `outs` write as the start keeps disjoint ranges disjoint, which is what makes the grouping in the follow-ups worth doing.

Indices are those of the domain block's own ops, so a user nested in a region has to fold to its enclosing block-level op — `Block::findAncestorOpInBlock`, as `PartitionPoolDomains.cpp` already does. Looking the nested op up directly returns index 0 and silently shortens the range, which would later let a live buffer be reused. `hipsr.pool_domain` is `IsolatedFromAbove`, so failing to find an ancestor means the IR broke that invariant: an assert, not a fallback branch.

`emit-pool-report` is the option name the whole series uses, not test scaffolding: `#110`-`#118` add group, size and reuse figures to the same report.

## What

- `Passes.td`: `emit-pool-report` bool option, default false, worded after the legacy `hip-pool-allocs` `emit-fragmentation-report`.
- `HipsrPoolAllocPass.cpp`: `Lifetime` plus `computeLiveness(Block &)` returning a `Value` -> `Lifetime` map, and `emitPoolingReport()` which walks the block so remark order is block order. An alloc with no DPS `outs` write has no range and is absent from the map; the warning for that case belongs to `#118`.
- `test/lit/Dialect/Hipsr/pool_alloc_report.mlir`: one RUN line, `--verify-diagnostics` for the ranges and `FileCheck --implicit-check-not` for the unchanged module.

Not done: alias-aware last use. The legacy `BufferUtils.h` tracks view-like aliases through `BufferViewFlowAnalysis` because `hip` domains contain them; hipsr domain bodies do not yet. This has to be revisited before `#117` wires the rewrite into a bufferized pipeline, otherwise an aliased use shortens the range.

## Test plan

- [x] `lit --filter pool_alloc` — `pool_alloc_report.mlir` and `pool_alloc.mlir` PASS.
- [x] Full `check-hip-mlir-lit` — 356 passed, 2 unsupported, 0 failed.
- [x] `hip-mlir-opt --help` lists `--emit-pool-report` under `--hipsr-pool-alloc`.
- [x] Mutation check: changing one expected range fails the run, and injecting a `hipsr.get_pool` into a domain trips `--implicit-check-not`, so neither half of the RUN line is vacuous.

## Notes for reviewers

- The default path is unchanged from the skeleton: the analysis runs, nothing is rewritten, nothing is printed. The lit RUN line asserts the module is byte-for-byte the input while the report is on.
- `Lifetime` carries only `start`/`end`. The endpoint ops are only needed by `#118`'s debug trace, so they land with it.
- Follow-ups land in upstream order `#110`-`#118`, each on top of the previous one.

## Checklist

### DIFF
--- a/include/hip/Dialect/Hipsr/Transforms/Passes.td
+++ b/include/hip/Dialect/Hipsr/Transforms/Passes.td
@@ -167,6 +167,11 @@ def HipsrPoolAllocPass : Pass<"hipsr-pool-alloc", "mlir::ModuleOp"> {
     Pools each `hipsr.pool_domain`'s `memref.alloc`s by lifetime:
     non-overlapping ones share one `hipsr.get_pool` buffer as `memref.view`s.
   }];
+  let options = [
+    Option<"emitPoolReport", "emit-pool-report", "bool", /*default=*/"false",
+           "Debug-only: emit remarks describing the pooling decisions. "
+           "No effect on generated code.">
+  ];
   let dependentDialects = [
     "::mlir::hipsr::HipsrDialect",
     "::mlir::memref::MemRefDialect",
--- a/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
+++ b/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
@@ -6,10 +6,18 @@
 #include "hip/Dialect/Hipsr/Transforms/Passes.h"
 
 #include "hip/Dialect/Hipsr/IR/HipsrDialect.h"
+#include "hip/Dialect/Hipsr/IR/HipsrOps.h"
 
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/IR/BuiltinOps.h"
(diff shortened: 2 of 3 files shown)


wcy123 (approved):
LGTM

PR #594, opened 2026-07-30
fix(morphizen): guard empty output NodeArg in try_fuse

AMDmoore:
## Summary
Large hybrid/quantized models (e.g. NemotronH INT4-RTN) contain nodes with empty, non-existent `NodeArg` slots. Two distinct cases show up:

- **Empty first output slot** (the crashing case for NemotronH): a multi-output op whose first output is unused. Concretely, NemotronH's `TopK` consumes only its `Indices` output (output 1); the `Values` output (output 0) is unused, so output slot 0 is an empty `NodeArg`.
- **Empty optional input slot**: `DequantizeLinear`'s optional `zero_point` in INT4-quantized weights.

Reading such a slot via `node_arg_get_name` fatal-CHECKs in `node_arg.cpp` (`node_arg doesn't exist!`), aborting the entire model compilation.

Fix:
- Add `node_get_first_existing_output_name()` as a node utility next to `node_get_first_output_name` (in `node.hpp` / `node.cpp`), and use it at every `try_fuse` boundary that reads a node's output name (`calculate_return_values`, `check_loop`, meta_def node emission, and `Pass::fuse`). It skips empty/non-existent output slots and returns the first existing output name. A fused node must still produce at least one real output, so it `CHECK`s (matching `node_get_first_output_node_arg`) rather than emitting a bogus node name.
- The isolated-node initializer collection additionally guards the empty optional input slot with `node_arg_exists` before reading `zero_point`'s name.

## Test plan
- [x] Incremental build on `fix` branch succeeds; `hipgpu.dll` / `hip-compiler.dll` rebuilt and deployed.
- [x] Llama-3.1-8B (INT4) via OGA `run_oga_dynamic` compiles and runs with no `node_arg doesn't exist` crash (`MorphiZen EP Load ONNX Model Success` + `Compilation successful`).
- [x] Baseline comparison: same Llama run on `main` produces byte-identical output, confirming zero behavioral change for models whose output slots all exist.
- [ ] Nemotron INT4-RTN end-to-end compile (the crashing model) -- long-running (~40min), to confirm the original crash is resolved.

### DIFF
--- a/morphizen/morphizen-core/src/pass.cpp
+++ b/morphizen/morphizen-core/src/pass.cpp
@@ -49,27 +49,6 @@ static bool node_arg_is_initializer(const Graph &graph,
                      });
 }
 
-// Safety helper for the try_fuse boundary. In large hybrid / quantized models
-// some nodes carry an empty (non-existent) NodeArg in an output slot. Calling
-// node_arg_get_name on such a slot fatal-CHECKs in node_arg.cpp
-// ("node_arg doesn't exist!"). Return the first EXISTING output name instead of
-// unconditionally reading output slot 0. A node reaching fusion must still
-// produce at least one real output, so CHECK (like
-// node_get_first_output_node_arg) if none exists rather than emitting a bogus
-// node name.
-static const std::string &node_first_existing_output_name(const Node &node) {
-  const std::string *name = nullptr;
-  for (const NodeArg *arg : node_get_output_node_args(node)) {
-    if (arg != nullptr && node_arg_exists(*arg)) {
-      name = &node_arg_get_name(*arg);
-      break;
-    }
-  }
-  CHECK(name != nullptr) << "no existing output NodeArg on node: "
-                         << node_as_string(node);
-  return *name;
-}
-
 // return values are Node found by node_arg name and the node_arg names
 // that cannot find the node through node_arg
 // Cannot find the node are three scenarios where the node cannot be found using
@@ -306,9 +285,11 @@ check_loop(const Graph &graph, const std::vector<const Node *> &input_nodes,
         if (hit_output) {
           ret = true;
           auto route = map_route[current_node.ptr()];
-          for (size_t i = 0; i < route.size(); i++) {
-            auto node = route[i];
-            maybe_loop_path.push_back(node_first_existing_output_name(*node));
+          for (auto *node : route) {
+            auto node_ref =
+                morphizen_cxx::NodeConstRef::from_node(graph, *node);
+            maybe_loop_path.push_back(
+                node_get_first_existing_output_name(*node_ref.ptr()));
           }
         }
         return false; // leave callback return value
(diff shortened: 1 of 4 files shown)


morphizen/morphizen-core/src/pass_imp.cpp
```diff
@@ -411,7 +411,22 @@ MetaDefProto &Pass::fuse(Graph &graph, const std::string &name,
...
+    // output, so CHECK if none exists.
+    const std::string *picked = nullptr;
+    for (const NodeArg *a : node_get_output_node_args(nd)) {
+      if (a != nullptr && node_arg_exists(*a)) {
+        picked = &node_arg_get_name(*a);
+        break;
+      }
+    }
+    CHECK(picked != nullptr)
+        << "no existing output NodeArg on node: " << node_as_string(nd);
```
amd-mingw: reuse node_first_existing_output_name()?

morphizen/morphizen-core/src/pass.cpp
```diff
@@ -286,10 +308,7 @@ check_loop(const Graph &graph, const std::vector<const Node *> &input_nodes,
           auto route = map_route[current_node.ptr()];
           for (size_t i = 0; i < route.size(); i++) {
             auto node = route[i];
-            auto node_ref =
-                morphizen_cxx::NodeConstRef::from_node(graph, *node);
-            maybe_loop_path.push_back(
-                morphizen::node_arg_get_name(node_ref.first_output_node_arg()));
+            maybe_loop_path.push_back(node_first_existing_output_name(*node));
```
amd-mingw: use node_ref ?

amd-mingw (changes requested):

morphizen/morphizen-core/src/pass_imp.cpp
AMDmoore: Done in 3382305c. pass_imp.cpp now calls the shared helper: meta_def->add_nodes(node_first_existing_output_name(node_ref));. To share it across both files, I moved 
ode_first_existing_output_name next to 
ode_get_first_output_name in 
ode.hpp/
ode.cpp (it's a generic node utility), so pass.cpp and pass_imp.cpp both reuse it instead of duplicating the first-existing-output loop.

morphizen/morphizen-core/src/pass.cpp
AMDmoore: Done in 3382305c. check_loop now builds a 
ode_ref and passes it: uto node_ref = morphizen_cxx::NodeConstRef::from_node(graph, *node); maybe_loop_path.push_back(node_first_existing_output_name(node_ref)); (NodeConstRef implicitly converts to const Node&), consistent with the input_nodes_cxx conversion just above.

AMDmoore:
TopK op which has the first output not consumed in model is fused into custom op QMoE, so there is no NodeArg assert issue anymore. Close this PR.

PR #589, opened 2026-07-30
perf(linear_attention): chunk-parallel + WMMA gated_delta prefill (5.5x kernel, up to 1.9x TTFT)

BoarQing:
## Summary

Rewrites the `gated_delta` linear-attention **prefill** from a 587-deep serial chunk recurrence into a **three-pass chunk-parallel** form, with the dominant GEMM running on the **RDNA3 wave-matrix cores (WMMA)**. This is now the sole prefill implementation for the supported config.

The original kernel ran `grid = B*Hkv = 32` blocks on 40 CUs (25% occupancy, latency-bound) with a strict cross-chunk state dependency. This exploits the affine chunk-state map `S_out = L_c·S_in + b_c` to decouple the per-chunk work from the sequential scan:

- **Pass 1** (`grid = B*Hkv*nchunks`, ~18.7k blocks): `S_in`-independent locals — `Uloc = (I−trilT)⁻¹(β·v)`, `W = (I−trilT)⁻¹(diag(β·A)K)`, `Ktilde`, `a_last`.
- **Pass 2** (`grid = B*Hkv`, light serial scan): `U = Uloc − W·S0`, `S1 = a_last·S0 + Ktildeᵀ·U`, fp16-rounded per chunk to match the original per-chunk quantization.
- **Pass 3** (`grid = B*Hkv*nchunks`): `O = scale·(A·(Q·S0) + P_lower·U)`. The `[Q;W]@S0` GEMM (M=64, K=128, N=128) runs on `__builtin_amdgcn_wmma_f32_16x16x16_f16_w32` (fp16 in / fp32 accumulate); the small `P_lower@U` intra term stays scalar.

Numerically equivalent to the original (same fp32 math, same per-chunk fp16 rounding); differs only in fp32 op-order.

## Changes

- `483b5e28` — chunk-parallel + WMMA prefill implementation.
- `17581b03` — promote it to the only path: delete the serial chunked kernel and the scalar Pass-3 variant, remove all bring-up env gates (`HIPDNN_EP_LA_PARALLEL`/`_WMMA`/`_PROF`/`_SCANBLK`/`_P3BLK`) and per-pass event timing. On any launch/alloc failure the launcher returns nonzero, so the runtime still falls back to the per-token loop (unchanged safety net). WMMA requires `dk,dv` multiples of 16; otherwise it declines to the per-token loop.

## Correctness

[...]

PR #596, opened 2026-07-30
perf(linear_attention): session-scoped prefill scratch + drop Ktilde buffer

BoarQing:
## Summary

Follow-up to #589 (chunk-parallel + WMMA `gated_delta` prefill). Same kernel math; this trims the device scratch and fixes its lifetime. Numerically exact (same fp32 math / per-chunk fp16 rounding).

- **Session-scoped scratch.** Move the prefill scratch out of the process-static, never-freed `g_la_scratch` into the per-session `RuntimeState::la_scratch` pool (grow-on-demand, freed in `hipdnn_ep_state_cleanup`), mirroring `qmoe_scratch`/`conv_scratch`. The kernel now takes a caller-owned `scratch` + `size`; the wrapper sizes it via the new `hip_linear_attention_prefill_scratch_bytes()`. Footprint is bounded to the session and released on teardown instead of held for the whole process.
- **Drop the `Ktilde` buffer.** Store only the per-token `rlast` scalar (`C` floats/chunk) instead of the full `Ktilde = rlast*K` tile (`C*dk` floats/chunk); the scan rebuilds `Ktilde` from the original `key` tensor. Removes the `Kt_g` buffer (~1/dk the size) at bit-identical values.

## Test plan

- [x] Builds clean on gfx1151.
- [x] End-to-end on Qwen3.5-9B (gfx1151, `dog.jpg`): chunk-parallel prefill path exercised (`dk=dv=128`), coherent output, no regression.
- [ ] fp64 oracle cosine = 1.0 at T=256/512 (unchanged math; to re-confirm).
- [ ] Device-memory delta measurement at long sequence (host RSS does not reflect the GPU scratch saving).

## Note

Targets `main`. Since #589 is not yet merged, this PR also contains #589's two commits; the incremental change here is the final commit (`session-scoped prefill scratch + drop Ktilde buffer`). Merge/rebase onto #589 first if it lands.

### DIFF
--- a/lib/Runtime/Kernels/hip/linear_attention_kernel.hip
+++ b/lib/Runtime/Kernels/hip/linear_attention_kernel.hip
@@ -675,15 +675,15 @@ extern "C" int hip_linear_attention_decode(
 }
 
 //===----------------------------------------------------------------------===//
-// Chunked-parallel gated-delta PREFILL kernel
+// Gated-delta linear-attention PREFILL
 //===----------------------------------------------------------------------===//
 //
 // Replaces the per-token launch loop (one decode-kernel launch per timestep,
-// each streaming the whole recurrent state from global) with a SINGLE launch
-// that processes the entire sequence.  One block per (batch, kv_head); each
-// block loops over chunks of LA_CHUNK tokens, keeping the recurrent state S in
-// GLOBAL (read once / written once per chunk) and the per-chunk tiles + the
-// [C,C] coupling matrices in LDS.
+// each streaming the whole recurrent state from global) with a chunked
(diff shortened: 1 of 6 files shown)


BoarQing:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1585/

BoarQing:
new: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1585/
old: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1579/

I expect lower memory consumption

BoarQing:
slightly lower mem consumption. Lower TPS is confirmed to be noise.

PR #602, opened 2026-07-31
HIP EP Enablement for MI350

penglu12-quark:
## Summary

Enablement of hip-ep on MI350X (gfx950 / CDNA4 / wave64)

- GQA softmax launches now take their block-size floor from the runtime wave size
  instead of a hard-coded 64, which restores the original block geometry on
  wave32.
- The protobuf `dllexport_decl` workaround is narrowed to a new
  `MORPHIZEN_PROTO_DLL_SPEC`, so it no longer strips the visibility attribute
  from morphizen's hand-written public API headers.

## Why

**Softmax block floor.** The wave64 port raised the softmax block-size floor to a
hard-coded 64 for a real reason: a 32-thread block is a partial wave on CDNA, and
the inactive upper lanes get folded into the `__shfl_xor` reduction. But the floor
is unconditional, so it also applies on RDNA, where it forces a second wave for
every row count <= 32 — the common decode shape — with nothing to show for it.
The floor needs to be the *device's* wave size, not the larger of the two.

`num_waves` has to be derived from that same value, because it sizes the dynamic
LDS fan-in buffer and must match what the kernel computes internally
(`blockDim.x / WAVE_SIZE`). Previously it used `HIPDNN_WAVE_SIZE`, which the host
pass hardcodes to 32, so on gfx950 that buffer was over-allocated by 2x.

**Wave-size probe caching.** The probe cached a single answer for the whole
process. A host can drive GPUs from different families at once (e.g. a gfx1151 APU
alongside a discrete card), where that cached answer gets applied to the wrong
device after `hipSetDevice`.

**Export macro over-reach.** protobuf v34 emits
`class <decl> [[gnu::warn_unused]] Name`, and GCC rejects a GNU `__attribute__`
placed before a C++11 `[[...]]` attribute, so the decoration has to be empty on
ELF. The previous fix achieved that by emptying `MORPHIZEN_DLL_SPEC` itself — but
that macro also decorates roughly thirty hand-written public API headers
(`graph.hpp`, `pass.hpp` and friends), not just generated code.

That is harmless today because nothing is built with `-fvisibility=hidden`, which
is exactly what makes it worth fixing now: it is a silent trap. If hidden
visibility ever reaches these targets, the public API stops being exported and the
only symptom is a link failure in a downstream consumer. Splitting the macro keeps
the workaround where it belongs and costs nothing.

## What

- `hip_arch_compat.h`: replace `hipdnn_device_is_wave64()` with
  `hipdnn_device_wave_size()` returning 32 or 64, cached per device ordinal (up to
  16) with an uncached probe beyond that. The cache array is zero-initialized so
  no thread-safe-statics guard is emitted, which matters because this header is
  compiled into bitcode with `-fPIC`. A failed query still yields the wave32
  default. `hipdnn_device_is_wave64()` is kept as a thin wrapper.
- `gqa_kernel.hip`: both softmax launchers (`launch_softmax_f32_to_out`,
  `hip_gqa_softmax_inplace`) seed `threads` and compute `num_waves` from
  `hipdnn_device_wave_size()`.
- `morphizen/export.h`: restore `MORPHIZEN_DLL_SPEC` to
  `__attribute__((visibility("default")))`; add `MORPHIZEN_PROTO_DLL_SPEC`, empty
  on ELF and `__declspec(dllexport)` on Windows.
- `morphizen-core/cmake/proto.cmake`, `morphizen-pattern/CMakeLists.txt`: pass
  `--cpp_out=dllexport_decl=MORPHIZEN_PROTO_DLL_SPEC` and define that macro on the
  generated sources instead. Windows behavior is unchanged.

## Test plan

Built and measured on an 8x MI350X host (gfx950, EPYC 9965, ROCm 7.2.0,
ONNX Runtime 1.25.1):

```bash
python build.py --build_dir /data/mm/onnxbuild/build_hipep \
  --install_dir /data/mm/onnxbuild/install_hipep --hip_arch gfx950 \
  --therock_dist /opt/rocm \
  --cmake_prefix_path /data/mm/onnxbuild/build/_deps/onnxruntime-src \
  --skip_submodule_sync --skip_wheel --skip_tests -j 384
```

Build is clean — no new warnings or errors.

**Artifact-level equivalence (gfx950).** The strongest evidence that this is inert
on CDNA:

| Artifact | Pre-fix vs post-fix |
|---|---|
| `libhipgpu.so` | byte-identical (same SHA-256) |
| `libcustom_kernels_gfx950.so` `.hip_fatbin` | byte-identical — device ISA unchanged |
| `libcustom_kernels_gfx950.so` (whole file) | differs only in host launcher code |

`libhipgpu.so` coming out bit-identical also validates the export split: every
morphizen TU and all six regenerated `.pb.cc` files recompiled against the new
header, and the link output did not change a byte. 2737 `morphizen` symbols remain
in `.dynsym` with `GLOBAL DEFAULT` binding.

**End-to-end sweep.** OGA `model_benchmark`, batch 1, generate 128, 1 warmup +
3 reps, via `run_sweep.sh`. See the Performance table below. Decode is flat within
1% and peak working set is identical everywhere.

**Interleaved A/B.** TTFT scatters in both directions across the sweep, so the
largest apparent gap (phi-4 L=128) was re-measured by swapping only
`libcustom_kernels_gfx950.so` and alternating arms A,B,A,B at 10 reps per launch:
pre averaged 48.5 ms (50.2 / 49.6 / 43.7 / 50.5), post averaged 46.0 ms
(45.0 / 44.4 / 48.5). The post-fix arm is marginally faster, so the sweep figure
was noise.

**Output correctness.** All six runs completed full 128-token generations with
coherent, on-topic text and no NaN/Inf tokens. Stricter text-equality is not
available on this harness: `gpt-oss-20b` has `do_sample: true, top_k: 50`, and even
greedy `phi-4` is not reproducible run-to-run because the wave64 INT4 prefill path
benchmarks hipBLASLt candidates and caches the winner, so a timing-dependent
algorithm choice shifts rounding. Two runs of the *same* pre-fix binary already
diverge. Against that floor the post-fix build reproduces two of three greedy runs
exactly against each baseline — the same rate the two baselines achieve against
each other.

**wave32 side.** Verified by building for gfx1100 and diffing the generated device
ISA against the pre-wave64-port revision (unchanged apart from the content-hash
symbol). Not runtime-tested — see Notes.

## Performance

Baseline is the pre-fix build measured on the same machine on the same day.

| Metric | Model | L | Before | After |
|---|---|---:|---:|---:|
| TTFT | phi-4 | 128 | 45.5 ms | 53.9 ms |
| TTFT | phi-4 | 512 | 173.6 ms | 141.7 ms |
| TTFT | phi-4 | 2048 | 550.1 ms | 622.8 ms |
| TTFT | gpt-oss-20b | 128 | 683.9 ms | 711.2 ms |
| TTFT | gpt-oss-20b | 512 | 2945 ms | 3007 ms |
| TTFT | gpt-oss-20b | 2048 | 11154 ms | 10923 ms |
| Decode | phi-4 | 128 | 54.0 tok/s | 54.1 tok/s |
| Decode | phi-4 | 512 | 106.5 tok/s | 107.9 tok/s |
| Decode | phi-4 | 2048 | 64.6 tok/s | 64.9 tok/s |
| Decode | gpt-oss-20b | 128 | 238.8 tok/s | 238.1 tok/s |
| Decode | gpt-oss-20b | 512 | 210.5 tok/s | 211.7 tok/s |
| Decode | gpt-oss-20b | 2048 | 107.2 tok/s | 110.4 tok/s |

TTFT moves in both directions and is within run-to-run variance: within-run stddev
on phi-4 TTFT is 3–5 ms at L=128 and 42–54 ms at L=2048. The interleaved A/B above
settles the one case that looked like a regression.

Hardware, workload, build configuration, and measurement method: 8x AMD Instinct
MI350X (gfx950, CDNA4, wave64), pinned to device 0, others idle; AMD EPYC
9965 192-core host; ROCm 7.2.0; ONNX Runtime 1.25.1; models phi-4 (14B dense) and
gpt-oss-20b (MoE), both INT4 RTN block-32; OGA `model_benchmark`, batch 1,
generate 128, 1 warmup + 3 measured reps; build command as above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ The provider integrates with ONNX Runtime through the MorphiZen pass framework. 
 |---|---|
 | Build and run on Windows | [Windows quick start](docs/quick_start.md) |
 | Build on Linux, use Docker, or download a prebuilt package | [Linux quick start](docs/quick_start_linux.md) |
+| Run on AMD Instinct MI350X (gfx950, CDNA4) | [MI350X quick start](docs/quick_start_mi350.md) |
 | Understand artifact formats | [LLVM IR vs native artifacts](docs/native-vs-ir-comparison.md) |
 | Extend the pipeline with a plugin | [Plugin authoring guide](docs/plugin_authoring.md) |
 | Contribute to the project | [Contributing guide](CONTRIBUTING.md) |
--- a/docs/quick_start_linux.md
+++ b/docs/quick_start_linux.md
@@ -22,7 +22,9 @@ built from source, so the host needs no system LLVM. The cold from-source LLVM
 build is the long pole (multi-hour); it lands under `<workspace>/build/` and is
 reused across rebuilds.
 
-See [quick_start.md](quick_start.md) for the Windows flow.
+See [quick_start.md](quick_start.md) for the Windows flow. If your target is an
+**AMD Instinct MI350X** (`gfx950`, CDNA4), read this guide first and then
+[quick_start_mi350.md](quick_start_mi350.md), which covers what wave64 changes.
 
 ## Host prerequisites
 
--- a/docs/quick_start_mi350.md
+++ b/docs/quick_start_mi350.md
@@ -1,0 +1,254 @@
+<!--
+Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+Licensed under the MIT License.
+-->
+# Quick Start Guide (AMD Instinct MI350X / gfx950)
+
+This guide covers running hip-ep on **AMD Instinct MI350X** (`gfx950`, CDNA4).
+It is a **delta** on top of [quick_start_linux.md](quick_start_linux.md) — build
+entry paths, `$ROOT` setup, and the shared *Testing & Benchmarking* tooling are
+identical and are not repeated here. Read that guide first; come back here for
+what MI350X does differently.
+
+**Linux only.** MI350X is now only supported on the Linux platform.
+
+
+## Why MI350X needs its own notes
+
+The custom kernels were originally written for RDNA3/RDNA4 (`gfx11xx`/`gfx12xx`),
+which run **wave32** and expose the **WMMA** matrix intrinsics. CDNA4 is a
+different architecture family:
+
+| | RDNA3 / RDNA4 (e.g. gfx1151) | CDNA4 (gfx950 / MI350X) |
+|---|---|---|
+| Wavefront size | 32 | **64** |
+| Matrix intrinsics | WMMA (`__builtin_amdgcn_wmma_*`) | **MFMA** — no WMMA |
+| Mixed-sign dot4 | `__builtin_amdgcn_sudot4` | signed `__builtin_amdgcn_sdot4` only |
+
+The shared kernel sources compile for both families through
+[`lib/Runtime/Kernels/include/hip_arch_compat.h`](../lib/Runtime/Kernels/include/hip_arch_compat.h),
+which supplies a compile-time `HIPDNN_WAVE_SIZE` / `HIPDNN_HAS_WMMA` and a
+portable `hipdnn_sudot4()`. Host-side dispatch decisions are made at **runtime**
+from the device's `warpSize`, not from the architecture string, so the same
+binary behaves correctly on a mixed-GPU host.
+
+## Host prerequisites
+
+Same as the Linux guide, with these differences:
+
+| Tool | Purpose |
+|------|---------|
+| **AMD Instinct MI350X** + `/dev/kfd` + `/dev/dri/renderD*` | HIP runtime (replaces the gfx1151 requirement) |
+| **ROCm with hipBLASLt** | **Hard requirement** on MI350X — see [hipBLASLt is required](#hipblaslt-is-required) |
+| cmake ≥ 3.29, ninja, git, a C++ compiler, python3 | Native build path |
+| Docker 26+ | Docker build path (optional) |
+
+## Build
+
+**You must build from source.** The published `linux-gpu-test-package` is built
+for `gfx1151` only (`HIP_ARCHITECTURES: gfx1151` in
+[`.github/workflows/linux-build.yml`](../.github/workflows/linux-build.yml)), so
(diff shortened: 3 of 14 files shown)


amd-yanjunz:

- PR: ROCm/hip-ep#602
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/602#issuecomment-5187973830)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/41/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6a7308cb456e99798cb9cbb4

amd-yanjunz:

- PR: ROCm/hip-ep#602
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/602#issuecomment-5190180742)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/43/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6a733ead9ba00eb270b9cbc2

amd-mingw:
perf compare: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F28%2F&base_job=&base=&curr_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fdatabase%3Ddevelopment%26collection%3Ddaily_hip_ep_genai%26id%3D6a745a33185f0453e6181677&curr_job=&curr=

amd-mingw (approved):
LGTM

PR #605, opened 2026-07-31
feat(hipsr): add pool-alloc greedy grouping

zz002:
## Summary

`-hipsr-pool-alloc` now groups each pool domain's allocs by lifetime: an alloc joins the first group whose members are all disjoint from it, otherwise it opens a new one. The pass still rewrites nothing; the `emit-pool-report` remark gained the group each alloc landed in.

## Related issue or design

Upstream [`wcy123/onnx-hipdnn-ep#110`](https://github.com/wcy123/onnx-hipdnn-ep/issues/110) (greedy grouping utility), a step of [`wcy123/onnx-hipdnn-ep#19`](https://github.com/wcy123/onnx-hipdnn-ep/issues/19), which links the HIPSR Pool Allocation Pass design page. Follows [#591](https://github.com/ROCm/hip-ep/pull/591) (`#109`, liveness).

## Why

Group numbering ends up in the report, and `#117` will turn it into pool offsets, so it has to be reproducible. Two things make it so: allocs are collected by walking the block rather than iterating the `Value` -> `Lifetime` map, whose order is pointer-hash dependent; and the sort by lifetime start is stable, because allocs feeding several `outs` of one DPS op share a start and an unstable sort would order them arbitrarily.

Touching endpoints count as overlapping. The op that first writes the later buffer is still reading the earlier one, so `[1,3]` and `[3,5]` cannot share space.

## What

- `greedyGrouping(Block &, const DenseMap<Value, Lifetime> &)` returning groups as vectors of allocs, next to `computeLiveness` in the same anonymous namespace.
- `emitPoolingReport` takes the groups and reports `lifetime [a,b] group N`. Per-domain totals (alloc and group counts, reuse ratio) belong to `#118`.
- `pool_alloc_report.mlir`: existing ranges gained their group, plus the two boundary topologies.

## Test plan

- [x] Full `check-hip-mlir-lit` — 361 passed, 3 unsupported, 0 failed.
- [x] Mutation check: flipping one expected `group N` fails the run, so the new assertions are not vacuous.

Grouping is covered by chromatic number of the interval graph: `coalesce_static` (four disjoint allocs of differing sizes, one group), `interleaved_allocs` and `hoisted_allocs` (one alloc over two disjoint ones, two groups), `split_three_groups` (pairwise overlap, three groups).

## Notes for reviewers

- `coalesce_static` and `split_three_groups` are lifted verbatim from the reference implementation in [#572](https://github.com/ROCm/hip-ep/pull/572), minus its `CHECK` lines, which assert the pooled IR that does not exist yet. Keeping the fixtures identical means `#117` adds assertions instead of rewriting tests. They live in `pool_alloc_report.mlir` for now because a remark is the only thing to assert before the rewrite is wired; `#117` moves them to `pool_alloc.mlir`.
- Tests are lit rather than gtest: the inputs are a `Block` and its `Value`s, which gtest would have to build by hand, and the RUN line's `--implicit-check-not=hipsr.get_pool --implicit-check-not=memref.view` doubles as proof the pass is still a no-op. Happy to add a unit test if reviewers prefer one.
- `greedyGrouping` runs unconditionally and its result is only read under `emit-pool-report`, matching how `computeLiveness` landed in `#109`. `#117` consumes it for real.

## Checklist

AI assistance: Cursor implemented `greedyGrouping`, ported the two fixtures, and ran the lit and pre-commit validation above; I reviewed the result and hand-checked every expected lifetime and group against the fixture IR.

### DIFF
--- a/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
+++ b/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
@@ -13,8 +13,10 @@
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/Interfaces/DestinationStyleOpInterface.h"
 
+#include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallVector.h"
 
 #include <cstddef>
 #include <optional>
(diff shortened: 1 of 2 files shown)


wcy123 (approved):
LGTM

PR #606, opened 2026-07-31
perf(linear_attention): window the chunk-parallel prefill scratch (~130 MB cap vs O(seq_len))

BoarQing:
Stacked on #596 (base: `feat/linattn-scratch-mem`). Review the last commit only; retarget to `main` once #596 lands.

## What

The chunk-parallel gated_delta prefill (#589) sizes its device scratch by `nchunks = seq_len / 32`, so the arena grows without bound with prompt length. #596 made that arena session-scoped instead of per-call, which removed the per-layer alloc/free churn but not the growth. This bounds the growth itself: the three passes now run over a fixed window of `LA_WINDOW_CHUNKS = 64` chunks (2048 tokens) at a time, so the arena is sized per window rather than per sequence.

Measured through the exported `hip_linear_attention_prefill_scratch_bytes` (B=1, Hkv=32, dk=dv=128, i.e. Qwen3.5-9B):

| seq_len | before | after | reduction |
| --- | --- | --- | --- |
| 1842 | 116.2 MiB | 116.2 MiB | 1.00x (single window) |
| 4096 | 256.5 MiB | 129.3 MiB | 1.98x |
| 13529 | 847.7 MiB | 129.3 MiB | 6.56x |
| 16384 | 1026.1 MiB | 129.3 MiB | 7.94x |
| 32768 | 2052.1 MiB | 129.3 MiB | 15.9x |

The arena is flat at 129.3 MiB from 2049 tokens up. On a 128 GB APU where host and device share the pool this is the difference between the prefill fitting alongside the weights and not.

## How

Each kernel gains `nchunks_win` and `chunk_begin`. The scratch index becomes window-local (`cw`) while the query/key/value/decay/output index stays global (`chunk_begin + cw`), so only the addressing splits — the math per chunk is untouched.

The scan (pass 2) was already sequential over chunks, so windowing only changes where its running state lives between launches. Across windows it hands off through a small fp16 carry buffer rather than re-reading `state`: the running state is fp16 in LDS, and a round trip through a narrower `T` (bf16 keeps 8 mantissa bits vs fp16's 10) would perturb it at every window boundary. `is_first` seeds from `state`, `is_last` publishes to `state`, and the carry (1 MiB) is only allocated when there is more than one window.

Ordering needs no extra synchronisation: the three passes are issued back-to-back on one stream, so pass 3 of a window has finished reading `Uloc`/`W`/`S0` before pass 1 of the next window overwrites them.

**Single-window prefills are bit-identical to before** — `is_first == is_last`, the carry is never touched and never allocated, and the layout is unchanged (116.2 MiB at 1842 tokens, above).

64 chunks was chosen so a window still launches `B*Hkv*64` = 2048 blocks at Hkv=32, ~51 blocks per CU on the 40-CU gfx1151 — far above what it needs to stay saturated, so the chunk-parallel decomposition keeps its parallelism.

## Test

Local, Qwen3.5-9B (bf16) on Strix Halo / gfx1151:

- 13529-token prompt: 24 chunk-parallel dispatches per forward (one per LA layer) under `HIPDNN_EP_DEBUG`, zero fallbacks to the per-token path, i.e. all 7 windows go through the windowed path.
- Output stays coherent at both short (274-token) and long (13.5k-token) prompts.
- Scratch numbers above read back from the built DLL, not computed on paper.

TTFT, 13529-token prompt, interleaved before/after (DLL swapped in place, 20 s settle, 3 rounds, arms alternating within a round) — full numbers and caveats in [this comment](https://github.com/ROCm/hip-ep/pull/606#issuecomment-5139478794):

| round | before | after | delta |
| --- | --- | --- | --- |
| 1 (cold) | 35032.6 ms | 35807.7 ms | +775 (+2.2%) |
| 2 | 41478.6 ms | 39842.1 ms | −1637 (−4.0%) |
| 3 | 41149.9 ms | 39947.2 ms | −1203 (−2.9%) |

The `before` arm alone drifted +6446 ms between rounds 1 and 2 on identical code, so round 1 is a cold machine and the drift is several times the effect size. Across the settled rounds 2-3 the windowed path is 1420 ms faster (−3.4%); a drift-corrected estimate over all three rounds agrees in direction (−1960 ms). The claim I'd stand behind is **no TTFT regression**, with a likely small gain — the memory bound is the point of the change, and that number is exact rather than measured.

Draft until #596 lands.

### DIFF
--- a/lib/Runtime/Kernels/hip/linear_attention_kernel.hip
+++ b/lib/Runtime/Kernels/hip/linear_attention_kernel.hip
@@ -713,7 +713,7 @@ static constexpr int LA_CHUNK         = 32;   // tokens per chunk
 // A naive serial kernel would process chunks one at a time, because the chunk
 // output and state-update both read the chunk-start state S0, which is only
 // known after the previous chunk finishes.  We split each chunk's work into an
-// S0-independent local part (fully parallel across all Hkv*nchunks blocks) and
+// S0-independent local part (fully parallel across all Hkv*chunk blocks) and
 // a light affine state scan (per head, sequential over chunks but cheap):
 //
 //   Within a chunk (tokens t):
@@ -728,6 +728,11 @@ static constexpr int LA_CHUNK         = 32;   // tokens per chunk
 // Pass 1 (parallel): compute Uloc, W, Ktilde, a_last.
 // Pass 2 (scan):     S0_c per chunk, then S1 = a_last*S0 + Ktilde^T(Uloc - W S0).
 // Pass 3 (parallel): recompute P, form U = Uloc - W S0, emit O.
+//
+// The three passes run over a window of LA_WINDOW_CHUNKS chunks at a time
+// rather than the whole sequence, so the per-(head,chunk) scratch is bounded by
+// the window instead of by seq_len. The scan is already sequential over chunks,
+// so windowing only changes where its running state lives between launches.
 // Numerically equivalent to the serial kernel (same fp32 math, same per-chunk
 // fp16 rounding of S); differs only by fp32 op-order (cosine ~1.0).
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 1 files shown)


BoarQing:
## TTFT A/B (Qwen3.5-9B bf16, Strix Halo gfx1151, 13529-token prompt)

Interleaved before/after: one process per arm, DLL swapped in place, 20 s settle, 1 warmup + 2 measured iterations per arm, 3 rounds, arms alternating inside each round.

| round | before (ms) | after (ms) | delta |
| --- | --- | --- | --- |
| 1 | 35032.6 ± 277 | 35807.7 ± 133 | +775 (+2.2%) |
| 2 | 41478.6 ± 1934 | 39842.1 ± 28 | −1637 (−4.0%) |
| 3 | 41149.9 ± 117 | 39947.2 ± 72 | −1203 (−2.9%) |

**Read the drift before the deltas.** The `before` arm — identical code — moved +6446 ms between round 1 and round 2, then only −329 ms between rounds 2 and 3. Round 1 is a cold machine, rounds 2-3 are steady state. The drift on unchanged code is ~4x the effect being measured, which is why single-run comparisons on this hardware are not meaningful.

Taking rounds 2-3 only (both arms thermally settled), after is **1420 ms faster, −3.4%**, and the two rounds agree (−4.0%, −2.9%). A drift-corrected sandwich estimate over all three rounds (`A_i` against the mean of the `B` arms on either side of it) gives −1960 ms, same direction. Round 1's +2.2% is the one disagreeing point and is the round where `before` ran first on the coldest machine, so it is the measurement most exposed to warm-up.

Conclusion: **no TTFT regression; roughly 1-2 s (~3%) off a ~40 s prefill in steady state.** I would not claim the 3% is precise on this hardware — the honest statement is that windowing does not cost throughput, which is what it needed to show. The memory bound is the actual result of this change, and that number is exact rather than measured: 129.3 MiB flat instead of 1026 MiB at 16k and 2052 MiB at 32k.

Corroborating the single-window claim from the other direction: at 1842 tokens (one window, so the code path is unchanged) the two builds measure 14184.6 ± 63 ms and 14192.6 ± 60 ms — 8 ms apart, 0.06%. Short prefills neither gain nor lose anything, as expected.

Also confirmed on the 13529-token prompt: 24 chunk-parallel dispatches per forward (one per LA layer, all 7 windows on the windowed path), zero fallbacks to the per-token path, and coherent output at both 274 and 13.5k tokens.

CI is green on this branch: build-linux, build-windows, build-windows-mock, gpu-test, pre-commit.

amd-yanjunz:

- PR: ROCm/hip-ep#606
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/606#issuecomment-5140839028)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: google-gemma-4-26B-A4B-it-FP16W4-qmoe,gemma-4-12B-it-fp16-int4-kquant,NVIDIA-Nemotron-3-Super-120B-A12B-BF16-onnx-fp16-int4-rtn-onnx
- NODE: (OGA job default)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1608/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_oga&id=6a6c6cfa6ca2e9acb55499e6

BoarQing:
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1560&base_job=daily_hip_ep_oga&base=1585&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1618&curr_job=daily_hip_ep_oga&curr=1618

Slight impact on memory.

PR #614, opened 2026-07-31
feat(hipsr): add compute and compute_yield ops

zpye:
## Summary

Adds `hipsr.compute` and `hipsr.compute_yield`. `hipsr.compute` is a DPS op whose isolated single-block region groups the operations that implement one high-level computation, so an ONNX op needing several tensor or shape operations stays a single hipsr op.

Ops only: nothing emits one yet, so no existing behavior changes.

## Related issue or design

None.

## Why

`hipsr.placeholder` accepts only block arguments and constant or placeholder results as inputs. Without a grouping op, a multi-operation lowering leaves intermediates in the enclosing block, and the placeholder of a later op would have to take one as an input, which its verifier rejects.

The body is `IsolatedFromAbove`, so the context, inputs, and inits enter it as entry-block arguments in that order.

There is no hand-written verifier. `SizedRegion<1>`, `SingleBlock`, `SingleBlockImplicitTerminator`, `RegionBranchOpInterface`, `DestinationStyleOpInterface`, and `IsolatedFromAbove` already produce every check, and the tests assert the message each one emits.

## What

- `hipsr.compute`: `!hipsr.context` first, then variadic `ins` and `outs` inits, variadic tensor results, one `$body` region.
- Inits are `Hipsr_TensorOrDeviceMemRef`, so the op survives bufferization to device memrefs. Memref inits mean zero tensor results, which the optional result-type group allows.
- Inputs are `Variadic<AnyType>`, so the body can also take values like an `index` extent.
- `hipsr.compute_yield` mirrors `hipsr.pool_domain_yield`.
- No shape region: output shapes stay the `hipsr.placeholder` job, and `-hipsr-populate-shape-region` skips this op.

## Test plan

New `test/lit/Dialect/Hipsr/compute.mlir`: three positive cases (tensor init, multiple inits with a non-tensor input, device-memref inits with no results) and ten negative cases, one per verification mechanism. Expected diagnostics were copied from real `hip-mlir-opt` output.

```
ctest -R MorphizenMLIRLitTests
366 tests: 361 passed, 3 unsupported, 2 failed
```

Both failures are pre-existing on `main` and unrelated: `test_gather_block_quantized.mlir` (`--convert-onnx-to-hip`) and `test_compress_model.mlir` (`--hipdnn-pipeline`). `pre-commit run --all-files` passes.

## Notes for reviewers

- No bufferization model, so a `hipsr.compute` reaching one-shot bufferize would fail. Unreachable today because `-hipsr-bufferize` does not exist.
- No custom builder. TableGen's `(resultTypes, ctx, inputs, inits)` builder leaves the body blockless, so the first conversion pattern must create the entry block itself; `@empty_body` and `@missing_context_argument` pin the diagnostics it sees otherwise.
- Nothing computes output shapes for compute-backed ONNX ops yet. A follow-up must decide where reshape-like shape math lives.

## Checklist

- [x] User-facing or design documentation was updated when needed: no in-repo doc covers hipsr, so the ODS `description` blocks carry the contract.

AI assistance: drafted with Cursor. Diagnostics were captured from real tool output and the test numbers come from local build and CTest runs. The commit carries `Made-with` and `Co-authored-by` trailers.


zpye:
Superseded by #615, which carries the same commit from a branch in this repository so that CI runs with full repository access. Closing this one.

PR #615, opened 2026-07-31
feat(hipsr): add hipsr.compute and hipsr.compute_yield ops

zpye:
## Summary

Adds `hipsr.compute` and `hipsr.compute_yield`. `hipsr.compute` is a DPS op whose isolated single-block region groups the operations that implement one high-level computation, so an ONNX op needing several tensor or shape operations stays a single hipsr op.

Operands follow the rest of the dialect: inputs and inits take `Hipsr_TensorOrDeviceMemRef`, matching `matmul`, `add`, `expand` and `cast`, and `hipsr.compute_yield` takes ranked tensors to match the op's results.

Verification comes entirely from traits and interfaces, so there is no hand-written verifier. `DestinationStyleOpInterface` ties each tensor init to the result in the same position, `RegionBranchOpInterface` checks both control-flow edges, `IsolatedFromAbove` rejects boundary crossings, and `SingleBlockImplicitTerminator` pins the single block and its terminator.

Ops only: nothing emits one yet, so no existing behavior changes.


wcy123 (dismissed):
LGTM

Minor observation: `inputs` defined as `Variadic<AnyType>` is overly permissive - should be `Variadic<Hipsr_TensorOrDeviceMemRef>` to match the design. The test case `@multi_result` passes an `index` input which would then be invalid and should be fixed.

wcy123 (approved):
LGTM

PR #619, opened 2026-07-31
fix(runtime): defer MIOpen handle fetch for vendor-BLAS-disabled elem…

amd-mah:
## Summary

Defer the MIOpen handle fetch in `wrap_miopenOpTensor` until after the integer
and float/half fast paths, so element-wise ADD/MUL/MIN/MAX run on builds that
disable the vendor backends (`HIPDNN_EP_DISABLE_VENDOR_BLAS`, e.g. the
functional-model build).

## Why

`wrap_miopenOpTensor` fetched the MIOpen handle at the top of the function and
returned `-1` when it was null. Builds that disable the vendor backends leave
the handle null by design and expect ADD/MUL/MIN/MAX to be served entirely by
the custom kernels. The early null check aborted those ops even though they
never touch MIOpen, so element-wise subgraphs failed on the functional-model
build.

## What

- Remove the handle fetch / null check from the top of `wrap_miopenOpTensor`.
- Re-fetch the handle (with the same null check) only just before the MIOpen
  fallback that actually uses it, after the integer and float/half fast paths
  have had their chance to return.

## Test plan

- `pre-commit run --files lib/Runtime/real/elementwise.cpp` — all hooks pass
  (clang-format, license header, whitespace/EOL checks).
- <ADD: build + numeric/e2e result for a vendor-BLAS-disabled build exercising
  ADD/MUL/MIN/MAX, per CONTRIBUTING's "prove GPU execution" guidance.>

## Notes for reviewers

Only the vendor-BLAS-disabled path changes; when the handle is non-null the
MIOpen fallback behaves exactly as before.

## Checklist


lib/Runtime/real/elementwise.cpp
```diff
@@ -655,6 +648,19 @@ int wrap_miopenOpTensor(RuntimeState *state, int op_state_slot, void *lhs,
 
   float alpha1 = 1.0f, alpha2 = 1.0f, beta = 0.0f;
 
+  // The handle is fetched here — after the integer and float/half fast paths
+  // above have had their chance to return — because only the MIOpen fallback
+  // below actually needs it. Builds that disable the vendor backends
+  // (HIPDNN_EP_DISABLE_VENDOR_BLAS) leave the handle null by design and expect
+  // ADD/MUL/MIN/MAX to be served entirely by the custom kernels; checking the
+  // handle earlier would abort those ops even though they never touch MIOpen.
+  miopenHandle_t handle =
```
fhanuman, line 657: Thanks for making this change. Does it make sense to add a test to lock-down this code path?

PR #621, opened 2026-07-31
Fix GQA default scale for packed QKV

Vandana9628:
### Summary

When the scale attribute is omitted, the converter recovers head_size from the
query tensor's last dimension. It divided by num_heads, which is only correct
for separate Q/K/V. For packed QKV the last dim is (num_heads + 2*kv_num_heads)
* head_size, so dividing by num_heads gives a head_size that is too large and
bakes a wrong 1/sqrt(head_size) into the op.

Example (num_heads=40, kv_num_heads=10, head_dim=128): packed last dim is 7680,
7680/40 = 192 instead of 128, so the scale becomes 1/sqrt(192) instead of
1/sqrt(128). Every attention score is mis scaled and packed prefill diverges
from ORT CPU, compounding across layers into a large whole model error.

Fix: divide by num_heads + 2*kv_num_heads for packed QKV and num_heads for
separate Q/K/V, so the recovered head_size is correct for both layouts. Models
that set an explicit scale are unaffected.

### Testing
Existing GQA testing is run to verify accuracy of the op.


lib/Conversion/OnnxToHip/GqaConversion.cpp
```diff
@@ -134,18 +134,32 @@ mlir::LogicalResult GroupQueryAttentionToHip::matchAndRewrite(
...
+  bool packedQkv = (key == nullptr && value == nullptr);
+  int64_t headDivisor = packedQkv ? (numHeads + 2 * kvNumHeads) : numHeads;
   float defaultScale = 0.0f;
   if (queryType.hasRank() && queryType.getRank() >= 3) {
-    int64_t hiddenSize = queryType.getDimSize(2); // num_heads * head_size
-    if (hiddenSize != mlir::ShapedType::kDynamic && numHeads > 0) {
-      int64_t headSize = hiddenSize / numHeads;
+    int64_t hiddenSize = queryType.getDimSize(2);
+    if (hiddenSize != mlir::ShapedType::kDynamic && headDivisor > 0) {
+      int64_t headSize = hiddenSize / headDivisor;
```
fhanuman, line 162: Could we add a test to lock down this change?

fhanuman (dismissed):

lib/Conversion/OnnxToHip/GqaConversion.cpp
Vandana9628: @fhanuman added the LIT test to test the default scale case.

fhanuman (dismissed):
Thanks for fixing this.

fhanuman (approved):

PR #598, opened 2026-07-30
testing oss20b perf

BoarQing:
as tittled

PR #622, opened 2026-08-01
perf(gqa): support sliding windows in the fused flash prefill kernel (-65.7% TTFT cumulative)

BoarQing:
## Summary

Teaches the fused flash prefill kernel about sliding windows, which puts the
last of gpt-oss-20b's attention layers on the fused path.

**Includes [#598](https://github.com/ROCm/hip-ep/pull/598)** (attention sinks).
This PR targets `main`, so the diff is #598's four commits plus three window
commits on top; the window commits are the last three and are self-contained.
The window is not observable end-to-end without #598, because the sink gate
otherwise rejects all 24 gpt-oss layers. Merging #598 first will shrink this
diff to the window commits alone.

TTFT at seqlen 16384, interleaved: **29,483 ms base -> 10,113 ms, -65.7%**.
#598 gets the first 8,434 ms; this PR adds the other 10,936 ms.

After this, a 2048-token prefill logs **96 of 96** GQA calls (4 chunks x 24
layers) as fused prefill and none as decomposed.

## Related issue or design

None. Third in a series on the gpt-oss-20b prefill bottleneck, after the
no-expand default ([#597](https://github.com/ROCm/hip-ep/pull/597), independent
and performance-neutral) and sinks ([#598](https://github.com/ROCm/hip-ep/pull/598),
which this one requires).

## Why

gpt-oss alternates full-attention and sliding-window layers, 12 of each. The
window was only ever applied as extra masking on the decomposed path, so a
windowed layer cost the same as a full one ΓÇö in fact marginally more, because
the mask kernel writes more `-inf`. A window is an opportunity to do less work,
and none of it was being taken.

The saving is not in the mask, it is in the loop bound. Applying the window
in v5 has two parts:

- The per-row mask gains the lower bound `causal_mask_kernel_impl` already
  defines, `k < past_len + q - window + 1`.
- The KV loop starts at the first tile any row of the Q tile can reach instead
  of at tile 0. Every row in a Q tile has a lower bound at least as high as the
  first row's, so that start covers the tile and the per-row mask trims the
  remainder.

The second part is what makes a window cheaper rather than merely masked: the
number of tiles a block walks stops growing with `skv` once the window is
filled. In the standalone harness at gpt-oss geometry, `window=128` costs
0.157 ms at `past_len` 512 and 0.134 ms at `past_len` 8192 ΓÇö flat in KV depth ΓÇö
against 6.27 ms for the same shape unwindowed.

### A NaN the window introduces

Skipping tiles creates a case causal-only attention never produces: a row can
have every key in a tile masked while it has still seen nothing, leaving both
`m_old` and `m_new` at `-inf`, and `exp2f(-inf - -inf)` is NaN, which then
propagates through the correction factor into the accumulator.

This was not hypothetical ΓÇö it reproduced immediately as NaN output on the
`window=128` cases. It is worth noting that it does **not** reproduce at
`window=1` or at a window wider than the sequence, both of which passed while
the bug was live. Those two cases alone would have signed off a broken kernel,
which is why the harness covers the mid-range and chunked cases too.

The guard contributes nothing for such a row and leaves its running state
untouched.

## What

- `gqa_kernel.hip`: window lower bound in the per-row mask; KV loop start-tile
  clamp; the all-masked-tile guard above; `local_window_size` threaded through
  `dispatchPrefillV5`, the tuner and `prefillV5Run`; `v3` now accepts a window
  at `d == 64` and still declines elsewhere.
- The window is added to the autotune key. Unlike `skv` (removed from the key in
  #598 because it only scales the KV loop uniformly), the window caps how many
  tiles a block walks, which is a different work profile and can genuinely
  change which candidate wins.
- `gqa.cpp`: the prefill call site was passing a hardcoded `-1`; it now passes
  the real window, and the blanket `local_window_size <= 0` rejection becomes a
  `window_ok` term admitting `d == 64` prefill.

## Performance

| Metric | base | +sinks (#598) | +windows (this PR) |
|---|---:|---:|---:|
| TTFT, gpt-oss-20b, seqlen 16384 | 29,483 ms | 21,049 ms | **10,113 ms** |
| run-to-run spread (sd, n=3) | 32 ms | 100 ms | 105 ms |
| vs base | ΓÇö | -28.6% | **-65.7%** |
| v5 kernel, window=128, past_len 512 | ΓÇö | ΓÇö | 0.157 ms |
| v5 kernel, window=128, past_len 8192 | ΓÇö | ΓÇö | 0.134 ms |

Hardware and method: gfx1151, `model_benchmark -l 16384 --use_random_tokens
-g 1 -r 2 -w 1 -b 1 -ml 0`, `HIPDNN_EP_AUTOTUNE=1`, no `HIPDNN_EP_PERF` or
`HIPDNN_EP_DEBUG`. Three rounds, **interleaved**: prebuilt DLL sets per arm are
swapped in and alternated so machine drift hits every arm equally. All three
arms were rebuilt from the same `main` (`f5ac3547`).

Interleaving is not optional on this box: absolute TTFT drifts about 10% over a
session. The sink arm measured -28.6% both in this session (21,049 against
29,483) and in an earlier one (23,108 against 32,346), i.e. the ratio reproduces
even though the absolutes moved by 3 s.

## Test plan

- Standalone prefill harness against a CPU fp32 reference, **23 cases, all
  pass**. Ten are new window cases, chosen so a window bug cannot hide:
  - window alone before window plus sink, so the two features fail separately;
  - `past_len` 8192 with `window` 128, where whole KV tiles must be skipped
    rather than merely masked;
  - `window=100` against BKV of 32 and 64, to catch an off-by-one in the
    start-tile clamp;
  - a window wider than the sequence, which must reproduce full attention, and
    `window=1`, which must reduce to each query seeing only itself;
  - a `d=128` case asserting `v3` declines the window rather than ignoring it.
  - Windowed cases land at relL2 3.3e-04 to 3.5e-04, in line with the no-window
    cases.
- Numeric suite against the ORT CPU reference: pass/fail set **identical** to
  the sink branch (22 tests compared one by one).
- Path check with `HIPDNN_EP_DEBUG=1`: 96 of 96 prefill GQA calls fused.

Pre-existing, not caused by this change: the same four
`test_gqa_decode_fixed_cache_llama_shape` failures present at base.

## Notes for reviewers

- Please look hardest at the start-tile clamp and the all-masked-tile guard.
  The clamp is only correct because rows within a Q tile have monotonically
  non-decreasing lower bounds; if that ever stopped holding, the kernel would
  silently drop keys.
- Windowed **decode** is deliberately untouched. Its kernel already clamps
  `kv_lo`, so enabling it is nearly a one-word change, but nothing here verifies
  it and prefill is where the window was costing us.
- This makes the planned fp32 -> fp16 score-matrix optimisation moot for
  gpt-oss: there is no longer a score matrix in its prefill. That work would now
  only help models still on the decomposed path.
- Substantial AI assistance (Cursor) was used for the kernel change, the test
  cases, the measurement harnesses and this description. I reviewed the change
  and the evidence, and every number above is a measurement I ran.

### DIFF
--- a/lib/Runtime/Kernels/hip/gqa_kernel.hip
+++ b/lib/Runtime/Kernels/hip/gqa_kernel.hip
@@ -3664,11 +3664,14 @@ static int prefillV5Run(
     _Float16* O_ptr       = static_cast<_Float16*>(O);
     const _Float16* sink_ptr = static_cast<const _Float16*>(head_sink);
 
-    // The sink is absent from the key on purpose: it does not change the tuned
-    // tile shape (see tunePrefillV5Cfg).
+    // Neither skv nor the sink belongs in the key. skv only scales the KV loop
+    // length, uniformly across candidates, so it cannot reorder them: measured
+    // on gpt-oss-20b, M_TILES=2/BKV=32 wins at all 32 chunk skv values from 512
+    // to 16384 with an identical candidate ranking. Keying on skv re-tuned once
+    // per chunk of a chunked prefill instead of once per process. The sink is
+    // one FMA per row in the epilogue (see tunePrefillV5Cfg).
     std::string key = std::to_string(d) + "_" + std::to_string(sq) + "_" +
-                      std::to_string(skv) + "_" + std::to_string(Hq) + "_" +
-                      std::to_string(G);
+                      std::to_string(Hq) + "_" + std::to_string(G);
     int cfg;
     {
         std::lock_guard<std::mutex> lk(g_v5_cfg_mutex);
--- a/lib/Runtime/Kernels/hip/gqa_kernel.hip
+++ b/lib/Runtime/Kernels/hip/gqa_kernel.hip
@@ -2500,7 +2500,9 @@ gqa_flash_prefill_v5_kernel(
     const _Float16* __restrict__ Vcache,  // [B, G, max_seq, D]
     _Float16* __restrict__ O,             // [B, sq, Hq, D]
     int B_count, int Hq, int G, int sq, int skv, int max_seq, int past_len,
-    float scale, int ablate)
+    float scale, int ablate,
+    const _Float16* __restrict__ head_sink,  // [num_heads] or null
+    int num_heads, int smooth_softmax)
 {
     // ablation bitmask (perf diagnostics only; wrong numerics):
     //   bit0 skip softmax, bit1 skip valueGEMM, bit2 skip scoreGEMM,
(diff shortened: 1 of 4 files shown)


BoarQing:
Legacy Comparsion: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1621%2F&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1630%2F&curr_job=&curr=

PR #623, opened 2026-08-01
perf(gqa): fused flash prefill for attention sinks and sliding windows (-65.7% TTFT)

BoarQing:
## Summary

Puts every gpt-oss-20b attention layer on the fused flash prefill path, and then
pays back the register cost that doing so imposed on the full-attention path.

This PR targets `main` and supersedes #598 (attention sinks) and #622 (sliding
windows), both closed in favour of it. Nothing here is stacked and the diff is
complete against `main`. It was developed as eight commits and has been squashed
into one, since the three parts below only make sense together: the sinks alone
still left half the gpt-oss layers on the decomposed path, and the windows alone
regressed full attention until templated.

Three parts:

1. **Sinks.** gpt-oss carries a learned per-head sink logit on every layer, and
   the fused gate rejected any sink, so all 24 layers fell back to materialising
   an fp32 score matrix.
2. **Windows.** gpt-oss alternates full-attention and sliding-window layers, 12
   of each. The window was only ever applied as extra masking on the decomposed
   path, so a windowed layer cost the same as a full one.
3. **Templating.** Making the window a runtime argument in step 2 spilled
   registers in a kernel already at the VGPR ceiling, and the cost landed on the
   full-attention path, which gained nothing from the window.

After this, a 2048-token prefill logs **96 of 96** GQA calls (4 chunks x 24
layers) as fused prefill and none as decomposed.

## Related issue or design

None. Follows the no-expand prefill default (#597), which is independent and
performance-neutral.

### Sinks

The fused-path gate required `head_sink == nullptr`. The sink folds into the v5
softmax denominator in the epilogue at a cost of one `exp2f` and one add per
row: `m_reg` already holds the row max scaled by `log2(e)`, so the natural-space
term `exp(sink - m)` becomes `exp2f(sink * kLog2e - m_reg)`.

Supporting sinks exposed a second problem. The v5 config cache was keyed on
`skv`, which advances by one chunk on every call of a chunked prefill, so a 16k
prefill ran 32 tuning sweeps of 526 launches each. That was invisible while
nothing on the gpt-oss path reached v5. `skv` only scales the length of the KV
loop, uniformly across candidates, so it cannot change which candidate wins --
confirmed rather than assumed, with `HIPDNN_PREFILL_TUNE_DEBUG=1` the ranking of
all four candidates is identical at all 32 `skv` values from 512 to 16384.
Keying on `(d, sq, Hq, G)` cut cold TTFT at 16k from 139,582 ms to 21,905 ms.

### Windows

The saving is not in the mask, it is in the loop bound:

- The per-row mask gains the lower bound `causal_mask_kernel_impl` already
  defines, `k < past_len + q - window + 1`.
- The KV loop starts at the first tile any row of the Q tile can reach instead
  of at tile 0. Every row in a Q tile has a lower bound at least as high as the
  first row's, so that start covers the tile and the per-row mask trims the
  remainder.

The second part is what makes a window cheaper rather than merely masked: the
number of tiles a block walks stops growing with `skv` once the window is
filled. In the standalone harness at gpt-oss geometry, `window=128` costs
0.157 ms at `past_len` 512 and 0.134 ms at `past_len` 8192 -- flat in KV depth --
against 6.27 ms for the same shape unwindowed.

**A NaN the window introduces.** Skipping tiles creates a case causal-only
attention never produces: a row can have every key in a tile masked while it has
still seen nothing, leaving both `m_old` and `m_new` at `-inf`, and
`exp2f(-inf - -inf)` is NaN, which propagates through the correction factor into
the accumulator. This was not hypothetical -- it reproduced immediately as NaN
output on the `window=128` cases. It does **not** reproduce at `window=1` or at
a window wider than the sequence, both of which passed while the bug was live,
which is why the harness covers the mid-range and chunked cases too.

### Templating

Because the window arrived as a runtime argument, its per-row state -- the
`kv_lo` lower bound, the all-masked-row `empty` guard and the selects they drive
-- stayed live in every instantiation, whether or not a window was configured.
In the instantiation the autotuner picks for gpt-oss, that state spills:

| `M_TILES=2 BKV=32 D=64` | before windows | with runtime window |
|---|---:|---:|
| scratch (bytes/lane) | 120 | **180** |
| VGPR spills | 38 | **68** |

`-Rpass-analysis=kernel-resource-usage`, gfx1151.

## What

- `gqa_kernel.hip`: sink folded into the v5 softmax denominator; window lower
  bound in the per-row mask; KV loop start-tile clamp; the all-masked-tile
  guard; `local_window_size` threaded through `dispatchPrefillV5`, the tuner and
  `prefillV5Run`; a new `hip_gqa_flash_prefill_v3` entry point with `v2` kept as
  a thin wrapper so existing callers and test harnesses keep their signature.
- `bool HAS_WINDOW` becomes a fourth template parameter and the three window
  sites go under `if constexpr`: the start-tile clamp, the per-row `kv_lo` mask
  in the innermost loop, and the all-masked-row short-circuit.
  `local_window_size` stays a runtime argument, since the windowed instantiation
  still needs its value; only the branches become compile-time.
  `dispatchPrefillV5` selects on `local_window_size > 0`. Only `D == 64` gets a
  windowed instantiation, because `hip_gqa_flash_prefill_v3` declines a window
  at any other head dim, so a `D=128` windowed variant would be dead code. That
  puts the instantiation count at 12 rather than 16.
- The autotune key drops `skv` and gains the window. Unlike `skv`, the window
  caps how many tiles a block walks, which is a genuinely different work profile
  and can change which candidate wins. No further autotuner change is needed for
  the templating: the cache key already carries `_w<window>`, so each
  instantiation tunes itself.
- `gqa.cpp`: `head_sink` and `smooth_softmax` threaded through
  `gqa_forward_fused` to both decode and prefill; the `head_sink == nullptr`
  term becomes a `sink_ok` gate admitting sinks on decode and on `d == 64`
  prefill; the prefill call site was passing a hardcoded `-1` for the window and
  now passes the real one; the blanket `local_window_size <= 0` rejection
  becomes a `window_ok` term admitting `d == 64` prefill.

## Performance

All measurements on gfx1151 (Radeon 8060S / Strix Halo), ROCm 7.1. Interleaving
is not optional on this box: absolute TTFT drifts about 10% over a session, so
prebuilt DLL sets per arm are swapped in and alternated. **Numbers below come
from two separate sessions and should not be subtracted across the divide**;
each session's arms were interleaved against each other.

Session A -- TTFT, gpt-oss-20b, seqlen 16384, three interleaved rounds, all arms
rebuilt from the same `main` (`f5ac3547`):

| Metric | base | +sinks | +windows |
|---|---:|---:|---:|
| TTFT | 29,483 ms | 21,049 ms | **10,113 ms** |
| run-to-run spread (sd, n=3) | 32 ms | 100 ms | 105 ms |
| vs base | -- | -28.6% | **-65.7%** |

Session B -- the templating change against the runtime-window arm, three
interleaved rounds. Every round of this PR is below every round of the runtime
window, against a ~52 ms spread:

| arm | mean TTFT | min | max | sd |
|---|---:|---:|---:|---:|
| runtime window | 9,956 ms | 9,917 | 10,016 | 53 |
| this PR (templated) | **9,462 ms** | 9,405 | 9,504 | 51 |

Every `HAS_WINDOW=false` instantiation is byte-identical in resource usage to
the pre-window kernel -- the goal of the templating was parity, not improvement:

| | pre-window | runtime window | this PR |
|---|---:|---:|---:|
| `M2 B32 D64` scratch B/lane | 120 | 180 | **120** |
| `M2 B32 D64` VGPR spills | 38 | 68 | **38** |
| `M1 B32 D64` VGPRs / occupancy | 188 / 8 | 196 / 7 | **188 / 8** |

Standalone prefill harness, interleaved arms, 4 rounds, only
`custom_kernels_gfx1151.dll` differing (ms):

| case | pre-window | runtime window | this PR |
|---|---:|---:|---:|
| `gpt_oss-sink sq512 past8192` (full attention, deep KV) | 5.16 | 6.29 | **5.16** |
| `gpt_oss-sink sq512 past512` | 0.57 | 0.60 | 0.56 |
| `llama-3.2-1b sq2048` | 1.45 | 1.57 | 1.42 |
| `llama-3.1-8b sq2048` (D=128, control) | 2.75 | 2.72 | 2.72 |
| `gpt_oss-win sq512 past8192 w128` (windowed) | n/a | 0.13 | 0.15 |

RGP capture, chunk 16, layer 1. The two layer types now dispatch different
instantiations, so they separate by name rather than by duration:

| dispatch | runtime window | this PR |
|---|---:|---:|
| full attention | 8,313 us (n=13, 8,033-8,613) | **5,970 us** (n=7, 5,865-6,105) |
| windowed | 229 us (n=13) | 233 us (n=6) |

## Known trade-off

One shape moves the other way under templating. On `gpt_oss-20b sq2048 past0` at
`M_TILES=2 BKV=32`, the spilled codegen of the runtime window is *faster* --
3.29 ms against 3.74 pre-window and 3.77 here (medians of 5 interleaved
autotuner sweeps). It reproduces, and it does not appear in the other three
no-window shapes, where the runtime window is 20-30% slower at that config.
Restoring the pre-window codegen restores its weakness on that shape along with
its strengths. Net TTFT still improves by 494 ms, but the ~13% there is
unexplained and worth a look on its own.

Related: the regression was partly masked by the autotuner. With the runtime
window the degraded `M_TILES=2 BKV=32` loses to `M_TILES=1 BKV=32` on three of
four no-window shapes, so end-to-end timings understated the per-config damage.

## Test plan

- [x] Standalone prefill harness against a CPU fp32 reference, **26 cases, all
  pass**. Ten are window cases, chosen so a window bug cannot hide:
  - window alone before window plus sink, so the two features fail separately;
  - `past_len` 8192 with `window` 128, where whole KV tiles must be skipped
    rather than merely masked;
  - `window=100` against BKV of 32 and 64, to catch an off-by-one in the
    start-tile clamp;
  - a window wider than the sequence, which must reproduce full attention, and
    `window=1`, which must reduce to each query seeing only itself;
  - a `d=128` case asserting `v3` declines the window rather than ignoring it.
  - Windowed cases land at relL2 3.3e-04 to 3.5e-04, in line with the no-window
    cases.
- [x] Three cases send a sink tensor **and** `smooth_softmax=1` together. That is
  the only combination the runtime actually produces -- `gqa.cpp` derives smooth
  from `(head_sink != nullptr || smooth_softmax == 1)` and then passes both --
  but every earlier case set exactly one of the two, so the real configuration
  was untested. `head_sink` takes precedence in both kernel and reference; a
  kernel that also added the smooth term would double-count the denominator and
  fail these.
- [x] Every `relL2` bit-identical before and after the templating change.
  Removing the `empty` guard from the no-window path is a behavioural change on
  paper, so the cases that caught the original NaN were checked specifically:
  `window=1`, `window > sequence`, and `past_len 8192`.
- [x] Numeric suite against the ORT CPU reference: pass/fail set **identical**
  to the sink branch (22 tests compared one by one).
- [x] Path check with `HIPDNN_EP_DEBUG=1`: 96 of 96 prefill GQA calls fused.
- [x] Resource usage compared instantiation by instantiation against the
  pre-window kernel (table above).
- [x] Interleaved harness A/B, interleaved TTFT, and RGP dispatch-level capture
  (tables above).
- [x] Rebased onto `main` at `fcfdf0cf` and rebuilt; harness re-run, 26/26 pass.
- [ ] CI.

Pre-existing, not caused by this change: the same four
`test_gqa_decode_fixed_cache_llama_shape` failures present at base.

## Notes for reviewers

- Please look hardest at the start-tile clamp and the all-masked-tile guard.
  The clamp is only correct because rows within a Q tile have monotonically
  non-decreasing lower bounds; if that ever stopped holding, the kernel would
  silently drop keys.
- Tuning launches deliberately omit the sink: one FMA per row cannot reorder the
  `(M_TILES, BKV)` candidates, and leaving it out keeps tuned configs comparable
  with the non-sink case.
- Windowed **decode** is deliberately untouched. Its kernel already clamps
  `kv_lo`, so enabling it is nearly a one-word change, but nothing here verifies
  it and prefill is where the window was costing us. The fused decode call site
  passes an explicit `0` rather than forwarding the parameter, because
  `window_ok` routes every windowed decode to the decomposed path and
  forwarding it would only look like the path were supported.
- This makes the planned fp32 -> fp16 score-matrix optimisation moot for
  gpt-oss: there is no longer a score matrix in its prefill. That work would now
  only help models still on the decomposed path.
- Substantial AI assistance (Cursor) was used for the kernel change, the test
  cases, the measurement harnesses and this description. I reviewed the change
  and the evidence, and every number above is a measurement I ran.

### DIFF
--- a/lib/Runtime/Kernels/hip/gqa_kernel.hip
+++ b/lib/Runtime/Kernels/hip/gqa_kernel.hip
@@ -2492,7 +2492,11 @@ typedef float float8_t __attribute__((ext_vector_type(8)));
 // Each wave owns M_TILES 16-row q-subtiles x all BKV columns. Only LDS is
 // P_lds (C->A transpose); only sync is single-wave around it.
 // ---------------------------------------------------------------------
-template <int M_TILES, int BKV, int D>
+// HAS_WINDOW is a template parameter rather than a runtime test because this
+// kernel runs at the VGPR ceiling: carrying the window's per-row state live in
+// the full-attention instantiation pushed M_TILES=2 from 120 to 180 bytes/lane
+// of scratch (38 -> 68 spills) and cost that path 21% at deep KV.
+template <int M_TILES, int BKV, int D, bool HAS_WINDOW>
 __global__ void __launch_bounds__(32)
 gqa_flash_prefill_v5_kernel(
     const _Float16* __restrict__ Q,       // [B, sq, Hq, D]
@@ -2500,7 +2504,10 @@ gqa_flash_prefill_v5_kernel(
     const _Float16* __restrict__ Vcache,  // [B, G, max_seq, D]
     _Float16* __restrict__ O,             // [B, sq, Hq, D]
     int B_count, int Hq, int G, int sq, int skv, int max_seq, int past_len,
-    float scale, int ablate)
+    float scale, int ablate,
+    const _Float16* __restrict__ head_sink,  // [num_heads] or null
+    int num_heads, int smooth_softmax,
+    int local_window_size)                   // <= 0 for full attention
 {
     // ablation bitmask (perf diagnostics only; wrong numerics):
     //   bit0 skip softmax, bit1 skip valueGEMM, bit2 skip scoreGEMM,
@@ -2596,7 +2603,18 @@ gqa_flash_prefill_v5_kernel(
                               ? (skv - 1) : (past_len + q_start + ROWS - 1);
     const int num_tiles = kv_max / BKV + 1;
 
-    for (int t = 0; t < num_tiles; ++t) {
+    // With a window, every row in this Q tile has a lower bound at least as high
+    // as the first row's, so starting from the first row's bound covers the whole
+    // tile and the per-row mask trims the rest. This is what makes the window
+    // cheaper rather than merely masked: the number of tiles walked stops growing
+    // with skv once the window is filled.
+    int t_start = 0;
+    if constexpr (HAS_WINDOW) {
+        const int kv_lo_tile = past_len + q_start - local_window_size + 1;
+        t_start = (kv_lo_tile > 0) ? (kv_lo_tile / BKV) : 0;
+    }
+
+    for (int t = t_start; t < num_tiles; ++t) {
         const int kv0 = t * BKV;
 
         // Stage V tile into LDS with coalesced HALF16 loads along D.
@@ -2629,13 +2647,18 @@ gqa_flash_prefill_v5_kernel(
             for (int e = 0; e < 8; ++e) {
                 const int row    = mi * 16 + e * 2 + pair;        // local q row
                 const int q_pos  = q_start + row;
-                // causal mask + per-row partial max over this lane's columns
+                // causal mask, sliding-window lower bound, and per-row partial
+                // max over this lane's columns
+                const int abs_q = past_len + q_pos;
+                int kv_lo = 0;
+                if constexpr (HAS_WINDOW) kv_lo = abs_q - local_window_size + 1;
                 float rmax = -INFINITY;
                 #pragma unroll
                 for (int sj = 0; sj < S_TILES_J; ++sj) {
                     const int kv_pos = kv0 + sj * 16 + wmma_lane;
-                    if (kv_pos >= skv || kv_pos > past_len + q_pos)
-                        S_reg[mi][sj][e] = -INFINITY;
+                    bool masked = (kv_pos >= skv || kv_pos > abs_q);
+                    if constexpr (HAS_WINDOW) masked = masked || (kv_pos < kv_lo);
+                    if (masked) S_reg[mi][sj][e] = -INFINITY;
                     rmax = fmaxf(rmax, S_reg[mi][sj][e]);
                 }
                 rmax = fmaxf(rmax, __shfl_xor(rmax, 1));
@@ -2645,13 +2668,22 @@ gqa_flash_prefill_v5_kernel(
 
                 const float m_old = m_reg[mi][e];
                 const float m_new = fmaxf(m_old, rmax);
-                const float corr  = exp2f(m_old - m_new);
+                // A window can mask a whole tile for a row while that row has
+                // still seen nothing, leaving m_old and m_new both at -inf.
+                // exp2f(-inf - -inf) is NaN, so short-circuit: contribute
+                // nothing and leave the running state untouched. Causal-only
+                // never reaches this, since tile 0 always holds key 0, so the
+                // guard folds away in the full-attention instantiation.
+                bool empty = false;
+                if constexpr (HAS_WINDOW) empty = (m_new == -INFINITY);
+                const float corr  = empty ? 1.0f : exp2f(m_old - m_new);
                 corr_reg[mi][e]   = corr;
 
                 float rsum = 0.0f;
                 #pragma unroll
                 for (int sj = 0; sj < S_TILES_J; ++sj) {
-                    const float p = exp2f(S_reg[mi][sj][e] - m_new);
+                    const float p =
+                        empty ? 0.0f : exp2f(S_reg[mi][sj][e] - m_new);
                     rsum += p;
                     P_lds[row * P_STR + sj * 16 + wmma_lane] = static_cast<_Float16>(p);
                 }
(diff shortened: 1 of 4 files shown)


BoarQing:
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1635&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1633&curr_job=&curr=

BoarQing:
acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1645/

amd-mingw (approved):

PR #627, opened 2026-08-02
perf(linear_attention): chunk-parallel WMMA gated_delta prefill with windowed scratch

BoarQing:
## Summary

Replaces the serial gated_delta linear-attention prefill with a three-pass
chunk-parallel form, puts the dominant GEMM on the RDNA3 wave-matrix cores, and
bounds the device scratch it needs to a fixed window instead of the sequence.

This PR targets `main` and supersedes #589 (chunk-parallel + WMMA), #596
(session-scoped scratch) and #606 (windowed scratch), all closed in favour of
it. Nothing here is stacked and the diff is complete against `main`. It was
developed as six commits and has been squashed into one: the pieces are not
independently shippable, since the chunk-parallel form is what creates the
scratch problem that the windowing then solves, and the chunk size and the scan
split only pay together.

## Related issue or design

None. Supersedes #589, #596 and #606.

## Why

The prefill ran a recurrence over chunks with `grid = B*Hkv = 32` blocks on a
40-CU part: 25% occupancy and latency-bound, because every chunk's output and
state update read the chunk-start state `S0`, which is only known once the
previous chunk finishes.

That dependency is weaker than it looks. The chunk-state map is affine,
`S_out = L_c*S_in + b_c`, so each chunk's work splits into an `S0`-independent
local part that is fully parallel across all (head, chunk) pairs, and a light
affine scan that is still sequential in the chunk index but cheap.

Making the local part parallel then creates a second problem: it has to
materialise per-(head, chunk) tiles, and sized by `nchunks` that arena grows
without bound with prompt length -- 1,026 MiB at 16k tokens, 2,052 MiB at 32k.

And once both are fixed, what limits the result is LDS rather than arithmetic.
At `LA_CHUNK=32` pass 3 asks for 61,696 B, over half of a CU's 64 KB, so one
block is resident per CU and a stall has nothing to hide behind.

## What

**Three passes.**

- Pass 1 (`grid B*Hkv*nchunks`): `Uloc = (I-trilT)^-1 (beta.*v)`,
  `W = (I-trilT)^-1 (diag(beta.*A) K)`, `rlast`, `a_last`.
- Pass 2 (scan): `U = Uloc - W S0`, `S1 = a_last*S0 + Ktilde^T U`, fp16-rounded
  per chunk.
- Pass 3 (`grid B*Hkv*nchunks`): `O = scale*(A.*(Q S0) + P_lower U)`, with the
  stacked `[Q;W]@S0` GEMM issued as
  `__builtin_amdgcn_wmma_f32_16x16x16_f16_w32`.

`Ktilde` is never stored. It is `rlast .* K`, so the scan rebuilds it from the
key tensor for bit-identical values at 1/dk the scratch.

**Windowed scratch.** The passes run `LA_WINDOW_CHUNKS` at a time, so the arena
is sized per window and stops scaling with the sequence: a flat 49.1 MiB at
`B=1, Hkv=32, dk=dv=128` for any prompt past one window. Ordering needs no
extra synchronisation, because the passes are issued back-to-back on one
stream, so pass 3 of a window has finished reading `Uloc`/`W`/`S0` before pass 1
of the next overwrites them. Across windows the scan hands off through a small
fp16 carry buffer rather than re-reading `state`: the running state is fp16 in
LDS, so a round trip through a narrower `T` (bf16 keeps 8 mantissa bits against
fp16's 10) would perturb it at every boundary. A single-window prefill never
allocates or touches the carry, so it is bit-identical to the unwindowed form.

**Session-scoped arena.** It lives in `RuntimeState::la_scratch`
(grow-on-demand, freed in `hipdnn_ep_state_cleanup`, the same policy as
`conv_scratch` and `qmoe_scratch`) rather than a process-static buffer, so the
footprint is bounded to the session and released on teardown. The kernel takes
a caller-owned pointer and size; the wrapper sizes it through the new
`hip_linear_attention_prefill_scratch_bytes()`. A null or under-sized arena
makes the launcher decline, and the runtime falls back to the per-token loop.

**Chunk size and scan width.** `LA_CHUNK` 32 -> 16 halves every chunk-sized LDS
tile and drops pass 3 to 29,824 B, which fits two blocks per CU;
`LA_WINDOW_CHUNKS` follows 64 -> 32 to hold the window at 512 tokens. The extra
launches are not a real cost: at `LA_CHUNK=16` a 32-chunk window beats a
128-chunk one by 9.6% despite issuing 3.7x as many.

That leaves the scan as the tail, running one block per `(batch, kv head)` --
32 blocks against 40 CUs. Both of its inner products touch only column `j` of
`S0`, so the recurrence factors completely over `dv` and splits across blocks.
At `LA_SCAN_NSPLIT=4` that is 128 blocks, and the narrower 256-thread block a
`dv/4` slice wants is what keeps them resident. Each split re-reads the
`dv`-independent `W` tile; that duplication is what buys the parallelism, and it
only pays once `LA_CHUNK=16` has shrunk the tile -- applied on top of
`LA_CHUNK=32` the split is a 12.8% regression. `LA_SCAN_NSPLIT=1` restores the
unsplit scan exactly.

Only gated_delta with scalar log-decay is implemented, and WMMA needs `dk` and
`dv` to be multiples of 16; the launcher declines anything else so the runtime
falls back to the per-token loop.

## Performance

All on gfx1151 (Radeon 8060S, 40 CUs). **The two stages below were measured in
different sessions against different baselines and must not be chained**; each
is against the code immediately before it.

Chunk-parallel + WMMA against the serial kernel, Qwen3.5-35B, seq 18,778:

| Metric | serial | chunk-parallel + WMMA |
|---|---:|---:|
| prefill kernel | 604.8 ms | **110.0 ms** (5.5x) |
| 16k TTFT | 33,476 ms | **17,580 ms** (1.90x) |
| occupancy | 25% | 25-100% |
| wavefronts | 256 | 150,272 |

No regression at shorter sequences: 1.51x end-to-end at 128 tokens, 1.77x at
2048.

Chunk 16 plus the `dv`-split against the windowed form, Qwen3-Next-80B-A3B fp16,
30 layers, 18,778-token prompt, paired interleaved runs:

| Metric | windowed | this PR |
|---|---:|---:|
| isolated op | baseline | **-25.2%** (10/10 paired wins) |
| TTFT | baseline | **-1,069 ms, -5.4%** (5/6 paired rounds) |
| pass 1 / scan / pass 3 | baseline | -54.3% / -5.8% / -42.2% |
| scratch arena | 129.3 MiB | **49.1 MiB** |
| pass 3 LDS per block | 61,696 B | 29,824 B |
| scan blocks per launch | 32 | 128 |

Interleaving is not optional on this box: the same binary measured anywhere from
75 to 165 ms per call across a session, so only within-round paired deltas are
reported, with arm order rotated each round.

Scratch, from the exported `hip_linear_attention_prefill_scratch_bytes` at
`B=1, Hkv=32, dk=dv=128` -- deterministic, not timed:

| seq | sized by `nchunks` | this PR |
|---:|---:|---:|
| 4,096 | 256.5 MiB | **49.1 MiB** |
| 16,384 | 1,026.1 MiB | **49.1 MiB** |
| 32,768 | 2,052.1 MiB | **49.1 MiB** |

## Test plan

- [x] fp64 per-token gated_delta reference at 18,778 tokens, which crosses 36
  window boundaries: **cosine 0.999999960, trailing-half relative L2 2.749e-4,
  no NaN or Inf**.
- [x] The gate keys on the trailing-half relative L2 rather than cosine. A
  whole-tensor cosine is not a usable gate here: a deliberately corrupted build
  whose inter-window carry is scaled by 0.99 still passes `cosine > 0.9999`,
  because one perturbed boundary barely moves it. The trailing-half metric moves
  2.5x on that same corruption. The earlier reference only ran at `T <= 512`,
  which never crosses a boundary at all and so could not have caught any of the
  windowing work.
- [x] Single-window prefills are bit-identical to the unwindowed form
  (`is_first == is_last`, carry never read, written or allocated).
- [x] The chunked form is *more* accurate than the kernel it replaces, not just
  equivalent: it keeps `S` in fp32 across a whole chunk and rounds only at the
  chunk-end write, where the per-token kernel quantised `S` to fp16 every
  timestep.
- [x] Isolated-op A/B over 10 interleaved rounds with arm order rotated, and
  in-model TTFT over 6 paired rounds (tables above).
- [x] Rebased onto `main` at `fcfdf0cf`, rebuilt, gate re-run: cosine
  0.999999960, arena 49.07 MiB, unchanged.
- [ ] CI.

## Notes for reviewers

- Please look hardest at the cross-window carry. It is the only state hand-off
  between launches, it is fp16, and a single-window prefill never exercises it,
  so anything short of a multi-window sequence tests nothing about it.
- **A third optimisation was measured and rejected.** Storing `S0` transposed
  makes pass 3's WMMA B fragment one 32-byte vector load instead of 16 two-byte
  loads. The ISA confirms the mechanism exactly as predicted --
  `global_load_u16` per WMMA falls from 20.7 to 4.7 and the inner loop from 78
  instructions to 13 -- and it is still not worth taking, because the
  transposing store has to read LDS down a column and the scan loses more than
  pass 3 gains. The conclusion is that the B-fragment load was never the
  limiter. It is not in this PR and no dead code for it is left behind.
- **Coverage limit.** `LA_SCAN_NSPLIT=4` was verified at `B=1, Hkv=32`, where
  the scan was starved of blocks. On a shape where `B*Hkv` already exceeds the
  CU count the split has nothing to win and the duplicated `W` read may make it
  a small loss; worth a second look before relying on it for large-batch
  prefill.
- The decode path is untouched. `HIPDNN_EP_LA_PARALLEL` still exists for the
  separate decode-side G-split and is unrelated to anything here.
- Substantial AI assistance (Cursor) was used for the kernel work, the
  correctness harness, the measurement scripts and this description. I reviewed
  the change and the evidence, and every number above is a measurement I ran.

### DIFF
--- a/lib/Runtime/Kernels/hip/linear_attention_kernel.hip
+++ b/lib/Runtime/Kernels/hip/linear_attention_kernel.hip
@@ -16,6 +16,17 @@
 
 static constexpr int LA_THREADS = 256;
 
+// --- WMMA (RDNA3 wave32) support, mirrors gemm_wmma_kernel.hip ---
+typedef _Float16 la_half16 __attribute__((ext_vector_type(16)));
+typedef float    la_float8 __attribute__((ext_vector_type(8)));
+#if !defined(__has_builtin) ||                                                  \
+    !__has_builtin(__builtin_amdgcn_wmma_f32_16x16x16_f16_w32)
+static __device__ __forceinline__ la_float8
+la_wmma_unavailable(la_half16, la_half16, la_float8 c) { __builtin_trap(); return c; }
+#define __builtin_amdgcn_wmma_f32_16x16x16_f16_w32(a, b, c) la_wmma_unavailable((a),(b),(c))
+#endif
+static constexpr int LA_WT = 16;   // WMMA 16x16x16 tile dim
+
 static constexpr int LA_RULE_LINEAR      = 0;
 static constexpr int LA_RULE_GATED       = 1;
 static constexpr int LA_RULE_DELTA       = 2;
@@ -664,15 +675,15 @@ extern "C" int hip_linear_attention_decode(
 }
 
 //===----------------------------------------------------------------------===//
-// Chunked-parallel gated-delta PREFILL kernel
+// Gated-delta linear-attention PREFILL
 //===----------------------------------------------------------------------===//
 //
 // Replaces the per-token launch loop (one decode-kernel launch per timestep,
-// each streaming the whole recurrent state from global) with a SINGLE launch
-// that processes the entire sequence.  One block per (batch, kv_head); each
-// block loops over chunks of LA_CHUNK tokens, keeping the recurrent state S in
-// GLOBAL (read once / written once per chunk) and the per-chunk tiles + the
-// [C,C] coupling matrices in LDS.
+// each streaming the whole recurrent state from global) with a chunked
+// formulation that processes the whole sequence in a few launches, keeping the
+// per-chunk tiles and [C,C] coupling matrices in LDS. The per-chunk math is
+// stated here; the chunk-parallel decomposition that runs it (and breaks the
+// cross-chunk serial dependency) is documented below.
 //
 // Math (scalar gate a_t = exp(g_t), per kv-head; tokens 0..C-1 in a chunk;
 // chunk-start state S0):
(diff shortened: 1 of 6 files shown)


BoarQing:
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1635&base_job=daily_hip_ep_oga&base=1618&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1643&curr_job=daily_hip_ep_oga&curr=1643

BoarQing:
Accuracy: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1644

BoarQing:
Superseded by #631, which contains this change unchanged as its first commit and adds a second one putting all three of the prefill's matmuls on the matrix cores, for a further -31.6% on the op.

Closing here rather than stacking, so there is one PR to review against `main`.

Two things in this description are corrected there and are worth flagging rather than leaving buried:

- The note above says the transposed `S0` was measured and rejected. #631 adopts it. The reasoning here was right about *why* it lost -- the transposing store cost more than pass 3's vector load saved -- and what changed is that the scan now holds its state transposed natively, so that store becomes a straight copy. Measured alone on scalar code the layout is still a 26% regression.
- The performance tables there add a third stage and repeat the warning, which now matters more: the three stages were measured in different sessions, against different baselines, on two different models. They cannot be chained.

PR #631, opened 2026-08-03
perf(linear_attention): chunk-parallel gated_delta prefill on the matrix cores, windowed scratch

BoarQing:
## Summary

Replaces the serial gated_delta linear-attention prefill with a three-pass
chunk-parallel form, puts all three of its matmuls on the RDNA3 wave-matrix
cores, and bounds the device scratch it needs to a fixed window instead of the
sequence.

This PR targets `main` and supersedes #589 (chunk-parallel + WMMA), #596
(session-scoped scratch), #606 (windowed scratch) and #627 (all of the above,
squashed), all closed in favour of it. The diff is complete against `main` and
nothing is stacked.

It is **two commits**, and they are worth keeping apart because they were
measured against different baselines and neither number is valid against the
other's:

1. `perf(linear_attention): chunk-parallel WMMA gated_delta prefill with
   windowed scratch` -- the content of #627 unchanged. Itself developed as six
   commits and squashed, because the pieces are not independently shippable:
   the chunk-parallel form is what creates the scratch problem that the
   windowing then solves, and the chunk size and the scan split only pay
   together.
2. `perf(linear_attention): run the prefill's three matmuls on the matrix
   cores` -- new since #627. A further **-31.6%** on the op.

## Related issue or design

None. Supersedes #589, #596, #606 and #627.

## Why

The prefill ran a recurrence over chunks with `grid = B*Hkv = 32` blocks on a
40-CU part: 25% occupancy and latency-bound, because every chunk's output and
state update read the chunk-start state `S0`, which is only known once the
previous chunk finishes.

That dependency is weaker than it looks. The chunk-state map is affine,
`S_out = L_c*S_in + b_c`, so each chunk's work splits into an `S0`-independent
local part that is fully parallel across all (head, chunk) pairs, and a light
affine scan that is still sequential in the chunk index but cheap.

Making the local part parallel then creates a second problem: it has to
materialise per-(head, chunk) tiles, and sized by `nchunks` that arena grows
without bound with prompt length -- 1,026 MiB at 16k tokens, 2,052 MiB at 32k.

And once both are fixed, what limits the result is LDS rather than arithmetic.
At `LA_CHUNK=32` pass 3 asks for 61,696 B, over half of a CU's 64 KB, so one
block is resident per CU and a stall has nothing to hide behind.

**Then, with all of that done, arithmetic became the limit after all.** The op
split roughly 17 / 43 / 39 across the three passes, and each pass contained a
product of exactly the shape the matrix cores want, running as a scalar fp32
loop -- the scan at about 1.1 TFLOP/s on a part that does far better than that
in `v_wmma_f32_16x16x16_f16_w32`.

The blocker there was layout, not arithmetic. The fragments are
`a_frag[ele] = A[lane][k+ele]` and `b_frag[ele] = B[k+ele][lane]`, so A wants
row-major and B column-major. With the scan's running state held as `dk` rows
of `dv`, its first product had B strided and its second had A strided, and
scalar code was the only thing that could read it. Holding the state
transposed, as `dv` rows of `dk`, makes **every fragment in both of the scan's
products contiguous at once**, and gives pass 3's B fragment the same for free.

## What

**Three passes.**

- Pass 1 (`grid B*Hkv*nchunks`): `Uloc = (I-trilT)^-1 (beta.*v)`,
  `W = (I-trilT)^-1 (diag(beta.*A) K)`, `rlast`, `a_last`.
- Pass 2 (scan): `U = Uloc - W S0`, `S1 = a_last*S0 + Ktilde^T U`, fp16-rounded
  per chunk.
- Pass 3 (`grid B*Hkv*nchunks`): `O = scale*(A.*(Q S0) + P_lower U)`, with the
  stacked `[Q;W]@S0` GEMM issued as
  `__builtin_amdgcn_wmma_f32_16x16x16_f16_w32`.

`Ktilde` is never stored. It is `rlast .* K`, so the scan rebuilds it from the
key tensor for bit-identical values at 1/dk the scratch.

**Windowed scratch.** The passes run `LA_WINDOW_CHUNKS` at a time, so the arena
is sized per window and stops scaling with the sequence: a flat 49.1 MiB at
`B=1, Hkv=32, dk=dv=128` for any prompt past one window. Ordering needs no
extra synchronisation, because the passes are issued back-to-back on one
stream, so pass 3 of a window has finished reading `Uloc`/`W`/`S0` before pass 1
of the next overwrites them. Across windows the scan hands off through a small
fp16 carry buffer rather than re-reading `state`: the running state is fp16 in
LDS, so a round trip through a narrower `T` (bf16 keeps 8 mantissa bits against
fp16's 10) would perturb it at every boundary. A single-window prefill never
allocates or touches the carry, so it is bit-identical to the unwindowed form.

**Session-scoped arena.** It lives in `RuntimeState::la_scratch`
(grow-on-demand, freed in `hipdnn_ep_state_cleanup`, the same policy as
`conv_scratch` and `qmoe_scratch`) rather than a process-static buffer, so the
footprint is bounded to the session and released on teardown. The kernel takes
a caller-owned pointer and size; the wrapper sizes it through the new
`hip_linear_attention_prefill_scratch_bytes()`. A null or under-sized arena
makes the launcher decline, and the runtime falls back to the per-token loop.

**Chunk size and scan width.** `LA_CHUNK` 32 -> 16 halves every chunk-sized LDS
tile and drops pass 3 to 29,824 B, which fits two blocks per CU;
`LA_WINDOW_CHUNKS` follows 64 -> 32 to hold the window at 512 tokens. The extra
launches are not a real cost: at `LA_CHUNK=16` a 32-chunk window beats a
128-chunk one by 9.6% despite issuing 3.7x as many.

That leaves the scan as the tail, running one block per `(batch, kv head)` --
32 blocks against 40 CUs. Both of its inner products touch only column `j` of
`S0`, so the recurrence factors completely over `dv` and splits across blocks.
At `LA_SCAN_NSPLIT=4` that is 128 blocks, and the narrower 256-thread block a
`dv/4` slice wants is what keeps them resident. Each split re-reads the
`dv`-independent `W` tile; that duplication is what buys the parallelism, and it
only pays once `LA_CHUNK=16` has shrunk the tile -- applied on top of
`LA_CHUNK=32` the split is a 12.8% regression. `LA_SCAN_NSPLIT=1` restores the
unsplit scan exactly.

### Second commit: all three matmuls on the matrix cores

**Scan, -53.5%.** `Ssm` becomes `[dv][dk+8]`, `Ktilde` is staged `[dk][C]` and
`U` as `[dv][C]`, all fp16. `U = Uloc - W S0` and `S1 = a_last*S0 + Ktilde^T U`
become 16 WMMA each, replacing the two scalar loops. The per-chunk `S0` handoff
becomes a straight contiguous copy rather than a transposing store. Scan LDS
drops to 15,872 B, below the scalar version despite the padding, because every
tile it stages is now fp16, and the scan stops being occupancy-limited: 75% ->
100%.

**Pass 1, -31.0%.** `T = K K^T` on WMMA. No layout change needed. One 16x16
tile, one wave, replacing a loop every wave ran in full regardless of how few
lanes sat below the diagonal.

**Pass 3, -8.8%.** `P = Q K^T` on WMMA, same shape, also no layout change. Its
`QWS` B fragment picks up the transposed state for free. And the postprocess,
which ran `for (j = tid; j < dv; j += nthreads)` with `dv`=128 against a
256-thread block, now folds `U` into the `W S0` rows in place so the idle half
of the block takes half the output rows -- no extra LDS, no extra global
traffic.

Only gated_delta with scalar log-decay is implemented, and WMMA needs `dk` and
`dv` to be multiples of 16; the launcher declines anything else so the runtime
falls back to the per-token loop. The decode path is untouched.

## Performance

All on gfx1151 (Radeon 8060S, 40 CUs). **The three stages below were measured in
different sessions against different baselines and on two different models, and
must not be chained.** Each is against the code immediately before it.

Stage 1, chunk-parallel + WMMA against the serial kernel, Qwen3.5-35B, seq
18,778:

| Metric | serial | chunk-parallel + WMMA |
|---|---:|---:|
| prefill kernel | 604.8 ms | **110.0 ms** (5.5x) |
| 16k TTFT | 33,476 ms | **17,580 ms** (1.90x) |
| occupancy | 25% | 25-100% |
| wavefronts | 256 | 150,272 |

No regression at shorter sequences: 1.51x end-to-end at 128 tokens, 1.77x at
2048.

Stage 2, chunk 16 plus the `dv`-split against the windowed form,
Qwen3-Next-80B-A3B fp16, 30 layers, 18,778-token prompt, paired interleaved
runs:

| Metric | windowed | after |
|---|---:|---:|
| isolated op | baseline | **-25.2%** (10/10 paired wins) |
| TTFT | baseline | **-1,069 ms, -5.4%** (5/6 paired rounds) |
| pass 1 / scan / pass 3 | baseline | -54.3% / -5.8% / -42.2% |
| scratch arena | 129.3 MiB | **49.1 MiB** |
| pass 3 LDS per block | 61,696 B | 29,824 B |
| scan blocks per launch | 32 | 128 |

Stage 3, the matrix cores against everything above, i.e. commit 2 against commit
1. Qwen3.6-35B-A3B, 18,778-token prompt. Medians of 6 paired rotated rounds,
`la_check.exe` under `LA_TIME`, median of the 30 timed calls per round; base
read 74.0-74.3 ms in every round of every sweep in that session, so these
separations sit far outside the spread:

| Metric | commit 1 | commit 2 |
|---|---:|---:|
| isolated op | 74.20 ms | **50.76 ms, -31.6%** (6/6 paired wins) |
| pass 1 / scan / pass 3 | 14.60 / 30.63 / 29.02 ms | **10.02 / 14.22 / 26.46 ms** |
| in-model op, per prefill | 2,228 ms | **1,522 ms, -32%** |
| in-model TTFT | 19,060 / 18,726 / 18,226 ms | **18,215 / 18,128 / 17,335 ms** (3/3, median -4.4%) |
| scan LDS per block | 18,432 B | **15,872 B** |
| scan occupancy | 75% | **100%** |

TTFT moves far less than the op because LA is only about an eighth of prefill by
this point: the 706 ms the op saves per prefill takes its share of TTFT from
11.7% to 8.4%, and that fully accounts for the TTFT delta.

Leave-one-out from commit 2, the only test that shows a piece still pays once
the others are in: without the WMMA `T` -25.6%, without the WMMA `P` -27.8%,
without the widened postprocess -30.7%, without the scan work -12.2%. Every
piece still contributes.

Interleaving is not optional on this box: the same binary measured anywhere from
75 to 165 ms per call across a session, so only within-round paired deltas are
reported, with arm order rotated each round.

Scratch, from the exported `hip_linear_attention_prefill_scratch_bytes` at
`B=1, Hkv=32, dk=dv=128` -- deterministic, not timed:

| seq | sized by `nchunks` | this PR |
|---:|---:|---:|
| 4,096 | 256.5 MiB | **49.1 MiB** |
| 16,384 | 1,026.1 MiB | **49.1 MiB** |
| 32,768 | 2,052.1 MiB | **49.1 MiB** |

## Test plan

- [x] fp64 per-token gated_delta reference at 18,778 tokens, which crosses 36
  window boundaries: **cosine 0.999999960, trailing-half relative L2 2.749e-4,
  no NaN or Inf** for commit 1.
- [x] The gate keys on the trailing-half relative L2 rather than cosine. A
  whole-tensor cosine is not a usable gate here: a deliberately corrupted build
  whose inter-window carry is scaled by 0.99 still passes `cosine > 0.9999`,
  because one perturbed boundary barely moves it. The trailing-half metric moves
  2.5x on that same corruption. The earlier reference only ran at `T <= 512`,
  which never crosses a boundary at all and so could not have caught any of the
  windowing work.
- [x] **Commit 2 costs precision, and it is the one thing here that does.** At
  T=4096 the gate reads `rel_l2_tail` **3.154e-04** against commit 1's 2.749e-04
  and 6.468e-04 for the deliberately corrupted control, cosine 0.999999946, no
  NaN or Inf. Better than 2x margin on the metric that can actually see a bad
  carry.
- [x] **That cost is localised and attributed.** It is entirely the fp16 `U` the
  matrix cores require in the scan, which drives the state update. Dumping
  outputs at T=18,778 and diffing separates it: the pass 1 and pass 3 changes
  together move the output by `1.9e-05` rel L2, which is reassociation noise,
  while commit 2 as a whole moves it by `7.2e-04`. On the synthetic input, whose
  magnitudes span `3.7e-07` to `2.9e+15`, that is a mean element-wise relative
  error of `7.5e-04` and a p99 of `8.0e-03` over the upper half of the range. It
  buys 20 of the 31 points.
- [x] Single-window prefills are bit-identical to the unwindowed form
  (`is_first == is_last`, carry never read, written or allocated).
- [x] The chunked form is *more* accurate than the kernel it replaces, not just
  equivalent: it keeps `S` in fp32 across a whole chunk and rounds only at the
  chunk-end write, where the per-token kernel quantised `S` to fp16 every
  timestep.
- [x] ISA checked before any of commit 2's timings were believed, since this
  series already produced one arm whose ISA was perfect and whose speedup was
  zero. Per kernel, summed over the three dtype instantiations, commit 1 ->
  commit 2: WMMA 0 -> 3 in pass 1, 0 -> 6 in the scan, 3 -> 6 in pass 3; pass
  3's 16-bit global loads 62 -> 14 with 6 `global_load_b128` appearing, which is
  the transposed state arriving as vectors.
- [x] Isolated-op A/B over 10 (commit 1) and 6 (commit 2) interleaved rounds with
  arm order rotated, and in-model TTFT over 6 and 3 paired rounds respectively.
- [x] RGP capture on the same fence, same shape, same single LA-bearing decoder
  layer, before and after commit 2: 37 dispatches per pass unchanged,
  `avg_wavefronts` unchanged, pass 1 mean 377.4 -> 280.0 us, scan 952.0 -> 389.0
  us, scan no longer LDS-limited. Confirms the mechanism moved rather than the
  machine.
- [x] Rebased onto `main` at `38e7abbb`, rebuilt, gate re-run: cosine
  0.999999946, `rel_l2_tail` 3.154e-04, bit-identical to the pre-rebase build.
  Neither of main's two intervening commits touches any file in this PR.
- [ ] CI.
- [ ] Only gfx1151 was built and measured.

## Notes for reviewers

- Please look hardest at the cross-window carry. It is the only state hand-off
  between launches, it is fp16, and a single-window prefill never exercises it,
  so anything short of a multi-window sequence tests nothing about it.
- **#627 said the transposed `S0` was measured and rejected. Commit 2 adopts it,
  and the reversal is the most interesting thing in this PR.** The earlier
  attempt stored `S0` transposed purely to give pass 3's WMMA B fragment one
  32-byte load instead of 16 two-byte loads. The ISA confirmed the mechanism
  exactly as predicted and it was still a loss, because the transposing store
  had to read LDS down a column and the scan lost more than pass 3 gained. The
  conclusion drawn -- that the B-fragment load was never the limiter -- was
  correct. What changed is that the scan now holds its state transposed
  *natively*, so that store is a straight copy and its cost disappears, and the
  layout pays for itself inside the scan rather than needing pass 3 to justify
  it. **Measured alone, on scalar code, the layout change is still a 26%
  regression**; it only wins in combination with the WMMA that consumes it.
- **The row strides are measured, not derived, and this is where the earlier
  attempt died.** 16-byte fragment alignment needs the state row stride to be a
  multiple of 8 halves, which forces `gcd(stride/2, 32) >= 4`, so conflict-free
  banking is simply unreachable. `dk+8` hits that 4-way floor. `dk+0` is 32-way
  and costs the scan 59% (14.37 -> 22.88 ms) even with the matrix cores doing
  the arithmetic, and takes the scan to 6.0x base in the layout-only variant
  that still reads the state with scalar code. `dk+16` is 8-way and costs 5%.
  Please do not "clean up" these paddings.
- **One further change was measured and rejected, and is left as a comment in
  the source so it is not retried.** Pass 1's two triangular solves share the
  same triangular factor, so merging them into one of width `dv+dk` uses all 256
  threads and halves the 30 barriers per chunk. It needs both right-hand sides
  resident at once, 13,568 -> 21,760 B and 4 blocks/CU -> 3, and it is 4.7%
  slower. That is the third time on this kernel that occupancy bought with LDS
  has failed to pay, after the scan-LDS probe (-1.4%) and the layout change
  measured alone (+26%).
- **Coverage limit.** `LA_SCAN_NSPLIT=4` was verified at `B=1, Hkv=32`, where
  the scan was starved of blocks. On a shape where `B*Hkv` already exceeds the
  CU count the split has nothing to win and the duplicated `W` read may make it
  a small loss; worth a second look before relying on it for large-batch
  prefill.
- **Where the next win is, if anyone wants it.** Pass 3 is now 56% of the op, up
  from 39%, and the only pass still occupancy-limited, at 50% on 30,208 B of
  LDS, of which its fp32 `QWS` accumulator alone is 16,384 B. Given the three
  results above I would treat that as a cheap hypothesis to test rather than a
  plan.
- `HIPDNN_EP_LA_PARALLEL` still exists for the separate decode-side G-split and
  is unrelated to anything here.
- Substantial AI assistance (Cursor) was used for the kernel work, the
  correctness harness, the measurement scripts and this description. I reviewed
  the change and the evidence, and every number above is a measurement I ran.

## Checklist

### DIFF
--- a/lib/Runtime/Kernels/hip/linear_attention_kernel.hip
+++ b/lib/Runtime/Kernels/hip/linear_attention_kernel.hip
@@ -16,6 +16,17 @@
 
 static constexpr int LA_THREADS = 256;
 
+// --- WMMA (RDNA3 wave32) support, mirrors gemm_wmma_kernel.hip ---
+typedef _Float16 la_half16 __attribute__((ext_vector_type(16)));
+typedef float    la_float8 __attribute__((ext_vector_type(8)));
+#if !defined(__has_builtin) ||                                                  \
+    !__has_builtin(__builtin_amdgcn_wmma_f32_16x16x16_f16_w32)
+static __device__ __forceinline__ la_float8
+la_wmma_unavailable(la_half16, la_half16, la_float8 c) { __builtin_trap(); return c; }
+#define __builtin_amdgcn_wmma_f32_16x16x16_f16_w32(a, b, c) la_wmma_unavailable((a),(b),(c))
+#endif
+static constexpr int LA_WT = 16;   // WMMA 16x16x16 tile dim
+
 static constexpr int LA_RULE_LINEAR      = 0;
 static constexpr int LA_RULE_GATED       = 1;
 static constexpr int LA_RULE_DELTA       = 2;
@@ -664,15 +675,15 @@ extern "C" int hip_linear_attention_decode(
 }
 
 //===----------------------------------------------------------------------===//
-// Chunked-parallel gated-delta PREFILL kernel
+// Gated-delta linear-attention PREFILL
 //===----------------------------------------------------------------------===//
 //
 // Replaces the per-token launch loop (one decode-kernel launch per timestep,
-// each streaming the whole recurrent state from global) with a SINGLE launch
-// that processes the entire sequence.  One block per (batch, kv_head); each
-// block loops over chunks of LA_CHUNK tokens, keeping the recurrent state S in
-// GLOBAL (read once / written once per chunk) and the per-chunk tiles + the
-// [C,C] coupling matrices in LDS.
+// each streaming the whole recurrent state from global) with a chunked
+// formulation that processes the whole sequence in a few launches, keeping the
+// per-chunk tiles and [C,C] coupling matrices in LDS. The per-chunk math is
+// stated here; the chunk-parallel decomposition that runs it (and breaks the
+// cross-chunk serial dependency) is documented below.
 //
 // Math (scalar gate a_t = exp(g_t), per kv-head; tokens 0..C-1 in a chunk;
 // chunk-start state S0):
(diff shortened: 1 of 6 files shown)


BoarQing:
Perf: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1635&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1671%2F&curr_job=&curr=

BoarQing:
Acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1673/

amd-mingw (approved):

PR #633, opened 2026-08-03
fix(version): correct EP version info and rename XLNX_ONNX_EP_VERBOSE to HIP_EP_VERBOSE

zz002:
## Summary

`EXEC VERSION` logged `onnx-hipdnn-ep (N/A) :N/A` when built from a git worktree, and the remaining entries were either duplicated, empty, or reported the wrong project. This fixes the collection, and renames the verbose env var and its log prefix from `XLNX_ONNX_EP_VERBOSE` to `HIP_EP_VERBOSE`.

Before:

```
[XLNX_ONNX_EP_VERBOSE] EXEC VERSION: vai-rt (v1.0) :
[XLNX_ONNX_EP_VERBOSE] EXEC VERSION: onnxruntime (e48bfc) :e48bfc6b0f62e10238c4bd62e3cfc0b6c5f05518
[XLNX_ONNX_EP_VERBOSE] EXEC VERSION: onnx-hipdnn-ep (v0.3.0-9-ge48b) :e48bfc6b0f62e10238c4bd62e3cfc0b6c5f05518
[XLNX_ONNX_EP_VERBOSE] EXEC VERSION: morphizen (e48bfc) :e48bfc6b0f62e10238c4bd62e3cfc0b6c5f05518
```

After:

```
[HIP_EP_VERBOSE] EXEC VERSION: hip-ep (v0.3.0-12-ga481e) :a481ee3d83159f29e39aedf1c37551ec68500298
[HIP_EP_VERBOSE] EXEC VERSION: onnxruntime (1.27.0)
```

## Related issue or design

None

## Why

`morphizen_add_version_info` only ran git when `.git` was a directory. In a linked worktree `.git` is a file, so the function returned nothing and the entry fell back to `N/A`. The component directory came from the project-name variable `onnx-hipdnn-ep_SOURCE_DIR` with a hardcoded `../onnx-hipdnn-ep/` path as backup, both of which break as soon as the project or the checkout directory is renamed.

The other three entries were misleading rather than missing. `onnxruntime` ran `git rev-parse` against the vendored ORT header directory, which lives in this repository, so it reported hip-ep's own commit rather than an ORT version. `vai-rt` came from an environment variable that CMake passes empty. `morphizen` is an in-tree subtree and therefore always shares hip-ep's commit.

## What

- Query git unconditionally and fall back only when it produces no output; resolve the source directory from `CMAKE_SOURCE_DIR` and report the component as `hip-ep`.
- Take the ORT version from `onnxruntime_VERSION`; the version block moves below the ORT resolution in `deps.cmake` because that is where the variable is set.
- Drop the `vai-rt` and `morphizen` entries, and print an entry without a commit as `name (version)`.
- Rename `XLNX_ONNX_EP_VERBOSE` to `HIP_EP_VERBOSE` across the three `DEF_ENV_PARAM` definitions, the matching hardcoded log prefix, and every caller, build config and doc reference. The old name is no longer read.

## Test plan

- [x] Worktree regression reproduced and fixed: against a linked worktree the old `.git`-is-directory test yields `N/A`, the new code yields `hip-ep;a5b61f754...;v0.2.0-90-ga5b61`
- [x] Local build + `hip-onnx-runner` on a small model prints the two entries shown above, with the `[HIP_EP_VERBOSE]` prefix
- [x] Setting only the old `XLNX_ONNX_EP_VERBOSE` produces no verbose output, confirming the rename is a hard switch
- [x] `pre-commit run --all-files` clean
- [ ] CI build-windows and linux-build green

No automated test covers version-string collection; the checks above were run manually.

## Notes for reviewers

The first two commits are missing the AI attribution trailers; the third commit carries them.

## Checklist


amd-mingw (approved):
LGTM

PR #635, opened 2026-08-03
feat(hipsr): add parallel placeholder shape yield

zpye:
## Summary
- Add temporary `hipsr.shape_yield2`, yielding one `!shape.shape` per placeholder result.
- Use it for placeholder shape regions while keeping `hipsr.shape_yield` unchanged for DPS shape regions.
- Add focused LIT coverage for the new op.


wcy123 (approved):
LGTM.

PR #638, opened 2026-08-03
feat(hipsr): add Cast placeholder shape population

zpye:
## Summary
- Add a placeholder-owned shape-region population pass with an initial Cast recipe.
- Define the minimal Normal and Barrier block layouts used by population recipes.
- Keep precise Cast population coverage in the Cast LIT test.

## AI assistance
Cursor assisted with implementation and validation.


include/hip/Dialect/Hipsr/Transforms/Passes.td
```diff
@@ -41,6 +41,23 @@ def PopulateShapeRegionPass
   ];
 }
 
+def PopulateShapeRegion2Pass
```
wcy123: we don't need  `PopulateShapeRegion2Pass`, because `PopulateShapeRegionPass` was not in use any more. all tests are disabled.

lib/Dialect/Hipsr/IR/HipsrPlaceholderOp.cpp
```diff
@@ -37,6 +38,19 @@ Operation *PlaceholderOp::getDpsConsumer() {
   return nullptr;
 }
 
+SmallVector<Type> PlaceholderOp::getShapeRegionArgumentTypes() {
```
wcy123, line 41: I am not sure if we need this function, because element type is compile time constant.

lib/Dialect/Hipsr/Transforms/PopulateShapeRegionPass.cpp
```diff
@@ -4,62 +4,91 @@
...
+  Operation *consumer = placeholder.getDpsConsumer();
+  if (!consumer) {
+    return placeholder.emitOpError(
+        "shape-region population requires a DPS consumer");
+  }
+
+  Block &block = createPlaceholderShapeBlock(builder, placeholder);
+
+  if (auto castOp = dyn_cast<CastOp>(consumer)) {
+    if (placeholder.getPlaceholderType() != PlaceholderType::Normal) {
```
wcy123: we don't need this sanity check.

lib/Dialect/Hipsr/Transforms/PopulateShapeRegionPass.cpp
```diff
@@ -4,62 +4,91 @@
...
+  return block;
+}
+
+LogicalResult populatePlaceholderShapeRegion(OpBuilder &builder,
+                                             PlaceholderOp placeholder) {
+  if (!placeholder.getShapeRegion().empty()) {
+    return success();
+  }
+
+  Operation *consumer = placeholder.getDpsConsumer();
```
wcy123, line 58: we should move it to verifier?

wcy123 (approved):
LGTM

lib/Dialect/Hipsr/IR/HipsrPlaceholderOp.cpp
zpye: keep it

lib/Dialect/Hipsr/Transforms/PopulateShapeRegionPass.cpp
zpye: removed

lib/Dialect/Hipsr/Transforms/PopulateShapeRegionPass.cpp
zpye: fixed

PR #649, opened 2026-08-03
fix(promote-strided): select consumers by DPS interface

fhanuman:
## Issue

`hip-promote-strided-operands` only handled operations owned by `HipDialect`. Plugin operations can also use destination-passing style and the same bare-pointer runtime ABI, but the pass skipped them. A plugin operation receiving a subview could therefore read the parent buffer instead of the intended slice because the ABI does not carry the memref offset or strides.

## Fix

Select consumers through `DestinationStyleOpInterface` instead of their dialect. For every non-identity-layout DPS input, the pass creates an identity-layout temporary, copies the logical view into it, and passes that temporary to the consumer. DPS init operands remain unchanged.

```mlir
// Before
%view = memref.subview %source[...] [...] [...]
  : memref<...> to memref<..., strided<...>>
plugin.compute ins(%view : memref<..., strided<...>>)

// After
%tmp = memref.alloc() : memref<...>
memref.copy %view, %tmp
plugin.compute ins(%tmp : memref<...>)
memref.dealloc %tmp
```

### DIFF
--- a/docs/design/pool-allocs-memory-planning.md
+++ b/docs/design/pool-allocs-memory-planning.md
@@ -60,7 +60,7 @@ The default domain ID is zero and is omitted from textual IR. Additional domains
 
 ### Correctness and ABI preparation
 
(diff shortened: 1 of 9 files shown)


kirankumarmatam-amd (approved):
Looks good!

PR #62, opened 2026-03-26
GQA: Complete MS Specification Alignment (Phase 1 - Interface Only)

mingyueliuh:
## Summary

This PR completes **Phase 1: MLIR Interface Complete Alignment** with the Microsoft ONNX Runtime GroupQueryAttention specification. All interfaces are updated to support the full MS spec (14 inputs + 12 attributes + 4 outputs), while maintaining 100% backward compatibility.

**Important**: Phase 1 implements interface alignment only. New features' actual functionality (kernel implementation) will be added in Phase 2.

### Task 1: MLIR Operator Definition (HipOps.td)
- Added 7 new optional inputs: `cos_cache`, `sin_cache`, `position_ids`, `attention_bias`, `head_sink`, `k_scale`, `v_scale`
- Made 4 inputs optional: `key`, `value`, `past_key`, `past_value` (support packed QKV and first-time inference)
- Added 6 new attributes: `local_window_size`, `smooth_softmax`, `qk_output`, `k_quant_type`, `v_quant_type`, `kv_cache_bit_width`
- Added optional output: `output_qk`
- Used `AttrSizedOperandSegments` trait for multiple optional operands
- Added verifier for quantization parameter consistency

### Task 2: ONNX Conversion Layer (OnnxToHip.cpp)
- Support 7-14 variable inputs (removed hardcoded 9-operand check)
- Handle ONNX `NoneType` for optional inputs
- Extract all 12 attributes with proper defaults
- Generate `operand_segment_sizes` for variable operands

### Task 3: LLVM Lowering (HipToLLVM.cpp)
- Generate complete 37-parameter runtime function call
- Handle optional inputs (convert to nullptr)
- Convert string `quant_type` to enum (NONE=0, PER_TENSOR=1, PER_CHANNEL=2)

### Task 4: Runtime Wrapper Signature
- Updated `wrap_group_query_attention` to 37 parameters
- Added debug warnings for unimplemented features (Phase 2)
- Maintained existing kernel call (backward compatible)

### Task 5: IR Conversion Tests
- 6 test cases for ONNX→HIP conversion
- 4 test cases for HIP→LLVM lowering
- Coverage: all input/attribute combinations, optional handling, enum conversion

## Alignment with MS Specification

| Component | MS Spec | Hip GQA | Status |
|-----------|---------|---------|--------|
| Inputs | 14 | 14 | ✅ 100% |
| Attributes | 12 | 12 | ✅ 100% |
| Outputs | 4 | 4 | ✅ 100% |

[...]

PR #34, opened 2026-03-13
Port LLaMA operations from PR-17: rope, matmul, gqa, gather

mingyueliuh:
## Summary

Port four key LLaMA operations from PR-17 to main branch with improvements:
- **hip.rope** - Rotary position embedding
- **hip.matmul** - Matrix multiplication (renamed from hip.hipblaslt.matmul)
- **hip.gqa** - Group query attention (upgraded interface)
- **hip.gather** - Gather operation with axis attribute support

### 1. hip.rope (Rotary Position Embedding)

**Operation redesign:**
- **Rename**: `hip.miopen.rope` → `hip.rope` (remove backend prefix)
- **Signature**: Replace in-place Q/K modification with standard DPS (single input → output)
  - Old: `(ctx, start_pos, cos_cache, sin_cache, q, k)` - 6 params
  - New: `(ctx, input, position_ids, cos_cache, sin_cache, output)` - 6 params + 3 attributes
- **Attributes**: Add `interleaved`, `num_heads`, `rotary_embedding_dim`
- **Assembly format**: Declarative (not custom)
- **Kept**: `$ctx` parameter naming (not `$handle` as in PR-17)

**LLVM lowering:**
- Runtime function: `wrap_rotary_embedding` (was `hip_miopen_rope`)
- Parameters: 12 (6 pointers + 6 i64 values)
- Supports dynamic shapes via `computeNumElements()` utility

**ONNX conversion:**
- Pattern: `RotaryEmbeddingToHip` for `com.microsoft::RotaryEmbedding`
- Attribute inference: When `num_heads=0` or `rotary_embedding_dim=0`, infer from shapes:
  - `rotary_dim = cos_cache.shape[-1] * 2`
  - `num_heads = input.shape[-1] / rotary_dim`

**Runtime function signature:**
```cpp
int wrap_rotary_embedding(
    RuntimeState* state,
    void* input,
    void* position_ids,
    void* cos_cache,
    void* sin_cache,
    void* output,
    int64_t interleaved,
    int64_t num_heads,
    int64_t rotary_embedding_dim,
    int64_t input_num_elements,
    int64_t cos_cache_num_elements,
    int64_t element_size_bytes)
```

**Tests:**
- ✅ `test/lit/Conversion/hip-to-llvm/test_rope.mlir` - HIP→LLVM lowering
- ✅ `test/lit/Conversion/onnx-to-hip/test_rope.mlir` - ONNX→HIP conversion
  * Static shape support with attribute inference
[...]

PR #650, opened 2026-08-04
fix(gqa): size present KV init to past + current seq length

BoarQing:
## Summary

`GroupQueryAttentionToHip` sized the `present_key` / `present_value` DPS init tensors from the **past** operand alone. Present KV is `concat(past, current)` along the seq axis, so its seq extent is `past_seq + current_seq`. On a fresh prefill `past_seq == 0`, so the init collapsed to a **zero-length buffer**; the output allocator then hands the runtime a null `present_key`/`present_value`, `wrap_group_query_attention` rejects the call, and the attention output is left zero-filled.

On gpt-oss-20b this skips *every* GQA call in the graph (24 layers), and the model emits gibberish.

This ports the `%tot = %past + %cur` rule that `OnnxAttentionConversion.cpp` already applies to `onnx.Attention`.

## Root cause / when it was introduced

- Introduced in db7c51fc, "Port LLaMA operations from PR-17: rope, matmul, gqa, gather" (#34, 2026-03-15) — the very first GQA conversion, as `createEmptyTensor(rewriter, loc, presentKeyType, pastKey)`.
- #62 ("GQA: Complete MS Specification Alignment") later changed it to `pastKey ? pastKey : (key ? key : query)`, which only added a null fallback and kept the wrong dimension source.
- #98 moved the code into `GqaConversion.cpp` during the monolith split.

So this is a day-one bug in the GQA path, not a recent regression.

## Why lit never caught it

Every existing GQA lit test (`test_gqa.mlir`, `test_gqa_prefill_sentinel.mlir`) uses **fully static** present shapes. With a static result type the init is a plain `tensor.empty()` with no dynamic operands, so the sizing source is unobservable and the buggy and correct code emit identical IR. The bug only shows up when the present seq dim is dynamic, which is what the real gpt-oss export produces.

This PR adds `test_gqa_dynamic_present.mlir` with dynamic batch/seq. Verified it **fails** against a pre-fix `hip-mlir-opt` and **passes** after the fix.

Pre-fix IR (present sized from `%arg4` = `past_key` only, no `addi`):

```mlir
%dim_1 = tensor.dim %arg4, %c0 : tensor<?x8x?x128xf16>
%dim_2 = tensor.dim %arg4, %c2 : tensor<?x8x?x128xf16>
%1 = tensor.empty(%dim_1, %dim_2) : tensor<?x8x?x128xf16>   // past_seq == 0 -> empty
```

Post-fix IR:

```mlir
%dim_2 = tensor.dim %arg1, %c1 : tensor<?x?x4096xf16>       // current
%dim_3 = tensor.dim %arg4, %c2 : tensor<?x8x?x128xf16>      // past
%1 = arith.addi %dim_3, %dim_2 : index                      // total
%2 = tensor.empty(%dim_1, %1) : tensor<?x8x?x128xf16>
```

## Scope guard

Only rank-3 BSH query (seq at dim 1) with rank-4 BNSH present (seq at dim 2) is re-sized. Any other layout falls back to the previous positional behaviour rather than guessing which axis carries the sequence.

## Validation

Local gpt-oss-20b greedy decode, 125-token prompt, 128 new tokens, HipEP on gfx1151:

| | before | after |
|---|---|---|
| `GQA requires present_key/present_value` errors | 3072 (24 layers x 128 steps) | 0 |
| first token | `11` (`,`) | `20837` (`' AI'`) — matches CPU EP exactly |
| output | `, and la) (H] (H] (H] (H]...` | ` AI, fully or partially or fully?\n\nAMD has fully embraced generative AI...` |
| prefill logits cosine vs CPU EP | — | 0.99998 |

Lit: all `onnx-to-hip` conversion tests pass (107 pass; 4 negative tests need `not`, 2 need `%t`, both limitations of my ad-hoc local runner, not the tests).

## Test plan

- [ ] CI lit: `test/lit/Conversion/onnx-to-hip/test_gqa_dynamic_present.mlir`
- [ ] CI accuracy/E2E on gpt-oss-20b (this is the run to watch — please confirm the CI gibberish clears)
- [ ] Existing GQA/Attention conversion tests still green

### DIFF
--- a/lib/Conversion/OnnxToHip/GqaConversion.cpp
+++ b/lib/Conversion/OnnxToHip/GqaConversion.cpp
@@ -184,11 +184,58 @@ mlir::LogicalResult GroupQueryAttentionToHip::matchAndRewrite(
   // === Create DPS init tensors ===
 
   mlir::Value outputInit = createEmptyTensor(rewriter, loc, outputType, query);
-  mlir::Value presentKeyInit = createEmptyTensor(
-      rewriter, loc, presentKeyType, pastKey ? pastKey : (key ? key : query));
-  mlir::Value presentValueInit =
-      createEmptyTensor(rewriter, loc, presentValueType,
-                        pastValue ? pastValue : (value ? value : query));
+
+  // present KV is concat(past, current) along the seq axis, so its seq extent
+  // is past_seq + current_seq. Taking it from the past operand alone collapses
+  // present to zero length on a fresh prefill (past_seq == 0); the output
+  // allocator then hands the runtime a null present_key/present_value and
+  // wrap_group_query_attention rejects the call, leaving attention zero-filled.
+  // Same reasoning as OnnxAttentionConversion.cpp.
+  // Only rank-3 BSH query (seq at dim 1) and rank-4 BNSH present (seq at dim 2)
+  // are modelled here; for any other layout keep the positional behaviour
+  // rather than guessing which axis carries the sequence.
+  const bool canSizeFromTotal = queryType.getRank() == 3;
+
+  mlir::Value totalKvIdx;
+  auto getTotalKvIdx = [&]() -> mlir::Value {
+    if (totalKvIdx)
+      return totalKvIdx;
+    totalKvIdx = mlir::tensor::DimOp::create(rewriter, loc, query, 1);
+    if (pastKey) {
+      mlir::Value pastIdx =
+          mlir::tensor::DimOp::create(rewriter, loc, pastKey, 2);
+      totalKvIdx =
+          mlir::arith::AddIOp::create(rewriter, loc, pastIdx, totalKvIdx);
+    }
+    return totalKvIdx;
+  };
+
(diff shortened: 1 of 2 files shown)


BoarQing:
Closing: this change is wrong for the contract hip-ep actually supports.

`test/lit/e2e/test_gqa_output_allocator_present_dim.mlir` (added in #306) deliberately pins present.* to be sized from the matching past input's extent. Under OGA `past_present_share_buffer` that extent is the max_length capacity buffer, and `output_allocate_cb` passes the in-graph shape to `GetOutput` verbatim specifically so ORT returns the pre-bound OrtValue and the past==present pointer identity is preserved. Sizing present to `past + current` makes the requested shape `max_length + current`, which no longer matches the pre-bound value, so the KV-cache aliasing would be lost.

The empty-past case that motivated this only occurs with `past_present_share_buffer: false`, which OGA never uses, so this was never going to affect the gpt-oss-20b CI accuracy run either. Supporting the non-shared-buffer contract would be a deliberate change to the allocator invariant, not this patch.

PR #651, opened 2026-08-04
perf(gqa): fused v5 prefill for sinks and sliding windows, at 44% fewer instructions per KV tile

fhanuman:
## Summary

Two commits against the GQA prefill path, both on `gqa_flash_prefill_v5` (the `d==64` warp-private kernel):

1. **`perf(gqa): fused flash prefill for attention sinks and sliding windows`** — teaches v5 attention sinks, smooth softmax and sliding windows so gpt-oss no longer falls back to the decomposed path for its 12 sliding layers and 24 sink layers. `HAS_WINDOW` is a template parameter rather than a runtime test because the kernel runs at the VGPR ceiling: carrying the window's per-row state live in the full-attention instantiation pushed `M_TILES=2` from 120 to 180 bytes/lane of scratch and cost that path 21% at deep KV.

2. **`perf(gqa): cut v5 prefill KV-loop instruction count 44%`** — the loop was bound by instruction *issue*, not occupancy, bandwidth or the matrix units: 1,734 instructions per KV tile to retire 32 WMMA, so the matrix units saw 2% of the stream. Three changes take that to **967 on the 99.6% of tiles that need no causal mask**: native `exp2` in the softmax, the softmax denominator's cross-lane reduction deferred out of the loop into the epilogue, and a mask-free/masked tile split via a tag-dispatched tile body.

## Measured

Interleaved A/B on `gpt_oss-sink sq=512 past=8192`, each arm a separate build, 3 rounds:

| arm | vs parent |
|---|---|
| native exp2 alone | -0.5% |
| deferred reduction alone | -8.9% |
| both | -12.4% |
| **all three** | **-21.4%** |

`exp2` pays nothing on its own because the reduction chain it sits behind is what the loop waits on; it only becomes visible once that chain is gone. The tile split is likewise not separable from the restructure it needs — the same lambda without the split costs 32% on every `d=64` full-attention case, because the split is what lets the compiler keep a tile's live values in registers (in-loop scratch traffic 52 ops → 6).

In-model, 16k-token gpt-oss-20b prefill, RGP capture:

| | before | after |
|---|---|---|
| v5 full-attention dispatch mean | 5,989.7 us | **5,207.7 us** |
| share of capture window | 25.3% | **23.0%** |
| TTFT | 9,653 ms | **9,172 ms (-5.0%)** |

## Correctness

`test_gqa_prefill` covers 26 cases against an fp32 CPU reference: all four sink/smooth/both/none modes, windows at 1/100/128/4096, `past` at 0/512/1000/8192, `d` at 64 and 128, and the two `d=128` cases that assert v5 *declines* so the runtime falls back rather than silently ignoring a sink or a window. **All 26 pass**, worst `relL2` 5.74e-4, unchanged by commit 2 — the reordered summation in the deferred reduction is not observable at this tolerance.

Across the 24 cases the suite also times: 21 faster, 1 tie, no regressions. `d=128` routes to v7 and is unaffected (1.00x, 1.01x).

The new mask-free instantiation is exercised by every full-attention case with `past > 0`, and by the interior q-tiles of the `past = 0` cases.

## Notes for review

- The three A/B switches used to select these changes are **not** in the diff; they were collapsed to their enabled path before this PR. The collapse is provably free: the device image is byte-identical to the switched build across all 443,062 lines of gfx1151 ISA, differing only in the `__hip_cuid` symbol, which is a hash of the translation unit and moves when a comment does.
- Commit 2 adds two measurement-only env knobs, read once and inert unless set: `HIPDNN_PREFILL_V5_ABLATE` (per-phase timing bitmask; numerics are deliberately wrong under it, timing only) and `HIPDNN_PREFILL_V5_CFG` (pins `M_TILES,BKV` so one autotune candidate can be timed alone). Both exist because the questions this kernel raises are otherwise unmeasurable from outside the process. Happy to drop them if that is not wanted here.
- The RGP one-shot capture fence used to take the in-model numbers is deliberately **not** in this PR; it is unrelated profiling infrastructure and belongs in its own.

## What is deliberately left on the table

Re-profiling after the fact says the phase order inverted: the value GEMM is now the largest phase at 36.5% (was 19.8%) and softmax has fallen to 24.2% (was 39.1%). The loop is now only 28-41% issue-bound, so instruction count has largely stopped being the lever. Two candidates remain, both aimed at the LDS pipe and both sized from the ISA but **not built**: storing V transposed in LDS to turn 128 `ds_load_u16` into 16 `ds_load_b128` (worth -5.4% by instruction count, 3.6-7.1% by an LDS-pipe model, against an unquantified bank-conflict risk), and DPP `row_xmask` for the remaining 64 `ds_bpermute` (verified available on gfx1151, but it does not fold into the consuming `v_max`, so it is instruction-neutral and worth 0% or 4.6-9.1% depending on which LDS throughput assumption holds). Together roughly 2% of prefill at the optimistic end.

## Test plan

- [x] `test_gqa_prefill` — 26/26 pass, fp32 reference, worst `relL2` 5.74e-4
- [x] `test_gqa_prefill` timing sweep, 24 cases, 20 and 40 iters, no regressions
- [x] Byte-level ISA equivalence of the shipped source against the measured build
- [x] In-model 16k gpt-oss-20b prefill, RGP capture, TTFT -5.0%
- [ ] CI on this branch
- [ ] A reviewer's opinion on whether the two env knobs stay

### DIFF
--- a/lib/Runtime/Kernels/hip/gqa_kernel.hip
+++ b/lib/Runtime/Kernels/hip/gqa_kernel.hip
@@ -2492,7 +2492,11 @@ typedef float float8_t __attribute__((ext_vector_type(8)));
 // Each wave owns M_TILES 16-row q-subtiles x all BKV columns. Only LDS is
 // P_lds (C->A transpose); only sync is single-wave around it.
 // ---------------------------------------------------------------------
-template <int M_TILES, int BKV, int D>
+// HAS_WINDOW is a template parameter rather than a runtime test because this
+// kernel runs at the VGPR ceiling: carrying the window's per-row state live in
+// the full-attention instantiation pushed M_TILES=2 from 120 to 180 bytes/lane
+// of scratch (38 -> 68 spills) and cost that path 21% at deep KV.
+template <int M_TILES, int BKV, int D, bool HAS_WINDOW>
 __global__ void __launch_bounds__(32)
 gqa_flash_prefill_v5_kernel(
     const _Float16* __restrict__ Q,       // [B, sq, Hq, D]
@@ -2500,7 +2504,10 @@ gqa_flash_prefill_v5_kernel(
     const _Float16* __restrict__ Vcache,  // [B, G, max_seq, D]
     _Float16* __restrict__ O,             // [B, sq, Hq, D]
     int B_count, int Hq, int G, int sq, int skv, int max_seq, int past_len,
-    float scale, int ablate)
+    float scale, int ablate,
+    const _Float16* __restrict__ head_sink,  // [num_heads] or null
+    int num_heads, int smooth_softmax,
+    int local_window_size)                   // <= 0 for full attention
 {
     // ablation bitmask (perf diagnostics only; wrong numerics):
     //   bit0 skip softmax, bit1 skip valueGEMM, bit2 skip scoreGEMM,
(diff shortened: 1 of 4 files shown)


BoarQing:
Recreating under the correct account; superseded shortly.

PR #618, opened 2026-07-31
test(hipsr): gate lit tests during refactor

zpye:
## Summary
- Disable all 30 Hipsr LIT tests with `// UNSUPPORTED: true` during the refactor.
- Keep CMake, LIT configuration, and native unit tests unchanged.

PR #654, opened 2026-08-04
feat(hipsr): compute pool allocation sizes

zz002:
## Summary

Compute the byte size of each pooled allocation and report it through the existing `hipsr-pool-alloc` remark. Also restores the two pool-alloc LIT files so those assertions actually run.

## Related issue or design

`wcy123/onnx-hipdnn-ep#111`. Design: `pool-allocs-design-cn` HIPSR Pool Allocation Pass (Confluence v6). Reference implementation: #572.

## Why

Sizes come from an `arith` per-dimension product over the alloc's static shape and its `getDynamicSizes()` operands, not from the shape dialect. `shape.shape_of %alloc` consumes the alloc's result, which is exactly what the pass replaces with a `memref.view`, so it would be cyclic; reading the operands has no such problem. The upstream issue title still says "using shape dialect" and predates v6.

The two LIT files were gated by #618 while the hipsr refactor was in flight. The new assertions here are worthless while the files are skipped, and both were verified to pass unchanged against current `main`, so they are ungated in the same change that needs them.

## What

- `staticByteFactor` returns element bytes times every static dimension; `computeAllocationSize` pairs it with the alloc's dynamic-size operands. Both are pure functions that build no IR.
- The `#110` remark gains a size suffix: `size 8192` when the shape is fully static, `size 1024 x 1 dyn` when runtime extents remain.
- The 14 existing remark assertions carry the expected size, and four dynamic-shape cases (`dynamic_size`, `coalesce_mixed`, `coalesce_dynamic`, `split_two_groups_dynamic`) are ported from #572 as remark assertions.

## Test plan

- [x] `pool_alloc.mlir` and `pool_alloc_report.mlir` pass with `-verify-diagnostics`, so every `expected-remark` is matched exactly
- [x] Full `check-hip-mlir-lit` run shows no new failure: 338 passed, 1 pre-existing failure in `Conversion/onnx-to-hip/test_gather_block_quantized.mlir` (does not run this pass)
- [x] `pre-commit run --all-files` clean

## Notes for reviewers

- Only the two pool-alloc files are ungated; the other 28 files gated by #618 stay in that refactor's scope.
- The pass still does not modify IR. Turning this size into an `arith.muli` chain with alignment is `wcy123/onnx-hipdnn-ep#112`, and wiring the pass up is `#117`; until then the remark is the only observable output.
- `computeAllocationSize` returns the dynamic extents as `Value`s rather than a count so `#112` can feed them straight into the multiply chain.

## Checklist

### DIFF
--- a/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
+++ b/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
@@ -19,6 +19,7 @@
 #include "llvm/ADT/SmallVector.h"
 
 #include <cstddef>
+#include <cstdint>
 #include <optional>
 
 namespace mlir {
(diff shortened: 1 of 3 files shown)


wcy123 (approved):
LGTM

PR #114, opened 2026-04-14
feat: Add Softplus operator Support

mingyueliuh:
## Summary
-  Complete end-to-end implementation of the **Softplus** operator (`y = ln(exp(x) + 1)`) 
- `Sigmoid` align with ONNX spec (`input/output` → `x/y`)
-  Code Deduplication via generic `lowerMiopenActivation<OpType, activationMode>` template
    activationMode: kActivationSigmoid , kActivationSoftplus

### 4-Stage Pipeline
1. **Hip Dialect** - Added `Hip_SoftplusOp` with DPS semantics (HipOps.td, HipDialect.cpp)
2. **OnnxToHip** - Conversion pattern `onnx.Softplus` → `hip.softplus` (SoftplusConversion.cpp)
3. **HipToLLVM** - Lowering to LLVM IR with runtime calls (ActivationLowering.cpp)
4. **Runtime** - MIOpen activation wrapper supporting f32/f16/bf16 (activation.cpp)

### Test Coverage
- 2 OnnxToHip lit tests (static and dynamic shapes)
- 2 HipToLLVM lit tests (IR generation verification)
- 1 e2e lit test
- **All 81 project tests pass** ✅

PR #113, opened 2026-04-13
feat: PyTorch model support with hybrid GPU execution (Qwen3-0.6B & Qwen3.5-35B-A3B)

tejus26:
## Summary

This PR adds end-to-end PyTorch model support to the HIP MLIR compiler, enabling hybrid execution where compute-heavy ops run on compiled GPU DLLs while the rest executes on PyTorch.

### Architecture

```
                        PyTorch Model (HuggingFace)
                                  │
                          torch.export()
                                  │
                          FX Graph (ATen ops)
                                  │
                    ┌─────────────┴──────────────┐
                    │                            │
              Supported ops               Unsupported ops
          (linear, silu, mul,         (embedding, attention,
           rms_norm, matmul,           RMSNorm, masks, etc.)
           softmax, etc.)                     │
                    │                    PyTorch eager
              fx_to_mlir.py              execution
                    │                   (CPU or GPU)
            Torch dialect MLIR                │
                    │                         │
              hip-compiler.exe                │
                    │                         │
              GPU DLL (.dll)                  │
                    │                         │
              HipDllRunner                    │
              (ctypes load)                   │
                    │                         │
                    └─────────────┬───────────┘
                                  │
                          Model Output
                       (text generation)
```

### Key Components

| Component | Description |
|-----------|-------------|
| `scripts/fx_to_mlir.py` | Converts torch.export FX graphs to Torch dialect MLIR |
| `scripts/hip_backend.py` | torch.compile custom backend with op support classification |
| `scripts/hip_dll_runner.py` | ctypes-based in-process GPU DLL loader |
| `scripts/run_qwen_hybrid.py` | Top-level hybrid execution script |
| `scripts/run_qwen_pytorch.py` | PyTorch model validation |
| `scripts/test_numerical_accuracy.py` | GPU vs PyTorch numerical comparison |
| `scripts/test_dll_runner.py` | Full pipeline E2E test |
| `lib/Conversion/TorchToHip/*.cpp` | 12 new TorchToHip conversion patterns |
| `lib/Runtime/memref_copy.cpp` | GPU-aware strided memref copy |
| `lib/Runtime/real/silu.cpp` | SiLU activation runtime (MIOpen) |

### TorchToHip Conversion Patterns (30 ops)

**New patterns added in this PR:**
- `NormConversion.cpp`: `torch.aten.rms_norm` → `hip.rms_norm`
[...]

PR #111, opened 2026-04-13
feat: extend CI result tables and switch iGPU default to VRAM allocation

zz002:
### CI: extended result tables (`windows-build.yml`)
- **L2 accuracy table**: add `Total Elems` and `Skipped NaN/Inf` columns — matches the format already used in `gpu-perf-accuracy-test.yml`
- **Perf table**: add `Session (s)`, `1st Infer (ms)`, `CPU%`, `Mem (MB)` columns for richer at-a-glance diagnostics

### Runtime: iGPU default switched to hipMalloc (VRAM) path (`hipdnn_ep_runtime_state.cpp`)

The original `hipHostMalloc` path was designed to optimize **session creation time** by skipping the H2D copy — the GPU reads weights directly from pinned host memory (shared DRAM on iGPU). However, this means every inference reads weights from the slower host-side memory path, hurting throughput.

For production workloads (session created once, inference run thousands of times), the one-time copy cost at session init is negligible compared to the per-inference benefit of GPU reading from VRAM. Switching to `hipMalloc` (VRAM) + a single `hipMemcpy` at session init gives the GPU fast local access to weights on every inference.

Changes:
- Switch iGPU default from `hipHostMalloc` (pinned host) to `hipMalloc` (VRAM) + `hipMemcpy H2D`
- Add `HIPDNN_EP_IGPU_USE_HOST=1` env var to revert to the legacy path if session creation latency matters

## Performance Comparison (StrixHalo)

Baseline: PR #110 Run 661 (commit `80a921e`, `hipHostMalloc` path)
New: PR #111 Run 666 (commit `ca5312d`, `hipMalloc` VRAM path)

| Model | QPS (base) | QPS (new) | QPS Δ | Session Δ | 1st Infer Δ | Mem Δ |
|-------|----:|----:|----:|----:|----:|----:|
| full_model_seq128 | 5.85 | **7.32** | **+25.1%** | 40.57s → 41.60s | 303ms → 273ms (-9.9%) | 15693MB → 15586MB (-107) |
| GroupQueryAttention_seq128 | 1214.69 | **1216.75** | **+0.2%** | 1.08s → 1.07s | 113ms → 113ms | 330MB → 266MB (-64) |
| matmul_down_seq128 | 630.60 | 609.41 | -3.4% | 1.55s → 1.57s | 58ms → 58ms | 370MB → 341MB (-29) |

> Results are machine- and model-dependent.

## Test plan
- [x] CI step summary shows new table columns correctly
[...]

PR #656, opened 2026-08-04
feat(hipsr): locate the pool insertion point

zz002:
## Summary

`hipsr-pool-alloc` now locates the insertion point for the pool size arithmetic: the last poolable allocation in a `hipsr.pool_domain` body, reported as a domain-level remark under `emit-pool-report`. A domain holding no poolable allocation is rejected rather than silently skipped. The pass still does not modify IR.

## Related issue or design

`wcy123/onnx-hipdnn-ep#113`. Design: `pool-allocs-design-cn` HIPSR Pool Allocation Pass, Step 3. Reference implementation: #572. Follows #654 (`#111`).

## Why

**The point is the last *poolable* allocation, not the last `memref.alloc`.** It exists so that `#112`'s group size arithmetic can reference the `getDynamicSizes()` operands of the allocations that actually participate in pooling, and an allocation that participates in nothing is irrelevant to that. Taking the poolable one also keeps the point as early as possible, which matters for the dominance premise below. The design's Step 3 and #572's `findLastAlloc` both take the last allocation regardless of pooling. `@interleaved_allocs` distinguishes the two: `%unwritten` is the last allocation (op 6) but receives no destination-style write, while `%a3` is the last poolable one (op 4). The assertion there pins the choice down.

**An absent insertion point is an error.** Narrowing the search means the result can be absent, so the pass rejects such a domain. Keying the check on "no poolable allocation" rather than "no `memref.alloc`" is what makes it worth doing: the weaker condition still admits a domain whose allocations all lack a destination-style write, which yields no insertion point either and would force `#112` to check again. The stronger one turns "the insertion point exists" into an invariant `#112` can assume. This follows the design's Step 3 and diverges from #572, which emits a remark and returns when there is nothing to pool. Domain partitioning is not supposed to produce a domain with nothing poolable, and skipping silently would hide that structural error until `#114` wires the rewrite up.

## What

- `findInsertionPoint` scans the body's top level and returns the last `memref.alloc` that has a live range, or `nullptr`.
- `runOnOperation` calls it right after `computeLiveness`; on `nullptr` it emits an error on the `hipsr.pool_domain` and calls `signalPassFailure()`.
- `emitPoolingReport` takes the point and emits `hipsr-pool-alloc: insertion point after op N` on the domain, numbering ops in the same body top-level coordinates as the existing `lifetime [a,b]` remarks.
- `pool_alloc.mlir` is deleted. Both of its cases asserted that a domain with nothing poolable passes silently, which the precondition invalidates, and `@noalloc_noop` is tensor-level so it could never hold a `memref.alloc` at all. This leaves the plain no-report RUN path without LIT coverage until `#114`/`#116` make the pass rewrite IR and a FileCheck file is needed again.
- `pool_alloc_invalid.mlir` covers the error with an empty domain, and with a domain whose only allocation never receives a destination-style write.
- `pool_alloc_report.mlir` gains one domain-level assertion per function, plus `@alloc_inside_region`, which holds a `memref.alloc` inside an `scf.execute_region` and pins the search to the body's top level.

## Test plan

- [x] `check-hip-mlir-lit`: 338 passed, 1 failed -- the pre-existing `Conversion/onnx-to-hip/test_gather_block_quantized.mlir`, which does not run this pass. 370 discovered, unchanged: one file deleted, one added.
- [x] All 10 insertion-point remarks in `pool_alloc_report.mlir` match exactly under `-verify-diagnostics`; the indices come from a real run rather than by hand.
- [x] Both `expected-error`s in `pool_alloc_invalid.mlir` match exactly.
- [x] The invalid file is not vacuous: dropping its expectations, and separately corrupting only the second chunk's expected text, each make the run exit 1, and the second chunk's failure is reported on its own under `-split-input-file`.

## Notes for reviewers

- **Dominance premise for `#114`.** The design's Step 3 justifies a single insertion point by assuming `materialize-init-tensors` has hoisted every allocation to the start of the domain. `@interleaved_allocs` deliberately violates that shape, so its insertion point (op 4) sits after a user of `%a1`. Once `#114`/`#116` replace the allocations with `memref.view`, that view will not dominate its user and the IR becomes invalid. Nothing breaks here because this PR only reports, but the wiring in `#114` has to either rely on the hoisted shape or handle the interleaved one.
- The pass emits no IR yet. `#112` turns #654's sizes into an `arith.muli` chain and does one `alignUp(256)` per group at this point; `#117` wires the pass into a pipeline.
- Upstream lists `#112` before `#113`, but `#112` needs somewhere to put its arithmetic, so this lands first.
- Substantial AI assistance (Cursor) was used for this change and this description. I reviewed the result and understand it.

## Checklist


lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
```diff
@@ -132,9 +132,33 @@ greedyGrouping(Block &body, const llvm::DenseMap<Value, Lifetime> &lifetimes) {
   return groups;
 }
 
+Operation *
+findInsertionPoint(Block &body,
+                   const llvm::DenseMap<Value, Lifetime> &lifetimes) {
```
wcy123, line 137: it should be static function, minimize the scope

wcy123 (commented):
LGTM, but curly braces needed:

https://github.com/ROCm/hip-ep/blob/cfc0fd457f68b0e88ef915e3011cd98d4acb8e84/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp#L139-L142

Add braces per [LLVM standard](https://llvm.org/docs/CodingStandards.html) when the `if` body contains an early-return statement (line 144's `return`).

lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
zz002: currently using Anonymous namespaces.

wcy123 (approved):

PR #658, opened 2026-08-04
feat(onnx-to-hip): fold attention score prep Mul and broadcast Add

amd-mounikk:
## Summary
- Add pre-lowering patterns for Swin-style window-attention score preparation before lowerOnnxConstants.
- **FoldMatMulScaleMul**: rewrites MatMul(A,B) -> Mul(scores, scale) to Mul(A, scale) -> MatMul(A', B), avoiding a broadcast mul on large score tensors.
- **ExpandConstantBroadcastAdd**: compile-time expands broadcast constant biases so Add(activation, Constant[bias]) becomes same-shape add, avoiding runtime broadcast elementwise on large attention maps.
- **Review fixes (3bfb0919)**: MatMul scale fold guarded by K/N extent check; bias expansion capped at 1 Mi output elements.
- Includes LIT coverage in test_attention_score_prep_opt.mlir (positive + negative cases for both patterns).

## Test plan
- [x] pre-commit run --from-ref origin/main --to-ref HEAD (lintrunner + licenseheaders)
- [x] hip-mlir-opt test/lit/Conversion/onnx-to-hip/test_attention_score_prep_opt.mlir (FileCheck pass)
- [x] SwinV2 tiny 8x521x1024 end-to-end perf (local): avg ~274 ms/inference, matmul 27 calls / 31.8 ms GPU


lib/Conversion/OnnxToHip/AttentionScorePrepOptimizations.cpp
```diff
@@ -0,0 +1,347 @@
...
+      return failure();
+
+    Value a = matmulOp->getOperand(0);
+    Value b = matmulOp->getOperand(1);
+    auto aTy = dyn_cast<RankedTensorType>(a.getType());
+    auto scaleTy = dyn_cast<RankedTensorType>(scaleVal.getType());
+    if (!aTy || !scaleTy)
+      return failure();
+
+    auto scaledATy = broadcastShape(aTy.getShape(), scaleTy.getShape());
```
fhanuman, line 293: Could this rewrite change the result when the scale’s last dimension is not 1? For example, would moving the scale in the IR below make it scale the values being summed rather than the output columns?

```mlir
%scale = "onnx.Constant"() {
  value = dense<[10.0, 100.0]> : tensor<2xf16>
}
%scores = "onnx.MatMul"(%a, %b) // A, B, and output are 2x2
%out = "onnx.Mul"(%scores, %scale)
```

lib/Conversion/OnnxToHip/AttentionScorePrepOptimizations.cpp
```diff
@@ -0,0 +1,347 @@
...
+    if (!constTy || !isStaticShape(constTy))
+      return failure();
+    if (!broadcastsTo(constTy.getShape(), outTy.getShape()))
+      return failure();
+
+    auto denseOr = getConstantDenseElements(constOp);
+    if (failed(denseOr))
+      return failure();
+
+    auto expandedOr =
```
fhanuman, line 357: Could expanding constants become expensive for large static outputs? For example, would the IR below turn a 2-byte constant into ~1 GB of constants?

```mlir
%bias = "onnx.Constant"() {value = dense<0.0> : tensor<1x1x1x1xf16>}
%out = "onnx.Add"(%scores, %bias)
    : (tensor<1x32x4096x4096xf16>, tensor<1x1x1x1xf16>)
      -> tensor<1x32x4096x4096xf16>
```

lib/Conversion/OnnxToHip/AttentionScorePrepOptimizations.cpp
amd-mounikk: Good catch — agreed. Moving \(A * scale) @ B\ is not equivalent to \(A @ B) * scale\ when scale varies along MatMul's contraction (K) or output column (N) axis, including your \[10, 100]\ example on a 2×2 MatMul.

Fixed in 3bfb0919 by requiring scale extent to be 1 on both K (A's last dim) and N (result's last dim) after NumPy-style alignment. Swin per-head \[H,1,1]\ scales pass; column-scale patterns are rejected.

Added \	est_no_fold_column_scale\ in \	est_attention_score_prep_opt.mlir\ — expects \hip.matmul\ then \hip.mul\ (no fold).

lib/Conversion/OnnxToHip/AttentionScorePrepOptimizations.cpp
amd-mounikk: Yes — without a guard, materializing that \1×1×1×1\ bias to \1×32×4096×4096\ would be ~1 GB at compile time.

Fixed in 3bfb0919: \ExpandConstantBroadcastAdd\ now skips expansion when the output exceeds \kMaxConstantExpandElements\ (1 Mi elements, ~2 MiB fp16). Swin window attention maps stay well under this; huge static tensors keep the tiny broadcast constant and runtime broadcast add.

Added \	est_no_expand_huge_bias\ in \	est_attention_score_prep_opt.mlir\ — verifies the \1×1×1×1\ bias is not expanded. Lit + local Swin perf re-run passed.

fhanuman (approved):
I will approve the PR to unblock the merge. However, I am still not clear on how much this change improves performance, so in future PRs like this, it would be helpful to include an example showing where it helps and some before-and-after performance numbers.

PR #661, opened 2026-08-04
perf(matmul_nbits): lower the WMMA cutoff to M=8 (-2.3% TTFT on gpt-oss-20b 16k)

BoarQing:
Stacked on #652 — review that one first; this PR's diff is the top commit only.

## Summary

`MatMulNBits` routed every `1 < M < 16` call to the col-major GEMV path. In a chunked prefill that tier is not a decode leftover: the MoE experts that receive few routed tokens land in it on every layer of every chunk. In an RGP capture of a 512-token chunk, 30 of 155 expert blocks had M in 2..15, each costing 328 us against a 55 us weight-read floor — 43 GB/s effective on a 256 GB/s part, the worst-utilised bucket in the prefill.

This lowers the WMMA/GEMV boundary from 16 to 8, so those experts read B through the tiled, double-buffered WMMA pipeline instead of the GEMV's K-parallel reduction. No kernel changes: partial M tiles were already handled, because `need_bc` in `launchWmmaConfig` selects the bounds-checked variant whenever M is not a multiple of the tile, and that variant guards both the A loads and the C writes on M.

## Why 8

WMMA computes a whole tile of rows whatever M is, so the wasted rows grow as M shrinks and eventually outweigh the better read. 8 is the best of three cutoffs measured on gpt-oss-20b at 16k; taking the tier all the way down to 2 is a **regression**.

| cutoff | n | mean TTFT | paired vs 16 |
|---|---|---|---|
| 16 (before) | 16 | 9147 ms | — |
| **8 (this PR)** | 10 | **8944 ms** | **-211 ms (-2.3%)** |
| 2 | 12 | 9185 ms | +47 ms (+0.5%) |

Method: 16k prefill at `-r 4`, arms paired by round and interleaved, arm order reversed for half the rounds. Reversing moves the M=8 delta from -243 to -195 ms on one build and from -199 to -201 ms on a rebuild of this exact tree, so it is not position bias. Each arm gets its own `morphizen_wmma_cache_*.txt`, since that file holds a single build timestamp and a shared one would make every DLL swap a cold-tune run.

Two blocks of rounds were discarded and are not in the table: they sat 7-9% above the known 9042 ms baseline while the box shed heat from a build, and they reported the opposite sign.

## Caveat for anyone re-measuring with RGP

Under a capture the M=2 cutoff looks like a win — the 2..15 blocks fall 328 -> 287 us while the untouched 16..63 and 64..255 buckets drift only 3% and 0.4% — yet it is a regression in a normal run. RGP pegs clocks, which makes the WMMA path's redundant arithmetic free. For this cutoff, trust TTFT over the capture.

## Scope

Any `MatMulNBits` with batch 1, `K%32==0` and M in 8..15 changes path, so a short prompt or a small speculative batch is affected too. Those are covered for correctness but their timing is not part of the measurement above, which is dominated by MoE expert GEMMs.

4 and 12 are untested, so 8 is the best measured cutoff rather than a proven optimum.

## Test plan

- [x] `SEQ_LENS` in the MatMulNBits numeric tests was `[1, 128]`, leaving the entire `1 < M < 16` tier with no coverage at all. Extended to `[1, 2, 7, 15, 128]`: the row-major GEMV, the col-major GEMV on both sides of the new boundary, and a partial WMMA tile.
- [x] `pytest tests/test_matmul_nbits.py` — 70 passed, on both the experiment build and a clean rebuild of this tree.
- [x] `tests/test_qmoe.py::test_qmoe_gpt_oss_shape[1]` fails, but produces byte-identical values on the unpatched DLL, so it is pre-existing and unrelated. The M==1 path is untouched by this change.
- [x] Builds clean on this branch after rebase onto the current base.


BoarQing:
Acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/186/
Perf: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F190&base_job=daily_hip_ep_genai&base=19&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F202&curr_job=daily_hip_ep_genai&curr=202

amd-mingw (approved):

PR #663, opened 2026-08-05
feat(hipsr): add materialize-init-tensors pass with placeholder grouping

amd-genmingz:
## Summary
Adds the `-hipsr-materialize-init-tensors` pass and lands the transform end to end.
The pass rewrites each `hipsr.pool_domain` into the virtual three-region form — shape computation, tensor allocation, then data ops — so that `hipsr.placeholder` init tensors become plain `tensor.empty` before bufferization. The first submission carried phase 1 only; phases 2 to 4 have since landed on the branch, so the pass now runs to completion and no `hipsr.placeholder` survives in the output.

Four phases run per domain:
1. **Grouping** — every `hipsr.placeholder` in a domain moves to the front of the domain block, keeping the relative order SSA form already gives it.
2. **Shape computation** — each `shape_region` body moves into an `scf.execute_region` yielding `!shape.shape`. The region's block arguments are replaced by the shapes of the placeholder inputs: `shape.shape_of` for an input that outlives the pass, or the producing `scf.execute_region` result when the input comes from another placeholder, since that placeholder is erased later.
3. **Allocation** — one `tensor.empty` per placeholder result, its dynamic extents read out of the computed shape with `shape.get_extent`. Static dimensions are already carried by the result type and get no SSA value. All allocations are grouped after the last `scf.execute_region`.
4. **Cleanup** — every placeholder result use is rewired to its `tensor.empty`, then the placeholders are erased. The replacement has to precede the erase, and every placeholder in a domain has to go in the same run, because a placeholder result can feed a later placeholder's inputs.

Barrier placeholders are rejected with a diagnostic rather than materialized; that path comes later. The pass is not part of any pipeline yet; it runs only when requested explicitly through `hip-mlir-opt`.

Tests live in `test/lit/Dialect/Hipsr/Transforms/MaterializeInitTensors/`. The positive file covers the canonical single-placeholder domain, a fully static result, a result mixing dynamic and static dimensions, a cross-placeholder shape dependency, the same domain already grouped so that grouping is shown to be idempotent, and a domain with no placeholders. Each case spells out its whole function with `CHECK-NEXT`, so an unexpected extra or missing op breaks the chain rather than slipping past a spot check. The error file covers an unpopulated shape region and a barrier placeholder.

## Development plan
This PR is the first step of a series. Each step keeps the IR valid and ships with its own tests. What remains:
1. **Barrier placeholders** — their shape regions read data values, so block arguments pass through without `shape.shape_of`. Currently rejected with `barrier placeholders are not materialized yet`.
2. **Multi-result placeholders** — the pass is already generic over result count, with one `!shape.shape` and one `tensor.empty` per result mapped by result index, but no test exercises it yet.
3. **Remaining error paths** — two diagnostics are implemented but uncovered: a `shape_region` whose argument count disagrees with the placeholder inputs, and an input whose producing placeholder was not materialized first.
4. **Pipeline integration (optional)** — insert after `-hipsr-partition-pool-domains` and before bufferization.

## AI assistance
Substantial AI assistance (Cursor). It contributed the task breakdown for this series, the pass implementation, and the lit tests. The author reviewed all of it and owns every claim in this description; the change is validated by the lit suite described in the summary.


test/lit/Dialect/Hipsr/Transforms/MaterializeInitTensors/materialize-init-tensors.mlir
```diff
@@ -0,0 +1,289 @@
...
+// CHECK:             %[[B_SHAPE:.+]] = shape.shape_of %[[DB]] : tensor<256x512xf16> -> !shape.shape
+// CHECK:             %[[M:.+]] = shape.get_extent %[[A_SHAPE]], %{{.+}} : !shape.shape, !shape.size -> !shape.size
+// CHECK:             %[[N:.+]] = shape.get_extent %[[B_SHAPE]], %{{.+}} : !shape.shape, !shape.size -> !shape.size
+// CHECK:             %[[RESULT_SHAPE:.+]] = shape.from_extents %[[M]], %[[N]] : !shape.size, !shape.size
+// CHECK:             scf.yield %[[RESULT_SHAPE]] : !shape.shape
+// CHECK:           }
+// CHECK:           %[[EXTENT:.+]] = shape.get_extent %[[MATMUL_SHAPE]], %{{.+}} : !shape.shape, !shape.size -> !shape.size
+// CHECK:           %[[DIM0:.+]] = shape.size_to_index %[[EXTENT]] : !shape.size
+// CHECK:           tensor.empty(%[[DIM0]]) : tensor<?x512xf16>
+// CHECK:           %[[INIT:.+]] = hipsr.placeholder(%[[DCTX]]) ins(%[[DA]], %[[DB]] : tensor<?x256xf16>, tensor<256x512xf16>) {placeholder_type = #hipsr.placeholder_type<normal>} : tensor<?x512xf16>
```
wcy123: use CHECK-NEXT, so we can see the complete output.

lib/Dialect/Hipsr/Transforms/MaterializeInitTensorsPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+  OpBuilder builder(poolDomain.getContext());
+  FailureOr<DenseMap<PlaceholderOp, scf::ExecuteRegionOp>>
+      placeholderToExecuteRegion =
+          createShapeComputations(*placeholders, builder);
+  if (failed(placeholderToExecuteRegion)) {
+    return failure();
+  }
+
+  DenseMap<Value, Value> placeholderResultToInitTensor =
+      createTensorAllocs(*placeholders, *placeholderToExecuteRegion, builder);
```
wcy123: `createTensorAllocs` -> `createTensorEmpty()` ?

lib/Dialect/Hipsr/Transforms/MaterializeInitTensorsPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+      return nullptr;
+    }
+    return executeRegion.getResult(cast<OpResult>(input).getResultNumber());
+  }
+  return builder.create<shape::ShapeOfOp>(
+      loc, shape::ShapeType::get(builder.getContext()), input);
+}
+
+// Moves a placeholder's shape region body into a new scf.execute_region placed
+// just before it. hipsr.shape_yield2 is bound to hipsr.placeholder by HasParent
```
wcy123: shape_yield

test/lit/Dialect/Hipsr/Transforms/MaterializeInitTensors/materialize-init-tensors.mlir
```diff
@@ -0,0 +1,393 @@
...
+// CHECK-NEXT:        %[[B_SHAPE:.+]] = shape.shape_of %[[DB]] : tensor<256x512xf16> -> !shape.shape
+// CHECK-NEXT:        %[[M_INDEX:.+]] = shape.const_size 0
+// CHECK-NEXT:        %[[M:.+]] = shape.get_extent %[[A_SHAPE]], %[[M_INDEX]] : !shape.shape, !shape.size -> !shape.size
+// CHECK-NEXT:        %[[N_INDEX:.+]] = shape.const_size 1
+// CHECK-NEXT:        %[[N:.+]] = shape.get_extent %[[B_SHAPE]], %[[N_INDEX]] : !shape.shape, !shape.size -> !shape.size
+// CHECK-NEXT:        %[[RESULT_SHAPE:.+]] = shape.from_extents %[[M]], %[[N]] : !shape.size, !shape.size
+// CHECK-NEXT:        scf.yield %[[RESULT_SHAPE]] : !shape.shape
+// CHECK-NEXT:      }
+// CHECK-NEXT:      %[[DIM0_INDEX:.+]] = shape.const_size 0
+// CHECK-NEXT:      %[[DIM0_EXTENT:.+]] = shape.get_extent %[[MATMUL_SHAPE]], %[[DIM0_INDEX]] : !shape.shape, !shape.size -> !shape.size
```
wcy123, line 59: can we avoid size_to_index？

wcy123 (approved):
LGTM

PR #670, opened 2026-08-06
feat(hipsr): add hipsr.preserve_shape and its bufferization model

amd-genmingz:
## Summary
Adds hipsr.preserve_shape, an op that holds a !shape.shape next to the tensor or memref it describes, so that association can survive -one-shot-bufferize.
## Why
`-one-shot-bufferize` erases the only data-flow path between a shape computation and the buffer it sized. A DPS op bufferizes in place, so its init tensor and its result share one `memref.alloc` and the init's SSA value disappears; the shape computation that sized it is then unreachable and DCE deletes it, leaving nothing for `-hip-use-output-allocator` to recover a graph output's declared extents from. Holding both values in one op keeps the shape reachable.
## What
- `HipsrPreserveShapeOp.td`: two operands, no results, no verifier. The rank agreement the description states is documented, not enforced: `!shape.shape` is rank erased.
- `Hipsr_TensorOrMemRef` in `HipsrTypes.td`, a new and deliberately space-agnostic constraint. Unlike the four `Hipsr_TensorOr*MemRef` constraints beside it, it accepts a memref in any space including none, because the op has to hold after `-one-shot-bufferize`, whose `memref.alloc`s carry no `#hipsr.mem` space.
- `getEffects` reports a valueless `Write` on the default resource, the same encoding `hip.alloc_output` uses: enough that DCE and CSE leave the op alone, without naming a buffer it would then order against.
- `HipsrBufferize.h`: a hand-written `BufferizableOpInterface` model, since the op is not DPS and `DstBufferizableOpInterfaceExternalModel` does not apply. All three analysis queries report that nothing happens, so the op forces no copy and the DPS op it names still bufferizes in place; `bufferize` re-creates it with the `$data` buffer. Registered from `InitAllPasses.h` and `hip-mlir-opt.cpp`.
## AI assistance
Substantial AI assistance (Cursor): the op definition, the effects implementation, the bufferization model, and the lit tests
Made-with: Cursor


wcy123 (commented):
Missing `BufferizableOpInterface` - the design doc specifies it should be implemented.

**Issue:** https://github.com/ROCm/hip-ep/blob/680814909729a46f418b60efbfde263b413798af/include/hip/Dialect/Hipsr/IR/HipsrPreserveShapeOp.td#L15-L16

Current:
```tablegen
def Hipsr_PreserveShapeOp : Hipsr_Op<"preserve_shape",
    [DeclareOpInterfaceMethods<MemoryEffectsOpInterface>]> {
```

Should be (per design doc):
```tablegen
def Hipsr_PreserveShapeOp : Hipsr_Op<"preserve_shape",
    [DeclareOpInterfaceMethods<MemoryEffectsOpInterface>,
     DeclareOpInterfaceMethods<BufferizableOpInterface>]> {
```

The op needs `BufferizableOpInterface` to survive `-one-shot-bufferize` and transform from tensor form to memref form. Without it, the op will cause bufferization to fail with "op was not bufferized" (as the current ODS doc acknowledges). The implementation should follow the design doc pattern in `tech/hip-ep-doc/doc/op-preserve-shape.md`.

lib/Dialect/Hipsr/IR/HipsrPreserveShapeOp.cpp
```diff
@@ -0,0 +1,45 @@
...
+// to avoid being eliminated as a trivially dead operation.
+void PreserveShapeOp::getEffects(
+    SmallVectorImpl<SideEffects::EffectInstance<MemoryEffects::Effect>>
+        &effects) {
+  effects.emplace_back(MemoryEffects::Write::get(),
+                       ShapeMetadataResource::get());
+}
+
+LogicalResult PreserveShapeOp::verify() {
+  auto fromExtents = getShape().getDefiningOp<shape::FromExtentsOp>();
```
wcy123: cast<ShapeType>(getShape().getType()).getRank()

cast<ShapedType>(getData().getType()).getRank();

test/lit/Dialect/Hipsr/Transforms/bufferization/preserve_shape_bufferize.mlir
wcy123, line 1: test/lit/Dialect/Hipsr/preserve_shape_bufferize.mlir

maybe test/lit/Dialect/Hipsr/Transform/bufferization/preserve_shape.mlir is better, 
or test/lit/Dialect/Hipsr/Conversion/bufferization/preserve_shape.mlir , because tensor->memref.

because we might need to study the bufferization behaviors in the future.

test/lit/Dialect/Hipsr/Transforms/bufferization/preserve_shape_bufferize.mlir
```diff
@@ -0,0 +1,106 @@
...
+// CHECK-SAME: -> memref<?x2048xf32> {
+// CHECK-NEXT: %[[C2048:.+]] = arith.constant 2048 : index
+// CHECK-NEXT: %[[SHAPE:.+]] = shape.from_extents %[[D0]], %[[C2048]] : index, index
+// CHECK-NEXT: %[[ALLOC:.+]] = memref.alloc(%[[D0]]) {{.*}}: memref<?x2048xf32>
+// CHECK-NEXT: hip.matmul(%[[CTX]]) ins({{.+}}) outs(%[[ALLOC]] : memref<?x2048xf32>)
+// CHECK-NEXT: hipsr.preserve_shape %[[SHAPE]], %[[ALLOC]] : memref<?x2048xf32>
+func.func @dps_result(%ctx: !hip.context, %d0: index, %lhs: tensor<?x64xf32>,
+                      %rhs: tensor<64x2048xf32>) -> tensor<?x2048xf32> {
+  %c2048 = arith.constant 2048 : index
+  %shape = "shape.from_extents"(%d0, %c2048) : (index, index) -> !shape.shape
```
wcy123: `shape` should be known dialect.

lib/Dialect/Hipsr/IR/HipsrPreserveShapeOp.cpp
wcy123: if shape's rank is not known at compile-time, we don't need any verifier.

wcy123 (dismissed):
LGTM

wcy123 (approved):

PR #671, opened 2026-08-06
feat(hipsr): emit pool group size arithmetic

zz002:
## Summary

`hipsr-pool-alloc` rewrites IR for the first time: after the last poolable alloc it emits one `index` size chain per lifetime group — a byte size per member, `arith.maxui` across the group, and a single alignUp to 256 for the whole group.

## Related issue or design

wcy123/onnx-hipdnn-ep#112

## Why

- The static part is one compile-time constant, not an `arith.muli` per static dimension: `computeAllocationSize` already collapses the static extents into `staticByteFactor`, so only the dynamic-size operands need a multiply chain at runtime.
- alignUp runs once per group instead of once per member. Members share the group's base offset, so only the group extent has to land on a 256 B boundary; aligning each member would pad every one of them.
- Alignment is hardcoded to 256. No caller needs a different value, so there is no pass option for it yet.
- The chain is emitted unconditionally, including for all-static groups, and left for `canonicalize` to fold rather than folded in the pass.

## What

- `kPoolAlignment`, `emitAllocationSize`, `emitGroupSize` in `HipsrPoolAllocPass.cpp`.
- `runOnOperation` sets the insertion point after the anchor `findInsertionPoint` returns and emits one chain per group. The report still runs before emission so it observes the input IR.
- The group sizes are discarded for now; `hipsr.get_pool` becomes their consumer in wcy123/onnx-hipdnn-ep#114, so until then they are dead `arith` values. The verifier is fine with that and the pass is not in a pipeline yet.

## Test plan

- [x] `test/lit/Dialect/Hipsr/pool_alloc.mlir` passes: five cases carried over from #572 (`align_up_rounding`, `dynamic_size`, `dead_alloc_skipped`, `coalesce_static`, `split_two_groups`) with the CHECK chains trimmed to the end of the align chain.
- [x] `pool_alloc_report.mlir` and `pool_alloc_invalid.mlir` show no regression when run for real. The report file only asserts `CHECK-LABEL` plus `--implicit-check-not` on `hipsr.get_pool` / `memref.view`, and the new `arith` ops are on neither list; the invalid file returns before emission.
- [x] Full `check-hip-mlir-lit`: 352 passed, 24 unsupported, and the single failure is `Conversion/onnx-to-hip/test_gather_block_quantized.mlir`, which reproduces identically on clean `main` (an MLIR `BuiltinAttributes.cpp` assertion in an unrelated conversion).
- [x] `pre-commit run --all-files` converges at the hip-ep root.

## Notes for reviewers

- `pool_alloc.mlir` is the file name #656 deleted; it comes back here as this pass's single transform test, so git renders it as a same-name delete-then-add. The remaining steps of the upstream series — `wcy123/onnx-hipdnn-ep` [#114](https://github.com/wcy123/onnx-hipdnn-ep/issues/114) (pool emission), [#115](https://github.com/wcy123/onnx-hipdnn-ep/issues/115) (offsets), [#116](https://github.com/wcy123/onnx-hipdnn-ep/issues/116) (`memref.view` substitution) — keep extending the CHECKs in this same file instead of adding new ones: every stage rewrites the IR the previous stage's CHECKs matched, and by the last one even `memref.alloc` is gone, so one file makes each PR's diff show exactly how much further the transform now reaches.
- There is deliberately no `-canonicalize` RUN. With no consumer the whole chain is trivially dead, so canonicalize deletes it rather than folding it, and alignUp's numeric correctness only becomes observable once wcy123/onnx-hipdnn-ep#114 wires up `hipsr.get_pool`.
- Open question, not triggered by this PR: the insertion point is anchored after the last poolable alloc. Emitting sizes there is always safe, but wcy123/onnx-hipdnn-ep#116 has to confirm the `memref.view` dominates its users — the `@interleaved_allocs` shape in `pool_alloc_report.mlir` is the counterexample.

## Checklist

### DIFF
--- a/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
+++ b/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
@@ -40,6 +40,8 @@ struct AllocationSize {
   llvm::SmallVector<Value> dynamicDims;
 };
 
+constexpr int64_t kPoolAlignment = 256;
+
 int64_t staticByteFactor(MemRefType memTy) {
   int64_t factor = memTy.getElementTypeBitWidth() / 8;
   for (int64_t dim : memTy.getShape()) {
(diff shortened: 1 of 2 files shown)


wcy123 (approved):
LGTM

PR #673, opened 2026-08-06
feat(hipsr): emit the pool acquisition

zz002:
## Summary

`hipsr-pool-alloc` now sums the per-group aligned sizes it already emits and hands the total to `hipsr.get_pool`, tagged with the `domain_id` its `pool_domain` declares.

## Related issue or design

wcy123/onnx-hipdnn-ep#114

## Why

- Group sizes are summed, not maxed: groups are disjoint in time within a domain, but they still need disjoint storage inside one pool, so the pool has to hold all of them at once. The `arith.maxui` inside a group covers members that reuse the same bytes.
- `domain_id` belongs to the domain, not to the pass. The runtime keys pools by id, so the id decides which domains share a buffer — that is the author's call, and a pass counting its way through a walk would tie it to traversal order and to how many domains were skipped. `pool_domain` declares it, `hipsr-pool-alloc` passes it through, and duplicates are allowed because repeating an id is how two domains ask for one pool.
- Every `hipsr.get_pool` prints its `domain_id`. It used to default to 0, so single-domain IR read as if the op had no pool selector at all, and which pool a call draws from is not something to infer from an absence.
- Single-group domains emit no `arith.addi` — the pool size is that group's size.
- A poolable alloc no longer implies a usable context: `linalg.fill` is a DPS write that takes no context, so a `pool_domain` without a context block argument can reach the emission point with a group to place and nothing to pass to `hipsr.get_pool`. That is a diagnostic now rather than a crash.

## What

- `Hipsr_PoolDomainOp` gains a required `I64Attr:$domain_id`, and `Hipsr_GetPoolOp`'s `domain_id` drops its `DefaultValuedAttr` wrapper. `assemblyFormat` is untouched on both — the existing `attr-dict` prints them, so a domain reads `} -> tensor<?x512xf16> {domain_id = 0 : i64}` and a pool reads `hipsr.get_pool(%ctx, %size) {domain_id = 0 : i64}`.
- `findContext` / `emitPool` in `HipsrPoolAllocPass.cpp`; `runOnOperation` collects the `emitGroupSize` results it previously discarded and reads `domain.getDomainId()` for the pool it emits.
- `materializeDomains` in `PartitionPoolDomains.cpp` numbers its domains by position. It is the only `PoolDomainOp` builder call in tree and is currently unreached, so this is the minimal change that keeps it compiling.
- Being required, both attributes are now part of the parse, so every hand-written `pool_domain` and `hipsr.get_pool` in the LIT suite carries one: the files that run today, plus the `UNSUPPORTED: true` ones so re-enabling them does not hit a parse error.
- `hip.get_pool` in the `hip` dialect keeps its default — this PR does not touch the `hip-pool-allocs` path.
- The pool still has no consumer; wcy123/onnx-hipdnn-ep#116 rewrites the allocs to `memref.view` over it. `GetPoolOp` carries `MemoryEffects::Allocate`, so DCE keeps it alive until then.
- `pool_alloc.mlir` and `pool_alloc_report.mlir` drop `--implicit-check-not=hipsr.get_pool` from their RUN lines; the `memref.view` guard stays for #116.

## Test plan

- [x] `pool_alloc.mlir`: the five existing cases check `hipsr.get_pool` at the end of their size chains, and `@split_two_groups` checks the `arith.addi` that sums the two groups.
- [x] New `@split_two_groups_dynamic` covers the dynamic batch dim requested on #671: the `memref.dim` factor flows through one group's align chain into the pool-size sum.
- [x] New `@two_domains` declares ids 0 and 7, so the CHECK for `domain_id = 7 : i64` fails against any pass that numbers domains itself, and the CHECK for `domain_id = 0 : i64` on the other pool fails against a printer that elides a zero.
- [x] Full `check-hip-mlir-lit`: 353 passed, 24 unsupported, 1 failure (`Conversion/onnx-to-hip/test_gather_block_quantized.mlir`) that reproduces on `main` without this change.

## Notes for reviewers

Stacked on #671. Offsets (#115) and the `memref.view` rewrite (#116) keep extending the same CHECK blocks in `pool_alloc.mlir`.

The context diagnostic has no LIT case: `-hipsr-pool-alloc` runs after bufferization, where a `pool_domain` always carries the context argument, so the case had to be hand-built. Say the word if you want it back.

### DIFF
--- a/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
+++ b/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
@@ -218,11 +218,34 @@ Value emitGroupSize(OpBuilder &builder, Location loc,
   return arith::MulIOp::create(builder, loc, divided, alignVal);
 }
 
+Value findContext(Block &body) {
+  for (BlockArgument arg : body.getArguments()) {
+    if (isa<ContextType>(arg.getType())) {
+      return arg;
+    }
+  }
+  return nullptr;
+}
+
+Value emitPool(OpBuilder &builder, Location loc, Value ctx,
+               llvm::ArrayRef<Value> groupSizes, int64_t domainId) {
+  Value poolSize = groupSizes.front();
+  for (Value groupSize : groupSizes.drop_front()) {
+    poolSize = arith::AddIOp::create(builder, loc, poolSize, groupSize);
+  }
+  auto deviceSpace =
+      MemorySpaceAttr::get(builder.getContext(), MemorySpaceKind::Device);
+  auto poolType = MemRefType::get({ShapedType::kDynamic}, builder.getI8Type(),
+                                  MemRefLayoutAttrInterface{}, deviceSpace);
+  return GetPoolOp::create(builder, loc, poolType, ctx, poolSize, domainId);
+}
+
 struct HipsrPoolAllocPass : impl::HipsrPoolAllocPassBase<HipsrPoolAllocPass> {
   using impl::HipsrPoolAllocPassBase<
       HipsrPoolAllocPass>::HipsrPoolAllocPassBase;
 
   void runOnOperation() override {
+    int64_t domainId = 0;
     getOperation().walk([&](PoolDomainOp domain) {
       Block &body = domain.getBody().front();
       llvm::DenseMap<Value, Lifetime> lifetimes = computeLiveness(body);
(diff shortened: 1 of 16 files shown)


lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
```diff
@@ -218,11 +218,34 @@ Value emitGroupSize(OpBuilder &builder, Location loc,
   return arith::MulIOp::create(builder, loc, divided, alignVal);
 }
 
+Value findContext(Block &body) {
```
wcy123, line 221: findCtxArgument

wcy123 (approved):
LGTM

PR #675, opened 2026-08-06
perf(gqa): halve decode KV traffic and unify the dispatch path (up to 1.9x fp16, 2.3x INT8)

ZhanxingPu:
## Summary

Routes every fp16 and INT8 GQA decode geometry through one autotuned split-K
path, fixes a cache-line-width bug in its WMMA producer, and moves INT8
dequantization out of the KV loop. On gfx1151 (Radeon 8060S) decode is
1.3x-1.9x faster on fp16 and up to 2.3x on INT8, with no shape regressing.
Prefill is untouched, and the file ends up 63 lines shorter than main.

## Related issue or design

None. The work started from a report that gpt-oss-20b decode throughput
degraded as sequence length grew.

## Why

Sliding-window and attention-sink layers were rejected by the fused predicate
and fell back to a fixed 8-split legacy kernel, so gpt-oss-20b (24 layers,
all with a per-head sink, alternating full attention and a 128-key window)
never reached the tuned decode at all. That path also cannot use an INT8 KV
cache.

The WMMA producer then had a load-width bug that only hardware counters
expose. It prefetched one KV element per lane, so a wave's load covered 64 B
while a gfx1151 cache line is 128 B and a D=64 fp16 KV row is exactly one
line. Every line was requested in two halves, and with 384 waves each
streaming their own slice through a 2 MB L2 the second half missed again.
RGP measured 136.6 MB of DRAM traffic for a 67.1 MB working set, a 2.04x
over-fetch, at a 20.6% L0 hit rate. Reading 16 B per lane makes one
instruction cover four whole lines and brings traffic to 68.2 MB, 1.02x of
theoretical, with the L0 hit rate at 78%.

Measured achievable read bandwidth for this access pattern is 222 GB/s, not
the 256 GB/s paper figure, so fp16 decode now runs at about 90% of the
practical roofline.

INT8 was separately paying for dequantization inside the KV loop, scaling
every staged K and V element. The scalar decode already avoided that.

## What

- Unify decode on `hip_gqa_flash_decode`, now serving window, head sink,
  smooth softmax, MHA and GQA, fp16 and INT8. The legacy split-K decode
  kernels are removed.
- Widen the WMMA KV prefetch to 16 B per lane. Folding 32 scalar loads into
  4 vector loads also removes the address-register pressure that was
  spilling: 256 -> 170 VGPR and 136 B -> 0 B of scratch per lane.
- Autotune the WMMA KV tile height. BKV=32 halves the barriers per KV element
  but wastes more of the last tile, so it is offered only when a split spans
  at least 512 keys; below that the candidate set is unchanged, which keeps
  short-range tuning decisions identical.
- Move INT8 dequantization out of the KV loop, matching the scalar decode:
  the per-channel K scale folds into Q, and the per-channel V scale defers to
  the epilogue where O has accumulated in fp32. Accuracy improves as a side
  effect, relative L2 against the INT8 reference falling from 3.5e-04 to
  3.0e-04.
- Size the split count and candidate set from the device rather than fixing
  them, so the choice scales to wider parts.
- Share one autotune skeleton and one launcher between the two KV formats.
  They were near-identical copies, and the dispatch ladders that pick a
  kernel from (d, hpg, bkv) were written twice, so a newly supported geometry
  could end up reachable in one format and not the other.
- Replace the report documents in the GQA test directories with READMEs
  covering how to build and run each program, and fold the standalone decode
  probes into `test_gqa_decode` behind `--prod-only` and `--only`.

## Test plan

Hardware: AMD Radeon 8060S (Strix Halo, gfx1151, RDNA 3.5), ROCm 7.1,
`-O3 --offload-arch=gfx1151`, B=1.

Correctness:

- `test_gqa_decode` (fp16; MHA and GQA, HpG 1/2/3/4/5/8/16, D 64/128/256,
  len 512..32768, window/sink/smooth variants): ALL PASS, relative L2 versus
  the fp32 CPU reference 2.7e-04 to 3.5e-04.
- `test_gqa_decode_i8` (INT8 KV, 44 shapes): ALL PASS, relative L2 versus the
  INT8 reference 2.1e-04 to 3.0e-04.
- `test_gqa_prefill` and `test_gqa_prefill_i8`: ALL PASS.

Performance is measured two ways because one instrument does not cover both
scales. Steady-state wall clock is 500 calls enqueued back to back between
two events, five order-balanced process pairs, median, one model per process.
Microsecond-scale shapes are read from RGP hardware timestamps instead, since
wall clock there is dominated by launch overhead: the sliding-window shapes
read anywhere from 0.94x to 2.5x across sessions, while RGP repeats within
0.1 us.

fp16, steady-state wall clock:

| Shape | len | main (ms) | this PR (ms) | Speedup |
|---|--:|--:|--:|--:|
| gpt-oss-20b full, HpG8 D64, sink | 512 | 0.0064 | 0.0034 | 1.88x |
| gpt-oss-20b full | 2048 | 0.0269 | 0.0149 | 1.81x |
| gpt-oss-20b full | 8192 | 0.0738 | 0.0542 | 1.36x |
| gpt-oss-20b full | 32768 | 0.5369 | 0.3201 | 1.68x |
| llama-3.2-1b, HpG4 D64 | 32768 | 0.4263 | 0.3257 | 1.31x |
| llama-3.1-8b, HpG4 D128 | 32768 | 0.7002 | 0.6400 | 1.09x |

gpt-oss-20b sliding window, RGP exact kernel time (window 128, splits 8,
three runs each):

| | main | this PR |
|---|--:|--:|
| Split kernel | 8.5 us | 4.6 us |
| Reduce kernel | 3.6 us | 2.4 us |
| Total | 11.5 us | 7.0 us (1.64x) |

INT8 KV, steady-state wall clock, 44 shapes:

| Shape | len | main (ms) | this PR (ms) | Speedup |
|---|--:|--:|--:|--:|
| gpt-oss-20b, HpG8 D64 | 512 | 0.0167 | 0.0074 | 2.26x |
| gpt-oss-20b | 32768 | 0.3368 | 0.2171 | 1.55x |
| llama-3.2-1b, HpG4 D64 | 512 | 0.0078 | 0.0036 | 2.17x |
| mha-h20-d64, HpG1 | 32768 | 0.6810 | 0.5145 | 1.32x |
| mha-h20-d128, HpG1 | 32768 | 1.0659 | 0.8252 | 1.29x |

38 of the 44 INT8 shapes improve; the remaining six land at 0.976x-0.999x on
the scalar INT8 path, whose kernels are byte-identical to main, and a
previous run had a different six, so that spread is noise.

## Notes for reviewers

Change scope is verifiable rather than argued. Hashing each kernel's
instruction stream in the device assembly against main:

| Kernel family | Total | Changed | Added | Removed |
|---|--:|--:|--:|--:|
| `gqa_flash_decode_wmma_kernel` | 13 | 5 | 4 | 0 |
| `gqa_flash_decode_kernel` (scalar) | 45 | 0 | 0 | 0 |
| `gqa_flash_decode_reduce_kernel` | 5 | 0 | 0 | 0 |
| Prefill (v5 / v7 / v8) | 54 | 0 | 0 | 0 |
| `legacy_*` decode | 11 | 0 | 0 | 8 |

125 of main's 138 kernels are byte-identical here, including every prefill
kernel, so prefill cannot be affected. The D=128 fp16 gains come from
configuration selection alone, since those kernels did not change.

The vectorised load was additionally checked for bit exactness: under the
same forced configuration, fp16 output matches main byte for byte on every
shape tried, and so does the scalar path. Only the two configurations that
change arithmetic differ, both at rounding level.

The last two commits are pure refactors. All 134 kernels hash identically
across them, the autotuner walks the same candidate sequence, and the
configurations it settles on vary no more than the pre-refactor binary
varies against itself when run twice.

Two things worth knowing when re-measuring:

- Adding or removing kernels shifts code-object layout and moves the measured
  time of untouched kernels by roughly 1 us. Treat sub-microsecond
  differences on unchanged paths as layout noise.
- Order balancing removes bias between processes but not between shapes
  inside one process, which is what `--only` is for. Without it a D=128 shape
  appeared to regress by a flat 1.4 us purely because the D=64 shape ahead of
  it got faster.

Follow-ups, not in this PR:

- The INT8 path runs at 72% of achievable bandwidth against 90% for fp16. The
  remaining gap is likely the per-element int8-to-fp16 conversion.
- The three prefill autotuners share a skeleton and could be merged the same
  way the decode ones were.
- A single-dispatch fused kernel for sliding-window layers was prototyped and
  rejected: it is 1.4x-10x slower because the grid collapses to B*G
  workgroups and loses both machine occupancy and the barrier-free
  single-wave blocks that split-K enjoys.

AI assistance: the investigation, profiling harnesses, and this description
were prepared with Cursor. All kernel changes were reviewed by hand and
validated with the tests and measurements above.

## Checklist


ZhanxingPu:
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F121%2F&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_oga%2F1710%2F&curr_job=&curr=

amd-mingw (approved):

PR #678, opened 2026-08-07
refactor(hipsr): drop DPS from hipsr.compute

amd-genmingz:
## Summary
`hipsr.compute` is no longer a destination-passing-style op. An `outs` entry
now only supplies the destination buffer for the result in the same position,
and the two types are allowed to differ, so a compute body can reshape its
data without a round trip back into a same-shaped destination.
Developer-visible effects: the operand renames from `inits` to `outputs`,
`PlaceholderOp::getDpsConsumer` becomes `getConsumer`, and code that reached
hipsr destinations through `DestinationStyleOpInterface` must now use the two
shared helpers in `HipsrOps.h`. Existing hipsr DPS ops are unaffected.
bufferization will in next pull requests

## Why

`hipsr.compute` groups the ops that implement one high-level ONNX computation
so the group still looks like a single hipsr op. Being a DPS op worked against
that: `DestinationStyleOpInterface` requires `type(outs[i]) == type(result[i])`,
so a body could never change shape. A `tensor.collapse_shape` had to be paired
with a `tensor.insert_slice` back into a destination of the original shape,
which adds an op that exists only to satisfy the interface.
Dropping DPS makes `outs[i]` purely the destination buffer for `result[i]`,
and lets the two types differ.

## What

**Op definition.** Removed `DestinationStyleOpInterface`; renamed the operand
`$inits` to `$outputs`. Added `RecursivelySpeculatable`, which the DPS
interface used to provide, and `hasVerifier`. `hipsr.compute_yield` now
documents that it gives compute its result types, which need not match the
`outs` types.
**Verifier.** Only the result-to-output pairing is checked, because the rest is
already covered: `SizedRegion<1>` and `SingleBlock` enforce the single block,
`SingleBlockImplicitTerminator` the terminator, and `RegionBranchOpInterface`
both the operand-to-argument entry edge and the yield-to-result exit edge. That
leaves one rule: a non-empty result list must match the `outs` list in size,
since a result without a destination has nowhere to be written. The empty case
is the bufferized form, which writes into memref `outs` and has no results.
**Shared destination accessors.** Callers can no longer reach the destinations
through `DestinationStyleOpInterface`, so `HipsrOps.h` gains two helpers
covering both compute and the hipsr DPS ops:
- `getHipsrDestinationOperands(Operation *)`
- `isHipsrDestinationOperand(OpOperand &)`
Updated to use them:
- `PlaceholderOp::verify` and `PlaceholderOp::getConsumer`.
- `HipsrPoolAllocPass` liveness — a fix, not a rename. The old code only
  started a buffer's live range at a `DestinationStyleOpInterface` use, so a
  compute writing into a pool buffer no longer registered as a write and the
  range started too late.
- `OnnxToHipsr::finalizePlaceholderDependencies` — the walk covered DPS ops
  only, so a compute result feeding a downstream placeholder was never rewired
  into the shape graph.
**Placeholder.** The type equality check against the consumer's result is now
conditional on the consumer being DPS; for compute the destination type is
deliberately allowed to differ. The structural rules are unchanged: every
result fills exactly one `outs` slot, and all results fill slots of the same
op.

## AI assistance
This feature is achieved in collaboration with the cursor.
Made-with: Cursor Opus5


lib/Dialect/Hipsr/IR/HipsrOps.cpp
```diff
@@ -5,8 +5,46 @@
...
+  }
+  if (auto computeOp = dyn_cast<ComputeOp>(owner)) {
+    // use is in Outputs
+    size_t firstOutput = 1 + computeOp.getInputs().size();
+    size_t operandNumber = use.getOperandNumber();
+    return operandNumber >= firstOutput &&
+           operandNumber < firstOutput + computeOp.getOutputs().size();
+  }
+  if (auto dpsOp = dyn_cast<DestinationStyleOpInterface>(owner)) {
+    return dpsOp.isDpsInit(&use);
```
wcy123:
```c++
 OperandRange destinations = getHipsrDestinationOperands(use.getOwner());
  return llvm::is_contained(destinations, use.get());
```

lib/Dialect/Hipsr/IR/HipsrOps.cpp
```diff
@@ -5,8 +5,46 @@
...
 using namespace mlir;
 using namespace mlir::hipsr;
 
 #define GET_OP_CLASSES
 #include "hip/Dialect/Hipsr/IR/HipsrOps.cpp.inc"
+
+// return Outputs if op is a hipsr.compute op, or the DPS inits if op is a
+// hipsr DPS op.
+OperandRange mlir::hipsr::getHipsrDestinationOperands(Operation *op) {
+  OperandRange none = op->getOperands().take_front(0);
```
wcy123: OperandRange none = OperandRange()

wcy123 (approved):
LGTM

lib/Dialect/Hipsr/IR/HipsrOps.cpp
amd-genmingz: updated

lib/Dialect/Hipsr/IR/HipsrOps.cpp
amd-genmingz: updated: Using the same semantics

PR #679, opened 2026-08-07
refactor(hipsr): remove legacy barrier interfaces

zpye:
## Summary
- Remove the unused `StartBarrierInterface` and `EndBarrierInterface` definitions and build wiring.
- Detach `hipsr.expand` from the legacy start-barrier API.
- Keep placeholder-based barrier partitioning unchanged.

Pool-domain partitioning now uses `PlaceholderType::Barrier`, so the legacy interfaces are no longer needed.

## AI assistance
Cursor assisted with implementation and validation.


wcy123 (approved):
LGTM

PR #683, opened 2026-08-10
feat(hipsr): assign pool offsets per group

zz002:
## Summary

`hipsr-pool-alloc` now assigns each group a byte offset within the pool: the first group starts at a constant 0 and every later one starts at the running sum of the preceding group sizes.

## Related issue or design

wcy123/onnx-hipdnn-ep#115

## Why

- The offsets and the pool size are two points on one accumulation. Offsets are the exclusive prefix sums of the group sizes, the pool size is their total, so a single running chain yields both: each iteration records the accumulator as the next group's offset before adding the group size to it. Emitting them separately re-derived the same partial sums and opened with an add of zero.
- No alignment step of its own. #671 already rounds every group size up to the pool alignment, so summing aligned sizes yields aligned starts and consecutive groups land in non-overlapping intervals by construction.
- The offsets have no consumer yet. wcy123/onnx-hipdnn-ep#116 rewrites the allocs to `memref.view` over the pool and takes them then.

## What

- `emitPoolLayout` in `HipsrPoolAllocPass.cpp` builds that chain and returns the per-group offsets alongside the pool size. `emitPool` now takes the size as a `Value` instead of summing the group vector itself.
- Because the offsets are the chain's own partial sums, the only op this adds to the emitted IR is the `arith.constant 0` that starts group 0. The chain itself becomes observable in #116, where `memref.view %pool[%off]` binds each offset to the partial sum it reuses.
- `pool_alloc.mlir`: the existing cases expect that constant ahead of `hipsr.get_pool`, and `@two_domains` gets one per domain.
- New `@split_three_groups` covers a K3 interference clique, the smallest topology with more than one link in the chain.

## Test plan

- [x] `check-hip-mlir-lit`: 357 passed, 21 unsupported, and no failure that does not already reproduce on `main` without this change.
- [x] The three `pool_alloc*.mlir` files were confirmed green on `main` before this branch, since #674/#677/#679/#680 landed on top of #673.
- [x] `pre-commit run --all-files` converges in one pass.

## Notes for reviewers

Stacked on #673. #116 will extend these same CHECK blocks again and drop `--implicit-check-not=memref.view` from the RUN line.

## Checklist

### DIFF
--- a/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
+++ b/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
@@ -21,6 +21,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <optional>
+#include <utility>
 
 namespace mlir {
 namespace hipsr {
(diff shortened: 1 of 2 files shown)


wcy123 (dismissed):
LGTM

wcy123 (approved):
much better than I thought

PR #684, opened 2026-08-10
feat(hipsr): replace pooled allocs with views

zz002:
## Summary

Every poolable `memref.alloc` in a `hipsr.pool_domain` is now rewritten into a `memref.view` into the pool at its group's offset. This is the last piece of the transform: the group sizes, `hipsr.get_pool` and offsets emitted by the earlier patches in this series finally have a consumer.

## Related issue or design

Related issue: wcy123/onnx-hipdnn-ep#116

## Why

All members of a group share a single offset. Grouping already guarantees their lifetimes do not overlap, so they take turns on the same bytes — that reuse is the point of the pass.

Views are built at the pass insertion point, right after the pool, rather than in place at each alloc. The pool is anchored after the last poolable alloc, so a view built in place would precede the pool it reads.

## What

- `replaceAllocsWithViews` creates one `memref.view` per group member (original memref type, dynamic sizes forwarded from the alloc), replaces the alloc's uses, and erases it.
- `pool_alloc.mlir` drops `--implicit-check-not=memref.view` and each case now checks the views and the data ops that consume them, plus `CHECK-NOT: memref.alloc`. `@dead_alloc_skipped` is the exception: its alloc has no DPS write, is not pooled, and stays in place.
- `pool_alloc.mlir` also gains the three cases it was still missing relative to the reference fixture. `@coalesce_dynamic` and `@coalesce_mixed` previously had only remark coverage in the report test; `@split_with_coalesced_group` was not covered anywhere and is the widest topology in the set — three allocs of mixed element type and shape coalesced into one group folded by two `arith.maxui`, plus a fourth that overlaps all three and takes its own group.
- `pool_alloc_report.mlir` drops `@interleaved_allocs`. Allocations interleaved with data ops cannot be pooled under this anchor; `@hoisted_allocs` already covers the same lifetime topology, including a third alloc reusing the first group.

## Test plan

- [x] `check-hip-mlir-lit` green: 358 passed, and the single failure (`Conversion/onnx-to-hip/test_gather_block_quantized.mlir`) is unchanged from `main`
- [x] `pool_alloc.mlir`, `pool_alloc_report.mlir` and `pool_alloc_invalid.mlir` pass
- [x] Transformed report IR round-trips through `hip-mlir-opt --verify-each`, so views dominate their users — including the nested-region cases `@nested_region_user` and `@alloc_inside_region`
- [x] `pre-commit run --all-files` converges

## Notes for reviewers

The offsets consumed here are the ones from #683: the first group takes the constant 0 and later groups reuse the running pool-size sum, so a group's view is offset by the total size of the groups before it.

Next in the series: wiring the pass into the pipeline, which is the first time it sees real input.

## Checklist

### DIFF
--- a/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
+++ b/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
@@ -248,6 +248,18 @@ emitPoolLayout(OpBuilder &builder, Location loc,
   return {offsets, poolSize};
 }
 
+void replaceAllocsWithViews(OpBuilder &builder, llvm::ArrayRef<Value> group,
+                            Value pool, Value offset) {
+  for (Value alloc : group) {
+    auto allocOp = alloc.getDefiningOp<memref::AllocOp>();
+    auto view =
+        memref::ViewOp::create(builder, allocOp.getLoc(), allocOp.getType(),
+                               pool, offset, allocOp.getDynamicSizes());
+    allocOp.getResult().replaceAllUsesWith(view.getResult());
+    allocOp.erase();
+  }
+}
+
 struct HipsrPoolAllocPass : impl::HipsrPoolAllocPassBase<HipsrPoolAllocPass> {
   using impl::HipsrPoolAllocPassBase<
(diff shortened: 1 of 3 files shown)


wcy123 (approved):
LGTM

PR #685, opened 2026-08-10
refactor(hipsr): simplify placeholder input rewiring

zpye:
## Summary

`finalizePlaceholderDependencies` becomes a single walk over `hipsr.placeholder` ops that resolves one placeholder's inputs and writes them back with one `getInputsMutable().assign`. No behavior change for the IR the conversion produces.

## Related issue or design

None.

## Why

The old pass built two module-wide structures that never had to outlive a single placeholder: a `DenseMap` from every result to the outs operand it filled, and a vector of staged `OpOperand *` writes. `getHipsrDestinationOperands` does the same lookup on the spot, and rewiring a placeholder only reads other ops' operands.

It also carried a copy of `PlaceholderOp::verify()`'s diagnostic. The pass manager verifies after every pass, so a bad input still reports the same message at the same location. The resolver now starts from `isAllowedShapeGraphInput`, the verifier's own predicate, so the pass and the verifier cannot disagree about what a valid input is.

Dropping the trailing `destination.getDefiningOp<PlaceholderOp>()` check accepts more cases, not wrong ones. An outs operand that is a block argument or a constant has the same shape as the result and is a legal input. Anything else still fails verification.

The assign is unconditional. Comparing against the current inputs first would only avoid relinking unchanged operands in the use lists, which is O(1) per operand and order-insensitive, so it is not worth its own path.

## What

- `resolveShapeGraphInput(Value)` returns the shape-graph value for one input: an already-legal input comes back unchanged, a data result becomes the outs operand its producer writes into.
- `rewirePlaceholderInputs(ModuleOp)` maps it over each placeholder's inputs and assigns the result. It returns `void`, so `runOnOperation` loses its second failure branch.
- `llvm/ADT/DenseMap.h` and `llvm/ADT/STLExtras.h` are removed; `llvm/ADT/SmallVectorExtras.h` is added for `map_to_vector`.

## Compiler IR

<details>
<summary>Before / after</summary>

```mlir
// Before: the second placeholder's input points at the cast's data result.
%p0    = hipsr.placeholder(%ctx) ins(%in : tensor<?x8xf32>) ... : tensor<?x8xf16>
%cast0 = hipsr.cast(%ctx) ins(%in : tensor<?x8xf32>) outs(%p0 : tensor<?x8xf16>) : tensor<?x8xf16>
%p1    = hipsr.placeholder(%ctx) ins(%cast0 : tensor<?x8xf16>) ... : tensor<?x8xf32>
```

```mlir
// After: it points at the outs operand the cast writes into.
%p1    = hipsr.placeholder(%ctx) ins(%p0 : tensor<?x8xf16>) ... : tensor<?x8xf32>
```

</details>

## Notes for reviewers

`PlaceholderOp::verify()` is unchanged.

No test covers this transform yet, and none can: `--convert-onnx-to-hipsr` still passes an empty `RewritePatternSet` to `applyFullConversion`, and hand-written input whose placeholder points at a data result is rejected by the verifier before the pass runs. The only test that asserted the removed diagnostic is already `UNSUPPORTED: true`, like everything else in that directory.

## AI assistance

Cursor (Claude Opus 5) assisted with the refactor and the comment rewrites. Validated with a local incremental build, the MLIR LIT suite, and `pre-commit run --all-files`.

### DIFF
--- a/lib/Conversion/OnnxToHipsr/OnnxToHipsr.cpp
+++ b/lib/Conversion/OnnxToHipsr/OnnxToHipsr.cpp
@@ -34,7 +34,7 @@ namespace {
 
 // Conversion leaves each placeholder input pointing at a data result. Point it
 // at the outs operand its producer writes into, which has the same shape.
-static SmallVector<Value> resolveShapeGraphInputs(PlaceholderOp placeholder) {
+SmallVector<Value> resolveShapeGraphInputs(PlaceholderOp placeholder) {
   SmallVector<Value> resolved = llvm::to_vector(placeholder.getInputs());
   for (Value &input : resolved) {
     if (PlaceholderOp::isAllowedShapeGraphInput(input)) {
@@ -53,7 +53,7 @@ static SmallVector<Value> resolveShapeGraphInputs(PlaceholderOp placeholder) {
 
 // Placeholder inputs form the shape graph, not the data graph.
 // PlaceholderOp::verify reports any input this cannot fix.
-static void rewirePlaceholderInputs(ModuleOp module) {
+void rewirePlaceholderInputs(ModuleOp module) {
   module.walk([](PlaceholderOp placeholder) {
     SmallVector<Value> resolvedInputs = resolveShapeGraphInputs(placeholder);
     if (!llvm::equal(resolvedInputs, placeholder.getInputs())) {
--- a/lib/Conversion/OnnxToHipsr/OnnxToHipsr.cpp
+++ b/lib/Conversion/OnnxToHipsr/OnnxToHipsr.cpp
@@ -21,8 +21,8 @@
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/Transforms/DialectConversion.h"
 
-#include "llvm/ADT/STLExtras.h"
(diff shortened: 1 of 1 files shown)


wcy123 (dismissed):
LGTM. Two comments on the implementation:

1. [`resolveShapeGraphInputs`](https://github.com/ROCm/hip-ep/blob/928e6aeec775d50e321c29a6552d296c6c33552b/lib/Conversion/OnnxToHipsr/OnnxToHipsr.cpp#L37-L51) has early-continue branches but no curly braces. LLVM coding standards require braces for single-statement if/else when one branch uses early return/continue ([reference](https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return)).

2. [`rewirePlaceholderInputs`](https://github.com/ROCm/hip-ep/blob/928e6aeec775d50e321c29a6552d296c6c33552b/lib/Conversion/OnnxToHipsr/OnnxToHipsr.cpp#L56-L63) checks `!llvm::equal(resolvedInputs, placeholder.getInputs())` before calling `assign`. Is this optimization measurable? If not, simpler to unconditionally assign since it's already O(n).

wcy123 (approved):
LGTM

PR #686, opened 2026-08-10
refactor(hipsr): split the placeholder verifier

zpye:
## Summary

`PlaceholderOp::verify` becomes a list of named rules, one function each, and the LIT cases for those rules come back as a single file next to the op. What the verifier accepts and the messages it prints are unchanged.

## Related issue or design

None.

## Why

The verifier ran four unrelated rules in one body, so the input scan, the result-use bookkeeping and the shape region comparison all shared a scope.

One of the four was dead code. `DestinationStyleOpInterface` already reports a mismatch between an init operand and its tied result, and `hipsr.placeholder` is `IsolatedFromAbove`, so its own verifier runs after the consumer's and the custom diagnostic could never fire.

The tests for these rules were disabled. They sat behind `UNSUPPORTED: true` in two files, one of them outside the `IR` directory, alongside an onnx-to-hipsr case that no longer had a pattern to reach it.

## What

- `verifyShapeGraphInputs`, `verifyResultUses` and `verifyShapeRegionSignature` each hold one rule. `verify` keeps only the result-count precondition.
- Filtering the uses once with `make_filter_range` lets `hasSingleElement` and `all_equal` state the single-use and single-consumer rules directly, replacing the manual `outsUse` pointer and running consumer variable.
- The tied-result-type check and its `DestinationStyleOpInterface.h` include are removed.
- `test/lit/Dialect/Hipsr/IR/placeholder.mlir` covers every diagnostic the verifier emits, including the barrier shape region layout that had no invalid-IR case. The three stale files are deleted.

## Notes for reviewers

Round-trip cases that only restated the printed form are dropped. The one positive case kept is what no other test reaches: constants as shape-graph roots.

## AI assistance

Cursor (Claude Opus 5) assisted with the verifier split and the test consolidation. Validated with a local incremental build, the MLIR LIT suite, and `pre-commit run --all-files`.


wcy123 (approved):
LGTM

PR #691, opened 2026-08-11
test(hipsr): reorganize pool-alloc tests by topology

zz002:
## Summary

Pure test change: the `hipsr-pool-alloc` LIT tests move into `test/lit/Dialect/Hipsr/Transforms/PoolAlloc/` and are reorganized around lifetime topology instead of the old `coalesce_*` / `split_*` naming. No C++ touched.

## Related issue or design

Follows the pass implementation landed in #684 (steps #112-#116). Issue #117.

## Why

The three `pool_alloc*.mlir` files sat directly under `Dialect/Hipsr/`, while every other Hipsr pass keeps its tests in `Transforms/<PassName>/` with its own `invalid.mlir`. The old cases were also organized by chromatic number, which produced a lot of near-duplicates that each exercised one direction of the analysis: `@split_three_groups` is a clean K3, so an implementation that never reuses a slot passes it, and `@coalesce_dynamic` gave both allocs the same `%dim`, so it could not tell "different SSA extents still coalesce" from "same SSA value, so same bucket".

Organizing by topology fixes both. One composite case per topology exercises the split and reuse directions together, and the remaining cases are the ones genuinely orthogonal to topology.

## What

New layout under `Transforms/PoolAlloc/`:

- `line.mlir` — `@line` (dynamic) and `@static_line` (static), each a five-alloc chain `a(i+1) = f(a(i))`. Adjacent lifetimes overlap and non-adjacent ones do not, so the allocs alternate between two groups: `{a0,a2,a4}` at offset 0 and `{a1,a3}` at `G0`. Asserts group count, per-group membership and offsets, the `maxui` chain shape, and `CHECK-NOT: memref.alloc`.
- `diamond.mlir` — `A -> {B, C} -> D`. A, B and C get distinct offsets; D reuses A's slot, so group 0 sizes to `maxui(A, D)`.
- `basic.mlir` — the topology-independent properties: `align_up_rounding`, `dead_alloc_skipped`, `mixed_dtypes` (renamed from `coalesce_mixed`), `two_domains`, and a new `mixed_dynamic_dims` where the two allocs take their dynamic extents from different `memref.dim` values and still coalesce into one group.
- `report.mlir` — moved from `pool_alloc_report.mlir` and trimmed. Keeps the cases only the remark can observe (`hoisted_allocs`, `nested_region_user`, `alloc_inside_region`), keeps `mixed_dtypes` for the static-vs-dynamic size formats, and gains a `diamond` that pins the group numbers the way the transform file pins the offsets. Drops `coalesce_static`, `split_three_groups`, `coalesce_dynamic`, `split_two_groups_dynamic` and `dynamic_size`.
- `invalid.mlir` — moved from `pool_alloc_invalid.mlir`, plus a `@no_alloc` case (a domain with computation but no allocation) that hits the same diagnostic.

`pool_alloc.mlir` is deleted; `line.mlir` and `diamond.mlir` cover it.

## Test plan

- [x] `lit --param=build_mode=Release D:/ROCm/build/hip-ep/test/lit` — 361 passed, 18 unsupported, 1 failed. The failure is `Conversion/onnx-to-hip/test_gather_block_quantized.mlir`, which reproduces on an unmodified tree at the same commit (an `isSignedInteger()` assertion in the ONNX importer) and is unrelated to this change.
- [x] All five `PoolAlloc/` files pass individually.

## Notes for reviewers

Empty domains take `emitError` + `signalPassFailure` rather than the "emitRemark and skip" the issue #117 acceptance criteria describe; bad input should fail early, and `invalid.mlir` guards that behaviour.

Test bodies are written with the `dim` and `alloc` ops hoisted ahead of the computation, which is what `partition-pool-domains` produces. Interleaving them makes the pool's insertion point conflict with the requirement that a view dominate its first user, and the pass reports `operand #3 does not dominate this use`.

## Checklist


wcy123 (approved):
LGTM

PR #692, opened 2026-08-11
feat(hipsr): bufferize hipsr.compute and compute_yield

amd-genmingz:
## Summary
Registers `BufferizableOpInterface` models for `hipsr.compute`,
`hipsr.pool_domain`, and their yield terminators, so One-Shot Bufferize rewrites
both region-carrying ops, their bodies, and their terminators into device
memrefs. Both keep their results: a result buffer is the buffer of the value
yielded in the same position, never the buffer of an `outs` operand.

## Why
Neither op is DPS, so `DstBufferizableOpInterfaceExternalModel` — and the
`DpsBufferizableModel` this branch merges with — does not apply.
For `hipsr.compute` an `outs` operand is only the destination buffer for the
result in the same position, and the two types are allowed to differ; that
freedom is the reason the op exists. So the result buffer cannot come from
`outs`: there is nothing to hand a `tensor<6xf16>` result when the matching
`outs` is `tensor<2x3xf16>`. It comes from the yield operand instead, which is
also where the op's own aliasing chain ends.

`hipsr.pool_domain` has no `outs` at all: every operand is an input and every
result comes from the yield. Both ops bufferize the way `scf.execute_region`
does, with operands and entry block arguments added.

## What
All models live in `lib/Dialect/Hipsr/Transforms/BufferizableOpInterfaceImpl.cpp`,
beside `PreserveShapeBufferizableModel` and the `DpsBufferizableModel` that
bufferizes `hipsr.cast`, and are attached in the one registration function.
- `ComputeOpBufferization`
  - `bufferizesToMemoryRead` and `bufferizesToMemoryWrite` ask the body about
    the matching entry block argument instead of answering unconditionally. An
    `outs` buffer the body never touches then looks neither read nor written,
    which is what lets one value be passed as both `ins` and `outs` without the
    analysis allocating a second buffer for it.
  - `isWritable` returns true only for `outs` block arguments; `ins` stay
    read-only.
  - The aliasing layers tie an operand to its entry block argument, and a
    result to both its yield operand and its `outs` operand.
  - `getBufferType` takes a block argument's type from its operand and a
    result's type from the yield operand. The default would pick whichever
    equivalent alias comes first, which may be the `outs` operand with a
    different type.
  - `bufferize()` builds the new body with `createBlock`, one `to_tensor` per
    tensor argument, and `mergeBlocks`, instead of cloning the region and
    retyping arguments in place, which would leave existing users holding an
    operand that is no longer valid. It does not recurse into the region: the
    driver walks nested ops first, so the body is already memref by then.
- `ComputeYieldOpBufferization` rewrites its operands to their buffers and must
  bufferize in place; yielding out of place would leave an allocation inside the
  body for the parent to hand out as a result.
- `PoolDomainOpBufferization` is the same shape with the `outs` handling
  removed: no operand bufferizes to a write, no block argument is writable, and
  a result aliases only its yield operand. That single equivalent alias is
  enough for `defaultGetBufferType`, so only the block argument case is answered
  by hand.
- `PoolDomainYieldOpBufferization` mirrors `ComputeYieldOpBufferization`.
- ODS: `hipsr.compute`'s `$result_tensors` and `hipsr.compute_yield`'s
  `$operands` widen from `AnyRankedTensor` to `Hipsr_TensorOrDeviceMemRef`, so
  results can carry device memrefs after bufferization. Both descriptions are
  rewritten around that tensors-before, memrefs-after lifecycle.
- Tests
  - `Transforms/OneShotBufferize/compute.mlir` (new): a `pool_domain` body
    holding a `hipsr.cast` chain, a `hipsr.compute` that flattens its input
    while sharing one value between `ins` and `outs`, and the `preserve_shape`
    links; checks the domain allocates twice and emits no `memref.copy`.
  - `Transforms/OneShotBufferize/pool_domain.mlir` (new): a two-step
    `hipsr.cast` chain (f16 to f32 to i32) inside a domain; checks no
    `memref.copy` and no leftover `tensor` or `bufferization.to_*`.
  - Both RUN lines pass `use-encoding-for-memory-space` and every tensor carries
    `#hipsr.mem<device>`. That encoding is how the memory space the hipsr
    operand constraints require reaches the memrefs, including the ones a
    `tensor.empty` init becomes.
  - `IR/compute.mlir` gains `@memref_form_with_result`, a bufferized compute
    whose result type differs from its output, and drops `@host_memref_output`,
    which was only there while the operand constraints were in flux.

## AI assistance
The interface implementations, the lit cases, and this description were drafted
with AI assistance. Every design decision above was reviewed by hand, and the
behavior is pinned by the lit cases listed under Tests.
Made with: Cursor


test/lit/Dialect/Hipsr/Transforms/OneShotBufferize/compute.mlir
```diff
@@ -0,0 +1,216 @@
...
+// CHECK-NEXT: }
+// CHECK-NEXT: %[[SHAPE3:.+]] = scf.execute_region -> !shape.shape {
+// CHECK-NEXT: %[[S3:.+]] = shape.from_extents %[[FLAT_SIZE]] : index
+// CHECK-NEXT: scf.yield %[[S3]] : !shape.shape
+// CHECK-NEXT: }
+// CHECK-NEXT: %[[INIT1:.+]] = memref.alloc(%[[M]]){{.*}} : memref<?x512xf16>
+// CHECK-NEXT: hip.matmul(%[[DHIP]]) ins(%[[IN]], %[[W]] : memref<?x256xf16>, memref<256x512xf16>) outs(%[[INIT1]] : memref<?x512xf16>)
+// CHECK-NEXT: %[[INIT2:.+]] = memref.alloc(%[[M]]){{.*}} : memref<?x512xf16>
+// CHECK-NEXT: hip.add(%[[DHIP]]) ins(%[[INIT1]], %[[B]] : memref<?x512xf16>, memref<?x512xf16>) outs(%[[INIT2]] : memref<?x512xf16>)
+// CHECK-NEXT: %[[FLAT:.+]] = hipsr.compute(%[[DCTX]]) ins(%[[INIT2]] : memref<?x512xf16>) outs(%[[INIT2]] : memref<?x512xf16>) {
```
wcy123: If  `init2` and `flat` have different shapes, it can improve the test coverage.

wcy123 (approved):
LGTM

PR #693, opened 2026-08-11
refactor(hipsr): move bufferization models out of the public header

zpye:
## Summary

The hipsr `BufferizableOpInterface` models move from a public header to `lib/Dialect/Hipsr/Transforms/BufferizableOpInterfaceImpl.cpp`, behind one registration entry point. `hipsr.preserve_shape` is still the only op with a model and its bufferized output is unchanged.

## Related issue or design

None.

## Why

`include/hip/Dialect/Hipsr/IR/HipsrBufferize.h` defined the model inline in a public header that also pulled in `HipsrOps.h` and `BufferizableOpInterface.h`. Every translation unit that included it recompiled the model bodies, and the model name was reachable, so a caller could attach a partial set.

Upstream MLIR keeps these in `<Dialect>/Transforms/BufferizableOpInterfaceImpl.cpp` in all nine dialects that implement the interface.

## What

- The model moves into an anonymous namespace nested in `mlir::hipsr`, so only the registration function can attach it.
- `registerBufferizableOpInterfaceExternalModels` replaces `registerHipsrBufferizableOpInterfaceModels` and is declared in a header with no includes.
- `bufferize` uses upstream's `replaceOpWithNewBufferizedOp` instead of hand-rolling create-then-erase.
- The early exit for an already-memref `$data` is removed. `Bufferize.cpp` gates every `bufferize` call on `hasTensorSemantics`, and `hipsr.preserve_shape` has tensor semantics only when `$data` is a tensor, so the branch was unreachable.
- The LIT test moves to `test/lit/Dialect/Hipsr/Transforms/OneShotBufferize/preserve_shape.mlir` with its contents unchanged. It passes there with untouched RUN and CHECK lines, which is the evidence that the move is behavior-neutral.

## Notes for reviewers

The generic DPS model and the trait-filtered bulk attach are left out on purpose. hipsr DPS ops constrain operands to `Hipsr_TensorOrDeviceMemRef` while stock One-Shot Bufferize produces space-less memrefs, so they cannot be tested until memory space is settled.

`Conversion/onnx-to-hip/test_gather_block_quantized.mlir` fails in the full LIT suite. It is unrelated and pre-existing: it asserts inside `IntegerAttr::getSInt` during `--convert-onnx-to-hip` and reproduces on `main` with this branch stashed.

## AI assistance

Cursor (Claude Opus 5) assisted with the relocation and the cleanup. I reviewed the result and validated it with a local incremental build, the MLIR LIT suite, and `pre-commit run --all-files`.


wcy123 (approved):
LGTM

PR #694, opened 2026-08-11
feat(hipsr): report pool-alloc effect per domain

zz002:
## Summary

`hipsr-pool-alloc` now answers "how much did pooling save" instead of restating the grouping the IR already shows: `emit-pool-report` prints a per-domain summary (reuse ratio, before/after bytes, one row per group) and the pass gains statistics counters plus `LLVM_DEBUG` traces for the analysis. The transform itself is untouched.

## Related issue or design

wcy123/onnx-hipdnn-ep#118 (diagnostics for the rewritten pool-alloc pass).

## Why

The per-alloc remarks were scaffolding from the steps where the pass did not yet change IR, so printing was the only way to observe the analysis. Now that allocs become pool views, `line.mlir` and `diamond.mlir` assert the grouping directly through the view offsets, and `group N` / `size S` per alloc had become a second, weaker copy of that assertion.

What no remark answered is the question the option exists for: how much did pooling actually save, and where is the waste. That is a per-domain and per-group property, so the report is organized that way. The per-alloc detail is still worth having when a result looks wrong, but its audience is whoever is debugging the pass, which is what `LLVM_DEBUG` is for.

Two indicators are deliberately ratios rather than byte totals. Unused bytes per group grow with member count, so a 3-member group wasting 10 KB and a 10-member group wasting 10 KB read the same while the second is the better grouping; `avg unused = 1 - Σsize / (max × n)` normalizes that. Ratios also survive dynamic shapes, where the absolute numbers do not.

## What

- Four `STATISTIC` counters: `NumDomainsProcessed`, `NumAllocationsPooled`, `NumGroupsCreated`, `NumAllocationsReused` (pooled minus groups, i.e. the buffer slots reuse saved). Every number in the report maps onto one of them; the report is per domain, the counters accumulate per module.
- `DEBUG_TYPE` `hipsr-pool-alloc`, matching the pass registration name so one `--debug-only` does not also open the old `hip-pool-allocs` pass. Traces: each alloc's `[start, end]` with the ops selected as first write and last use, the grouping conflict pairs (which member's interval collided, in which group), each group's max with every member's slack against it, and the domain ordinal with the computed insertion point.
- A diagnostics-only size measurement over a set of allocs. Members are comparable when they multiply the same dynamic extents, in which case the common factor cancels out of every ratio; a set mixing extents reports `not comparable (mixed dynamic extents)` rather than inventing an order for symbolic products. Fully static sets report exact bytes, common-factor sets report ratios only since the absolute values depend on runtime extents. Comparability is decided per measured set, so a domain can be incomparable while each of its groups is fine.
- Byte figures exclude the 256-byte pool alignment: they measure what the grouping decision saves, while alignment is a fixed per-group allocator cost independent of who shares a group. Dropping the round-up also keeps the static and dynamic figures on one footing, since `alignUp` is not proportional to the dynamic factor. Reported bytes are therefore below the pool size `hipsr.get_pool` requests.

Report for the five allocations of `report.mlir`'s `@static_line`, sized 8192/4096/6144/2048/512:

```
hipsr-pool-alloc: domain 0: 5 allocs in 2 groups, 3 saved (60%)
hipsr-pool-alloc: domain 0: before vs after: 20992 bytes vs 12288 bytes, 41% saved
hipsr-pool-alloc: domain 0: group 0: 3 allocs, max 8192 bytes, 39% avg unused
hipsr-pool-alloc: domain 0: group 1: 2 allocs, max 4096 bytes, 25% avg unused
```

The three percentages measure different things (slots, bytes, per-use fill gap) and are normally unequal.

## Test plan

- [x] `ninja check-hip-mlir-lit`: 361 passed. The one failure, `Conversion/onnx-to-hip/test_gather_block_quantized.mlir`, reproduces on unmodified `main` in the same build (MLIR assertion `isSignedInteger()` in `BuiltinAttributes.cpp`) and is unrelated to this change.
- [x] `report.mlir` rewritten to assert the summary and group rows across all three size regimes: exact bytes when static, ratios only under a common dynamic factor, `not comparable` when extents differ, plus a domain that is incomparable while its groups are not.
- [x] `line.mlir`, `diamond.mlir`, `basic.mlir` and `invalid.mlir` unchanged, confirming the transform still emits identical IR and diagnostics.
- [x] Statistics and traces verified against a locally patched assertions build of the pass (see notes) and their numbers cross-checked against the report.

## Notes for reviewers

Empty domains still `emitError` and fail the pass, and dead allocs are still silently left in place. Turning either into a remark-and-skip changes behaviour, not diagnostics, so both stay out of this PR.

`STATISTIC` and `LLVM_DEBUG` are no-ops in a Release build, because both are gated on `NDEBUG` in the including translation unit and `CMAKE_BUILD_TYPE=Release` compiles the pass with `/DNDEBUG`. `--stats` and `--debug-only=hipsr-pool-alloc` therefore print nothing in the configuration CI and local builds use, which is also why no LIT test covers them and why none of the 20+ existing passes with these macros has such a test. I validated both layers by temporarily undefining `NDEBUG` for this translation unit and rebuilding. Worth deciding separately whether the tree wants an assertions build in CI, or `LLVM_FORCE_ENABLE_STATS` for the counters; either would make the second layer testable.

## Checklist

### DIFF
--- a/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
+++ b/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
@@ -16,11 +16,23 @@
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/Statistic.h"
+#include "llvm/Support/Debug.h"
 
+#include <algorithm>
 #include <cstddef>
 #include <cstdint>
 #include <optional>
 #include <utility>
+
+#define DEBUG_TYPE "hipsr-pool-alloc"
+#define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE "] ")
+
+STATISTIC(NumDomainsProcessed, "Number of pool domains processed");
+STATISTIC(NumAllocationsPooled, "Number of allocations backed by a pool view");
+STATISTIC(NumGroupsCreated, "Number of pool groups created");
+STATISTIC(NumAllocationsReused,
+          "Number of pool slots saved by reusing a group");
 
 namespace mlir {
 namespace hipsr {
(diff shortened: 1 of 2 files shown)


wcy123 (approved):
LGTM

PR #696, opened 2026-08-12
feat(hipsr): add pipeline scaffolding for the hipsr dialect

zpye:
## Summary

Adds a `Pipelines` component to the hipsr dialect and registers the `hipsr-pipeline` name.

## Why

The hipsr passes can only be named one at a time, so there is nowhere to record the order they must run in. This adds that place, following the layout upstream MLIR uses for `Dialect/Bufferization/Pipelines/`.

`convert-onnx-to-hipsr` is left out on purpose: it marks the `onnx` dialect illegal but calls `applyFullConversion` with no patterns, so it only succeeds on IR that has no ONNX ops.

## What

New:

- `include/hip/Dialect/Hipsr/Pipelines/Pipelines.h` — `HipsrPipelineOptions`, `buildHipsrPipeline`, `registerHipsrPipelines`.
- `lib/Dialect/Hipsr/Pipelines/Pipelines.cpp` — the builder and the `PassPipelineRegistration`.
- `lib/Dialect/Hipsr/Pipelines/CMakeLists.txt` — `HipsrDialectPipelines`.
- `test/lit/Dialect/Hipsr/Pipelines/embedding.mlir` — `--hipsr-pipeline` over a dynamically shaped graph.

Modified, one line each: `add_subdirectory(Pipelines)`, the `registerHipsrPipelines()` call in `registerAllPasses()`, and the library link in `LibHipCompiler` and `hip-mlir-opt`.

## Test

`embedding.mlir` is the ONNX-MLIR the EP emits for a text-embedding graph. Every batch and sequence extent stays symbolic, and `NonZero` makes the scatter index count data-dependent, so a pass that freezes an extent fails the checks.

## AI assistance

Cursor (Claude Opus 5) assisted with the design, the implementation and the test. I reviewed the result and understand it.

### DIFF
--- a/include/hip/Dialect/Hipsr/Pipelines/Pipelines.h
+++ b/include/hip/Dialect/Hipsr/Pipelines/Pipelines.h
@@ -17,12 +17,7 @@ namespace hipsr {
 struct HipsrPipelineOptions : public PassPipelineOptions<HipsrPipelineOptions> {
 };
 
-/// Build the hipsr pipeline into `pm`. Adds no passes yet.
-///
-/// The passes it will add, in order: hipsr-populate-shape-region,
-/// hipsr-partition-pool-domains, hipsr-materialize-init-tensors,
-/// bufferization, hipsr-pool-alloc, hipsr-externalize-constants. See
-/// Dialect/Hipsr/Transforms/Passes.td for the order rules.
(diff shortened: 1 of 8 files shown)


wcy123 (approved):
LGTM.

PR #699, opened 2026-08-12
feat(hipsr): bufferize hipsr.cast through a general DPS model

zpye:
## Summary

`hipsr.cast` now bufferizes, through a reusable `DpsBufferizableModel<OpTy>` that any hipsr destination-passing op can attach. `hipsr.cast` is the only op attached here.

## Related issue or design

None. Follows #693, which moved the hipsr bufferization models into `BufferizableOpInterfaceImpl.cpp` and deferred the generic DPS model.

## Why

`hipsr.cast` was the last hole between two neighbours that each assume the other side of bufferization: `hipsr-materialize-init-tensors` hands over `tensor.empty` inits, and `CastLowering` refuses tensors with "operands must be memrefs (run bufferization first)".

## What

Ported from upstream linalg's `bufferizeDestinationStyleOpInterface`, keeping its split: a file-local generic function that talks only to `DestinationStyleOpInterface`, plus a thin model over `DstBufferizableOpInterfaceExternalModel` that inherits the analysis answers and forwards `bufferize()`.

The only deliberate difference is dropping the region splice, since no `Hipsr_DpsOp` carries a region, which lets the bufferized op be built in one step. `replaceOpWithNewBufferizedOp` does not apply the way it does for the neighbouring `hipsr.preserve_shape` model: the tensor results become the init buffers, which are operands of the new op rather than results of it.

The device memory space comes from the `#hipsr.mem<device>` tensor encoding. `Hipsr_TensorOrDeviceMemRef` rejects any memref whose space is not `#hipsr.mem<device>` and stock One-Shot Bufferize allocates space-less memrefs, so `use-encoding-for-memory-space` carries the encoding onto every memref a tensor becomes and a `tensor.empty` init allocates in device space. No hipsr driver, `defaultMemorySpaceFn`, or post-bufferization rewrite of allocations is needed. Function boundaries use the identity layout map, as the production pipeline does, because HIP runtime calls consume contiguous buffers.

Before and after, from the static case of the new `cast.mlir`:

```mlir
// Before
%init = tensor.empty() : tensor<4x8xf16, #hipsr.mem<device>>
%0 = hipsr.cast(%ctx) ins(%in : tensor<4x8xf32, #hipsr.mem<device>>)
    outs(%init : tensor<4x8xf16, #hipsr.mem<device>>)
    : tensor<4x8xf16, #hipsr.mem<device>>

// After
%alloc = memref.alloc() {alignment = 64 : i64} : memref<4x8xf16, #hipsr.mem<device>>
hipsr.cast(%arg0) ins(%arg1 : memref<4x8xf32, #hipsr.mem<device>>)
    outs(%alloc : memref<4x8xf16, #hipsr.mem<device>>)
```

## Notes for reviewers

`invalid.mlir` pins the contract behind the encoding: a tensor without it bufferizes to a space-less memref that the operand constraint rejects, once on the input path through the function boundary and once on the init path through the allocation. The mixed buffer/tensor case is pinned there too, since the DPS verifier accepts an op the rewrite has no answer for.

The encoding has to be on the tensor types before bufferization runs, so `hipsr-materialize-init-tensors` will have to set it on the inits it creates, and `buildHipsrPipeline`, still an empty scaffold, will have to turn the switch on once it gains a bufferize step.

## AI assistance

Cursor (Claude Opus 5) assisted with the upstream port, the tests, and the memory-space investigation. I reviewed the result and validated it with a local incremental build, the MLIR LIT suite, and `pre-commit run --all-files`.

### DIFF
--- a/test/lit/Dialect/Hipsr/Transforms/OneShotBufferize/cast.mlir
+++ b/test/lit/Dialect/Hipsr/Transforms/OneShotBufferize/cast.mlir
@@ -1,74 +1,62 @@
 // Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
 // Licensed under the MIT License.
 
-//===----------------------------------------------------------------------===//
-// Positive coverage for one-shot-bufferize on hipsr.cast, the first hipsr DPS
-// op with a bufferization model.
-//
-// Both cases hand the pass device buffers through bufferization.to_tensor, so
-// the bufferized cast keeps the #hipsr.mem<device> space its operands require.
-// A tensor.empty init bufferizes to a space-less memref instead, which is an
-// error today and lives in invalid.mlir.
-//
-// Each case spells out its whole function, so the CHECK block reads as the
-// expected output and the absences are load bearing without a CHECK-NOT: an
-// inserted memref.copy, a surviving bufferization.to_tensor, or a result left
-// on the cast all break the CHECK-NEXT chain.
-//===----------------------------------------------------------------------===//
+// The #hipsr.mem<device> tensor encoding is what supplies the memory space that
+// hipsr operands require: use-encoding-for-memory-space carries the encoding of
+// a ranked tensor onto the memref it becomes, so a tensor.empty init allocates
+// in device space. Function boundaries use the identity layout map, as the
+// production pipeline does, because HIP runtime calls consume contiguous
+// buffers.
 
-// RUN: hip-mlir-opt --split-input-file --one-shot-bufferize %s | FileCheck %s
+// RUN: hip-mlir-opt --split-input-file --one-shot-bufferize="bufferize-function-boundaries function-boundary-type-conversion=identity-layout-map use-encoding-for-memory-space" %s | FileCheck %s
 
-// The init buffer is writable, so the cast is rewritten in place: it takes the
-// buffers the to_tensor ops came from and keeps no result.
 // CHECK-LABEL: func.func @cast_static(
 // CHECK-SAME: %[[CTX:.+]]: !hipsr.context,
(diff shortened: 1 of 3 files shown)


wcy123 (dismissed):
LGTM

test/lit/Dialect/Hipsr/Transforms/OneShotBufferize/cast.mlir
```diff
@@ -0,0 +1,74 @@
...
+// CHECK-SAME: %[[IN:.+]]: memref<4x8xf32, #hipsr.mem<device>>,
+// CHECK-SAME: %[[OUT:.+]]: memref<4x8xf16, #hipsr.mem<device>>) {
+// CHECK-NEXT: hipsr.cast(%[[CTX]]) ins(%[[IN]] : memref<4x8xf32, #hipsr.mem<device>>)
+// CHECK-SAME: outs(%[[OUT]] : memref<4x8xf16, #hipsr.mem<device>>)
+// CHECK-NEXT: return
+// CHECK-NEXT: }
+func.func @cast_static(%ctx: !hipsr.context,
+                       %in: memref<4x8xf32, #hipsr.mem<device>>,
+                       %out: memref<4x8xf16, #hipsr.mem<device>>) {
+  %input = bufferization.to_tensor %in restrict
```
wcy123: try alloc_tensor

wcy123 (approved):
LGTM

test/lit/Dialect/Hipsr/Transforms/OneShotBufferize/cast.mlir
zpye: using "encoding" attr in ranked tensor type

PR #700, opened 2026-08-12
fix(onnx-to-hip): take multi-dynamic expand_shape extents from the shape operand (-64% TTFT on Gemma-4 26B-A4B)

BoarQing:
## Summary

`buildExpandShapeOutputShape` derived every dynamic output dimension of a `tensor.expand_shape` from the source dimension it came from. That is correct only while a reassociation group covers at most one dynamic dimension. When a `Reshape` splits one dynamic source dimension into several dynamic output dimensions, the source extent is their *product*, so each of those output dimensions was assigned the whole product.

This makes the split silently wrong for any `[bs*ss, H] -> [bs, ss, H]` reshape, and every consumer inherits the inflated shape. On Gemma-4 26B-A4B it cost 78% of decoder prefill and made 4K prompts unrunnable.

## Related issue or design

None.

## Why

The per-layer `Reshape([bs*ss, 2816] -> [bs, ss, 2816])` in Gemma-4 26B-A4B lowered to

```mlir
%dim_289 = tensor.dim %2525, %c0 : tensor<?x2816xf16>
%dim_290 = tensor.dim %2525, %c0 : tensor<?x2816xf16>   // same dim, twice
%expanded_291 = tensor.expand_shape %2525 [[0, 1], [2]]
    output_shape [%dim_289, %dim_290, 2816] : tensor<?x2816xf16> into tensor<?x?x2816xf16>
```

With `bs = 1` the source dimension already equals `ss`, so the expansion claimed `[ss, ss, 2816]` rather than `[1, ss, 2816]`. The 30 layer input norms consuming it ran over `ss^2` rows.

Two things kept it silent. The wrong shape is numerically equal to the right one at `bs = 1`, which is the batch size everything is tested at. And the norm runtime ABI passes element counts rather than shapes: `simplified_layer_norm` recovers `num_rows` as `input_num_elements / scale_num_elements`, which divides evenly for any `ss`, so an `ss`-times-too-large buffer is indistinguishable from a well-formed larger batch. At 2,049 tokens that was 5,036 ms of the 6,486 ms decoder prefill; at 4,097 tokens the N-squared tensor needed 88 GB against 63.6 GB of shared APU memory, and the host rebooted.

`ReshapeShapeFold` already rewrites `Reshape(_, Shape(x))` into a host-visible `tensor.from_elements` carrying the correct per-dimension values, and its comment said this was so the conversion could consume them — but the consuming branch was never written and the conversion never read operand 1. The fix is to finish that path rather than add a new mechanism.

## What

- `ReshapeConversion.cpp`: `buildExpandShapeOutputShape` now detects reassociation groups with more than one dynamic output dimension and takes those extents from the shape operand, walking `tensor.from_elements` through `arith.index_cast`, `tensor.cast`, and constants. Extents are resolved lazily per dimension, so an unresolvable entry elsewhere in the operand — an ONNX `-1` placeholder at a statically known position, say — no longer sinks the whole expansion.
- Groups without a multi-dynamic split are delegated to upstream `tensor::ExpandShapeOp::inferOutputShape` instead of being re-derived here. Upstream declines the multi-dynamic case by design; the in-tree copy had dropped that bail-out, which is what let this through in the first place.
- The helper returns `std::nullopt` when a multi-dynamic group's extents are not host-readable, and all three call sites (`Reshape` expand, same-rank split/combine, `Unsqueeze`) fall back to `tensor.reshape` rather than emit an `expand_shape` whose `output_shape` cannot be justified.
- `ReshapeShapeFold.cpp`: correct the stale comment that claimed the operand was already consumed and that no change to `ReshapeConversion` was required. The fold is inert without the consuming branch, so the two halves need to describe each other.

No verifier or shape assertion is added. `[ss, ss, 2816]` is a legal shape and only the value is wrong, so nothing downstream can distinguish it; the fix has to be at the point the shape is constructed.

## Test plan

`test/lit/Conversion/onnx-to-hip/test_reshape.mlir`:

- `test_reshape_expand_multi_dyn` pins the `[?, 2816] -> [?, ?, 2816]` split to two *distinct* extents read from the shape operand, at dimension indices 0 and 1. Before the fix both came from index 0.
- `test_reshape_expand_multi_dyn_infer_static_dim` pins the lazy per-dimension resolution: a `-1` placeholder at a static position still permits the expansion.
- `test_reshape_expand_multi_dyn_opaque` pins the `tensor.reshape` fallback when the shape operand is opaque.

388 lit tests pass, 12 unsupported (pre-existing Morphizen plugin tests), 0 failures. The norm e2e tests (`test_simplified_layer_norm_model.mlir`, `test_skip_simplified_layer_norm_model.mlir`, `test_rms_normalization_model.mlir`) pass unchanged. 21 numeric layer-norm tests pass on gfx1151.

Correctness on Gemma-4 26B-A4B FP16W4-qmoe: generated text at 513 and 2,049 tokens matches the pre-fix build, and `HIPDNN_EP_DEBUG` confirms all 181 rank-3 norms report `2049x2816` with no remaining `4198401`.

## Performance

Gemma-4 26B-A4B FP16W4-qmoe, gfx1151 (Strix Halo, 63.6 GB shared), Release build, 2,049-token prompt, median of 5 after warmup.

| Metric | Before | After |
|---|---:|---:|
| TTFT | 7,865 ms | 2,859 ms |
| Decoder prefill | 6,486 ms | 1,520 ms |
| Layernorm within prefill | 5,036 ms | 50 ms |
| Norm rows dispatched | 4,198,401 | 2,049 |

TTFT at other lengths: 1,905 ms at 513, 2,183 ms at 1,025, and 4,983 ms at 4,097. Prefill is now near-linear in prompt length; 4,097 tokens previously exhausted shared memory and rebooted the host.

## Notes for reviewers

- Scope was narrowed after the first revision. An earlier push also added a conversion-time norm axis verifier, divisibility guards in two runtime norm wrappers, and a `hip-to-llvm` norm test. None of them could catch this bug — `4198401 * 2816` is exactly divisible by 2816 — and none are needed for the result above, so they were dropped to keep the change to one mechanism. The numbers were measured on a build that still carried them; they are assertions and preconditions only and do not affect generated code.
- The vision encoder path is unaffected; its 1,446 ms is now the largest fixed component of VLM TTFT.
- Worth confirming the fallback direction is the one you want: an unreadable shape operand on a multi-dynamic group now yields `tensor.reshape`, which is correct but blocks fusion. I found no such case in the models tested, so the branch is exercised only by the lit test.
- Routing single-dynamic groups through upstream `inferOutputShape` is the one change that touches existing expand-shape sites. The lit suite pins their IR and is unchanged, but it is the part worth a second look.

## Checklist

AI assistance: the diagnosis (IR dumps, roofline check confirming the kernel was at 91% of achievable bandwidth and so not the problem), the patch, the tests, and the benchmark runs were produced with Cursor. All measurements above are from real runs on the machine named, and I reviewed the change.

### DIFF
--- a/lib/Conversion/OnnxToHip/ReshapeConversion.cpp
+++ b/lib/Conversion/OnnxToHip/ReshapeConversion.cpp
@@ -51,51 +51,124 @@ validateSqueezeUnsqueezeOp(mlir::Operation *op, mlir::PatternRewriter &rewriter,
   return mlir::success();
 }
 
-/// Build output shape for expand_shape operations.
+/// Count the dynamic output dims a reassociation group covers.
+static int64_t countDynOutDims(mlir::RankedTensorType outputType,
+                               const mlir::ReassociationIndices &group) {
+  return llvm::count_if(
+      group, [&](int64_t idx) { return outputType.isDynamicDim(idx); });
+}
+
+/// Resolve output dim \p outDim's extent from a Reshape's `shape` operand.
+///
+/// Only a host-visible shape vector is honoured, i.e. the
+/// `tensor.from_elements` that ReshapeShapeFold leaves behind for the
+/// `Reshape(_, Shape(x))` idiom, whose elements are already host SSA values. A
+/// shape tensor that is still device-resident yields nullopt instead of paying
+/// for a readback; the caller then falls back to `tensor.reshape`, which
+/// consumes the runtime shape vector directly.
+///
+/// Resolution is per-dim rather than whole-vector because ONNX's `-1` ("infer")
+/// and `0` ("keep the input dim") are not usable as extents but normally sit on
+/// the feature dim -- a position the result type already pins as static, so the
+/// caller never asks about it. Resolving the whole vector up front would let a
+/// `-1` at such a position veto a split that is otherwise fully determined,
+/// which is the usual spelling of `[bs*ss, H] -> [bs, ss, -1]`.
+static std::optional<mlir::OpFoldResult>
+resolveShapeOperandExtent(mlir::PatternRewriter &rewriter, mlir::Location loc,
+                          mlir::Value shapeOperand, int64_t outDim,
+                          int64_t outputRank) {
+  if (!shapeOperand)
+    return std::nullopt;
+  // Shape-refining casts do not change the elements.
+  while (auto castOp = shapeOperand.getDefiningOp<mlir::tensor::CastOp>())
+    shapeOperand = castOp.getSource();
+
+  auto fromElements =
+      shapeOperand.getDefiningOp<mlir::tensor::FromElementsOp>();
+  if (!fromElements ||
+      static_cast<int64_t>(fromElements.getElements().size()) != outputRank)
+    return std::nullopt;
+
+  mlir::Value element = fromElements.getElements()[outDim];
+  // ReshapeShapeFold emits `arith.index_cast %tensor.dim`. Reusing the pre-cast
+  // index keeps the extent on the same SSA value the rest of the conversion
+  // derives dims from, instead of a round trip through i64.
+  if (auto castOp = element.getDefiningOp<mlir::arith::IndexCastOp>();
+      castOp && mlir::isa<mlir::IndexType>(castOp.getIn().getType()))
+    return mlir::OpFoldResult(castOp.getIn());
+  if (std::optional<int64_t> known = mlir::getConstantIntValue(element)) {
+    if (*known <= 0)
+      return std::nullopt;
+    return mlir::OpFoldResult(rewriter.getIndexAttr(*known));
+  }
+  return mlir::OpFoldResult(mlir::arith::IndexCastOp::create(
+      rewriter, loc, rewriter.getIndexType(), element));
+}
+
+/// Build the `output_shape` operand list for a `tensor.expand_shape`.
 /// Used by both Reshape and Unsqueeze when expanding dimensions.
 ///
-/// For static dimensions: use compile-time size from outputType.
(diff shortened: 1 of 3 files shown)


BoarQing:
perf: http://xsjnts.xilinx.com/ROCm/compare?base_url=genaihttp%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F161%2F&base_job=daily_hip_ep_genai&base=161&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F163&curr_job=daily_hip_ep_genai&curr=163(Gemma)
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F232&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F244%2F&curr_job=&curr= (General)

acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/241/(Gemma)
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/242/(General)

BoarQing:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/256/ last verification after trimming 
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/257/

amd-mingw (approved):

PR #701, opened 2026-08-12
feat(gqa): resolve autotune configs from an offline LUT

ZhanxingPu:
## Summary

GQA now resolves its launch configuration from a lookup table packaged with
EPContext instead of benchmarking candidates on the first call for each shape.
Resolution is three tiers: an exact key on the measured shape, a bucket key on
power-of-two length ceilings, then a fixed heuristic. Online tuning remains
available behind `HIPDNN_GQA_AUTOTUNE_MODE=online`.

## Related issue or design

None.

## Why

The online tuner times every candidate the first time it sees a shape. Measured
on gfx1151, that costs 526–1518 kernel launches per shape: 47 s for a 16k
d=128 prefill and 1766 s for a 32k d=256 one. A session hits each cache key a
few dozen times (once per layer), so the cost never amortises — it is pure
added latency before the first token.

The configurations themselves are stable enough to ship: sweeping 372 shapes
showed `prefill_v8` picking the same config 41/41 times, and `prefill_v7`
using only 2 of its 12 candidates.

## What

- `gqa_autotune.{h,cpp}`: policy resolving Exact → Bucket → Heuristic. Missing
  or incompatible LUTs yield an empty policy rather than an init failure.
- `gqa_autotune.fbs` + `lut/gfx1151.{json,fb}`: schema and the measured table
  (66 exact + 58 bucket entries). JSON is the reviewed source; `flatc` produces
  the binary the runtime loads.
- `hip_gqa_flash_decode_v2_configured` / `hip_gqa_flash_prefill_v3_configured`:
  kernel entries taking an explicit config, validated against the same
  candidate lists the tuner uses.
- Everything above moved into `lib/Runtime/Kernels/hip/autotune/gqa/`, which
  previously spanned `schemas/`, `lib/Runtime/real/` and `etc/`.
- `gqa_cost_model.{h,cpp}` + `tools/gqa-autotune-lut`: **offline only.** Scores
  candidates analytically so LUT rows can be generated for geometries nobody
  measured. Not linked into the EP or the kernel library.
- `Kernels/test/example/gqa/autotune/`: the sweep harness that produced the
  measurements, which also records what the cost model would have picked.

## Test plan

Local (gfx1151, hipcc 7.1.51803):

- `gqa_kernel.hip` and `gqa_cost_model.cpp` compile clean under `-Wall -Wextra`.
- Row generator built and run: 190 rows over 3 geometries, valid JSON; its picks
  match the measured winners (`prefill_v8` → `ND4_MT1_BKV32` throughout).
- LUT integrity: `gfx1151.fb` 8436 bytes, `gfx1151.json` 124 entries.

Needed on CI (not runnable locally — flatbuffers is fetched at configure time):

- Full configure + build, and `GqaAutotuneUnitTest`.
- GQA numeric tests with a packaged LUT.
- Perf comparison: LUT-packaged vs `HIPDNN_GQA_AUTOTUNE_MODE=online`.

## Notes for reviewers

**A missing LUT is a performance regression, not a fallback to online tuning.**
`mode` defaults to `Lookup`, so without a packaged table every shape lands on
the tier-3 heuristic. Measured against the per-shape optimum, that heuristic
costs +40.5% mean on `prefill_v7` and +24.5% on `prefill_v8` (120 ms on one 32k
prefill). Benchmarking this branch without `gqa_autotune_lut=<path>/lut/gfx1151.fb`
will therefore look like a regression. Whether a missing LUT should instead fall
back to online tuning is worth deciding before this merges.

Size: 24 files / ~5900 insertions, but 2985 of those are the generated
`gfx1151.json` table and ~1000 are the offline harness and tooling. The runtime
change is the policy plus two kernel entry points.

Bucket coverage is deliberately conservative: sliding-window decode has no
bucket entries, and `max_seq` stays in the exact key because two windowed decode
shapes with the same effective KV length select different winners when their
cache strides differ.

AI assistance: the cost model, the offline tooling, the sweep harness and this
description were prepared with Cursor. The measurements are from real runs on
gfx1151; I reviewed the code and own the claims above.


CaymanYang (commented):
@ZhanxingPu , As we talked yesterday, don't need cost_model cpp/h, what's your concern to keep these files?

ZhanxingPu:
PR701 VS Origin/Main（both on xsjnucstrhalo09）：
[PR701 vs main on xsjnucstrhalo09](http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F351&base_job=daily_hip_ep_genai&base=351&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F374&curr_job=daily_hip_ep_genai&curr=373)

Main->xsjnucstrhalo09 and PR701-->xconucstrhalo19：
[PR701_on_xconucstrhalo19 vs main_on_xsjnucstrhalo09](http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F351&base_job=daily_hip_ep_genai&base=351&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F373&curr_job=daily_hip_ep_genai&curr=374)
It appears that the TPS performance of Llama-3.1-8B-awq-g128-int4-asym-fp16-onnx-dml-pruned_lm_head has decreased, but after retesting, the performance is comparable, so it should be normal fluctuation.
[retest PR701 on xconucstrhalo19 for llama3.1-8b](http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F351&base_job=daily_hip_ep_genai&base=351&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F375&curr_job=daily_hip_ep_genai&curr=374)

qianglin-amd (approved):

PR #660, opened 2026-08-04
feat(hipsr): populate Expand placeholder shapes

zpye:
## Summary
- Populate Expand placeholder shape regions for normal attribute and barrier runtime layouts.
- Register the required pass dialects and add focused IR and population coverage.

## AI assistance
Cursor assisted with implementation and validation.

PR #702, opened 2026-08-12
test(hipsr): move tests into the nested layout

zpye:
## Summary

Moves the eight hipsr LIT tests that still sat flat in `test/lit/Dialect/Hipsr/` into the nested layout, and drops the legacy `test_` filename prefix from five conversion tests. Pure `git mv`; no test content changes.

## Related issue or design

First of a 14-PR series adding the ONNX operations the hipsr embedding pipeline needs. The series merges one PR at a time, in order; see **Notes for reviewers**.

## Why

#660 split the flat `Dialect/Hipsr/expand.mlir` into `IR/expand.mlir` plus `Transforms/PopulateShapeRegion/expand.mlir`, establishing where each kind of test belongs. Eight files never made that move, so the directory mixes both layouts and gives no signal about where a new test should go. The rest of the series adds tests, so settling the layout first keeps them from landing in the old shape.

## What

Into `IR/`, all of them round-trip or verifier runs: `compute.mlir`, `constant.mlir`, `get_pool.mlir`, `register.mlir`, `memory-space-attr.mlir` to `memory_space_attr.mlir`, `pool-domain.mlir` to `pool_domain.mlir`, and `pool-domain-isolation.mlir` to `pool_domain_isolation.mlir`. The last runs `-cse`/`-canonicalize` against the op isolation traits rather than a hipsr pass, so it is an IR test.

`test_externalize_constants.mlir` is the only transform-pass test in the batch and moves to `Transforms/ExternalizeConstants/externalize-constants.mlir`.

Underscores match the existing `IR/` names (`preserve_shape.mlir`, `shape_yield.mlir`, `placeholder_compute.mlir`).

Dropping the `test_` prefix leaves every conversion test named after its op, as `onnx-to-hipsr/cast.mlir` already was: `onnx-to-hipsr/test_constant.mlir` and `hipsr-to-llvm/test_{add,cast,constant,matmul}.mlir`.

No CMake change: lit discovers `.mlir` by suffix and there is no `lit.local.cfg` in these directories.

## Notes for reviewers

`// UNSUPPORTED: true` markers are deliberately left as-is. Un-gating the `onnx-to-hipsr` tests belongs with the next PR, which installs the conversion patterns they exercise.

This is PR 1 of 14. The remaining PRs are implemented and each will be opened against the updated default branch once its predecessor merges. They are sequential rather than parallel because five files (`OnnxToHipsr.cpp`, `OnnxToHipsr.h`, both `CMakeLists.txt`, and `HipsrOps.td`) are edited at the same location by nearly every one of them.

## AI assistance

Cursor (Claude Opus 5) assisted with the layout audit and the moves. I reviewed the result and understand it.

## Checklist


wcy123 (approved):
LGTM - pure file reorganization with no content changes.

PR #628, opened 2026-08-03
refactor(hipsr): enforce strict ONNX conversion legality

zpye:
## Summary
- Replace greedy rewriting with a strict MLIR conversion target for ONNX-to-HipSR.
- Require all ONNX operations to convert, allow scalar `arith.constant` operations at top level, and allow other helper operations inside `hipsr.compute`.
- Preserve placeholder shape-graph finalization and add focused legality tests.

## AI assistance
Cursor assisted with implementation and validation.

PR #703, opened 2026-08-12
fix(hipsr): install the ONNX conversion patterns

zpye:
## Summary

`convert-onnx-to-hipsr` built an empty pattern set, so every ONNX operation reached the strict conversion target unconverted. This adds the four `populate*` calls back and un-gates the four LIT tests that were supposed to catch this.

## Related issue or design

PR 2 of a 14-PR series adding the ONNX operations the hipsr embedding pipeline needs. Stacked on #702; see **Notes for reviewers**.

## Why

`ConvertOnnxToHipsrPass` marks the `onnx` dialect illegal and calls `applyFullConversion` with no patterns, so the pass only succeeded on IR that had no ONNX ops at all. Constant, Cast, MatMul and Expand each have a working pattern file, but nothing ever added them to the set, making all four dead code.

e48bfc6b (#628) dropped the calls when it swapped `applyPatternsGreedily` for `applyFullConversion`. Nothing caught it because #618 had already marked every test in `test/lit/Conversion/onnx-to-hipsr/` with `// UNSUPPORTED: true`, so the whole directory was gated at the time.

## What

`OnnxToHipsr.cpp` calls `populateOnnxToHipsrConstantPatterns`, `populateCastConversionPatterns`, `populateMatMulConversionPatterns` and `populateExpandConversionPatterns` before `applyFullConversion`.

The four tests come off `UNSUPPORTED` and are refreshed against what the conversion actually emits:

- The placeholder attribute is `placeholder_type`, not `type`. Both `cast.mlir` and `matmul.mlir` still expected the older spelling.
- `constant.mlir` gains `--split-input-file`. It already had `// -----` separators, but without the flag they were inert comments and all four cases parsed as one module.
- Expand's six negatives asserted that the ONNX op survived untouched. Under `applyFullConversion` an unmatched op fails the pass instead, so they move to `expand-invalid.mlir` as `failed to legalize operation 'onnx.Expand'` diagnostics, following `conversion-target-invalid.mlir`. That file now covers all seven rejection paths in the pattern: the three shape-operand guards, an unranked input, an element-type mismatch, a bad result rank, and a function whose argument 0 is not `!hipsr.context`.
- `expand.mlir` keeps the positive case and gains one for a shape longer than the input rank, which is the path that raises the output rank.

## Notes for reviewers

Stacked on #702, so the base is `test/hipsr-nested-test-layout` and the diff here is only this change's six files. GitHub retargets this PR to `main` when #702 merges. The dependency is one file: this PR edits `onnx-to-hipsr/constant.mlir`, which #702 creates by renaming `test_constant.mlir`. There is no textual conflict between the two.

`mlir::hip` declares identically named `populateCastConversionPatterns`, `populateMatMulConversionPatterns` and `populateExpandConversionPatterns` for the legacy pipeline. These calls resolve inside `mlir::hipsr` and the file includes only the hipsr header, so they reach the intended overloads.

The remaining 11 gated tests are outside this change: `hipsr-to-llvm` and the `Dialect/Hipsr/IR` files #618 disabled for other reasons.

## AI assistance

Cursor (Claude Opus 5) assisted with the root-cause analysis, the fix and the test refresh. I reviewed the result and understand it.

## Checklist

### DIFF
--- a/lib/Conversion/OnnxToHipsr/OnnxToHipsr.cpp
+++ b/lib/Conversion/OnnxToHipsr/OnnxToHipsr.cpp
@@ -71,6 +71,11 @@ struct ConvertOnnxToHipsrPass
     });
 
     RewritePatternSet patterns(&getContext());
+    populateOnnxToHipsrConstantPatterns(patterns);
+    populateCastConversionPatterns(patterns, &getContext());
+    populateMatMulConversionPatterns(patterns, &getContext());
+    populateExpandConversionPatterns(patterns, &getContext());
+
     if (failed(applyFullConversion(module, target, std::move(patterns)))) {
       signalPassFailure();
       return;
--- a/test/lit/Conversion/onnx-to-hipsr/cast.mlir
+++ b/test/lit/Conversion/onnx-to-hipsr/cast.mlir
@@ -1,8 +1,6 @@
 // Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
 // Licensed under the MIT License.
 
-// UNSUPPORTED: true
-//
 //===----------------------------------------------------------------------===//
 // Converts onnx.Cast to hipsr.cast, leaving the shape region empty (a later
 // pass populates it).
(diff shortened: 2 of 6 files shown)


wcy123 (approved):
LGTM

PR #705, opened 2026-08-13
refactor(hipsr): store constant bytes in DenseResourceElementsAttr

zz002:
## Summary

`hipsr.constant` now stores its bytes in MLIR's `DenseResourceElementsAttr` instead of the custom `#hipsr.file_source` / `#hipsr.mem_source` attributes. The op has a single required `value` attribute in one of two forms: an inline `DenseElementsAttr`, or a resource blob keyed `file|<path>|<offset>` (on-disk weights) / `mem|<address-hex>` (an ORT tensor pointer).

## Related issue or design

None.

## Why

Carrying `value` and `source` as two mutually exclusive optional attributes forced every consumer through a three-way dispatch and forced the verifier to police a value-XOR-source state machine. `DenseResourceElementsAttr` is the upstream mechanism for exactly this — an `ElementsAttr` whose bytes live outside the attribute — so the blob manager owns the pointer/length bookkeeping and the op keeps one attribute with one meaning.

Both external forms are non-owning `UnmanagedAsmResourceBlob`s: the `mem|` blob points at the ORT tensor and the `file|` blob at the dialect's memory-mapped weight file, so building the attribute copies no weight bytes.

## What

- `HipsrConstantOp.td`: `value` becomes a required `ElementsAttr`; `source` is dropped; the op gains `ConstantLike` and a folder. `isInline()` / `hasExternalSource()` are gone.
- `HipsrAttrs.td`: `Hipsr_FileSourceAttr` / `Hipsr_MemSourceAttr` deleted.
- `HipsrOps.h`: `ConstantOp::getDataValues<T>()` deleted — the blob manager already exposes the bytes.
- `ConstantConversion.cpp`: both branches build a key plus an `ArrayRef<char>` view and share one `DenseResourceElementsAttr::get`.
- `ExternalizeConstants.cpp`: dispatches on the attribute kind and, for resources, on the `file|` key prefix; file-backed entries still record path/offset only so the pass never reads the weight file.
- Verifier rejects a splat with more than one element and anything that is neither `dense<...>` nor `dense_resource<...>`.

## Test plan

- [x] `llvm-lit test/lit/Dialect/Hipsr test/lit/Conversion/onnx-to-hipsr test/lit/Conversion/hipsr-to-llvm test/lit/Pipeline` — 49 passed, 8 unsupported (all pre-existing).
- [x] `test-hipsr-externalize-constants` — all checks pass, covering the bytes actually written for both blob kinds.
- [x] `pre-commit run --all-files` — clean.

Coverage, all three files enabled by this PR or by its base:

- `Conversion/onnx-to-hipsr/constant.mlir` runs the conversion end to end now that #703 installed the patterns, so the emitted resource keys are checked against a real weight file the `RUN` line creates.
- `Dialect/Hipsr/IR/constant.mlir` round-trips all four value forms and pins every verifier rejection.
- `Dialect/Hipsr/Transforms/ExternalizeConstants/externalize-constants.mlir` pins the module-scoped cumulative offsets for inline and file-backed constants.

## Notes for reviewers

Rebased onto #703, which installed the ONNX conversion patterns. That is what makes `onnx-to-hipsr/constant.mlir` live coverage for `ConstantConversion.cpp` rather than a gated file; its `RUN` line gains `--split-input-file` from #703 plus `--mlir-elide-resource-strings-if-larger=0` so the blob keys print without their bytes.

The second commit renames that file's `mem_source_const` / `file_source_const` cases, which named the attribute this PR deletes.

## Checklist

### DIFF
--- a/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
+++ b/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
@@ -22,10 +22,6 @@ namespace hipsr {
 #define GEN_PASS_DECL_CONVERTONNXTOHIPSRPASS
 #include "hip/Conversion/Passes.h.inc"
 
-// Populates patterns that convert `onnx.Constant` into `hipsr.constant`
-// (rank>=1 inline / mem_source / file_source) or `arith.constant` (rank-0
-// scalar). Pure IR rewrite -- no file I/O and no size-threshold policy (that
-// is layered on in the externalization pass).
 void populateOnnxToHipsrConstantPatterns(::mlir::RewritePatternSet &patterns);
 
 void populateCastConversionPatterns(::mlir::RewritePatternSet &patterns,
--- a/include/hip/Dialect/Hipsr/IR/HipsrAttrs.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrAttrs.td
@@ -31,24 +31,4 @@ def Hipsr_PlaceholderTypeAttr :
   let assemblyFormat = "`<` $value `>`";
 }
 
-// A constant whose bytes live in an external file: absolute path + byte offset
-// + byte size. Read on demand (streaming) via the dialect's file-map cache.
-def Hipsr_FileSourceAttr : AttrDef<Hipsr_Dialect, "FileSource"> {
-  let mnemonic = "file_source";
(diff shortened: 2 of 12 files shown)


wcy123 (dismissed):
LGTM

wcy123 (approved):

PR #706, opened 2026-08-13
feat(hipsr): legalize onnx.Return and onnx.NoValue

zpye:
## Summary

Handles the two ONNX operations that carry graph structure rather than a computation, so `convert-onnx-to-hipsr` can run on IR as an importer actually emits it.

## Why

The target marks the whole `onnx` dialect illegal, so every ONNX operation needs handling, not just the compute ones. Every imported `func.func` ends in `onnx.Return`, and `onnx.NoValue` stands in wherever the exporter omitted an optional operand. Without both, the pass fails on any real graph before reaching a single compute operation.

## What

`ReturnConversion.cpp` rewrites `onnx.Return` to `func.return`. onnx-mlir does this in its own `StandardFuncReturnPass`, which is not in this pipeline. The rewrite is unconditional because `func.return` carries `HasParent<FuncOp>`, so that trait is what rejects an `onnx.Return` outside a function.

`onnx.NoValue` stays legal through the conversion so each consumer's conversion can drop the operand it stands for. `eraseDeadNoValue` then erases the placeholders left dead.

## Notes for reviewers

`onnx.NoValue` dominates its consumers, and `applyFullConversion` legalizes in dominance order and returns on the first failure. Making the placeholder illegal would therefore abort on it before reaching the consumer that drops the operand, which is why the erasure runs after the conversion instead of as a pattern.

DCE cannot take the sweep's place: an unregistered operation carries no effect information for `isMemoryEffectFree` to work with.

## AI assistance

Cursor (Claude Opus 5) assisted with the implementation and the tests. I reviewed the result and understand it.

### DIFF
--- a/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
+++ b/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
@@ -33,6 +33,9 @@ void populateMatMulConversionPatterns(::mlir::RewritePatternSet &patterns,
 void populateExpandConversionPatterns(::mlir::RewritePatternSet &patterns,
                                       ::mlir::MLIRContext *ctx);
 
+void populateReturnConversionPatterns(::mlir::RewritePatternSet &patterns,
+                                      ::mlir::MLIRContext *ctx);
+
 } // namespace hipsr
 } // namespace mlir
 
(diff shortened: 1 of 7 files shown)


wcy123 (approved):
LGTM

PR #707, opened 2026-08-13
refactor(hipsr): define the memory space with EnumAttr

zpye:
## Summary

`Hipsr_MemorySpaceAttr` is now defined with `EnumAttr` instead of a hand-written `AttrDef`, matching `Hipsr_PlaceholderTypeAttr` beside it in the same file. The generated class is still `mlir::hipsr::MemorySpaceAttr` and the syntax is still `#hipsr.mem<device>`.

## Related issue or design

None.

## Why

`EnumAttr` is upstream's wrapper for an attribute holding one enum value. The old definition hand-rolled the same thing from `AttrDef` plus an explicit `EnumParameter`, spelling out `mnemonic`, `summary`, and `parameters` to arrive at what `EnumAttr` gives by default.

Renaming the enum from `MemorySpaceKind` to `MemorySpace` is what keeps the C++ name unchanged, rather than a cosmetic choice. `EnumAttr` derives from `AttrDef<dialect, enumInfo.className>`, and `AttrDef` sets `cppClassName = name # "Attr"`, so `MemorySpaceKind` would have produced `MemorySpaceKindAttr`. This mirrors upstream `gpu::AddressSpace` / `gpu::AddressSpaceAttr`.

## What

- `HipsrEnums.td`: enum renamed to `MemorySpace`. Cases and their AMDGPU address-space values are unchanged.
- `HipsrAttrs.td`: becomes `EnumAttr<Hipsr_Dialect, Hipsr_MemorySpace, "mem">`, keeping `assemblyFormat` so it still prints `#hipsr.mem<device>` rather than `EnumAttr`'s default bare `$value`.
- `getKind()` becomes `getValue()`; `MemorySpaceKind::Device` becomes `MemorySpace::Device`.
- Two comments claimed the `Hipsr_*MemRef` constraints apply "once ops start using them". Seven op definitions use them today, so both are corrected.

The one externally visible change is the parse diagnostic, `parameter 'kind'` to `parameter 'value'`, because `EnumAttr` names its parameter `$value`. `test/lit/Dialect/Hipsr/IR/memory_space_attr.mlir` is updated to match. That test is gated by `UNSUPPORTED: true` from #618, so it was checked by running `hip-mlir-opt --split-input-file --verify-diagnostics` on it directly, which exits 0.

## AI assistance

Cursor (Claude Opus 5) assisted with the conversion and the comment corrections. I reviewed the result and validated it with a local incremental build, the full MLIR LIT suite (373 passed, 0 failed), and `pre-commit run --all-files`.

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrAttrs.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrAttrs.td
@@ -10,9 +10,8 @@ include "hip/Dialect/Hipsr/IR/HipsrDialect.td"
 include "hip/Dialect/Hipsr/IR/HipsrEnums.td"
 include "mlir/IR/AttrTypeBase.td"
 
-def Hipsr_MemorySpaceAttr : AttrDef<Hipsr_Dialect, "MemorySpace"> {
-  let mnemonic = "mem";
-  let summary = "hipsr memory space attribute";
+def Hipsr_MemorySpaceAttr :
+    EnumAttr<Hipsr_Dialect, Hipsr_MemorySpace, "mem"> {
   let description = [{
     Says where a buffer lives, used as a `memref` memory space:
     `device` (VRAM), `host` (pageable), `pinned` (page-locked host),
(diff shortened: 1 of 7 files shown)


wcy123 (approved):
LGTM

PR #599, opened 2026-07-30
test: update OGA version to supports features of gemma4-unified.

amd-genmingz:
OGA pr2194+pr2286 
and update onnxruntime-extension to pr1091

PR #708, opened 2026-08-13
testing PR700

BoarQing:
Testing PR: the changes from #700 stacked on top of #599 so they can be exercised together.

- Base branch is 	est/gemma4 (head of #599) instead of main.
- Contains the two commits from #700 (`dc6f87c`, `8b68f4c`) cherry-picked; the resulting diff against 	est/gemma4 is byte-for-byte identical to #700's diff against its base.
- No new work here. Review #700 for the actual change.


BoarQing:
Superseded by branch `test/pr700-on-main`, which replays PR700's reshape/norm fix directly on latest main (71c1a681) instead of stacking on test/gemma4.

PR #711, opened 2026-08-13
feat(hipsr): let compute hold buffers in any memory space

zpye:
## Summary

`hipsr.compute` and `hipsr.compute_yield` accepted only device memrefs, so a bufferized compute could not hold a buffer anywhere else. Both now accept a ranked tensor or a memref in any of the four `#hipsr.mem` spaces.

## Why

Not every value a compute produces belongs on the device. An extent vector is read on the host, which is where `hipsr.expand` already takes its `shape` operand from, so a compute that computes extents had nowhere to put them.

The looser constraint deliberately still requires a space to be named. `AnyMemRef` would also accept a memref with no space at all, and `OneShotBufferize/invalid.mlir` shows that rejection is load-bearing: a tensor that reaches bufferization without a `#hipsr.mem` encoding becomes a space-less memref, and the operand constraint is what reports it.

This only widens what verifies. No pass or conversion produces a non-device space yet.

## Notes for reviewers

`isAnySpaceMemRef` backs the new constraint. It and the four per-space predicates it generalizes now share one helper, so a single place decides which types name a hipsr space.

Cursor (Claude Opus 5) assisted with the constraint, the predicate and the tests; validated with the LIT suite. I reviewed the result and own every claim here.

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrComputeOp.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrComputeOp.td
@@ -32,9 +32,10 @@ def Hipsr_ComputeOp : Hipsr_Op<"compute", [
     type need not match that result's, so the body may yield a reshaped or cast
     view of it.
 
-    Operands and results are tensors before bufferization and device memrefs
-    (`#hipsr.mem<device>`) after. A bufferized compute has no results and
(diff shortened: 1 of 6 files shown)


wcy123 (approved):
LGTM.

PR #713, opened 2026-08-14
feat(hipsr): require host/device memory space on hipsr op tensor operands

amd-genmingz:
## Summary
- Tighten the type constraints of computation operations in the hipsr dialect 
  so that they accept only device-resident data. This enables errors involving 
  host-resident operands to be diagnosed much earlier in the compilation pipeline.
- Add `isHostRankedTensor` / `isDeviceRankedTensor` predicates (and a shared
  `tensorInSpace` helper) that read `RankedTensorType::getEncoding()` and match
  a `MemorySpaceAttr`, wired into the new `AnyHostRankedTensor` /
  `AnyDeviceRankedTensor` / `Hipsr_HostTensorOrMemRef` /
  `Hipsr_DeviceTensorOrMemRef` type definitions.
- Update all affected lit tests to annotate tensors according to operational
  context (device for data flowing through compute/cast/matmul/add/expand/
  constant/preserve_shape, host for shape operands), and refresh the
  corresponding `expected-error` / CHECK descriptions.
- OneShotBufferize tests run with `use-encoding-for-memory-space` so the tensor
  encoding is carried onto the resulting memref's memory space; the
  `preserve_shape` case uses `hipsr.cast` (matmul bufferization is not yet
  supported).

## Test plan
- [x] `llvm-lit test/lit/Dialect/Hipsr` passes
- [x] `llvm-lit test/lit/Conversion/onnx-to-hipsr` passes
- [x] Full lit suite shows no regressions

## AI assistance
All the changes have been fully understood.
Made with: Cursor


wcy123 (dismissed):
LGTM

wcy123 (approved):

PR #715, opened 2026-08-14
feat(MIOpen): Use Find API for Convolution

ilango100:
## Summary

Current Conv wrapper calls miopenFindConvFwdAlgorithm, which is expensive, because it looks up the cache entry from disk. Since this is done for every single invocation, it will not be launching the kernels fast enough. And it's not possible to run miopenConvFwd without the FindConvFwdAlgorithm.

The new Find 2.0 API in MIOpen allows for us to find the algorithm and keep a reference to the solution. I've migrated the MIOpen invocation to use the new API, saving the solution in op-slot and reusing it when possible.

## Why

miopenFindConvolutionForwardAlgorithm + miopenConvolutionForward is expensive operation because of loading disk-cache. Example kernel launch and execution profile:

image

And it is not possible to call miopenConvolutionForward without the first Find call.

## What

- Introduce op-slot for ConvOp
- Create ConvState that contains the lookup table for the given parameters
- Use miopenFindSolutions() API to get a solution on the first inference run
- Autotuning enabled always by default

## Performance

Latency (ms)

Model|Before|After
-----|-----:|-----|
DETR (ResNet50 backbone)|23.81|13.38
Mobilenet V2|17.55|2.13
SSD-Mobilenet V1|17.42|3.02
YOLO5 - Large|38.63|12.44
Resnet50 V1.5|17.46|2.95
SSD Resnet34|33.72|24.21

## Checklist

### DIFF
--- a/include/hip/Dialect/IR/HipOps.td
+++ b/include/hip/Dialect/IR/HipOps.td
@@ -623,7 +623,7 @@ def Hip_HipblasltGraphOp : Hip_Op<"hipblaslt.graph", [SingleBlock, NoTerminator]
 
 // ===== MIOpen DNN Operations =================================================
 
-def Hip_ConvOp : Hip_DpsOp<"conv", /*traits=*/[], /*outsAccessor=*/"Output",
+def Hip_ConvOp : Hip_DpsOp<"conv", /*traits=*/[OpStateOpInterface], /*outsAccessor=*/"Output",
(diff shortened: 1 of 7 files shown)


asparkhi (dismissed):
Thanks Ilango. Changes boost the conv speeds. LGTM.

fhanuman (approved):

PR #716, opened 2026-08-14
fix(pool-alloc): extend lifetimes through buffer aliases

zz002:
## Summary

`hipsr-pool-alloc` now follows buffer aliases when computing lifetimes, so two allocations that are simultaneously live through a view are no longer merged into the same pool slot.

## Related issue or design

[wcy123/onnx-hipdnn-ep#136](https://github.com/wcy123/onnx-hipdnn-ep/issues/136) — Step 1 of `doc/pass-09-pool-alloc.md`.

## Why

`computeLiveness` only inspected the direct users of each `memref.alloc` result. Once a buffer is wrapped in `memref.subview` / `memref.cast` / `memref.collapse_shape`, the real reads and writes behind the view are invisible and the lifetime ends early. That is a correctness problem rather than a missed optimization: two allocations that do overlap in time get judged disjoint and end up sharing the same pool bytes.

## What

- Build one `BufferViewFlowAnalysis` per `hipsr.pool_domain` and walk the users of every alias `resolve()` reports. `resolve()` includes the buffer itself, so the previously covered direct users stay covered.
- The DPS first-write probe compares against the alias instead of the original buffer, and `memref.dealloc` users no longer extend a lifetime.
- Link `MLIRBufferizationTransforms`, which is where `BufferViewFlowAnalysis` lives.

The analysis is rooted at the domain rather than the enclosing function: `Hipsr_PoolDomainOp` is `IsolatedFromAbove`, so aliases cannot escape it, and rebuilding per domain keeps the analysis valid across the alloc-to-view rewrite this pass performs on its way out.

Error handling and diagnostics are unchanged. An allocation with no DPS write is still excluded silently, because `memref.copy` writes fall into that bucket and `hip-promote-strided-operands` produces exactly that shape.

## Test plan

- [x] New `test/lit/Dialect/Hipsr/Transforms/PoolAlloc/alias.mlir`: in each of the `subview` / `cast` / `collapse_shape` cases the first buffer is read only through its alias, after the second buffer has been written. All three now produce two pool groups with distinct offsets and an `arith.addi` pool-size accumulation.
- [x] Reverting `computeLiveness` to the direct-users-only walk collapses all three cases into a single group sharing offset 0, which confirms the new tests catch the bug.
- [x] `ninja check-hip-mlir-lit`: 380 of 393 passed. The single failure, `Conversion/onnx-to-hip/test_gather_block_quantized.mlir`, reproduces on `main` without this change (assertion in `BuiltinAttributes.cpp:371`).

## Notes for reviewers

Scope is Step 1 of the design doc only; Steps 2-7 are already in tree and untouched.

## Checklist

### DIFF
--- a/lib/Dialect/Hipsr/Transforms/CMakeLists.txt
+++ b/lib/Dialect/Hipsr/Transforms/CMakeLists.txt
@@ -21,6 +21,7 @@ target_link_libraries(HipsrDialectTransforms PUBLIC
   HipsrDialectIR
   HipConstantsIO  # writeConstantsBinToFileSystem / ConstantEntry
   MLIRBufferizationDialect  # BufferizableOpInterface models
+  MLIRBufferizationTransforms  # BufferViewFlowAnalysis
   MLIRPass
   MLIRTransformUtils
   MLIRFuncDialect
--- a/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
+++ b/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
@@ -120,10 +120,15 @@ computeLiveness(PoolDomainOp domain) {
 
   LLVM_DEBUG(DBGS() << "Computing liveness for " << idx << " operations\n");
 
+  struct FoldState {
+    std::optional<size_t> firstWrite;
+    std::optional<size_t> lastUse;
+  };
+
   auto processUser =
-      [&body, &opIndices](Operation *user, Value alias, memref::AllocOp allocOp,
-                          const FailureOr<std::optional<Lifetime>> &current)
-      -> FailureOr<std::optional<Lifetime>> {
+      [&body, &opIndices](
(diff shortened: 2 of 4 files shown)


zz002:
## The error

The cleanest case is `basic.mlir:136` in `@mixed_dynamic_dims`, where `%a1` has no view, subview, cast or alias of any kind, it is written once and then read once:

```mlir
%a1 = memref.alloc(%d0) : memref<?x512xf16, #hipsr.mem<device>>
%a2 = memref.alloc(%d1) : memref<?x256xf16, #hipsr.mem<device>>
hipsr.add(%dctx) ins(%da, %da : ...) outs(%a1 : ...)   // writes %a1
hipsr.add(%dctx) ins(%a1, %a1 : ...) outs(%da : ...)   // reads %a1
```

```
basic.mlir:136:11: error: buffer used before written - invalid IR (first use is not a DPS destination)
```

wcy123 (approved):
LGTM

PR #719, opened 2026-08-15
perf(gqa): fold the attention bias into the softmax's score read, -11.7% TTFT at 16K

BoarQing:
## What

`add_attention_bias_kernel` was a full read-modify-write pass over the fp32 score buffer -- 17.2 GB per layer at a 16K prompt -- immediately followed by `softmax_f32_to_out_kernel` reading the same buffer again. The bias pass existed only to hand the softmax a value the softmax was about to read anyway.

This adds a `HAS_BIAS` template parameter to the softmax kernel and reads the bias through a `score_at()` accessor, deleting the separate pass. A block owns one query row and the bias runs along the key axis in the same order the softmax reduces, so the fold is one extra contiguous stream. The kernel's three passes re-read it, but a row is 32-64 KB and stays resident, so the cost is one stream from memory against the two the separate pass spent on the scores alone.

Two details that are load-bearing rather than incidental:

**The bias is added before the column max is taken.** Taking the max of unbiased scores and then adding the bias could leave a masked entry (bias `-65504`) as the column max and put every exponent off scale.

**The bias is indexed on its own full extents**, carrying the same four arguments `hip_gqa_add_attention_bias_f32` takes for it (`bias_sq`, `bias_row_offset`, `bias_total_seq`, `bias_col_offset`). The score buffer is a sub-block of the logical `[sq, total_seq]` matrix -- a chunk of query rows since #787 tiled the prefill, and a narrowed key range since #795 recovered the sliding window -- while the bias always describes the whole thing. Folding either offset into the pointer would mis-stride every batch/head plane after the first.

This also inverts the documented order of the causal triangle and the bias, which is safe only because the triangle writes `-INFINITY`: adding any finite bias to it leaves `-INFINITY`, and elsewhere the sum is the same either way. Called out in a comment at the site.

## Related issue or design

Rebased onto `main` and retargeted from `feat/rgp-capture-fence`, which is not going to land. The branch is now one commit on top of `main`.

Worth knowing for review: this is a **port, not a replay**. The original was written against that base branch's head-grouping of the score buffers, which `main` never took -- `main` solved the same allocation problem with #787's query-row chunking and then narrowed the key range in #795. Every line the original commit touched in `gqa.cpp` had changed shape, and the bias indexing genuinely had to be reworked rather than re-applied: the old kernel resolved the bias plane as `cols * rows` and offset it by `col * rows`, which is only correct when the score block *is* the whole matrix. Under chunking and narrowing it is not, so the plane is now `bias_sq * bias_total_seq` with the two offsets applied inside it.

## What is not here

There is **no environment variable**. The fold is unconditional: with a bias it is taken, and with none the softmax runs the instantiation it always did. The previous revision of this PR carried `HIPDNN_EP_GQA_FUSE_BIAS_SOFTMAX=0` to restore the separate pass from one binary, which is measurement scaffolding rather than something a merged kernel should ship.

One consequence to flag: with the switch gone, nothing calls `hip_gqa_add_attention_bias_f32` any more. The entry and its kernel are left in place -- they are an exported `HIP_KERNEL_API` symbol and still correct -- but a reviewer who wants them deleted should say so, since that is a separate change to the kernel library's surface rather than part of this one.

## Measurement

**Re-measured on `main`, on the model's published config.** Two earlier revisions of this section have been superseded. The first carried a 40,716 -> 33,981 ms (-16.6%) result taken on the pre-rebase branch, whose baseline predated #787 and #795. The second was taken on `main` but against a **stale `genai_config.json`** belonging to an older revision of the Gemma-4 export, and it claimed a config workaround was needed to run the model at all; both of those points were wrong and are corrected below. The numbers here are `f1a82124` against `f1a82124` + this commit, differing only by this commit, with both `hipgpu.dll` and `custom_kernels_gfx1151.dll` swapped together, on the export's `genai_config.json` exactly as published.

### TTFT, Gemma-4 26B-A4B 16K VLM prefill, gfx1151

`vlm_benchmark.py`, 16,241 prompt tokens (15,987 text + 254 image), 3 reps + 1 warmup per block, blocks run base/cand then cand/base so drift cannot pass for an arm effect:

| arm | block 1 | block 2 | mean p50 |
|---|---|---|---|
| `main` | 46,647.8 ms | 46,766.4 ms | 46,707.1 ms |
| `main` + this commit | 41,166.8 ms | 41,276.9 ms | 41,221.8 ms |

**-5,485 ms (-11.7%).** Each arm's two blocks land within 119 ms of each other against a 5,485 ms arm separation -- about 46x the largest within-arm spread -- and reversing the block order did not move the result.

### Per-op GPU time at the same operating point

`HIPDNN_EP_PERF` on the decoder graph at `sq=16384`, mean of two warm reps per arm (within-arm spread under 0.15%). These run the ONNX graph directly through `hip-onnx-runner` rather than through genai, so they are unaffected by the config correction above:

| op | `main` | + this commit | delta |
|---|---|---|---|
| `gqa` (30 layers) | 29,857.6 ms | 25,087.7 ms | **-4,769.9 ms (-16.0%)** |
| -- 25 local-attention layers (d=256) | 22,977.1 ms | 19,012.9 ms | -3,964.2 ms |
| -- 5 global layers (d=512) | 6,880.6 ms | 6,074.7 ms | -805.9 ms |
| `qmoe` (control) | 3,948.7 ms | 3,940.9 ms | -0.2% |
| `matmul_nbits` (control) | 1,646.5 ms | 1,627.2 ms | -1.2% |
| total GPU | 36,934.8 ms | 32,112.7 ms | -4,822.1 ms (-13.1%) |

The controls sit inside run-to-run spread while `gqa` moves 16%, and the whole-graph saving is accounted for by `gqa` on its own, so the effect is isolated to the op this commit changes.

### The saving is the whole cost of the deleted pass

The pass moved `2 x 17.2 GB` of fp32 scores plus `0.54 GB` of fp16 bias per layer, or 34.9 GB. Dividing that by the measured per-layer saving gives **220 GB/s** on the local-attention layers and **217 GB/s** on the global ones -- 86% and 85% of this part's 256 GB/s roofline.

Those two groups differ in `head_dim`, KV head count and absolute cost, so their agreeing to within 2% is what identifies the mechanism as "one fewer full stream over the score buffer" rather than something incidental. A pass running that close to bandwidth is almost entirely irreducible traffic, and this commit recovers it in full. It also means the three passes' bias re-reads cost nothing measurable, which is the residency assumption the kernel comment makes.

### On the #795 caveat

An earlier revision warned that #795's window narrowing might have cut the win substantially, on the reasoning that 25 of 30 layers would attend only 1024 keys. A later revision claimed to have disproved that from `skv=16385` on all 30 layers. **Both statements should be disregarded**; the second was measured on a config with the `sliding_window` block deleted, so it could not have observed narrowing either way.

What is actually true is simpler: **this export ships no `sliding_window` block at all**, so #795's narrowing is not active for it, and the measurement above is at full key width on every layer. Whether #795 reduces this commit's win for an export that does configure a window is **unmeasured** and remains an open question for review.

The gap between the original -16.6% and this -11.7% is largely denominator: the absolute saving is 5,485 ms against the original 6,735 ms, while this baseline is 46.7 s rather than 40.7 s -- a different code base (post-#787, post-#795) as well as a different config.

## Verification

Real-model output was **token-for-token identical** between the two arms over a 32-token greedy decode. That is the expected result rather than a lucky one: both paths add the bias to the score in fp32, so they evaluate the same rounding, and the separate pass merely stored the sum to memory in between. This was measured pre-rebase, using the switch, and **has not been re-run on the rebased branch** -- it is the main thing still outstanding here. The 16K runs above completed and generated tokens on both arms, which is a liveness check, not an equivalence check.

`custom_kernels_gfx1151.dll` and `hipgpu.dll` both build clean for gfx1151 with no new warnings, and `pre-commit` passes.

### Retracted: the "config workaround" caveat

A previous revision of this section reported that Gemma-4 could not run through genai on `main` at all, quoting an output-allocator failure on `present.0.key` (`{1,8,1024,256}` vs `{1,8,3264,256}`), attributing it to the EP not honouring `past_present_share_buffer`, stating it reproduced across five commits, and proposing it deserved its own issue.

**That was an artifact of a stale `genai_config.json`, not an EP defect.** The config in use belonged to an older revision of the export and set `past_present_share_buffer: true` with a `sliding_window` block; the published config sets it `false` and ships no window block, and the model runs unmodified. There is no bug here to file, and the earlier claim that the workaround "inflates absolute TTFT, which is why the baseline reads 46.5 s" was also wrong -- re-measuring on the published config moved the baseline by 0.4%, from 46,514.7 ms to 46,707.1 ms.

### RGP capture

Still not possible, for an unrelated reason: `model_benchmark` rejects Gemma-4's mixed KV geometry, and while `hip-onnx-runner` does reach GQA, it runs one inference and exits, so only 1.8 MB of a 142.6 MB trace streamed out before the process died. The bias/softmax split above is therefore byte accounting against op-level timing, not SPM counters.

## What is left after this

`gqa` is still 78% of a 16K prefill. The score buffer costs about 86 GB per layer once the bias pass is gone, and almost 60% of that is the softmax streaming the score matrix three times -- max, then exp+sum, then exp+normalise. Collapsing that to a single online pass is worth roughly 4.7 s, about as much again as this commit wins.

That is out of scope here and not small: two independent gates in `gqa.cpp` force this model onto the decomposed pipeline that materialises scores at all -- an additive attention bias forces it (lines 35-37), and windowed flash prefill is implemented only for `head_dim == 64` (line 25) while Gemma-4 is d=256.

## Notes for reviewers

- New entry `hip_gqa_softmax_f32_to_out_biased` dispatches the `(TOut, TBias)` instantiation from the output and bias dtypes; the existing callers select `HAS_BIAS=false` through an overload and are unchanged.
- Mind the axis naming when checking the indexing: the softmax kernel reduces along `rows`, so its `rows` is the key extent and its `cols` the query count -- the transpose of the names `hip_gqa_add_attention_bias_f32` uses for the same buffer. Both the header and the entry comment say so.
- `bias_element_size_bytes` is passed `elem_sz`, i.e. the GEMM/cache element size, exactly as the bias-add call did. That equates the mask dtype with the GEMM dtype; it is pre-existing behaviour carried over unchanged, not a new assumption.

### DIFF
--- a/lib/Runtime/Kernels/hip/gqa_kernel.hip
+++ b/lib/Runtime/Kernels/hip/gqa_kernel.hip
@@ -1168,16 +1168,65 @@ extern "C" int hip_gqa_add_attention_bias_f32(
 // Output-type-templated: reads fp32 scores, writes TOut ∈ {__half, float}.
 // fp16 output (TOut=__half) feeds the fp16 Value GEMM; fp32 output (TOut=float)
 // feeds the fp32 Value GEMM on the Whisper no_causal fp32 decomposed path.
-template <typename TOut>
+//
+// HAS_BIAS folds the external additive mask in here instead of running
+// add_attention_bias_kernel first. That kernel was a full read-modify-write
+// over the score buffer -- 2.14 GB per head at a 16K prompt -- purely to hand
+// this kernel a value it is about to read anyway. A block owns one query row,
+// and the bias runs along the key axis in the same order the reduction below
+// does, so the fold is one extra contiguous stream to read. The three passes
+// re-read it, but a row is 32-64 KB and stays resident across them, so the cost
+// is one stream from memory against the two the separate pass spent on the
+// scores alone.
+//
+// The bias is indexed on its own full extents rather than through a bumped
+// pointer, for the reason add_attention_bias_kernel gives at length: `input`
+// can be a sub-block of the logical [bias_sq, bias_total_seq] matrix -- `cols`
+// query rows from bias_row_offset when the caller tiles the prefill, and `rows`
+// key columns from bias_col_offset when it narrows to a sliding window -- while
+// the bias always describes the whole thing. Folding either offset into the
+// pointer would mis-stride every batch/head plane after the first.
+//
+// The bias is added to the score BEFORE the max is taken, which is required
+// rather than incidental: taking the max of unbiased scores and then adding the
+// bias could leave a masked entry (bias -65504) as the column max and put every
+// exponent off scale.
+template <typename TOut, bool HAS_BIAS = false, typename TBias = __half>
 __global__ void softmax_f32_to_out_kernel(
     const float* input, TOut* output,
     int rows, int cols, int input_batch_stride, int output_batch_stride,
-    const __half* head_sink, int num_heads, int use_smooth_softmax) {
+    const __half* head_sink, int num_heads, int use_smooth_softmax,
+    const TBias* __restrict__ bias, int bias_batch, int bias_heads,
+    int bias_sq, int bias_row_offset, int bias_total_seq, int bias_col_offset) {
   const int linear = __builtin_amdgcn_readfirstlane(blockIdx.x);
   const int head = __builtin_amdgcn_readfirstlane(linear / cols);
   const int col  = __builtin_amdgcn_readfirstlane(linear % cols);
   const float* in_col = input + (size_t)head * input_batch_stride + (size_t)col * rows;
   TOut* out_col = output + (size_t)head * output_batch_stride + (size_t)col * rows;
+
+  // Start of this query row inside the bias, resolved exactly as
+  // add_attention_bias_kernel resolves it. `col` is the query row (this kernel
+  // reduces along `rows`, the key axis), so it indexes the bias's row axis.
+  const TBias* bias_row = nullptr;
+  if (HAS_BIAS) {
+    const int b = head / num_heads;
+    const int h = head % num_heads;
+    const int bias_b = (bias_batch == 1) ? 0 : b;
+    const int bias_h = (bias_heads == 1) ? 0 : h;
+    const size_t bias_plane = (size_t)bias_sq * bias_total_seq;
+    bias_row = bias + (size_t)bias_b * bias_heads * bias_plane +
+               (size_t)bias_h * bias_plane +
+               (size_t)(bias_row_offset + col) * bias_total_seq +
+               bias_col_offset;
+  }
+
+  // Score with the bias already folded in, or the raw score when there is none.
+  auto score_at = [&](int i) -> float {
+    float s = in_col[i];
+    if (HAS_BIAS)
+      s += static_cast<float>(bias_row[i]);
+    return s;
+  };
 
   const int tid = threadIdx.x;
   const int wave_id = tid / WAVE_SIZE;
@@ -1188,7 +1237,7 @@ __global__ void softmax_f32_to_out_kernel(
   // Pass 1: find max (fp32, no overflow risk)
   float local_max = -INFINITY;
   for (int i = tid; i < rows; i += blockDim.x)
-    local_max = fmaxf(local_max, in_col[i]);
+    local_max = fmaxf(local_max, score_at(i));
 
   local_max = warp_reduce_max(local_max);
   if (tid % WAVE_SIZE == 0)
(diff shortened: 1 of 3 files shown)


BoarQing:
Gemma Perf: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F769%2F&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F805%2F&curr_job=&curr=
Gemma Acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1772/

General: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F763&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F806&curr_job=&curr=
QWen's TTFT degradation is just noise(unreachable by this PR).

amd-mingw (approved):

amd-mingw (approved):
LGTM

PR #721, opened 2026-08-16
perf(gqa): cut the d=256 flash-prefill kernel 43% at long prompts

BoarQing:
## Summary

`gqa_flash_prefill_v8_kernel` is the largest single kernel in a 16K Qwen3.6-35B-A3B prefill. This makes it **43% faster** and more accurate at the same time. A second commit halves the d=256 flash-*decode* LDS tile, which is neutral-or-better on the same model.

| | before | after |
|---|---:|---:|
| per call, uninstrumented (runtime autotune, sq=18646) | 361.4 ms | **203.9 ms** |
| per call, standalone microbenchmark | 348.1 ms | **203.4 ms** |
| config (ND, MT, BKV) | 4, 1, 32 | 2, 1, 16 |
| VGPRs / LDS bytes | 192 / 26112 | 240 / 11264 |
| occupancy / limiter | 25.0% / LDS | 31.2% / LDS |
| `relL2` vs CPU reference @ sq=2048 | 5.7e-04 | **3.3e-04** |

## Occupancy was the wrong target

The kernel was classed `latency/low-occupancy` (25%, LDS-bound), and its design note says the D-split exists "to lift occupancy". Measured at sq=18646, occupancy is *inversely* correlated with speed:

| cfg | vgpr | spill | occupancy | ms/call |
|---|---:|---:|---:|---:|
| 2,1,16 | 239 | 0 | 37.5% | **204** |
| 4,1,16 | 135 | 0 | 62.5% | 219 |
| 4,1,32 | 175 | 0 | 50.0% | 249 |
| 8,1,16 | 82 | 0 | 100.0% | 276 |

The fully-occupant config is 35% slower than the winner. Occupancy here is a function of `ND`, and each extra wave repeated the entire softmax — which ablation showed was 136 ms of 348 ms. The shipped kernel is still LDS-limited at 31.2%, and that is fine. Six of the eleven instantiated configs also **spill** (up to 768 registers), which no occupancy figure reveals; those had been sitting in the autotune candidate list.

## What changed

1. **`BKV=16` candidates**, absent from the table. Halves `V_lds`. 348 → 248 ms.
2. **Split the softmax rows `ND` ways** instead of recomputing them in every wave. Each wave publishes its per-row correction and its share of `l`. Handing the whole softmax to one wave was tried first and is worse — the others then idle at the barrier while that copy sits on the critical path. 248 → 215 ms.
3. **fp32 accumulator.** The WMMA accumulator is already fp32, so the fp16 store only bought a pack/unpack per element per tile. 215 → 204 ms, and it is the accuracy gain above.
4. **Pruned the tuned candidate list** to non-spilling configs, which also cuts the first-call sweep from 11 candidates to 4 (~40 s each at this shape).

## Second commit: d=256 flash-decode LDS tile

The split-K decode is bandwidth-bound and its grid always supplies more blocks than fit on the part, so block residency — set by the `2*BKV*D*2` byte K/V tile — is what bounds latency hiding. d=256 was sized to the same ~16 KB/block as d=128 and d=64; dropping it to 8 KB (`BKV=16 → 8`) doubles resident blocks per WGP.

`test_gqa_decode --iters 500`, base/final interleaved x10 (absolute numbers drift several % with clocks, so only within-run comparisons are meaningful):

| shape | before | after |
|---|---:|---:|
| Qwen3.5-9B H16 G4 d256 skv3860 | 0.0424 ms | **0.0403 ms** (+5.0%, median +7.1%) |
| Qwen3.6-35B H16 G2 d256 skv18646 | 0.1770 ms | **0.1734 ms** (+2.0%, median +1.0%) |

`BKV` is a real optimum rather than a direction: `BKV=32` costs 37% on the 9B (halved residency) and `BKV=4` costs 6% on the 35B (the tile stops amortizing the per-tile `__syncthreads` and staging setup).

**Scope note:** GQA decode is only ~2.8% of per-token decode GPU time on the 9B (RGP, 1834 dispatches), so this is worth ~0.2% end-to-end TPS — real at the kernel level but well below `vlm_benchmark`'s run-to-run noise. It should not be claimed as a TPS win.

## Measured and rejected

Recorded in comments so they are not retried blind:

- **Reducing the 3 barriers per KV tile** — ablating all three saves 5.5 ms of 203, so the restructuring cannot pay.
- **Skipping the `O` rescale when `corr == 1.0`** (most tiles, once the running max settles) — 0.3%, inside run noise.
- **Compacting `Sp_lds`** — aimed at occupancy headroom that is not worth buying, and only ~10% of LDS traffic against `V_lds`'s 16 KB/tile.
- **`ND=8` with `__launch_bounds__`** to force `vgpr <= 128` — reaches 100% occupancy and is the slowest config measured.

## Heads-up: SQTT per-call time no longer A/Bs this kernel

The instrumented cost moved 386.6 → 380.8 ms while the real cost fell 43%. SQTT overhead scales with instruction issue, and `BKV=16` doubles the KV-tile trip count (583 tiles/block instead of 292). The uninstrumented measurements agree and RGP does not: the microbenchmark and the runtime's own autotune inside the real process.

## Measurement scaffolding is not part of this change

Three things used to develop this are deliberately excluded:

- **`HIPDNN_PREFILL_V8_ABLATE`** — OR'd an environment variable into the in-kernel ablation mask, which *disables* parts of the computation (softmax, value GEMM, score GEMM, V staging, K loads). That is a hazard rather than clutter: a stray variable could make a shipped build return wrong results. The mask itself predates this PR and is already reachable deliberately via `hip_gqa_flash_prefill_v8_set_ablate`, which is an explicit opt-in; an ambient environment variable is not.
- **`HIPDNN_PREFILL_V8_CFG`** — pinned the autotuner's config so an A/B wasn't muddied by the tuner moving underneath it. Removing it costs nothing: the tuner independently picks the `2,1,16` winner.
- **`--dump`/`--ref`, `--sq`/`--only`, and a timed 16K case** — a bisection aid, not a test. With no `--ref` file supplied, which is how it always actually ran, the 16K case verified nothing and still printed `PASS`, so the suite's own headline count overstated its coverage. Coverage of the long-shape code paths comes from the ragged `sq=1000` case instead, which has a real CPU reference.

## Note on the end-to-end TTFT figure

An earlier revision of this description claimed 19,526 → 17,587 ms at 16K. **Those numbers should not be used.** Both arms were measured with `HIPDNN_EP_TRACE_FILE` set in the environment, which enables EP instrumentation and inflates TTFT by ~21% (the clean uninstrumented figure for the same code was 14,455 ms). The harness function that was supposed to clear profiling state did not remove that variable. Since instrumentation overhead scales with dispatch count and this change alters KV-tile trip counts — the same effect documented for SQTT above — the instrumented delta cannot be assumed to carry to a clean build.

The kernel-level numbers in this PR are unaffected: they come from the standalone microbenchmark and from the runtime's own autotune, neither of which depends on EP instrumentation. A clean end-to-end TTFT A/B against current `main` has not been run for this branch.

## Test plan

- [x] `test_gqa_prefill` — **ALL PASS, 0 failing**, every case CPU-referenced, including the untouched d=64 (v5) and d=128 (v7) paths. Re-verified after rebase onto `main`.
- [x] The v8 kernel had **no correctness coverage at all** before this: every existing case was d=64 or d=128. Added three CPU-referenced Qwen d=256 shapes at sq=512, 1000 and 2048.
- [x] `sq=1000` is deliberately not a multiple of either the 16-row Q tile or the 16-key KV tile, so it covers the partial tiles that 512 and 2048 both skip — which matters here because `BKV=16` is one of the changes.
- [x] Accuracy improves rather than trades: `relL2` 5.7e-04 → 3.3e-04, against a 2e-3 gate.
- [x] Autotune independently selects `2,1,16` at the 16K shape, and per-shape winners differ (`4,1,16` at sq=512), so the sweep is doing real work.
- [x] RGP fence capture confirms the resource change (LDS 26112 → 11264, occupancy 25.0% → 31.2%, still LDS-limited).
- [x] No spills at the shipped config (239 VGPRs), verified with `-Rpass-analysis=kernel-resource-usage`.
- [x] Decode: `test_gqa_decode` default suite passes (0 failing), plus the two d=256 shapes above run explicitly via `--d 256` and match the CPU fp32 reference (relL2 2.78e-04 and 2.90e-04).

Measured on gfx1151 (Strix Halo), ROCm 7.1.


BoarQing:
perf: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F203&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F211%2F&curr_job=&curr=

acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/212/

BoarQing:
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F213&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F228&curr_job=&curr=

http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/229

BoarQing:
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F246%2F&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F245&curr_job=&curr=

amd-mingw (approved):

BoarQing:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/296/ run it one more time as it modified gqa as since #780

PR #722, opened 2026-08-16
perf(causal_conv): absorb the Transpose pair into a channels-last convolution, -1.77 s TTFT at 16K

BoarQing:
## What

ONNX Conv is channels-first, so the exporter brackets every causal convolution
with a Transpose pair to reach `[B, C, L]`. Neither endpoint wants that layout:
the `in_proj` matmul upstream and linear attention downstream are both
token-major. The pair is pure data movement in service of an ABI.

Worse, `[B, C, L]` is also what held the convolution to ~8% of achievable
bandwidth. One thread owned each channel, so adjacent lanes read addresses
37,292 bytes apart and coalescing was impossible.

Four commits: parallelise the prefill across the sequence, add a channels-last
prefill kernel, fold the surrounding Transposes into a `channels_last`
attribute via a new canonicalization, and guard that fold to the shapes the
kernel actually accepts.

## Measured

Qwen3.6-35B-A3B VLM, 18,646 prompt tokens (16,823 text + 1,823 image),
gfx1151. Interleaved order-reversed A/B, 3 reps per run, no EP instrumentation:

| round | order | base | this branch | delta |
|---|---|---:|---:|---:|
| 1 | base first | 17,217 | 15,253 | -1,964 |
| 2 | base first | 17,221 | 15,495 | -1,726 |
| 4 | conv first | 16,959 | 15,341 | -1,618 |
| 5 | conv first | 17,118 | 15,337 | -1,781 |

**-1,772 ms (-10.35%), 95% CI [-2,002, -1,542].** Faster in all four rounds and
in both orderings. Confirmed at the op level too: the per-op profile goes from
100 `transpose` calls to 40, with `causal_conv` still at 30.

The absolute numbers here are ~19% above this box'\''s previously recorded clean
baseline of 14,455-14,610 ms. That shift is machine state, not this branch: it
appeared across an unrelated reboot, is present identically on both arms, and
also moves decode (19.70 -> 23.27 ms/token) and CPU-side preprocessing
(154 -> 191 ms). The paired A/B is what the delta rests on.

## Correctness

- `test/numeric` causal-conv suite: 18 passed, including
  `test_ccws_channels_last_bit_identical`, which asserts the channels-last path
  is byte-identical to channels-first, and
  `test_ccws_channels_last_matches_reference` at seq_len 1, 128 and 257.
- lit: 4 passed, including the new `hip-causal-conv-channels-last.mlir`.

Three things the fold has to get right, all covered by negative lit cases:

- **The MIOpen fallback is channels-first.** Its 4D descriptors are
  `(B, C, 1, virtual_len)` and the pitched copies assume each `(b, c)` plane is
  contiguous, so a `[B, L, C]` buffer read through them yields plausible
  garbage rather than an error. `wrap_causal_conv_with_state` now refuses
  `channels_last` on that path, and the fold only fires inside the custom
  kernel'\''s envelope (k in 1..8, fp16/fp32) so the refusal is unreachable.
  Without that guard a k > 8 model would have folded and then failed at
  runtime -- that is the last commit here.
- **Decode shares the graph.** At `seq_len == 1` the flat index of `(b, c)` is
  `b*channels + c` under both layouts, since the permuted axis has extent 1, so
  the decode kernel is layout-agnostic by construction rather than by accident.
- **Both Transposes must die.** If either has a second reader it still runs,
  and folding would buy a layout change without removing the traffic that
  motivated it. Both are checked single-use.

The carry state is `[B, C, k-1]` under both layouts and is forwarded untouched,
so the fold cannot invalidate a state buffer carried in from a previous step.

### DIFF
--- a/include/hip/Dialect/IR/HipOps.td
+++ b/include/hip/Dialect/IR/HipOps.td
@@ -2969,10 +2969,19 @@ def Hip_CausalConvWithStateOp : Hip_DpsOp<"causal_conv_with_state",
     the last spatial dimension) and depthwise (each channel is convolved
     independently with its own kernel).
 
-    Input layout is channels-first: (batch_size, channels, ...).
-    Weight layout: (channels, 1, k_1, ...) for depthwise convolution.
-    The carry state stores the last (k-1) positions along the causal axis
-    for incremental decode.
+    Input layout is channels-first, (batch_size, channels, ...), unless
+    `channels_last` is set, in which case 1D input is (batch_size, seq_len,
+    channels). Weight layout: (channels, 1, k_1, ...) for depthwise
+    convolution. The carry state stores the last (k-1) positions along the
+    causal axis for incremental decode, and is (batch, channels, k-1) in both
+    layouts -- only input and output are permuted, so a layout change never
+    invalidates a carried state.
+
+    `channels_last` exists because the surrounding graph is channels-last: a
+    channels-first convolution has to be bracketed by a Transpose pair, and at
+    the Qwen3.6 footprint that pair costs more than the convolution it serves.
+    The canonicalizer sets the attribute while absorbing both Transposes.
+    Only ndim=1 is supported channels-last.
 
     Example (1D):
     ```mlir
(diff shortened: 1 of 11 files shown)


BoarQing:
Perf: 
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F213&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F222%2F&curr_job=&curr=

Acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/223/

amd-mingw (approved):

PR #723, opened 2026-08-16
perf(qmoe): bucket MoE tokens in chunks, -348 ms TTFT at 16K

BoarQing:
## What

`bucket_tokens_kernel` runs as a single workgroup with one thread per expert, and
every one of those threads walks every `(token, slot)` pair twice. Its cost is
therefore `num_tokens * k` long no matter how many CUs are idle: 1 of 40 CUs
busy, the rest stalled behind it.

That is invisible at decode, where there are only `k` pairs, and dominant at
prefill. This commit replaces it with a chunked stable counting sort -- histogram
each chunk of pairs, scan the per-chunk histograms into per-`(chunk, expert)`
write bases, then have each chunk write its own pairs.

## Measured

Entry-point microbenchmark, unchanged from the original commit:

| tokens | experts | k | serial | chunked | speedup |
|---:|---:|---:|---:|---:|---:|
| 3985 | 256 | 8 | 1.7738 ms | 0.0382 ms | 46.5x |
| 2048 | 256 | 8 | 0.9226 | 0.0371 | 24.9x |
| 512 | 256 | 8 | 0.2369 | 0.0273 | 8.7x |
| 3985 | 32 | 4 | 1.3145 | 0.0496 | 26.5x |

End-to-end at 16K, which the original commit expected to be unresolvable and is
not. Qwen3.6-35B-A3B VLM, 18,646 prompt tokens, gfx1151, interleaved
order-reversed A/B, 4 reps per run, no EP instrumentation:

| round | order | base | this branch | delta |
|---|---|---:|---:|---:|
| 1 | base first | 15,197 | 14,880 | -317 |
| 2 | base first | 15,266 | 14,874 | -392 |
| 3 | branch first | 15,451 | 15,097 | -354 |
| 4 | branch first | 15,417 | 15,087 | -330 |

**-348 ms (-2.27%), 95% CI [-401, -296]**, sd of paired differences 33 ms.
Faster in all four rounds and in both orderings.

The original commit measured at 3,985 tokens, where this defect costs ~71 ms
against a noise floor of roughly +-200 ms, and concluded correctly that no
end-to-end A/B could see it. At 18,646 tokens the same defect costs 353 ms by
RGP while the noise floor is unchanged, which is what makes it measurable. RGP
predicted 353 ms and the A/B returned 348 ms; the agreement is expected rather
than lucky, because this kernel was dispatch-limited rather than
bandwidth-limited and so had no clock-dependent floor to fall short of.

Baseline for this A/B is the channels-last causal convolution branch
(#722), since these stack.

## Correctness

The reason this is safe is ordering, not speed. The downstream weighted
`scatter_add` needs each expert'\''s slice to stay token-major for reproducibility.
Pair index `i == t*k + s`, so ascending `i` is exactly the `(t, s)` nesting the
serial kernel used; chunks are contiguous ascending ranges of `i`; and the scan
hands out bases in chunk order. All four outputs come back byte-identical to the
single-workgroup kernel at every shape in the table, including the `-1` slots
`topk_routing` emits for an unroutable token, which both paths must drop rather
than count.

Shapes that would not benefit keep the old path: below one chunk'\''s worth of
pairs (decode, and anything under 512) three launches would cost more than the
scan they save, so the launcher falls through to the single workgroup. Same when
the caller passes no scratch, so callers without an arena keep working.

`test/numeric/tests/test_qmoe.py`: 3 passed, `test_qmoe_gpt_oss_shape[1]` failed.
That failure is pre-existing and not from here -- verified by running the same
test against the parent build, which fails with bit-identical numbers
(`max_diff=6.152344e-01`, `cosine=0.999961`). It is also a decode shape, which
takes the single-workgroup fallback, so the chunked path is not even reached.

## Not addressed

The D2H of the counts plus `hipStreamSynchronize` right after this call. It sizes
a host loop that issues gather/fc1/swiglu/fc2/scatter per active expert, and at
prefill essentially all 256 experts are active every layer -- on the order of
1,500 launches per layer. That launch count, not the sync, is the remaining cost
here, and it wants its own change.

### DIFF
--- a/lib/Runtime/Kernels/hip/qmoe_kernel.hip
+++ b/lib/Runtime/Kernels/hip/qmoe_kernel.hip
@@ -309,6 +309,140 @@ __global__ void bucket_tokens_kernel(
   }
 }
 
+// ---------------------------------------------------------------------------
+// Chunked bucketing: the same three phases, but over ranges of (token, slot)
+// pairs so the work is spread across the machine instead of one workgroup.
+//
+// bucket_tokens_kernel above walks every pair in every thread, so its cost is
+// num_tokens*k long no matter how many CUs are idle. That is invisible at
+// decode (k pairs) and dominant at prefill: 3985 tokens x top-8 is 31,880 pairs
+// scanned twice by one workgroup, ~1.75 ms per layer on Qwen3.6-35B-A3B, 70 ms
+// across its 40 layers.
+//
+// Splitting it is a stable counting sort. Phase 1 histograms each chunk
+// independently. Phase 2 turns those per-chunk histograms into a per-(chunk,
+// expert) write base: expert_offsets[e] plus the counts of the same expert in
+// every earlier chunk. Phase 3 has each chunk write its own pairs from that
+// base, in ascending pair order.
+//
+// Ordering is what makes this safe to swap in. Pair index i == t*k + s, so
+// ascending i is exactly the (t, s) nesting of the serial version, chunks are
+// contiguous ascending ranges of i, and phase 2 hands out bases in chunk order.
+// Every expert's slice therefore comes out token-major, byte for byte what the
+// single-workgroup kernel produces -- which the downstream weighted scatter_add
+// depends on for reproducibility.
+// ---------------------------------------------------------------------------
+
+// Pairs per chunk. Small enough that a prefill splits into enough chunks to
+// fill the machine, large enough that the per-chunk histogram and the phase-2
+// scan over chunks stay cheap.
+#define QMOE_BUCKET_CHUNK 512
+// Chunk cap, which bounds the scratch matrix at nchunks*num_experts int32.
+#define QMOE_BUCKET_MAX_CHUNKS 256
+
+__global__ void bucket_count_chunks_kernel(
+    const int32_t* __restrict__ expert_indices,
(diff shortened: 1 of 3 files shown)


amd-yanjunz:

- PR: ROCm/hip-ep#723
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/723#issuecomment-5505330033)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/657/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6a97e1e8b38f5467b2650391
- Compare with latest main branch results: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fids%3D6a971ab5f10c2b1fd363294f%26collection%3DHIP-EP-Dynamic&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=657

PR #749, opened 2026-08-17
feat(hipsr): add hipsr-add-context-arg pass

mingyueliuh:
## Summary

Add `hipsr-add-context-arg`, which inserts a `!hipsr.context` argument at position 0 into every `func.func`, and make it the first pass of `buildHipsrPipeline`.

## What

- Functions that already have `!hipsr.context` as argument 0 are skipped.
- Inserting the argument changes the function type, so every `func.call` to an updated function is rewritten to forward the caller's own context. 
- `buildHipsrPipeline` was an empty stub. The new pass has to run before any conversion, since the conversion patterns depend on argument 0 already being the context.

## Test plan

- [x] New `test/lit/Dialect/Hipsr/Transforms/AddContextArg/add-context-arg.mlir`: argument insertion, skipping a function that already has the context, and call-site forwarding.
- [x] `test/lit/Dialect/Hipsr/Pipelines/embedding.mlir` updated because `--hipsr-pipeline` is no longer a no-op.

### DIFF
--- a/include/hip/Dialect/Hipsr/Transforms/Passes.td
+++ b/include/hip/Dialect/Hipsr/Transforms/Passes.td
@@ -13,6 +13,21 @@
 
 include "mlir/Pass/PassBase.td"
 
+def AddContextArgPass : Pass<"hipsr-add-context-arg", "mlir::ModuleOp"> {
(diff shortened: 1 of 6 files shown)


wcy123 (approved):
LGTM

PR #750, opened 2026-08-17
feat(onnx-to-hip): lower axis-0 scalar Gather to extract_slice

amd-mounikk:
## Summary
- Lower Swin-style axis=0 `onnx.Gather` with a compile-time scalar (or len-1) index to `tensor.extract_slice` instead of `hip.gather`, producing zero-copy subviews when the offset is static.
- Add `getCompileTimeScalarInt` helper and tag externalized single-element integer globals with `hip.compile_time_scalar` so indices stored in `constants.bin` still fold at compile time (avoids runtime host loads from GPU-resident constants).
- Add lit test covering externalized index globals.

## Performance (SwinV2-Tiny 8×521×1024 fp16, gfx1151, hipgpu EP)

Benchmark: `onnxruntime_perf_test -I -m times -r 10 -c 1 -s` with `session.disable_cpu_ep_fallback|1`. Model: `swinv2-tiny-8x521x1024-fp16` (batch=8).

| Metric | Before | After | Delta |
|--------|--------|-------|-------|
| Avg inference latency (10 runs) | 252.1 ms | **240.3 ms** | **−11.8 ms (−4.7%)** |
| `gather` kernel calls (axis-0 Q/K slices) | 36 | **0** | **−36** |

Op-level profiling confirms all 36 axis-0 `gather` calls are eliminated (lowered to zero-copy `extract_slice`).

## Test plan
- [x] `hip-mlir-opt test/lit/Conversion/onnx-to-hip/test_gather_axis0_extract_slice.mlir --hip-add-context-arg --convert-onnx-to-hip`
- [x] SwinV2-Tiny 8×521×1024 latency before/after on hipgpu EP (see Performance table)
- [x] Existing gather lit tests (`test_gather.mlir`, `test_gather_shape_fold.mlir`, etc.) still pass


asparkhi (changes requested):
LGTM @amd-mounikk. We need tests for attachCompileTimeScalarAttr()

PR #751, opened 2026-08-17
feat(hipsr): convert onnx.Shape to a hipsr.compute

zpye:
## Summary

Convert `onnx.Shape` to a `hipsr.compute` that reads the input's extents into a host-space i64 tensor, and fill the destination placeholder's shape region during the conversion.

## Why

- No hipsr operation matches `onnx.Shape`. The work is a few tensor and arith operations rather than a library call, so it belongs in a `hipsr.compute` body.
- The extents are read on the host, which is also the space `hipsr.expand` takes its `shape` operand from, so the pattern names `#hipsr.mem<host>` for its own result. Data keeps the `#hipsr.mem<device>` the pass gives a tensor that names no space.
- `hipsr-populate-shape-region` cannot fill the region. It derives a shape from the placeholder's consumer, and a compute's result shape follows whatever its body does, so there is nothing to dispatch on. The conversion fills the region while it builds the body, and that pass already skips a region that is not empty.

## What

- New `ShapeConversion.cpp`: resolves the optional `start`/`end` attributes against the input rank, clamping into `[0, rank]` as the spec defines; emits a `normal` placeholder with a constant shape region, since the rank fixes the result length; builds the body as one `tensor.dim` per dynamic axis and an `arith.constant` per static one, packed with `tensor.from_elements`.
- `deviceTensorType` becomes `tensorTypeInSpace(type, space)`, one `RankedTensorType::cloneWithEncoding` for either space.
- `createPlaceholderShapeBlock` moves out of `PopulateShapeRegionPass.cpp` into `HipsrShapeRegionPopulationUtils.h`, since both callers build the same entry block.
- Helper operations are now legal in a placeholder's shape region as well as in a compute body, so the `shape.from_extents` this conversion emits no longer rolls the rewrite back.
- New `shape.mlir` checks every line each case produces; new `shape-invalid.mlir` covers the seven forms the pattern rejects.

## Notes for reviewers

- A graph whose extents are read has to name `#hipsr.mem<host>` on the ONNX result. A consumer pattern that carries the pass converter asks for each operand in `convertType(...)`, which is device for a tensor naming no space, and the pass then ends with an unresolved host-to-device materialization. Letting consumers take the space their producer chose is a follow-up.
- The destination the placeholder allocates is never written, because the body yields a fresh `tensor.from_elements`. Both buffers are host, so the result is correct, but bufferization allocates twice. Making the body store into its destination is a separate change.
- No HipsrToLLVM lowering yet, as for every operation in this series, so `--hipsr-pipeline` still cannot run this end to end.

## AI assistance

Cursor (Claude Opus 5) assisted with the conversion and the tests, and ran the build, the LIT suites and `pre-commit`. I reviewed the result and understand it.

### DIFF
--- a/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
+++ b/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
@@ -39,6 +39,10 @@ void populateExpandConversionPatterns(
     const ::mlir::TypeConverter &typeConverter,
     ::mlir::RewritePatternSet &patterns, ::mlir::MLIRContext *ctx);
 
+void populateShapeConversionPatterns(const ::mlir::TypeConverter &typeConverter,
+                                     ::mlir::RewritePatternSet &patterns,
+                                     ::mlir::MLIRContext *ctx);
+
 void populateReturnConversionPatterns(::mlir::RewritePatternSet &patterns,
                                       ::mlir::MLIRContext *ctx);
 
--- a/include/hip/Dialect/Hipsr/IR/HipsrShapeRegionPopulationUtils.h
+++ b/include/hip/Dialect/Hipsr/IR/HipsrShapeRegionPopulationUtils.h
@@ -9,6 +9,7 @@
 #include "hip/Dialect/Hipsr/IR/HipsrOps.h"
 
 #include "mlir/IR/Block.h"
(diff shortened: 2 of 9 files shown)


wcy123 (approved):
LGTM.

PR #760, opened 2026-08-17
fix(hipsr): keep an ONNX-to-hipsr operand in its producer's memory space

zpye:
## Summary

An ONNX-to-hipsr operand now arrives in the memory space its producer chose. No
per-operation pattern carries the pass type converter any more, and each pattern
names the space its own results need instead.

## Why

`remapValues` hands a pattern that carries a converter each operand in
`convertType(operandType)`, and that converter names `#hipsr.mem<device>` for a
tensor naming no space. The host extents `onnx.Shape` produces could not reach
`onnx.Expand`, which reads its `shape` operand on the host, so the pass ended
with an unresolved host-to-device materialization. This is the follow-up #751
named.

## What

- The cast, matmul, expand, constant and shape patterns no longer pass the
  converter to `ConversionPattern`, so each operand arrives as its producer
  built it.
- Each pattern names its result space through `tensorTypeInSpace`: device for
  cast, matmul, expand and constant, host for shape. A rank-0 result now names
  device too, which its ODS constraint required all along.
- `shape.mlir` drops the `#hipsr.mem<host>` its ONNX types had to spell;
  `expand-invalid.mlir` covers the graph argument that still has to name it.
- `OnnxToHipsr.cpp` is unchanged. The func and return patterns still use the
  converter at the graph boundary.

## AI assistance

Cursor (Claude Opus 5) assisted with the changes and the tests, and ran the
build, the LIT suites and `pre-commit`. I reviewed the result and understand it.

### DIFF
--- a/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
+++ b/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
@@ -23,6 +23,9 @@ namespace hipsr {
 #define GEN_PASS_DECL_CONVERTONNXTOHIPSRPASS
 #include "hip/Conversion/Passes.h.inc"
 
+// These patterns keep the converter out of ConversionPattern: a carried
+// converter forces every operand through convertType, which overwrites the
+// memory space its producer chose.
 void populateOnnxToHipsrConstantPatterns(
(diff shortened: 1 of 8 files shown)


wcy123 (approved):
LGTM

PR #771, opened 2026-08-18
feat(hipsr): add hipsr.alloc_output op

amd-genmingz:
## Summary
Register the `hipsr.alloc_output` operation. It obtains a graph-output
buffer from the EP output allocator at the point in the graph where the
output shape is known, so output buffers are owned by the EP/runtime
instead of being pooled or deallocated by the graph

## What
- Add `HipsrAllocOutputOp.td` defining `hipsr.alloc_output`:
  - operands: `!hipsr.context` + one `Index` per dynamic result dim
    (same operand convention as `hip.alloc`)
  - `out_idx` attribute identifying which graph output this buffer is
  - result: `AnyMemRef`
  - assembly format: `(%ctx, %d0, %d1) {out_idx = N} : memref<...>`
- Add `HipsrAllocOutputOp.cpp`:
  - `getEffects` reports only a `MemoryEffects::Write` (the EP owns the
    buffer, so it is not modeled as an allocation)
  - `verify` checks the dynamic-size operand count matches the number of
    dynamic dims in the result memref
- Register the op in `HipsrOps.td` and build it in the IR `CMakeLists.txt`.
This only adds the op. The pass that rewrites graph-output `memref.alloc`
into `hipsr.alloc_output` lands separately.

Made with Cursor

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrAllocOutputOp.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrAllocOutputOp.td
@@ -1,0 +1,30 @@
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// Licensed under the MIT License.
+//
+
+def Hipsr_AllocOutputOp : Hipsr_Op<"alloc_output",
+    [DeclareOpInterfaceMethods<MemoryEffectsOpInterface>]> {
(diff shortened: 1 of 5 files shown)


wcy123 (approved):

wcy123 (approved):
LGTM

PR #772, opened 2026-08-18
feat(onnx): add an opt-in registered onnx dialect and match hipsr conversions through it

zpye:
## Summary

Adds an `onnx` dialect covering every standard ONNX operation it can spell, and migrates the six ONNX-to-hipsr patterns onto it, so they match a typed operation class and read generated accessors instead of name strings and attribute lookups.

It is opt-in and does not replace the header-only stub:

```bash
hip-mlir-opt --onnx-dialect=stub      # default; ops stay unregistered
hip-mlir-opt --onnx-dialect=modeled   # ODS ops, verified on parse
```

`stub` stays the default, so `registerAllDialects` and the Execution Provider are unchanged. `lib/Conversion/OnnxToHip/` matches by name and gains nothing from a typed op set; the hipsr track wants it and asks for it explicitly.

`include/hip/Dialect/Onnx/IR/Ops/` holds one file per operation, 191 of them, which `OnnxOps.td` includes. Each schema operation carries its ONNX documentation as the `description`. `If`, `Loop` and `Scan` carry their subgraphs as regions.

## Notes for reviewers

- Read the commits in order. The first adds the dialect and touches no existing test; the second migrates the patterns and accounts for every test change.
- These `.td` files are the source of truth. Nothing generates them, so a new opset is an edit per file, and a wrong constraint is caught by review rather than by a checked regeneration.
- Fifteen operations are not modeled. Fourteen need an ONNX sequence, optional or map container, or a string tensor, none of which has an MLIR type here. `Constant` is hand-written because the importer also writes an external-data form no schema attribute covers. `allowUnknownOperations` keeps all fifteen parsing.
- The dialect has no LIT file of its own, since every rule in it is declarative ODS. The `onnx-to-hipsr` suite exercises the definitions by running under `--onnx-dialect=modeled`.
- The dialect asks two things of the importer, neither affecting anything today: a stand-in for a trailing omitted optional operand, and `i1` rather than `ui8` for an ONNX bool. `Pipelines/embedding.mlir` is hand-conformed in both places meanwhile.

Cursor (Claude Opus 5) assisted with the dialect and the migration. I reviewed the result and understand it.


wcy123 (approved):
LGTM

PR #774, opened 2026-08-19
feat(hipsr): add hipsr-use-output-allocator pass

amd-genmingz:
## Summary

Adds the `hipsr-use-output-allocator` compiler pass, which rewrites graph-output `memref.alloc` operations into `hipsr.alloc_output`. Output buffers are then allocated through the EP output-allocator callback at the point where their shape is known, and remain EP/runtime-owned instead of being pooled or deallocated by the graph.

This is the HIPSR counterpart to the existing `hip-use-output-allocator` pass and builds on the `hipsr.alloc_output` op added in #771.

## Why
ORT requires the tensors at the graph outputs to have the same shape when IOBinding is used. Therefore, when allocating the memory, the requested shape must be consistent with the ONNX graph output shape.

## What changes

- **New pass `hipsr-use-output-allocator`** (`lib/Dialect/Hipsr/Transforms/HipsrUseOutputAllocator.cpp`)
  - Operates on public graph-entry `func.func` whose argument 0 is `!hipsr.context`
  - Uses `BufferViewFlowAnalysis` to trace returned values (directly or through view-like aliases) back to underlying `memref.alloc`
  - Handles outputs produced via `hipsr.pool_domain` / `hipsr.pool_domain_yield`
  - Uses `hipsr.preserve_shape` metadata to materialize external vs. internal dynamic dimensions
  - Inserts `memref.reinterpret_cast` when the returned (external) type/shape differs from the internal allocation
  - Removes associated `memref.dealloc` ops on rewritten allocs
  - Validates single `func.return`, unique aliased alloc per output, and dominance of preserved shapes

- **`findVisibleContext` helper**
  - Walks outward through nested regions to locate a visible `!hipsr.context`, including across `hipsr.pool_domain` bodies
  - Fixes context lookup failures caused by `OpTrait::IsIsolatedFromAbove` on enclosing ops

- **Pass registration** in `include/hip/Dialect/Hipsr/Transforms/Passes.td` and `lib/Dialect/Hipsr/Transforms/CMakeLists.txt`

- **LIT coverage** (`test/lit/Dialect/Hipsr/Transforms/UseAllocOutput/hipsr_use_alloc_output.mlir`)
  - `dynamic_output`: simple dynamic alloc → `hipsr.alloc_output` with preserved `out_idx`
  - `test_return_shape`: end-to-end scenario with `pool_domain`, matmul/add/compute/cast, `preserve_shape`, and a rank-reducing collapse before return

## Test plan

- [ ] LIT: `hip-mlir-opt test/lit/Dialect/Hipsr/Transforms/UseAllocOutput/hipsr_use_alloc_output.mlir -hipsr-use-output-allocator | FileCheck`

## AI assistant

The code was completed collaboratively with CU, and all code has been reviewed


wcy123 (approved):
LGTM

PR #779, opened 2026-08-19
fix(hipsr): keep a graph output's memory space at the function boundary

zpye:
## Summary

`onnx.Return` now lowers through an `OpConversionPattern` that builds `func.return` from the adaptor operands and retypes the enclosing function's results to match. A graph output keeps the memory space its producer chose all the way to the function boundary, so a graph no longer has to declare that space in its signature.

## What

- `ReturnToFunc` is an `OpConversionPattern<onnx::ReturnOp>`. It takes the pass converter without handing it to the base, sets the function type from the adapted operand types, and replaces the terminator.
- `populateReturnOpTypeConversionPattern` and the `FuncConversions.h` include are dropped. `func::ReturnOp` keeps its dynamic legality on `converter.isLegal(op)`, so the pass legalizes `onnx.Return` only: a `func.return` already naming its spaces stays legal, and one that does not is now a legalization failure instead of a converter-forced rewrite.
- The positive lit cases that ended in a plain `func.return` over an unspaced type now use `onnx.Return`, and `return.mlir` checks the exact full output IR.

### DIFF
--- a/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
+++ b/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
@@ -46,8 +46,9 @@ void populateShapeConversionPatterns(const ::mlir::TypeConverter &typeConverter,
                                      ::mlir::RewritePatternSet &patterns,
                                      ::mlir::MLIRContext *ctx);
 
-void populateReturnConversionPatterns(::mlir::RewritePatternSet &patterns,
-                                      ::mlir::MLIRContext *ctx);
+void populateReturnConversionPatterns(
+    const ::mlir::TypeConverter &typeConverter,
+    ::mlir::RewritePatternSet &patterns, ::mlir::MLIRContext *ctx);
 
 } // namespace hipsr
 } // namespace mlir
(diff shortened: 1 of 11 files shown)


wcy123 (commented):
LGTM. The core change is correct: [ReturnToFunc](https://github.com/ROCm/hip-ep/blob/398c54812e36214bbd3db81298991e64940815b9/lib/Conversion/OnnxToHipsr/ReturnConversion.cpp#L27-L31) takes `TypeConverter` but doesn't pass it to the base, so operands keep the memory space their producers chose, and [setFunctionType](https://github.com/ROCm/hip-ep/blob/398c54812e36214bbd3db81298991e64940815b9/lib/Conversion/OnnxToHipsr/ReturnConversion.cpp#L39-L42) writes that space to the signature. The [extents_leave_the_function test](https://github.com/ROCm/hip-ep/blob/398c54812e36214bbd3db81298991e64940815b9/test/lit/Conversion/onnx-to-hipsr/shape.mlir#L196-L219) shows a `tensor<2xi64>` graph output becoming `tensor<2xi64, #hipsr.mem<host>>` at the boundary without annotation, which was the bug.

wcy123 (approved):
LGTM

PR #782, opened 2026-08-20
fix(onnx-to-hip): compute dynamic hip.slice extents from the slice bounds (2.2x decode TPS on Gemma-4 26B-A4B at 12K)

BoarQing:
## Summary

`SliceToHip` sized every dynamic output dimension of a `hip.slice` from the data
dimension at the same position, on the grounds that ONNX Slice cannot widen an
axis so the data dim is a safe upper bound. It is a safe bound, but it is not a
safe *extent*: the size placed on the `tensor.empty` init is the output shape as
far as everything downstream is concerned, because `tensor.dim` of a DPS result
resolves through to the init.

A Slice that keeps one row of a 12k-row tensor therefore handed every consumer a
12k-row shape. On Gemma-4 26B-A4B this inflated the decode causal mask from
`[1, 1, S]` to `[1, S, S]` and cost 55% of every generated token.

Two commits: the fix, and a review follow-up (a missing bound on the shape-index
walk, a test for the one branch of it nothing covered, and the rationale that was
previously only in this description). The `ci(windows)` OGA prerequisite that was
originally in this PR has been dropped from it and is tracked in #599; until that
lands CI cannot build an OGA that runs Gemma-4 unified, so every model number
below is a local run on the machine named.

## Related issue or design

None. Same bug class as #700 (`buildExpandShapeOutputShape` deriving a dynamic
extent from the wrong source), one op over.

## Why

Gemma-4's decoder builds its causal mask from the current query positions:

```mlir
%ids_len  = onnx.Shape(%input_ids)     {start = 1, end = 2}
%attn_len = onnx.Shape(%attention_mask){start = 1, end = 2}
%start    = onnx.Sub(%attn_len, %ids_len)
%pos      = onnx.Slice(onnx.CumSum(%attention_mask), %start, %attn_len, axes=[1])
```

`ends - starts` is `Shape(input_ids)[1]`, so during decode the slice keeps
exactly one element. `starts` is not a compile-time constant, so
`SliceDecompose` bails and `SliceToHip` fires and sizes the output `12277`
instead of `1`. Every consumer of the broadcast against it then ran on `S x S`.

An RGP dispatch capture, fenced onto the first op of the mask chain in the first
decode step, shows the inflation appearing in the dispatch immediately after the
slice:

| kernel | threads | duration |
|---|---:|---:|
| `hip_cumsum_kernel<i64>` | 256 | 796 us |
| `elementwise_sub_i64_kernel` | 256 | 11 us |
| `hip_slice_kernel<i64>` | 12,288 | 2.8 us |
| `hip_expand_kernel<i64>` | 150,724,864 | **6,161 us** |

150,724,864 is `12277^2`; 1.2 GB of i64 materialized in one dispatch, for a mask
that needs 98 KB. Per-op profiling puts the whole chain
(`not`/`less`/`cast`/`sub`/`and`/`or`) at 126 ms of a 210 ms decode step.

The clearest symptom in hindsight: `not` cost 27.2 ms in a 12,276-token prefill
and 27.9 ms in a *one-token* decode step. Identical, because both were building
the same square mask.

Numerics were never wrong. The runtime slice writes the correct value at element
0, garbage after it, and GQA with `sq=1` reads only row 0 -- so the output was
right and merely 12,277x overpriced. That is also why nothing downstream could
detect it: `[1, S, S]` is a legal shape and only the value is wrong, so the fix
has to be at the point the shape is constructed.

## What

- `SliceConversion.cpp`: `SliceToHip` now computes each dynamic output extent as
  `ends - starts` when both bounds are host-resolvable, applying the ONNX
  negative-index and clamping rules with `arith` select/max/min against the data
  dim and `ceildiv` by a positive constant step.
- New `resolveHostIndex` walks a shape operand to a host `index` value through
  `tensor.from_elements` / `arith.index_cast` / `tensor.dim`, `tensor.cast`,
  compile-time constants, `onnx.Shape`, and add/sub/mul. Both the pre- and
  post-conversion spelling are accepted (`onnx.Shape` or the
  `from_elements(index_cast(dim))` that `ShapeToTensorDims` leaves, `onnx.Sub`
  or `hip.sub`), because the greedy driver gives no ordering guarantee between
  this pattern and the ones rewriting its operands.
- No device readback is added: the bounds here are host arithmetic over
  `onnx.Shape` entries, so the extent is an `index` computation.
- Unresolvable bounds, a non-constant `axes`/`steps`, or a non-positive step
  keep the previous data-dim upper bound, so no currently-working model changes
  shape.

Review follow-up, no behavior change on any path that already resolved:

- `resolveHostIndex` took `idx` on trust. Every branch that indexes a list
  checked it, but the `onnx.Shape` branch computed `dim(start + idx)` and only
  bounded the *result* against the operand rank, so an element past the end of a
  shape narrowed by its `end` attribute resolved to a real but wrong dimension.
  `idx` is now bounded against the value's own extent once at entry. That is the
  same class of silently-wrong extent this PR exists to fix, so it should not be
  reachable from here.
- The data-dim fallback now emits an `LLVM_DEBUG` line naming the op and axis.
  The original bug was invisible until an RGP capture; the fallback is a known
  performance cliff and should be findable without one. It reports only axes the
  op actually slices: an axis `axes` does not name is not sliced, so the data dim
  is its exact extent and not a fallback at all. On Gemma-4 that distinction is
  the difference between four spurious warnings per compile and none.
- Two comments corrected. The file header claimed `hip.slice`'s runtime is a
  throwing stub and that models reaching this path are unsupported; it has been
  implemented in `Runtime/real/slice.cpp` since #284 and Gemma-4 executes it
  every step. And I had justified not using the `hip.ReadbackDimOp` that
  `CompressConversion` uses by the cost of a host sync per step -- which the
  runtime contradicts, since `wrap_slice` already D2Hs its own bounds and
  synchronizes the stream on every call. The narrower real reason (a readback
  would add a second sync point ahead of the slice, for a case nothing in scope
  reaches) is now in the comment, along with the note that the cost is lower
  than it looks if a model ever does land there.
- `INT64_MIN` needing no more care than `INT64_MAX` in the bound normalization
  is now written down.

## Test plan

`test/lit/Conversion/onnx-to-hip/test_slice.mlir`:

- `test_slice_native_shape_sub_extent` pins the Gemma-4 idiom: extent comes from
  the clamped bounds, and the axis `axes` does not touch still forwards the data
  dim.
- `test_slice_native_opaque_starts` pins the fallback when `starts` is a graph
  input.
- `test_slice_native_dyn_axis` previously asserted `tensor.empty(%dim)` -- it was
  pinning the bug -- and now asserts the clamped slice length.
- `test_slice_native_lowered_shape_sub_extent` (new, from the follow-up) feeds
  the same idiom already lowered, pinning the post-conversion branch of the walk.

402 lit tests: 390 passed, 12 unsupported (pre-existing Morphizen plugin tests),
0 failures.

That last test exists because the coverage claim in this PR was wrong.
Instrumenting `resolveHostIndex` to report every spelling it visits shows that
what reaches it on Gemma-4 26B-A4B is the ONNX spelling -- `onnx.Shape` six
times, `onnx.Sub` twice, and the lowered form never -- and the same
instrumentation over the lit suite shows `test_slice_native_shape_sub_extent`
takes that same ONNX path, not the `hip.sub` of two `from_elements` its comment
claimed. Both branches emit `tensor.dim` + `arith.subi`, so the CHECKs passed
either way and the lowered branch had no coverage at all, in the suite or on a
model. Nothing pins the driver's order, so it is kept and now tested rather than
deleted.

Correctness on Gemma-4 26B-A4B FP16W4-qmoe: the 48-token completion for a
12,276-token prompt plus image is byte-identical before and after.

## Performance

Gemma-4 26B-A4B FP16W4-qmoe, gfx1151 (Strix Halo, 63.6 GB shared), Release,
`vlm_benchmark.py`, no EP instrumentation. 64 tokens generated, 3 iterations
after 1 warmup at 12,276; 32 tokens, 2 iterations after 1 warmup on the sweep.

| Metric | Before | After | |
|---|---:|---:|---|
| Token generation p50 | 250.06 ms | 111.85 ms | **2.24x** (4.00 -> 8.94 tok/s) |
| Token generation avg | 266.49 ms | 127.68 ms | **2.09x** (3.75 -> 7.83 tok/s) |
| Token generation p90 | 322.51 ms | 172.80 ms | -46% |
| TTFT p50 | 25,434 ms | 25,329 ms | unchanged |
| Peak memory | 16,234 MB | 16,210 MB | unchanged |

TTFT is unchanged by design: during prefill the mask is genuinely square, so the
inflated extent cost nothing there.

The win scales quadratically with context, because the discarded work is an
`S x S` mask where `1 x S` was needed. Full sweep, p50 ms/token:

| Prompt | Before | After | | saved |
|---|---:|---:|---|---:|
| 513 | 20.70 | 20.88 | 0.99x | -0.18 |
| 1,025 | 25.66 | 24.70 | 1.04x | 0.96 |
| 2,240 | 36.93 | 32.75 | 1.13x | 4.18 |
| 4,097 | 59.74 | 45.26 | 1.32x | 14.48 |
| 8,181 | 137.92 | 71.22 | 1.94x | 66.70 |
| 12,276 | 257.08 | 106.60 | 2.41x | 150.48 |
| 16,241 | 405.37 | 169.97 | **2.38x** | 235.40 |

At 16,241 that is 2.47 -> 5.88 tok/s. Scaling the 2,240 saving by `S^2` predicts
219.7 ms at 16,241 against 235.40 measured (+7.1%), and the 8,181 saving
predicts 150.2 at 12,276 against 150.48 (+0.2%), so the quadratic model holds
across a 7.9x span of context. The endpoints bracket it rather than sitting on
it -- 4,097 -> 8,181 runs 15.5% above the prediction and 12,276 -> 16,241 runs
10.6% below -- which is the expected shape once the per-arm std reaches 12-48 ms
at the long end. The 512 arm is a wash (std 0.26 and 0.28 against a 0.18 delta),
as it should be: there is nothing to save when `S` is small.

### Is there anything left here

Two checks say no, and both are worth having on the record.

Per-op GPU profiling of a decode step at 16,241 tokens, before and after, sums
to 371.2 -> 133.5 ms (-237.7), which independently reproduces the 235.40 ms
wall-clock saving above to within 1.0%. The ops this PR was aimed at -- the
`cast`/`not`/`less`/`or`/`sub`/`and` mask chain -- go from 236.8 ms to 1.0 ms,
63.8% of the step down to 0.75%:

| op | before | after |
|---|---:|---:|
| `cast` | 49.3 | 0.7 |
| `not` | 47.5 | 0.0 |
| `less` | 47.1 | 0.1 |
| `or` | 36.2 | 0.0 |
| `sub` | 28.7 | 0.1 |
| `and` | 27.9 | 0.0 |
| **chain** | **236.8** | **1.0** |

The call counts are identical on both sides (66 casts, 3 nots, and so on): the
ops still run, they just run on `[1, 1, S]` instead of `[1, S, S]`. Removing the
chain outright would now be worth 0.75% of a decode step, so there is no
remaining headroom in it. GQA goes from 27.3% of the step to 74.5% purely by
subtraction, and all non-GQA work put together is 34.0 ms of 133.5.

The second check is whether any extent is still an upper bound, since the
fallback is where this fix stops. Instrumenting the classification and compiling
Gemma-4: every dynamic axis the op slices resolves from its bounds, four times
out of four, and the only data-dim sizings are the untouched batch axis where
the data dim is exact. So on this model the fix is not partial -- there is no
slice left whose consumers see an inflated shape, which is also why the debug
line added here prints nothing on it.

Reproducing the 16k arm needs #787's prefill tiling in the tree; on this branch
alone 16k prefill exhausts the workspace and the process fast-fails before it
generates a token. The 2,240 arms have std 0.51 and 0.39 ms against a 5.02 ms
delta, so even the smallest real win is well clear of noise, and it matches the
model: per-op profiling put the mask chain at 126 ms at `S = 12277`, and
`126 * (2270 / 12277)^2 = 4.3 ms` predicted against 5.0 ms measured.

## Notes for reviewers

- The interesting judgement call is the fallback direction. An unresolvable bound
  keeps the old upper bound, which is wrong in the same way as before but is what
  every model on this path already runs with. `CompressConversion` solves the
  same problem exactly with a `hip.ReadbackDimOp`, and I originally justified not
  copying it on cost grounds -- wrongly, as it turns out: `wrap_slice` already
  D2Hs its bounds and syncs the stream every call, so this op pays a host sync
  per execution no matter what this pattern does. The honest position is that a
  readback would add a *second* sync point, ahead of the slice, to serve a case
  no model in scope reaches, and that host arithmetic beats a readback wherever
  it is available. If you would rather have the exact extent unconditionally,
  that is a smaller change than I first claimed and I am happy to make it. The
  fallback is also now visible under `-debug-only=convert-onnx-to-hip`.
- `resolveHostIndex` accepting both dialect spellings of the same arithmetic is
  the part I would still most like a second opinion on, but it is no longer a
  guess: measured on the model, the ONNX spelling is what fires, and the lowered
  spelling is now held up by a lit test rather than by an assumption. The
  alternative is to force an ordering between this pattern and
  `ShapeToTensorDims`, which still seems worse.
- Ops materialized by a resolution attempt that then fails are pure and left for
  DCE rather than tracked and erased.
- Peak memory does not improve because prefill, where the mask really is square,
  sets the high-water mark.
- Where the earlier note about GQA went. This PR made GQA 70% of the decode step,
  and per-op profiling reported it at 68 ms against 48 ms before at nearly
  identical `skv`, which I could not explain and still cannot -- it does not
  appear in the uninstrumented wall-clock numbers above. What profiling it did
  settle is what that 70% consists of, which was worth knowing: at 16k the KV
  cache copy is 79.5 ms/token of it, measured by ablating the copy (143.9 ->
  64.5 ms/token), because Gemma-4 exports `present` as
  `past_sequence_len + sequence_len`, so past and present are distinct buffers
  and every layer rewrites the whole cache every token. #780 (merged) cuts that
  copy directly. The sliding-window read, which looked like the obvious suspect
  because the local layers were reading the entire cache to mask 15/16ths of it
  away, is worth only ~18 ms/token by comparison (#795, draft). Neither changes
  anything in this PR; recording it so the next person does not re-derive it.

## Checklist

- [x] Substantial AI assistance is disclosed, and I reviewed and understand the
      result.

AI assistance: the diagnosis (IR dumps, the fenced RGP dispatch capture, the
prefill-vs-decode per-op comparison that isolated the square mask), the patch,
the tests and the benchmark runs were produced with Cursor. All measurements are
from real runs on the machine named, and I reviewed the change.

### DIFF
--- a/lib/Conversion/OnnxToHip/SliceConversion.cpp
+++ b/lib/Conversion/OnnxToHip/SliceConversion.cpp
@@ -35,8 +35,21 @@ namespace {
 //     steps. Produces a native `hip.slice` DPS op whose runtime function is
 //     a stub today (throws); models that hit this path are unsupported until
 //     the runtime is implemented, but the conversion / bufferization pipeline
-//     can still verify and link the IR. Dynamic output dims are sourced from
-//     `tensor.dim` on `data` (an upper bound — Slice cannot widen any axis).
+//     can still verify and link the IR. Dynamic output dims are computed from
+//     the slice bounds when those are host-resolvable, and otherwise fall back
+//     to `tensor.dim` on `data` (an upper bound — Slice cannot widen any axis).
+//
+// That fallback is only an upper bound, and an upper bound is not a safe
+// stand-in for the extent: the extent this pattern puts on the `tensor.empty`
+// init IS the output shape as far as everything downstream is concerned, since
+// `tensor.dim` of a DPS result resolves through to the init. A Slice that keeps
+// one row of a 12k-row tensor therefore hands every consumer a 12k-row shape.
+// Gemma-4 26B-A4B decode does exactly this — `Slice(CumSum(attention_mask),
+// Shape(attn)[1] - Shape(ids)[1], Shape(attn)[1])` takes the single current
+// query position — and the inflated extent turned the causal mask from
+// [1, 1, S] into [1, S, S], 60% of a decode step. Hence resolveHostIndex:
+// `ends - starts` is host arithmetic over `onnx.Shape` entries, so the true
+// extent is computable without any device readback.
 
 /// Return the dense-elements attribute backing \p value if it can be
 /// determined at compile time. Recognizes arith constants, inspectable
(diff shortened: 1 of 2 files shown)


BoarQing:
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F275%2F&base_job=daily_hip_ep_genai&base=275&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F294&curr_job=daily_hip_ep_oga&curr=294 Gemma4
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F324%2F&base_job=daily_hip_ep_genai&base=324&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F334&curr_job=daily_hip_ep_genai&curr=332) General

http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/310/ general acc
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/334/ general perf

BoarQing:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1747/
http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1746/

amd-mingw (approved):

PR #786, opened 2026-08-20
diag(runtime): env-gated allocation tracing for every growable buffer

BoarQing:
## Summary

One commit. Env-gated allocation tracing, added because a memory-growth
investigation on Gemma-4 26B-A4B could not be done with what the runtime reports
today.

Every grow-on-demand buffer in `hipdnn_ep_runtime_state.cpp` prints the size the
caller asked for next to the size actually requested from the driver, plus a
per-site and an all-site high-water mark. The gap between those two numbers is
the whole question when a `hipMalloc` fails: today a failure is indistinguishable
between "demand genuinely rose" and "a growth policy inflated the request", and
the existing error message prints only the inflated figure. None of these buffers
ever shrink, so the all-site sum is the standing residency a failing allocation
has to fit alongside — which no individual site can report.

The decomposed attention path additionally prints its per-call workspace
composition, because the workspace total cannot separate the quadratic score
terms from the linear ones, and cannot say which layer's geometry triggered a
grow.

Off unless `HIPDNN_EP_ALLOC_TRACE=1`. No behaviour change when unset: the gate is
a function-local static read once, and every call site is inside it.

## Why this is a separate PR

It is the measurement apparatus for #TBD_CHUNK and #TBD_QUANTUM, which are based
on it, and it is what makes the numbers in those two reproducible. It stands
alone and can be merged, held, or dropped independently — neither fix needs it at
runtime.

## Output format

```
ALLOC_TRACE site=<name> event=cold|grow needed=<bytes> alloc=<bytes> old=<bytes> hwm=<bytes> hwm_all=<bytes>
GQA_WS layer=<n> H=.. G=.. d=.. sq=.. total_seq=.. qtrans=.. kexp=.. vexp=.. sf32=.. sf16=.. o=.. gemm_ws=.. temp_end=.. total=..
```

An example of what it resolves, from a 16K Gemma-4 prefill — three workspace
allocations in one prefill, the third asking for 11.5 GiB more than any caller
requested:

```
workspace event=cold needed=4220321296  alloc=4220321296
workspace event=grow needed=25854112864 alloc=25854112864
workspace event=grow needed=26386297952 alloc=38781169296   <- 1.5x of the previous buffer
```

## Test plan

- [x] Swept 2K/4K/8K/12K/14K/15K/16K prompts on Gemma-4 26B-A4B; trace lines
      parse cleanly and the reported per-site sums reconcile with the totals
- [x] Score-buffer figures cross-checked against a closed form: `sf32 + sf16`
      is exactly `96 * S^2` bytes at every length measured
- [x] Unset-by-default confirmed: runs with the variable absent produce no
      output and no measurable TTFT change

## Notes for the reviewer

Redirect stderr at the process level when collecting these. PowerShell wraps a
native process's stderr into ErrorRecords, word-wraps them at the console width
and prefixes each with `python.exe :`, which splits trace lines mid-field.

### DIFF
--- a/lib/Runtime/hipdnn_ep_runtime_state.cpp
+++ b/lib/Runtime/hipdnn_ep_runtime_state.cpp
@@ -32,6 +32,82 @@ static int per_entry_load_constants(RuntimeState *state,
                                     morphizen::FileSystem *fs,
                                     const char *constants_filename);
 
+//===----------------------------------------------------------------------===//
+// Allocation tracing (diagnostic, HIPDNN_EP_ALLOC_TRACE=1)
+//===----------------------------------------------------------------------===//
+//
+// Every grow-on-demand buffer in this file reports both the size its caller
+// asked for and the size actually requested from the driver. The two differ
+// wherever a growth-amortisation rule applies, so an inflated request is
+// attributable to the policy rather than to a genuine rise in demand -- which
+// is not otherwise distinguishable from the failure message alone.
+//
+// Off unless the env var is set; when off this costs one relaxed load of a
+// function-local static per allocation, and allocations here are rare by
+// construction (every one of these buffers never shrinks).
+static bool alloc_trace_enabled() {
+  static const bool enabled = [] {
+    const char *v = std::getenv("HIPDNN_EP_ALLOC_TRACE");
+    return v && std::strcmp(v, "0") != 0;
+  }();
+  return enabled;
+}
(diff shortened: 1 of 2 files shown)


BoarQing:
Folding this out: the fixes should sit on merged code and carry only the fix itself. Reopening later if the tracing is wanted on its own.

PR #787, opened 2026-08-20
fix(gqa): tile the decomposed prefill over query rows (16K prefill allocatable, -89% workspace)

BoarQing:
## Summary

One commit on `main`. Makes the decomposed attention prefill's memory linear in
sequence length instead of quadratic, which is what stops a Gemma-4 26B-A4B
prefill above 12K from allocating at all on a 64 GB shared-memory part.

The decomposed path materialises the whole `[B*H, sq, total_seq]` score matrix
**twice**: once in fp32, which the bias add, the causal mask and the softmax
reduction need, and once in the GEMM input dtype as the probabilities the Value
GEMM consumes. Measured, the pair is exactly `96 * S^2` bytes - 23.58 GiB at a
16K prompt, or 97.9% of the workspace. Every other buffer on this path (Q, K, V,
O) is linear in `sq` and they come to under half a gigabyte between them. A
single block that size is not something the driver will hand out while the
weights are already resident, so the op does not run slowly, it fails.

Nothing here needs all query rows resident at once. **The softmax reduces along
`total_seq`, so every query row is independent of every other**: a block of rows
can be scored, biased, masked, softmaxed and multiplied by V to completion before
the next block starts. This is a tiling of the same arithmetic, not an
approximation - and unlike a fused flash kernel it needs no running maximum and
no rescaling, because each row still sees its whole key range inside one chunk.

Only the score buffers shrink. Q, K, V and O stay whole, so Q is read and O is
written through a per-chunk offset while their batch strides keep describing the
full `sq`.

Chunk size comes from a byte budget rather than a row count, because what has to
fit is bytes. `HIPDNN_EP_GQA_SCORE_BUDGET_MB` defaults to 1024, well above
anything that already runs: nothing tiles below roughly 3.5K tokens on a 16-head
model, so shorter sequences keep their existing launch counts and descriptor
cache entries and take a single-iteration loop whose pointers and strides all
reduce to what they were. `=0` restores the untiled path for A/B.

## Measurements

Gemma-4 26B-A4B, `max_length=16512` held constant so the KV cache is fixed and
prompt length is the only variable. "peak" is the largest single workspace block
requested from the driver; "allocs" counts workspace allocations in one prefill.

| prompt | base allocs | base peak | chunk | allocs | peak | peak drop |
|---|---|---|---|---|---|---|
| 12K | 3 | 20.77 GiB | 896 | **1** | 2.25 GiB | -89.2% |
| 14K | 3 | 27.06 GiB -> **abort** | 768 | **1** | 2.93 GiB | -89.2% |
| 15K | 3 | 30.93 GiB -> **abort** | 640 | **1** | 3.36 GiB | -89.1% |
| 16K | 3 | 36.12 GiB -> **abort** | 640 | **1** | 3.93 GiB | -89.1% |

14K, 15K and 16K go from aborting to completing. The score pair at 12K drops from
14.47 GB to 1006 MiB.

The allocation count falling to 1 is a side effect worth calling out: once the
quadratic term is gone, attention's workspace demand lands below the size the
workspace already had from an earlier consumer, so it never grows. The
three-allocations-per-prefill churn disappears without touching the allocator.

**Provenance of the two allocation columns.** Per-allocation request and granted
sizes are not printed by any build in this PR; I read them from an env-gated
`hipMalloc`/`hipFree` tracing patch applied locally over the same tree. That
patch is deliberately not part of this PR. Everything else here (pass/fail, TTFT,
generated text) is observable from a stock build of this branch.

**Re-verified on this PR's actual merge base.** The table above was collected on a
tree that also carried that tracing patch and the unmerged slice fix from #782.
Since neither is on `main`, I rebuilt this branch on the merge base alone
(`6c4544c2`) and re-ran the outcome that matters:

| prompt | budget | exit | result |
|---|---|---|---|
| 12K | 0 (untiled) | 0 | completes |
| 12K | 1024 (tiled) | 0 | completes, TTFT 26.5 s, 16 tokens |
| 16K | 1024 (tiled) | 0 | **completes** |
| 16K | 0 (untiled) | `0xC0000409` | **abort** |

So the unblock is a property of this commit on `main`, not of anything it was
measured on top of. #782 is not a prerequisite.

### TTFT

Paired A/B on one build at a 12K prompt, toggling only the budget. Warmup
iteration discarded, arm order alternated between rounds so thermal drift does
not land on the same arm twice:

| round | untiled | tiled | delta |
|---|---|---|---|
| 1 | 25272.9 ms | 25436.4 ms | +0.65% |
| 2 | 25528.9 ms | 25543.8 ms | +0.06% |
| mean | 25400.9 ms | 25490.1 ms | **+0.35%** |

At 14K/15K/16K there is no comparison arm - the untiled path does not allocate.

:warning: Do not use per-length sweep TTFT for this. Each sweep arm is a fresh
process with one measured iteration, so it carries model compilation; the untiled
sweep is non-monotonic in prompt length (4K slower than 12K) for that reason. An
earlier reading of 245 s for tiled 12K was that artefact, not the change.

Note also that the harness's "Peak Memory Usage" line is host RSS sampled once at
exit (`psutil` `memory_info().rss`), so it does not move with this change and is
not the metric in the tables above.

## Implementation notes

- **Bias row offset is passed to the kernel, not folded into the pointer.** The
  bias plane stride is set by the full query extent while the score plane stride
  is set by the chunk; with `bias_batch` or `bias_heads` above 1 those differ, so
  folding would silently mis-address every plane after the first. This is the one
  kernel signature change (`hip_gqa_add_attention_bias_f32` gains `bias_sq` and
  `bias_row_offset`); it has a single caller.
- **The mask needs nothing new.** It already derives each row's absolute query
  position as `past_len + row`, so advancing `past_len` by the chunk start puts
  both the causal triangle and the sliding window where they belong.
- **A ragged final chunk is a second GEMM descriptor.** Both shapes are resolved
  before the workspace is sized, since the GEMM workspace region has to cover
  whichever algorithm asks for more.
- **The no-expand flavour is excluded.** Its Q and O operands are laid out
  `[B, G, HPG, sq, d]` with `n = HPG*sq`, so a range of query rows is not a
  contiguous range of `n` and the offset trick does not hold. That path is off by
  default for prefill.

## Test plan

- [x] Builds clean on `main` (`6c4544c2`); only pre-existing kernel warnings
- [x] 12K/14K/15K/16K on Gemma-4 26B-A4B, allocation trace verified per arm
- [x] 12K and 16K, tiled and untiled, re-run on the merge base with no other patches
- [x] Paired TTFT A/B at 12K (above)
- [x] Equivalence against the untiled path at 4K, three arms in one script
- [ ] **No unit-level numeric test** - see below

## Correctness evidence, and its limit

I could not run `test/numeric`: it needs `onnxruntime_morphizen_ep.dll`, which
has no target in this build tree. Equivalence is therefore model-level, greedy
decoding, 32 tokens, one 4K prompt, three arms differing only in budget:

| arm | chunks | result |
|---|---|---|
| `budget=0` | 1 (untiled) | reference |
| `budget=1024` | 2, offset 2688 in use | identical for 26 of 32 words; differs only on a trailing token |
| `budget=17` | 92, ragged tail | diverges at word 8, stays coherent, and still reads an exact fact out of the prompt ("repeats the same paragraph 28 times") |

Why I read the 92-chunk divergence as accumulation order rather than a
positional error: only prefill is tiled (decode runs at `sq==1`), and the last
query row is the one the first token is read from. Its mask row sits at the very
end of the bias, in chunk 91 at local row 1. If the row offset were being dropped
or misapplied, that row would attend to the first two positions instead of all
4096 and the first token would be garbage. It is not - the first token is
identical across all three arms, and the arm still extracts a precise count from
the prompt body. Divergence appears later, which is where fp differences
propagate through the KV cache that subsequent layers write.

That is an argument, not a unit test. If you want the stronger check before this
lands, the numeric suite needs its EP shim buildable here, or a standalone
`onnx.Attention` model with a bias at a length long enough to force tiling.

## Prior art you should know about

Commit `579ad66a` ("fix(gemma4): make the 16K VLM prefill correct and
allocatable", on `perf/qwen-ttft-stack`, never merged) already solved this by
**head grouping** - splitting steps 8-10 over groups of heads sized to a budget,
with benchmarked TTFT (39.8 s for 8 groups of 2 heads against 38.1 s for 2 groups
of 8 at 16K).

I chose query-row tiling over extracting that, for two reasons. It declines on
`B > 1`, a per-head bias, a head sink, or smooth softmax, because grouping
renumbers the head index the kernels see - so gpt-oss, which uses head sinks,
would never get the fix. And it can only divide by divisors of `B*H`, which
floors the score buffers at one head's worth (1.58 GiB at 16K), where a byte
budget goes as low as asked. Happy to be argued out of this if the benchmarked
implementation is preferred; the two are alternatives, not complements.

### DIFF
--- a/lib/Runtime/Kernels/hip/gqa_kernel.hip
+++ b/lib/Runtime/Kernels/hip/gqa_kernel.hip
@@ -1064,11 +1064,16 @@ extern "C" int hip_gqa_causal_mask_f32(
   GQA_RETURN_LAUNCH_STATUS();
 }
 
+// `sq` counts the query rows present in `scores`, which is a chunk of the full
+// query range when the caller is tiling the prefill. `bias_sq` is the full query
+// extent of the bias tensor and `bias_row_offset` locates the chunk inside it --
+// the two cannot be folded together because the bias plane stride is set by the
+// full extent while the score plane stride is set by the chunk.
 template <typename TBias>
 __global__ void add_attention_bias_kernel(
     float* __restrict__ scores, const TBias* __restrict__ bias,
     int total_heads, int num_heads, int bias_batch, int bias_heads, int sq,
-    int total_seq, int score_batch_stride) {
+    int total_seq, int score_batch_stride, int bias_sq, int bias_row_offset) {
   int head = blockIdx.y;
   int tid = blockIdx.x * blockDim.x + threadIdx.x;
   int n = sq * total_seq;
@@ -1080,9 +1085,10 @@ __global__ void add_attention_bias_kernel(
   int h = head % num_heads;
   int bias_b = (bias_batch == 1) ? 0 : b;
   int bias_h = (bias_heads == 1) ? 0 : h;
-  size_t bias_idx = static_cast<size_t>(bias_b) * bias_heads * sq * total_seq +
-                    static_cast<size_t>(bias_h) * sq * total_seq +
-                    static_cast<size_t>(row) * total_seq + col;
+  size_t bias_plane = static_cast<size_t>(bias_sq) * total_seq;
+  size_t bias_idx = static_cast<size_t>(bias_b) * bias_heads * bias_plane +
+                    static_cast<size_t>(bias_h) * bias_plane +
+                    static_cast<size_t>(bias_row_offset + row) * total_seq + col;
   scores[static_cast<size_t>(head) * score_batch_stride + tid] +=
       static_cast<float>(bias[bias_idx]);
 }
@@ -1090,10 +1096,13 @@ __global__ void add_attention_bias_kernel(
 extern "C" int hip_gqa_add_attention_bias_f32(
     void* stream_ptr, void* scores, const void* bias, int total_heads,
     int num_heads, int bias_batch, int bias_heads, int sq, int total_seq,
-    int score_batch_stride, int bias_element_size_bytes) {
+    int score_batch_stride, int bias_element_size_bytes, int bias_sq,
+    int bias_row_offset) {
   if (!scores || !bias)
     return -1;
   hipStream_t stream = static_cast<hipStream_t>(stream_ptr);
+  if (bias_sq <= 0)
+    bias_sq = sq;
   int n = sq * total_seq;
   int blocksX = (n + GQA_THREADS - 1) / GQA_THREADS;
   (void)hipGetLastError();
@@ -1102,13 +1111,13 @@ extern "C" int hip_gqa_add_attention_bias_f32(
         <<<dim3(blocksX, total_heads), GQA_THREADS, 0, stream>>>(
             static_cast<float*>(scores), static_cast<const __half*>(bias),
             total_heads, num_heads, bias_batch, bias_heads, sq, total_seq,
-            score_batch_stride);
+            score_batch_stride, bias_sq, bias_row_offset);
   } else if (bias_element_size_bytes == 4) {
     add_attention_bias_kernel<float>
         <<<dim3(blocksX, total_heads), GQA_THREADS, 0, stream>>>(
             static_cast<float*>(scores), static_cast<const float*>(bias),
             total_heads, num_heads, bias_batch, bias_heads, sq, total_seq,
-            score_batch_stride);
+            score_batch_stride, bias_sq, bias_row_offset);
   } else {
     fprintf(stderr,
             "hip_gqa_add_attention_bias_f32: unsupported bias elem size %d\n",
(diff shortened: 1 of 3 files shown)


BoarQing:
Perf Comparsion on Gemma http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F332&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F336%2F&curr_job=&curr=

BoarQing (commented):
1

BoarQing:
Acc:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/315/
http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1755/
Perf:
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F324&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F335&curr_job=&curr=

amd-mingw (approved):

PR #789, opened 2026-08-20
feat(hip-ep): register com.amd::QMoE as a schema-only custom op

AMDmoore:
## Summary
- Adds `backend-mlir-compiler/custom-op-schema/`, a self-registering `Ort::CustomOpDomain` module: a single concrete `hipep::SchemaOnlyCustomOp` class (parameterized only by op name; `UNDEFINED` element type + `VARIADIC` input/output characteristics skip ORT's type/count validation) plus a `kSchemaOnlyOps` table in `schema_only_custom_op.cpp`. Adding a new schema-only op is a single-line table entry -- no new file, no CMakeLists.txt edit.
- Registers `com.amd::QMoE` via this table, with a schema-only stub kernel (throws `ORT_NOT_IMPLEMENTED` if ever invoked) so ORT can recognize and `Resolve` graphs containing this op ahead of a real HIP dialect/conversion/kernel implementation.
- Wires `morphizen-core`'s `CollectCustomOpDomains()` into `MorphiZenEpFactory::GetNumCustomOpDomainsImpl`/`GetCustomOpDomainsImpl` so the new plugin-EP-v2 ABI surfaces these domains to ORT (the old-ABI `initialize_onnxruntime_morphizen_ep()` path is untouched and not exercised by hip-ep). The domains are collected once into a `MorphiZenEpFactory::custom_op_domains_` instance member in the constructor, tying their lifetime to the factory instance instead of a process-level static cache.
- Adds a Python test that builds a minimal single-node `com.amd::QMoE` graph and verifies ORT can build a session against `hipgpu.dll`.

## Test plan
- [x] `test/python/test_qmoe_amd_schema_resolve.py` passes locally against a built `hipgpu.dll` (single-node graph resolves and session creation succeeds).
- [ ] CI build + existing test suite green.


AMDmoore:
Rebased onto latest \main\ (was out-of-date). Also added a follow-up commit (\ix(hip-ep): give schema-only ops a concrete output type\) found while validating this PR against the real Nemotron-3-Super-120B QMoE graph: \SchemaOnlyCustomOp\ defaulted every op's output to \UNDEFINED\+\VARIADIC\, which made \Graph::UpdateShapeInference\ fail with a type-inference error on the first real graph where the op's output tensor wasn't already explicitly typed (QMoE's output feeds an untyped \Add\). Parameterized the class by output type (QMoE -> \FLOAT16\) to fix it; verified locally against the full 120B model that \Model::Load\ now succeeds and the graph proceeds past QMoE schema resolution.

amd-mingw (approved):

PR #791, opened 2026-08-20
feat: fuse conv + ReLU/ReLU6 via parametric MIOpen activation

amd-mounikk:
## Summary
- Add `hip-fuse-conv-relu6` pass to fold direct `hip.conv` + `hip.max` (+ optional `hip.min`) into `hip.conv { fused_activation = true, activation_clip_lo, activation_clip_hi }`.
- Support **plain ReLU** (ResNet50: clip_lo=0, clip_hi=0) and **ReLU6** (MobileNet: clip_hi=6) through one parametric runtime path instead of a fixed activation enum.
- Extend `wrap_miopenConvolutionForward` with `fused_activation` + clip bounds; apply MIOpen ReLU / ClippedReLU after conv+bias, with optional FusionPlan (Conv+Bias+Activation) when supported.
- Run the pass **after** `hip-externalize-constants` and resolve externalized splat clip constants (`memref.get_global` + `bufferization.to_tensor`) so ResNet50 ReLU zeros fuse correctly.
- FusionPlan is auto-disabled on gfx115x (MIOpen limitation on gfx1151); set `HIPDNN_EP_CONV_FUSION=1` to force attempt or `=0` to force off.

## ResNet50 v1.5 fp16 results (gfx1151, batch=1, 224×224)
| Metric | Before | After | Change |
|--------|--------|-------|--------|
| Average inference time | 3.15 ms | 3.04 ms | ~3.3% faster |
| Elementwise kernel calls | 65 | 32 | −33 (−51%) |
| Direct Conv→ReLU fused | 0 | ~33 | ReLU folded into conv |

Remaining 32 elementwise ops are residual **Add** + **ReLU after Add** (not covered by this pass).

## Test plan
- [x] Pre-commit hooks pass on all changed files
- [x] LIT: `test/lit/Dialect/fuse-conv-relu.mlir`
- [x] LIT: `test/lit/Dialect/fuse-conv-relu6.mlir`
- [x] LIT: `test/lit/Dialect/fuse-conv-relu-externalized.mlir`
- [x] ResNet50 v1.5 fp16 ONNX compiles and runs on gfx1151
- [x] MobileNet v2 fp16 ONNX compiles and runs on gfx1151

## Known limitations
- FusionPlan not used on gfx115x by default (compile/execute failures on gfx1151 with MIOpen 3.5.1)
- Bias remains a separate `miopenOpTensor(ADD)` on gfx1151 fallback path
- Residual Add + ReLU after Add are not fused (future work: custom add+relu kernel)

### DIFF
--- a/include/hip/Dialect/IR/HipOps.td
+++ b/include/hip/Dialect/IR/HipOps.td
@@ -639,8 +639,13 @@ def Hip_ConvOp : Hip_DpsOp<"conv", /*traits=*/[OpStateOpInterface], /*outsAccess
     - pads: Array of padding values [top, left, bottom, right]
     - dilations: Array of dilation values [H, W]
     - group: Number of groups for grouped convolution
-    - activation: Optional fused post-conv activation. 0 = none (default),
-      1 = ReLU6 (Clip to [0, 6] via MIOpen FusionPlan CLIPPEDRELU).
+    - fused_activation: When true, apply a fused post-conv activation via MIOpen
+      after conv (+ bias). When false (default), no activation is applied.
+    - activation_clip_lo: Lower clip bound (default 0.0). Used when
+      fused_activation is true.
+    - activation_clip_hi: Upper clip bound (default 0.0). 0.0 means no upper
+      clip (plain ReLU when clip_lo is 0). ReLU6 uses clip_hi = 6.0.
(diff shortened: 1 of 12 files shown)


asparkhi:
MIOpen will be phased out. It is better to write a custom kernel. What do you suggest @amd-mingw ?

PR #795, opened 2026-08-21
perf(gqa): recover onnx.Attention's sliding window from its mask and narrow the KV cache to it

BoarQing:
## Summary

A windowed attention layer only ever attends to the last `W` key positions, but
the decomposed GQA path read, expanded and copied the entire KV cache on every
step regardless. This recovers the sliding window from `onnx.Attention`'s
additive mask subgraph — the only place a pre-opset-25 export can put one — and
then narrows the KV read, the expand and the separate-buffer copy to that window
on every call, prefill and decode alike.

On Gemma-4 26B-A4B-it (25 of 30 decoder layers windowed at `W = 1024`) this is
**1.60x steady-state decode at a 12K prompt** and **1.47x at 8K**. Initial
prefill and TTFT are unchanged by construction.

## Related issue or design

None. There was no design discussion before implementation, which CONTRIBUTING
asks for at this size — see the scope note under *Notes for reviewers*.

## Why

`onnx.Attention` gained window attributes only in opset 25. A model exported
below that with sliding-window layers — Gemma-4 among them — arrives with
`is_causal = 0` and the window expressed entirely inside its additive mask, so
nothing downstream knows the layer is windowed. The mask is applied faithfully
and the result is correct, but every step still pays for the full cache: at a
12K prompt a windowed layer scores 12K keys to keep 1024, and the other ~92% are
multiplied, exponentiated and then annihilated by the mask's large-negative
entries.

Attention is where that shows up on this model. The window is a hard promise
about which keys can matter, and the whole change is about pushing that promise
one level down, from "masked to zero after the fact" to "never read".

## What

Three commits, in dependency order.

**1. `feat(onnx-to-hip)`: recover the window from the mask.** A new
`AttentionWindowFold.cpp` pre-lowering pass matches the mask's boolean
keep-condition, `And(pad, Or(win, seg))`, recovers `W` from
`win = And(GreaterOrEqual(P, Q), Less(Sub(P, Q), W))`, and stamps
`hipdnn.local_window_size`, which `OnnxAttentionConversion` puts on `hip.gqa`.

The window leg has to satisfy two separate requirements, and the second is the
one worth reviewing closely.

*Identity.* The compare and the subtraction must reference the same two tensors
in the same order, so a mask that merely resembles the shape does not match.
That fixes which side is the query and which the key.

*Provenance.* Identity alone only says `P - Q < W` bounds the difference of two
tensors. The runtime then converts `W` into `kv_lo = abs_q - W + 1`, an
**absolute key position**. For that step to follow, `P` and `Q` have to be
positions on one common scale — same origin, unit stride. Structurally identical
arithmetic over tensors scaled by a block size or re-based per segment would
match just as well and yield a `W` in the wrong units, and the narrowing would
then drop keys the model wanted. So both sides must reach a **common**
`onnx.CumSum` or `onnx.Range` within a bounded walk. Requiring a shared node
rather than one per side is deliberate: two unrelated numberings each have a
position source, and subtracting across them measures nothing.

That requirement was written against the export, not guessed at. Walking the
Gemma-4 decoder shows the key side is `Unsqueeze(CumSum(attention_mask))` and
the query side is `Unsqueeze(Slice(CumSum(attention_mask)))` — the same CumSum
node, at depths 1 and 2.

The `seg` leg is Gemma-4's same-image exception, `And(Equal(bq, bk), bidx >= 0)`,
which lets tokens of one image attend to each other past the window. Recovering
a window in its presence is only sound because that leg is bounded, and what
bounds it is that the block index is a `CumSum` — a running count of block
starts, so equal values are necessarily within one block. The identical
expression tree over a document ID or a prefix-LM boundary has *unbounded*
reach. So `matchSegmentTerm` requires a `CumSum` ancestor of one of the `Equal`
operands within `kSegmentProvenanceDepth = 6`, rather than trusting the shape.

Both legs therefore ask for a `CumSum`, and — worth knowing when reading a dump
— not the same one. The export has exactly two: a block-index CumSum over the
image-block starts, which the segment leg reaches at depths 3 and 4, and a
position CumSum over `attention_mask`, which the window leg's `P` and `Q` share.
Finding the position CumSum behind a segment leg would say nothing about block
extent, which is why the segment check looks above the `Equal` operands
specifically rather than anywhere in the mask region.

Reading a declared `left_window_size` where an export has one is a separate and
strictly easier change, deliberately **not** made here — see the scope note.

**2. `perf(gqa)`: narrow the decode KV read.** `kv_lo` / `kv_span` replace `0` /
`total_seq` in the score and value GEMM descriptors, the expand source offsets,
the buffer sizing and the chunk budget.

**3. `perf(gqa)`: narrow the copy, and the read bound at every `sq`.** The
separate-buffer concat skips `[0, kv_lo)`, and the read bound generalizes from
decode to all `sq`:

```cpp
int64_t kv_span = total_seq;
int64_t kv_lo = 0;
if (local_window_size > 0 && !bidirectional_no_past && present_key &&
    present_value) {
  const int64_t lo = past_len - local_window_size + 1;
  if (lo > 0) { kv_lo = lo; kv_span = total_seq - kv_lo; }
}
```

### Why skipping the write is safe

A call's query rows occupy absolute positions `[past_len, past_len + sq)`, so
the oldest key any row may attend to is `past_len - W + 1`. One global lower
bound serves the whole call — rows above the first are left keys they do not
use, which the mask drops exactly as before, so no per-row state is needed.

`copy_lo` is `kv_lo`, deliberately the same value and not a second bound derived
in parallel. That equality is what closes the invariant:

- a call writes `[kv_lo, total_seq)` where `kv_lo = past_len - W + 1`;
- the next call has `past_len' = total_seq > past_len`, so its
  `kv_lo' = past_len' - W + 1 > kv_lo`;
- so **every read range is a strict subset of the previous write range**, for
  decode and prefill alike.

Applying the bound to prefill as well as decode is what makes that hold. Leaving
prefill unnarrowed — reading from 0 while decode writes from `kv_lo` — is
precisely the case that breaks it: a continuation prefill would read memory a
narrowed decode never wrote. Unwritten memory can hold a non-finite bit pattern,
and no finite mask suppresses it. The softmax's first pass reduces with `fmaxf`,
which ignores NaN, but the second pass accumulates it, so a single such key
poisons its whole output row.

At `sq == 1` the bound reduces to `total_seq - W`. At a first prefill
`past_len = 0` makes it non-positive and nothing narrows, so **initial prefill
and TTFT are bit-identical by construction**.

### Restrictions

- **`!bidirectional_no_past`.** The bound is derived from
  `total_seq = past_len + sq`, and `bidirectional_no_past` (Whisper encoder and
  cross-attention) is the one branch where that identity does not hold —
  `total_seq = skv`, `past_len = 0`. With mask recovery as the only producer of
  a window, that pairing cannot currently occur: a window implies a mask, a mask
  means `attention_bias` is non-null, and `bidirectional_no_past` is
  `no_causal && !attention_bias`. The gate is there anyway, because that is a
  property of the converter rather than of the runtime function's contract — a
  future producer reading opset 25's `left_window_size` would need no mask at
  all and would arrive with both set.
- **A present cache is required.** `kv_lo` indexes key positions by absolute
  position in a BNSD cache page; without `present_key` / `present_value` the
  GEMMs read the raw operands, where a `past_len` derived from `skv - sq` does
  not describe an offset into anything.
- **`no_causal` is deliberately not a gate.** It was one while a window could
  only arrive as an attribute, where it would have been ignored rather than
  applied. A window read out of the mask is enforced by that mask, so gating on
  `no_causal` would reject the only shape that needs narrowing.

## Test plan

Hardware: AMD Radeon 8050S (Strix Halo, gfx1151), Windows, HIP 7.2.53150.

**Lit** — `check-hip-mlir-lit`: 470 tests, 458 passed, 12 unsupported, 0 failed.

The six window-recovery cases are four negative controls and two positives, each
isolating one requirement:

| case | expected |
| --- | --- |
| windowed mask, indices from one CumSum, segment block index from another | `local_window_size = 1024` |
| global mask, identical but for the window leg | nothing |
| unrecognized OR leg above a valid window leg | nothing |
| segment leg with its CumSum removed, nothing else changed | nothing |
| window leg whose indices come straight from graph inputs | nothing |
| window leg whose indices come from two *different* CumSums | nothing |

The last three each pass if the corresponding requirement is dropped, which is
why they are there. The final one is what distinguishes "each side has a position
source" from "both sides share one".

**Numeric** (`test/numeric`, `--backend ort_ep --ep-name hipgpu`):

| suite | result |
| --- | --- |
| whisper cross / encoder / self + MHA (97) | **97 pass** |
| `test_gqa.py` (22) | 14 fail, the identical set as on main |

The 97 whisper/MHA cases are the regression cover for the
`!bidirectional_no_past` gate: they are exactly the `total_seq = skv`,
`past_len = 0` geometry it excludes. `test_gqa.py`'s 14 failures were diffed
against a main-only arm and are the same set, so they are pre-existing and
unrelated (a `seqlens_k` geometry problem in the converter).

**No numeric test for the narrowing itself is included**, and that is a real gap
rather than an oversight — see the scope note. No existing numeric test drives a
windowed layer through `gqa_forward_hipblaslt`: both windowed cases in
`test_gqa.py` use `d == 64` and route to the fused kernels instead.

**What the local validation found, for whoever closes that gap.** A harness at
Gemma-4's geometry (H=16, G=8, d=256) was used while developing this, not
included here. Two results from it are worth recording because they constrain
what a future test has to assert:

- Poisoning `[0, kv_lo)` with NaN after a narrowed decode at `past_seq = 2048`,
  then running a continuation prefill at `past_len = 2049`, gives an output that
  is **entirely NaN** without commit 3's generalized bound and NaN-free with it.
  That is the unwritten-prefix hazard, driven directly.
- Rebuilt with the copy bound one position above the read bound
  (`copy_lo = kv_lo + 1`), the attention-output comparison **passes** at cosine
  0.999884 (max diff 2.7e-3) while the in-window slice of `present` is wrong in
  2047 of 2,111,488 elements. One missing key out of 1024 does not move a
  softmax enough to trip a 2e-2 tolerance, so **an output-only test would not
  catch this**. Whatever covers this later has to assert `present` from the same
  `kv_lo` the runtime narrows to — and deliberately not the dead region, since
  asserting anything about memory nobody promised anything about, even that it
  is zero, would be asserting on the allocator.

**Recovery and narrowing on the real model** — `HIPDNN_EP_DEBUG=1`, 1800-token
prompt plus generation, 270 `[REAL] GQA hipBLASLt` lines:

| phase | calls | narrowed | `kv_span` |
| --- | ---: | ---: | --- |
| initial prefill `sq = 2068` | 30 | 0 | full range |
| decode `sq = 1` | 240 | **200** | **1024** |

The 200 / 40 split of the decode calls is exactly the export's 25 windowed
against 5 global layers over 8 steps; `kv_span` is pinned at 1024 on every
narrowed call and `kv_lo` tracks `past_len - 1023` (`kv_lo = 1045` at
`total_seq = 2069`). All 30 initial-prefill calls run their full range, which is
the measured form of "TTFT is untouched". The same run on an arm built before the
provenance check was added produces an identical narrowing profile, so the check
does not cost any recovery on this export.

## Performance

Interleaved three-arm A/B against `52886b28`, round-robin with the order
reversed on even rounds, one process per measurement so no arm inherits
another's warm caches, 48 tokens per iteration.

Steady-state decode, p20 of per-step times:

| prompt | main `52886b28` | this PR | speedup |
| --- | ---: | ---: | ---: |
| 8192 | 95.0 ms | 64.8 ms | **1.47x** |
| 12288 | 129.5 ms | 81.1 ms | **1.60x** |

TTFT, median over post-compile iterations — flat to within 3%, as designed:

| prompt | main | this PR |
| --- | ---: | ---: |
| 8192 | 19.2 s | 19.0 s |
| 12288 | 34.8 s | 34.2 s |

Two measurement notes, both of which forced the statistic:

- **Iteration 1 of every process is 4-6x slower than iteration 2** (main at 12K:
  554 ms then 131 ms) in every arm, because graph compilation lands on the first
  one. A pooled median is therefore bimodal by construction. The p20 above
  tracks the iteration-2 median to within 1 ms. The pooled median over all
  samples, including that regime, is the conservative floor: 223.0 → 196.8 ms
  at 8K (1.13x) and 321.8 → 269.9 ms at 12K (1.19x).
- **48 tokens, not 100.** A 100-token run at 12K measured ~66 ms for 10 steps,
  ~81 ms through step 54, then jumped to ~520 ms and stayed there. That cliff is
  a memory-pressure regime change, not attention scaling, and any statistic
  spanning it reports which side of it the run landed on.

Run-to-run spread within an arm is under 1%: main `[95.3, 95.0]` and
`[130.4, 129.5]`, this PR `[64.9, 64.8]` and `[80.8, 81.8]`.

Whether the correctness work costs anything was measured separately, four paired
rounds at 8192: 64.8 ms with the `sq == 1` restriction against 64.9 with the
generalized bound. Parity — the `bidirectional_no_past` gate is inert on this
model and the prefill bound is inert at `sq == 1`.

The branch has since been rebased onto `f219dd21`. The commits between that and
the measured base are hipsr dialect work, a move of the `HIPDNN_EP_DATATYPE_*`
defines into a shared header with values unchanged, and shape-fold patterns;
none touch `gqa.cpp`, the GQA kernels or the attention path, so the A/B was not
re-run against them. Lit and the numeric suites above were re-run on the rebased
tree.

## Notes for reviewers

**The 1.90x at 16K in an earlier version of this description is withdrawn.** It
was baselined against `6f334726` and cannot be reproduced against current main —
not because the optimization changed, but because 16K does not complete on
current main at all. `bench_seqlen.py 16384` dies with `STATUS_HEAP_CORRUPTION`
(`0xC0000374`) about 130 s in, around the end of prefill, on **plain
`52886b28`** with none of this PR's code present. `HIPDNN_EP_DEBUG=1` masks it,
which is the signature of a race rather than a deterministic overrun. 8192 and
12288 were stable across every run, so the numbers above are at those lengths.
The 16K crash is a separate pre-existing bug and is not addressed here.

**The unwritten-prefix hazard is real at the operator level but latent on this
model.** The multi-turn path that would produce `sq > 1` with `past_len > 0`,
`og.Generator.append_tokens`, fails before reaching the EP's attention — and
fails identically on an arm without this PR's provenance work, re-confirmed
today:

```
The output OrtValue provided for output 'present.0.key' has shape
{1,8,2154,256} but the computed output shape for this run is {1,8,2153,256}
```

genai pre-allocates `past + sq` while ORT computes one less. So the generalized
bound closes the hole *before* users can reach it, rather than repairing a
regression they can hit today. Claiming an observable model-level fix would be
false.

**Scope.** 1662 insertions and 97 deletions across 9 files, under CONTRIBUTING's
2000-line threshold, of which 630 lines are lit tests. Four things were removed
from an earlier revision of this PR to get there and to keep the diff to what
the measurement actually justifies:

- the opset-25 `left_window_size` / `right_window_size` attribute path and its
  three lit cases. Gemma-4 exports at opset 24, so that path never fired in any
  measurement here and could not be tested against a real export;
- a `DEBUG_TYPE` move (briefly its own PR, #880, now closed). It was warning
  hygiene only — MSVC C4005, non-fatal — and current main has three sibling
  files raising the same warning, so it belongs in a cleanup of its own rather
  than here. This PR's new file uses the header's `DEBUG_TYPE` and adds no
  warning;
- a `window=` field on the runtime debug line;
- `test/numeric/tests/test_attention_sliding_window.py`. This is the removal
  worth arguing about: it is the only thing that exercised the narrowing
  numerically, and dropping it leaves the gap documented under *Test plan*.
  Restoring it is a straightforward follow-up and I will do that if a reviewer
  would rather have the 690 lines than the threshold.

The three commits are individually reviewable and in dependency order — recover
the window, use it for the read, then extend it to the copy and to prefill — but
they are not independently *shippable*: commit 1 alone stamps an attribute
nothing consumes, and commit 2 alone narrows the read while the copy still
writes the full range, which is the inconsistent state commit 3 resolves.

**Where to look first.** The two provenance checks in
`AttentionWindowFold.cpp`, `hasSharedPositionSource` and `matchSegmentTerm`'s
`CumSum` requirement, are the places where a wrong answer is silent rather than
loud. Each decides whether a mask actually enforces the window being recovered,
and a false accept would narrow against a promise the model does not make. Both
were written after walking the real export rather than from the shape alone:
both `Equal` operands in the segment leg trace to the same block-index `CumSum`
at depths 3 and 4, in both the windowed and the global mask and in both copies
of the export, and the window leg's two indices trace to a single position
`CumSum` at depths 1 and 2.

## AI disclosure

Substantial AI assistance (Cursor), disclosed per CONTRIBUTING and trailered on
each commit. The mask-topology matcher, the narrowing arithmetic and this
description were AI-assisted. Validation was run locally and is reported above
rather than asserted: the mask shape, the segment `CumSum` provenance and the
position provenance of the window leg's indices were each confirmed by walking
the actual export before the predicate was written; the `present`-assertion
coverage was confirmed by deliberately breaking the copy bound and checking that
an output-only comparison would have missed it; and the performance claims come
from the interleaved A/B above, with the earlier 16K claim withdrawn because it
could not be reproduced. I reviewed the result and own every claim here.

## Checklist

### DIFF
--- a/lib/Conversion/OnnxToHip/AttentionWindowFold.cpp
+++ b/lib/Conversion/OnnxToHip/AttentionWindowFold.cpp
@@ -1,0 +1,632 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+//===- AttentionWindowFold.cpp - Recover a sliding window from the mask ---===//
+//
+// `onnx.Attention` had no way to declare a sliding window until opset 25, which
+// added `left_window_size` / `right_window_size`. At opset 23 and 24 the op's
+// attributes are only `is_causal`, `kv_num_heads`, `q_num_heads`,
+// `qk_matmul_output_mode`, `scale`, `softcap` and `softmax_precision`, so a
+// model whose attention is windowed has nowhere to say so and bakes the window
+// into the additive float mask instead: the mask is 0 for the key positions a
+// query may attend to and a large negative value everywhere else. The op then
+// reaches the runtime with `local_window_size == -1`, so the decomposed path
+// scores and copies the whole key range even though everything older than the
+// window is already -inf in the scores.
+//
+// So this rewrite recovers the window from the mask's producing subgraph and
+// stamps it as `hipdnn.local_window_size`, which OnnxAttentionConversion reads
+// onto `hip.gqa`. This is inference about what the model happens to have built,
+// not the model telling us, and it is here to serve the opset-24 exports that
+// have no way to tell us. Reading a declared `left_window_size` where one
+// exists is a separate and strictly easier change, deliberately not made here:
+// no export on hand has the attribute, so it could not be tested against
+// anything real.
+//
+// `hipdnn.local_window_size = N` means N attended keys *including* the current
+// one, matching `com.microsoft.GroupQueryAttention` (whose own definition moved
+// to include the current token in onnxruntime PR 25927) and the runtime's
+// `kv_lo = abs_q - local_window_size + 1`. A HuggingFace `sliding_window: 1024`
+// is `hipdnn.local_window_size = 1024`, and the mask's `P - Q < 1024` below
+// carries that same count, so W transfers with no adjustment.
+//
+// What is matched
+// ---------------
+// Gemma-4's 25 windowed layers and its 5 global layers share one mask shape and
+// differ by exactly one node, which is what makes this recognizable:
+//
+//   windowed:  keep = And(pad, Or(win, seg))
+//   global:    keep = And(pad, Or(GreaterOrEqual(P,Q), seg))
+//   both:      mask = Where(keep, 0.0, -65504.0)
+//
+//   with  win = And(GreaterOrEqual(P,Q), Less(Sub(P,Q), 1024))
+//         seg = the same-image-block term described below
+//
+// The window leg is `And(P >= Q, P - Q < W)` where the compare and the
+// subtraction reference the SAME two values in the SAME order. That identity
+// fixes the two roles relative to each other: the causal conjunct `P >= Q`
+// makes P the query side and Q the key side, since the reverse reading would
+// keep only future keys, which no decoder mask does. So `P - Q < W` reads
+// `q - k < W`, i.e. `k >= q - W + 1`, which is exactly the runtime's
+// `kv_lo = abs_q - local_window_size + 1`.
+//
+// Identity alone is not enough, though, and this is the part worth being
+// careful about. It says `P - Q < W` bounds the difference of two tensors; the
+// runtime then narrows KV cache reads by ABSOLUTE POSITION. For the second to
+// follow from the first, P and Q have to be positions on one common scale --
+// same origin, unit stride. Structurally identical arithmetic over tensors that
+// were scaled or re-based would match just as well and yield a W in the wrong
+// units, and the narrowing would then drop keys the model wanted.
+//
+// So P and Q must additionally be drawn from the SAME position sequence: both
+// must reach a common `onnx.CumSum` or `onnx.Range` within
+// `kPositionProvenanceDepth`. That is the standard way a decoder builds
+// positions -- a cumulative count over the attention mask, or a plain range --
+// and sharing the node is what makes the subtraction a genuine distance rather
+// than a difference of two unrelated quantities. On the Gemma-4 export the key
+// side is `Unsqueeze(CumSum(attention_mask))` and the query side is
+// `Unsqueeze(Slice(CumSum(attention_mask)))`, the same CumSum reached at depths
+// 1 and 2, so the bound is small. Where both index tensors are recognizable
+// unsqueezes their axes are checked as further corroboration.
+//
+// How the legs combine
+// --------------------
+// The keep-condition is a monotone And/Or tree, and the bound propagates
+// through it the only way it can:
+//
+//   * AND: a conjunct can only remove keeps, so an unrecognized conjunct is
+//     harmless and any bounded conjunct bounds the whole node (min of the
+//     bounds). This is what absorbs the padding leg for free.
+//   * OR: a disjunct can add keeps, so EVERY disjunct must be bounded or the
+//     node is unbounded (max of the bounds).
+//
+// The segment exception
+// ---------------------
+// Gemma-4's mask ORs in a same-image-block bidirectional term,
+// `And(Equal(a,b), a >= 0)`. It is identically false for a text-only prompt, so
+// narrowing is then exact. With images it keeps keys in the query's own image
+// block, and can therefore reach past the window only if a single image block
+// is longer than the window -- 256 soft tokens per image against a 1024 window
+// here, a 4x margin. That bound is a property of the input rather than of the
+// graph, so it cannot be proven here; this rewrite recognizes the term
+// explicitly and accepts it, rather than accepting anything it does not
+// understand. A leg that is neither windowed nor this exact shape makes the
+// whole match fail and the op keep its full range.
+//
+// `And(Equal(a,b), Cmp(a, const))` on its own is too weak to accept, because it
+// is also how a same-document mask or a prefix-LM mask is spelled, and those
+// reach arbitrarily far back. What separates the block-index reading is where
+// the compared values come from: a block index is a running count, so an
+// `onnx.CumSum` is an ancestor of the equality's operands. Requiring that
+// ancestor is what makes the leg identifiable rather than merely plausible --
+// a document id read straight out of an input tensor has no CumSum behind it
+// and is declined. On the Gemma-4 export the CumSum sits 3 and 4 nodes above
+// the two `Equal` operands respectively, so `kSegmentProvenanceDepth` only has
+// to be a small constant.
+//
+// Both legs therefore ask for a CumSum, for different reasons and -- worth
+// knowing when reading a dump -- not the same one. The export has exactly two:
+// a block-index CumSum over the image-block starts, which is the one the
+// segment leg reaches, and a position CumSum over `attention_mask`, which is
+// the one the window leg's P and Q share. Finding the position CumSum behind a
+// segment leg would say nothing about block extent, which is why the segment
+// check looks above the `Equal` operands specifically rather than anywhere in
+// the mask region.
+//
+// Implementation notes
+// --------------------
+//   * Roots on `onnx.Attention`. Idempotent: bails if the attribute is already
+//     set, so the greedy `ExistingOps` pre-lowering loop quiesces.
+//   * Must run BEFORE `convertComputeOps`, whose greedy pattern set would have
+//     rewritten the mask's producers into `hip.*` ops in an order this cannot
+//     rely on. Pre-lowering still sees generic `onnx.*` with inline constants.
+//   * Recognition is conservative throughout: every unhandled shape declines
+//     with a breadcrumb and leaves the op scoring its full key range, which is
+//     what it does today.
+//
+//===----------------------------------------------------------------------===//
+
+#include "OnnxToHipUtils.h"
+
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "llvm/ADT/SmallPtrSet.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/Statistic.h"
+#include "llvm/Support/Debug.h"
+
+#include <algorithm>
+#include <optional>
+
+// DEBUG_TYPE comes from OnnxToHipUtils.h ("convert-onnx-to-hip"), shared with
+// the rest of this directory rather than defined per file here, so that this
+// change does not also have to move the header's definition.
+
+STATISTIC(NumAttentionWindowStamps,
+          "Number of onnx.Attention ops stamped with a sliding window "
+          "recovered from the additive mask subgraph");
+
+namespace mlir {
+namespace hip {
+
+namespace {
+
+/// Recursion and work caps. The real subgraphs sit at depth 7 with a few dozen
+/// nodes; these are loose enough not to bind and tight enough that a
+/// pathological graph cannot make the pre-lowering loop quadratic.
+constexpr int kMaxDepth = 16;
+constexpr int kMaxVisited = 128;
+
+/// The mask value must be negative enough to be a "drop" marker rather than a
+/// bias. fp16's most negative finite value (-65504) is the usual choice; -inf
+/// and float's -3.4e38 also appear.
+constexpr double kDropThreshold = -1.0e4;
+
+static llvm::StringRef opName(mlir::Value v) {
+  mlir::Operation *def = v ? v.getDefiningOp() : nullptr;
+  return def ? def->getName().getStringRef() : llvm::StringRef();
+}
+
+/// Return `v`'s value when it is an inline integer scalar: `onnx.Constant` with
+/// either a rank-0/single-element `value` tensor or the `value_int` attribute
+/// (both forms appear in the same graph -- initializers import as the former,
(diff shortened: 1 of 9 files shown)


BoarQing:
Force-pushed: rebased onto current `main`, retargeted onto #880, and the three
commits were rewritten rather than appended to, so previously reviewed SHAs are
gone. What changed since the last version, beyond the rebase:

**Three correctness fixes**, folded into the commits whose code they correct:

1. **The narrowing is now gated on `!bidirectional_no_past`.** That is the one
   branch where `total_seq != past_len + sq`, so the bound was being computed
   against the wrong absolute query position. Reachable rather than theoretical:
   the attribute path stamps a window before it looks at the mask operand.
2. **The read bound now applies at every `sq`, not just `sq == 1`.**
   `kv_lo = max(0, past_len - W + 1)`. The previous version narrowed the decode
   read and the copy but left prefill reading from 0, so a continuation prefill
   could read cache the narrowed decode never wrote. The new numeric case fails
   on the old code with every output element NaN.
3. **`matchSegmentTerm` now requires `CumSum` provenance.** It previously
   accepted any `And(Equal(x, y), Cmp(x, const))`, which is also the spelling of
   a document-ID or prefix-LM mask whose reach is unbounded. Verified against
   the actual export before writing the predicate: both `Equal` operands trace
   to the same `CumSum` at depths 3 and 4.

**The 1.90x at 16K is withdrawn.** 16K does not complete on current `main` at
all — `STATUS_HEAP_CORRUPTION` about 130 s in, reproducing on plain `52886b28`
with none of this PR's code present. Re-baselined at 8K and 12K instead:
1.47x and 1.60x steady-state decode. Details and the measurement method are in
the description.

Also corrected two claims in the previous description that were wrong: that no
producer pairs `no_causal` with a window (the attribute path does), and the
"nobody reads the unwritten prefix" argument (replaced with the closed
invariant, which is what fix 2 is for).

BoarQing:
Force-pushed again: unstacked from #880 (now closed) and rebased directly onto `main` at `49308bec`. This PR is no longer stacked and reviews standalone.

#880's two lines are folded into the first commit here, which is the one that adds the file needing them. Rationale for the change of approach: the `DEBUG_TYPE` redefinition is MSVC C4005, a *warning*, and the object builds regardless — I verified by restoring the header's `#define` and compiling. So a separate stacked PR was more process than the problem warranted.

Re-verified after the rebase: `AttentionWindowFold.cpp` and `SliceConversion.cpp` compile with no C4005 and no new warnings, `check-hip-mlir-lit` is 470 / 458 passed / 12 unsupported / 0 failed, and `pre-commit run --all-files` is clean. The four files the numeric and perf validation ran against are byte-identical to what was measured, so those results carry over unchanged.

`main` moved from `52886b28` (the perf baseline) to `49308bec`. The three commits in between are hipsr dialect work plus a move of the `HIPDNN_EP_DATATYPE_*` defines into a shared header with values unchanged; none touch `gqa.cpp`, the GQA kernels or the attention path, so the A/B was not re-run.

BoarQing:
Perf General: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F599&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F596&curr_job=&curr= (tie)
Acc General: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?id=6a95a8eebf0bd6c7b64cfca1&database=development&collection=daily_hip_ep_genai&build_id=597
Perf Gemma: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F605&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F614&curr_job=&curr= (+~100% at 16K)
Acc Gemma: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1771/

amd-mingw (approved):

PR #801, opened 2026-08-21
ci(windows): build LLVM and ORT once in a prerequisite job

zz002:
## Summary

Splits the Windows CI into one dispatcher plus four reusable workflows, and moves every heavy from-source dependency — LLVM, ONNX Runtime, amdgpu-ep and OGA — into a prerequisite workflow that builds each one once per ref. `windows-build.yml` goes from 1366 lines to 48 and now holds nothing but the trigger set and the job graph. Rebased onto current `main` (includes #856/#859 push triggers and artifact retention). Dispatcher caller jobs renamed for clearer PR check names (`prebuild`, `build_real`, `build_mock`, `gpu_test`).

## Related issue or design

Same cache-pressure area as #785, but independent of it. See Notes for reviewers for how this interacts with the restore-only follow-up that #785 proposes.

## Why

`windows-build.yml` and `windows-build-mock.yml` each carried their own copy of the LLVM and ORT from-source recipes, roughly 130 lines apiece, kept byte-identical by hand so both would resolve to the same cache keys. That shares the cache *entry* but not the *work*: the two workflows are independent, so on a cold cache they start in parallel, both miss, and each runs the same multi-hour LLVM build.

Making the dependency build a prerequisite job fixes that structurally. The build jobs cannot start before it finishes, so a cold cache pays for LLVM once per ref rather than once per build job, and there is now a single copy of each recipe with the pins that drive its cache key sitting next to it.

Splitting the rest into their own files is the other half of it: with the heavy steps gone the dispatcher is just the dependency graph, and each job is readable without scrolling past the other three.

## What

- **`windows-deps.yml`** (new): four jobs. `llvm` and `ort` are independent; `amdgpu` and `oga` both `needs: ort` because they link against the patched `ort-install`, and they run in parallel with each other. Each job probes its own cache with `lookup-only: true`, so a hit costs one cache API call instead of a multi-GB download, and every step below the probe — including restoring `ort-install` — is conditional on the miss. Exports the four cache keys. Holds the `LLVM_TAG` / `ONNXRUNTIME_*` / `OGA_*` pins, since a callee's `env` cannot be supplied through `with`.
- **amdgpu-ep and OGA moved out of the real build.** Both used to be built inside the real job, so their cost sat on the critical path of every real build and the two could not overlap. As deps jobs they run in parallel, and `build_real` only restores their artifacts. `build_real` consequently drops the `onnxruntime_version` / `onnxruntime_pr_patches` inputs, which existed only to compose those two cache keys.
- **amdgpu-ep caches two DLLs instead of a build tree.** The cache entry was the whole `build-amdgpu` directory, hundreds of MB of object files, while the only things any downstream step consumes are `amdgpu-ep.dll` and `hip-backend.dll` (via `HIP_WHEEL_EXTRA_DLLS` and the staging step). It now caches an `amdgpu-artifacts/` directory holding exactly those two, matching the shape `oga-artifacts` already had, and the consumers reference them by fixed path instead of `find`. Its key also takes the `windows-` prefix that the Linux workflow already uses for its own entries; the OGA key is deliberately left as-is so existing entries keep hitting.
- **`windows-build-real.yml`** and **`windows-gpu-test.yml`** (new), **`windows-build-mock.yml`** (converted to `workflow_call`): the LLVM and ORT build steps are replaced by `actions/cache/restore` with `fail-on-cache-miss`, which makes the contract explicit — a green `prebuild` run means both entries exist, so a miss is a wiring bug rather than a cue to rebuild.
- The `llvm` job gets its own sccache namespace (`sccache-windows-deps-llvm-`), and its sccache restore is gated on the install-prefix miss so a warm run does not download a compiler cache it will never use. The `ort` job has no sccache at all: its `build.bat` call passes no compiler launcher.
- `prebuild` does not cancel in progress. A cold LLVM build killed mid-flight leaves the cache unpopulated and the next run pays for it again.
- Concurrency groups keep their existing names and move to job level, which is mandatory for the mock: workflow-level `concurrency` has no effect in a reusable workflow, so leaving it at the top would have silently dropped its cancel-on-new-push behaviour.
- Dispatcher caller jobs renamed (`deps`→`prebuild`, `real`→`build_real`, `mock`→`build_mock`, `gpu`→`gpu_test`) so PR checks read `Windows Build / prebuild / llvm` instead of `Windows Build / deps / llvm`.
- Docs that pointed at `windows-build.yml` for a step that moved now point at the file holding it. The ORT source-build recipe in `whisper_quick_start.md` also picks up the current `1.27.0` pin and loses its PR-patch step, `ONNXRUNTIME_PR_PATCHES` being empty now; PyPI's `onnxruntime-directml` still tops out at 1.24.4, so the source build itself is still needed.

## Test plan

- [x] `pre-commit run --all-files` passes with nothing rewritten
- [x] All five workflows parse, every caller `with` block matches the callee's declared `inputs` exactly, and every `needs.prebuild.outputs.*` reference is declared by the deps workflow
- [x] Every migrated step compared step-by-step against the pre-split files and byte-identical, including the amdgpu-ep and OGA checkout / patch / build steps now in deps. Intentional differences only: each deps job's own sccache namespace, the amdgpu `dep-` to `windows-` key rename, and the amdgpu build appending a copy of its two DLLs into `amdgpu-artifacts/`
- [x] The OGA cache key is character-for-character what it was, so the move does not invalidate it
- [x] Every `steps.*`, `env.*` and `inputs.*` reference resolves within its own job — migrated bodies keep their original step ids, so their `if:` conditions needed no rewriting
- [x] Warm path measured (pre-rebase): `prebuild / llvm` and `prebuild / ort` hit and returned in 4 s and 6 s; `prebuild / oga` hit in 7 s; `prebuild / amdgpu` missed once on the renamed key, rebuilt in 12m32s, after which `build_real / build` restored all four prefixes and `build_mock / build` restored only the two it needs
- [ ] `build_real / build`, `build_mock / build` and `gpu_test / test` finish green after rebase
- [ ] Cold path: bump a pin temporarily and confirm each dependency builds exactly once and is consumed by both downstream jobs

## Notes for reviewers

**Required status checks must be reconfigured before this can merge.** They are now two-level (`<caller job> / <callee job>`), and `Windows Build (Mock)` disappears as its own workflow entry:

- gone: `build-windows`, `build-windows-mock`
- new, all under `Windows Build`: `prebuild / llvm`, `prebuild / ort`, `prebuild / amdgpu`, `prebuild / oga`, `build_real / build`, `build_mock / build`, `gpu_test / test`

**This split constrains the restore-only follow-up floated in #785.** Saving `dep-llvm` / `dep-ort` on `main` only works inside a single job: a PR misses, builds, uses the result, and skips the save. Across a job boundary it does not, because the install prefix can only reach the build jobs through the cache, so a PR that misses and does not save leaves `fail-on-cache-miss` with nothing to restore. Cache behaviour here is unchanged from before the split — a PR that misses still writes to its own scope, exactly as both build jobs already did — so this is not a regression, but that follow-up will need artifact hand-off or a different split.

`cmake/deps.cmake:390` still points at a `windows-build.yml` step named "Build protobuf from source" that no longer exists anywhere. It predates this PR and is left alone.

The three deps pins live in `windows-deps.yml` while `linux-build.yml` keeps its own copies, so the LLVM and ORT pins are duplicated across two files rather than three. Unifying them across platforms is out of scope here.

## Checklist


amd-mingw (approved):

amd-mingw:
image

PR #802, opened 2026-08-21
Pack static high-rank elementwise into the 4-D HIP path

tuhinp-amd:
## Summary
- Pack static high-rank ONNX Add, Sub, Mul, Div, Less, and Greater into rank <= 4 with `tensor.collapse_shape` / `tensor.expand_shape` when each collapsed group is either fully broadcast or fully present.
- Do this as a convert-onnx-to-hip **pre-lowering** rewrite on the ONNX ops, then map the packed rank-<=4 op to HIP. Rank-0 scalars stay scalars; shorter ranks are right-aligned with leading 1s.
- Leave the original rank when packing is unsafe or dynamic, so HIP-to-LLVM can still reject rank > 4 today and a future high-rank backend needs no frontend change. Rank 7 packs when grouping is safe.

### Packing procedure

HIP binary elementwise lowerings only accept a 4-D broadcast descriptor. Before compute conversion, `populatePackBroadcastTo4DPatterns` rewrites a static rank > 4 ONNX binary op into a rank-<=4 one without changing ONNX Numpy-style broadcasting.

1. Require a static result rank greater than 4 and static operand shapes. Otherwise packing fails and the original rank is kept.
2. Right-align both operands to the result rank by inserting leading size-1 dims (`tensor.expand_shape`). A rank-0 scalar is left as-is because it already broadcasts through the 4-D path.
3. Search contiguous partitions of the rank axes, preferring 4 groups then 3, 2, and 1. `collapse_shape` cannot reorder axes.
4. Accept a group only if each operand is either fully broadcast in that group (product 1) or fully present (product equals the result group product). Mixing a broadcast axis with a data axis in one group would change which values participate after collapse.
5. Collapse lhs and rhs with that reassociation, emit the same ONNX op at the packed rank, then expand the result back to the original ONNX shape. Compute conversion then creates `hip.*` at rank <= 4.

If no valid partition exists, convert-onnx-to-hip emits `hip.*` at the original rank.

## Test plan
- [ ] LIT: `test_pack_broadcast_to_4d.mlir`, `test_elementwise_high_rank_fallback.mlir`, and the rank-5/6 cases in Add/Sub/Mul/Div/Less/Greater
- [ ] Full MorphizenMLIRLitTests


asparkhi (changes requested):
LGTM @tuhinp-amd. Can we please add following types of tests?

- Neg: High-rank broadcasts that cannot be safely packed (should remain unchanged)
- Neg: Dynamic shapes (should bail out)
- Neg: Non-static ranks

Also, would it make sense to have C API tests for covering variety of shapes with rightAlignTensor()?

asparkhi (approved):

PR #806, opened 2026-08-24
feat(hipsr): convert onnx.Reshape with a compile-time destination

zpye:
## Summary

Converts `onnx.Reshape` into a `hipsr.compute` holding a `tensor.collapse_shape`, a `tensor.expand_shape`, or both. Both bufferize to memrefs that alias their source, so the reshape costs nothing at runtime.

This is the alternative to #781: the same conversion, cut down to the compile-time destination.

## What it does

The target shape operand has to fold to a constant, so every result dimension is known at compile time apart from the one written as `-1`.

The result type is inferred from the operands rather than taken from the type ONNX declared, which shape inference is not required to have filled in. The folded operand states the dimensions, and the input gives the element type and the memory space the result aliases. Nothing widens the type back afterwards, so a consumer's adaptor and a graph output's signature both see the sharper one.

The shape region resolves the destination from the input's shape and reads no memory. A `-1` becomes `shape.num_elements` divided by the dimensions the operand states, unless the input is static, in which case it is resolved while the type is inferred.

Every regroup goes through the 1-D form rather than searching for a single reassociation, because `ComposeExpandOfCollapseOp` already finds that grouping.

Rejected with a TODO: a target shape that only holds its value at runtime, a `0` entry, and a nonzero `allowzero`. Rejected for good: a second `-1`, which has no answer, and an unranked input, which has no dimensions to build the 1-D form from.

## Against #781

#781 also handles the `0` entry and the host read behind a runtime target shape. Leaving both out takes `ReshapeConversion.cpp` from 676 lines to about 300, and drops the barrier placeholder, the runtime resolution behind it, and the shared-helper lift into `OnnxToHipsrUtils.h`.

Two differences beyond the narrowing. A target shape that does not fold is rejected here rather than falling back to the declared result type, so a graph building its target through Shape/Gather/Concat needs that chain folded before this pass. And the identity fold runs on the inferred type, so a reshape whose operand does not fold is rejected rather than replaced by its input.

## Notes for reviewers

Stacked on #779, which the inferred return type depends on. Review that first; this rebases onto `main` once it lands.

No HipsrToLLVM lowering yet, so `--hipsr-pipeline` cannot run this end to end.

Cursor (Claude Opus 5) assisted with the conversion and the tests. I reviewed the result and understand it.


wcy123 (approved):
LGTM

PR #819, opened 2026-08-25
feat(hipsr): convert onnx.Unsqueeze to a hipsr.compute

zpye:
## Summary

Converts `onnx.Unsqueeze` into a `hipsr.compute` holding a `tensor.expand_shape`, with the placeholder's shape region populated inline. The expansion bufferizes to a memref that aliases its source, so the unsqueeze costs nothing at runtime.

## What it does

The result type is inferred from the operands rather than taken from the type ONNX declared, which shape inference is not required to have filled in. The `axes` operand places the unit dimensions and the input gives the rest, along with the element type and the memory space an alias has to share with its source. This is the model #806 uses for its target shape. Reading the axes also resolves a dimension the declared type leaves dynamic: a declared `tensor<?x?x?xi1>` with axes `[1]` over a `tensor<?x?xi1>` input infers `tensor<?x1x?xi1>`.

`getReassociationIndicesForReshape` recovers the grouping from the two shapes and `tensor::ExpandShapeOp::inferOutputShape` builds the result dimensions from the input's, so the region needs no dimension arithmetic of its own. The body's expand takes its `output_shape` off the destination, as #806's does, so those dimensions are resolved once.

Rejected with a TODO: axes that only hold their value at runtime, which is the TODO #806 carries for a runtime target shape. Rejected for good: a repeated axis and an axis outside the result rank, both of which ONNX itself calls errors, and an unranked input, which has no dimensions for the result to take.

A rank-0 input is out of scope. The pass leaves a host root without a memory space and an aliasing expansion cannot name one, so the result names no space and the pass's own type rules turn it away. The form converts once its producer names a space.

## Notes for reviewers

Rebased onto `main` after #806 landed, so the single commit sits directly on it.

No HipsrToLLVM lowering yet, so `--hipsr-pipeline` cannot run this end to end.

Cursor (Claude Opus 5) assisted with the conversion, the tests, and the comments. I reviewed the result and understand it.

### DIFF
--- a/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
+++ b/include/hip/Conversion/OnnxToHipsr/OnnxToHipsr.h
@@ -50,6 +50,10 @@ void populateReshapeConversionPatterns(
     const ::mlir::TypeConverter &typeConverter,
     ::mlir::RewritePatternSet &patterns, ::mlir::MLIRContext *ctx);
 
+void populateUnsqueezeConversionPatterns(
+    const ::mlir::TypeConverter &typeConverter,
+    ::mlir::RewritePatternSet &patterns, ::mlir::MLIRContext *ctx);
+
 void populateReturnConversionPatterns(
     const ::mlir::TypeConverter &typeConverter,
     ::mlir::RewritePatternSet &patterns, ::mlir::MLIRContext *ctx);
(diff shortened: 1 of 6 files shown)


wcy123 (commented):
LGTM

wcy123 (approved):

PR #815, opened 2026-08-24
fix(onnx-to-hip): revert #782 slice dynamic extent computation

zz002:
## Summary

Revert #782 (`fix(onnx-to-hip): compute dynamic hip.slice extents from the slice bounds`).

## Related issue or design

Reverts #782

## Why

The dynamic `hip.slice` extent computation from slice bounds is being rolled back pending further investigation.

## What

Revert merge commit 86cc6bb0ba0f8888455dec5a2353d436d7b9e989, restoring the prior upper-bound sizing from the data dimension for unresolved dynamic extents in `SliceConversion.cpp` and the pre-#782 lit coverage in `test_slice.mlir`.

## Test plan

- [ ] `build-windows` green on cold cache
- [ ] `onnx-to-hip` lit suite passes (including `test_slice.mlir`)

## Notes for reviewers

None

## Checklist

- [x] The change is focused, or links a design/series explaining its scope.
- [ ] Relevant tests were added or updated and the results are documented.
- [ ] User-facing or design documentation was updated when needed.
- [x] Substantial AI assistance is disclosed, and I reviewed and understand the result.

PR #820, opened 2026-08-25
fix(onnx-to-hip): re-land #782 with a zero-length slice extent guard (2.99x decode TPS on Gemma-4 26B-A4B at 12K)

BoarQing:
## Summary

Re-lands #782, which #815 reverted, with the zero-extent regression fixed.

#782 sized every host-resolvable `hip.slice` extent exactly. That is right for the Gemma-4 decode mask it targeted and wrong for one value: zero. The extent lands on the `tensor.empty` init, so it is also the allocation, and `Slice(x, k, k, axis)` -- the ONNX empty-tensor idiom -- seeds a growing-Concat accumulator that a `hip.loop` appends into. `FixLoopAccumulatorOffset.cpp` states that dependency in its own header ("v_init is pre-sized to full capacity (the canonical `Slice(x, k, k, axis) -> Loop` pattern)"), but nothing on the producing side did, so the change never surfaced it. Qwen3.6-VL's windowed attention died in `memrefCopy(rank-2 strided) failed: invalid pitch argument`, the append subview's destination pitch being 0.

Commit 1 is `git cherry-pick 86cc6bb0`, verbatim, so the re-land stays reviewable as "exactly what #782 did". Commit 2 is the fix.

`SliceToHip` now takes the data dim whenever the slice length is zero: directly when both bounds are the same SSA value, and via an `arith.select` on the computed length otherwise. Gemma-4's length is `Shape(attn)[1] - Shape(ids)[1]`, 1 at decode and S at prefill, so the select never fires and the 12,277 -> 1 shrink the PR exists for is intact.

The rest is about not losing the invariant again. It lived in a file the original change never opened, so it is now stated from both ends: `SliceConversion.cpp` explains why the allocation is an upper bound and points at the pass, and the pass points back. The pass also refuses a statically empty accumulator axis rather than rewriting offsets into it -- a backstop, since the capacity is dynamic in the case that actually broke. `wrap_slice` warns once when a sliced axis gets a zero-capacity output from a non-empty input, which is the signature the failing model presented.

## Gemma-4 26B-A4B at 12,276 prompt tokens

5 runs, 2 warmup, 68 tokens generated on both arms. Baseline is `origin/main` at `0704691b`.

|                | baseline              | this branch           |         |
| -------------- | --------------------- | --------------------- | ------- |
| decode p50     | 359.59 ms/tok (2.78 TPS) | 120.16 ms/tok (8.32 TPS) | **2.99x** |
| TTFT p50       | 41,582 ms             | 40,693 ms             | unchanged |
| peak memory    | 31.08 GB              | 29.97 GB              | -1.11 GB |

Better than the 2.24x #782 measured, because #787's prefill tiling is in main now.

## Test plan

- [x] `check-hip-mlir-lit`: 391 passed, 0 failed. New test 11 pins the seed slice keeping `tensor.dim`; tests 7, 8 and 10 grow the capacity select.
- [x] All 5 VLM models via `vlm_benchmark.py` on gfx1151, coherent output, no pitch error:
  - `Qwen3.6-27B` (the regression case) 11.93 TPS
  - `Qwen3.6-35B-A3B` 48.90 TPS
  - `Qwen3.5-35B-A3B` 48.80 TPS
  - `Qwen3.5-9B` 32.87 TPS
  - `google-gemma-4-26B-A4B` 40.92 TPS
- [x] All 7 text models via `model_benchmark -l 512 -g 128 -r 3 -w 1`: Llama-3.1-8B 38.36, Mistral-7B 42.08, phi4-14B 21.52, Qwen2.5-14B 22.52, Qwen2.5-Coder-14B 22.89, gpt-oss-20b 60.30, DeepSeek-R1-Distill-Llama-70B 4.93 TPS.

### DIFF
--- a/lib/Conversion/OnnxToHip/SliceConversion.cpp
+++ b/lib/Conversion/OnnxToHip/SliceConversion.cpp
@@ -52,6 +52,18 @@ namespace {
 // [1, 1, S] into [1, S, S], 60% of a decode step. Hence resolveHostIndex:
 // `ends - starts` is host arithmetic over `onnx.Shape` entries, so the true
 // extent is computable without any device readback.
+//
+// The one extent that must NOT shrink is zero. The extent here is also the
+// allocation, and one consumer depends on that allocation being the upper
+// bound: FixLoopAccumulatorOffset.cpp rewrites a growing-Concat accumulator in
+// an outlined `hip.loop` body and, as its header states, "assumes ... v_init is
+// pre-sized to full capacity (the canonical `Slice(x, k, k, axis) -> Loop`
+// pattern)". `Slice(x, k, k, axis)` is the ONNX empty-tensor idiom used to seed
+// such an accumulator; its exact extent is 0, and a zero-capacity init leaves
+// the loop appending chunks into nothing -- on Qwen3.6-VL's windowed attention
+// that surfaced as `memrefCopy(rank-2 strided) failed: invalid pitch argument`,
+// because the append subview's destination pitch was 0. So an extent that
+// evaluates to zero falls back to the data dim; see buildAllocCapacity.
 
 /// Return the dense-elements attribute backing \p value if it can be
 /// determined at compile time. Recognizes arith constants, inspectable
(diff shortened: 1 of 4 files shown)


BoarQing:
baseline perf(done on 19): 
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F395%2F&base_job=daily_hip_ep_genai&base=398&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F398%2F&curr_job=daily_hip_ep_genai&curr=395
general acc : http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/399/

gemma perf:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/400/
gemma acc:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1765/

amd-mingw (approved):

PR #827, opened 2026-08-25
feat(hipsr): add the equal op, its ONNX conversion and its lowering

zpye:
## Summary

Adds `hipsr.equal`, the elementwise equality operation, with the `onnx.Equal` conversion, the shape-region recipe and the HipsrToLLVM lowering, so ONNX -> hipsr -> memref -> LLVM works end to end for it.

Equality against a token id is what produces the placement mask in the embedding graph. It is a broadcasting elementwise operation backed by a library call, so it is a dialect operation rather than a `hipsr.compute` body, and it follows the `hipsr.add` template. Part of the hipsr operation series, after #806 and #819.

## Why

**The mask is `ui8`, not `i1`.** The runtime writes one byte per element, so the byte states what the buffer holds, it is what the hip path already carries, and it keeps sub-byte rounding out of every pass that sizes the buffer. `onnx.Equal` keeps the bool its schema declares, so the conversion is the one place the two spellings meet.

**A host constant operand gets a device constant.** The embedding graph compares its ids against one id held as a rank-0 constant, which the constant conversion keeps on the host as a compile-time root. hipsr cannot decide that at the constant, because a rank-0 host root is what an axis or an index reads at compile time, so the use decides instead, as the hip dialect does by giving every `onnx.Constant` a carrier whatever its rank. The shape is left alone: `wrap_equal` takes each shape as 4D left-padded with 1, so a rank-0 operand already broadcasts. Only a constant is taken; a host operand without a compile-time value has nothing to carry, and two constants compare on the host.

## Notes for reviewers

`hipsr.constant` has no bufferization model, so a graph holding any constant still stops at One-Shot Bufferize, which is why `Pipelines/embedding.mlir` stays unsupported. The lowering rejects rank > 4, which is what the 4D broadcast descriptor API accepts.

Cursor (Claude Opus 5) assisted with the operation, the conversion, the lowering and the tests. I reviewed the result and understand it.


wcy123 (approved):
LGTM

PR #828, opened 2026-08-25
feat(hipsr): add the transpose op, its ONNX conversion and its lowering

zpye:
## Summary

Adds `hipsr.transpose`, the axis-permutation operation, with the `onnx.Transpose` conversion, the shape-region recipe and the HipsrToLLVM lowering, so ONNX -> hipsr -> memref -> LLVM works end to end for it.

The embedding graph transposes NonZero's `[rank, count]` index matrix into the `[count, rank]` layout ScatterND takes. Permutation is a library call rather than a few tensor operations, so it is a dialect operation, and it follows the `hipsr.equal` template. Part of the hipsr operation series, after #806, #819 and #827.

## Why

**`perm` is always spelled out.** ONNX leaves it out to mean the reverse permutation. Requiring it here keeps every later pass from re-deriving that default, so the conversion is the one place the two spellings meet.

**The input shape and `perm` go on the stack.** `wrap_transpose` reads both on the host. It also moves whole elements, so it takes an element width in bytes rather than a data type.

## Notes for reviewers

A host-resident input is rejected rather than copied, since the conversion emits no copy. The lowering rejects rank 0, as the runtime does.

Cursor (Claude Opus 5) assisted with the operation, the conversion, the shape region, the lowering and the tests. I reviewed the result and understand it.


wcy123 (approved):
LGTM

PR #834, opened 2026-08-25
perf(conv): take MIOpen out of every convolution path

BoarQing:
## Summary

No convolution in the EP calls MIOpen any more. `real/miopen.cpp` is deleted, along with `wrap_miopenConvolutionForward` and `wrap_miopenConvolutionTranspose`. MIOpen stays found and linked — elementwise, gemm, norm, activation, softmax and pow still use it — so this removes MIOpen from convolution, not from the EP.

Three paths moved, and one rewrite in here has a blast radius well past convolution:

1. **Forward `onnx.Conv`** — a new `hip_conv` kernel, plus two rewrites that divert most nodes to paths that already existed.
2. **`onnx.ConvTranspose`** — a new in-tree gather kernel, no atomics.
3. **Causal convolution past 8 taps** — dynamic-K kernels instead of the MIOpen fallback.
4. **`hip_transpose` is canonicalised before it picks a kernel.** This began as a fix for the patch-embed gap but applies to every `Transpose` in every graph, including models with no `Conv` at all. Reviewers should weigh it on its own.

## 1. Forward Conv

Scanning the 28 ONNX files in the local corpus: 8 contain `Conv`, 34 nodes in total, landing in three places.

| nodes | where they are | routed to |
|---|---|---|
| 24 | E2B + E4B audio encoders, `W=[1024,1,5]` `group=1024` `pads=[4,0]` | `hip_causal_conv_prefill` |
| 6 | 5 Qwen rank-5 patch embeds + the gemma3 rank-4 one | Gemm |
| 4 | E2B + E4B audio encoders, 3x3 stride-2 | `hip_conv` |

Only 4 nodes in the corpus reach the new kernel; the rest go to code that was already there.

- **The 24 depthwise causal convs** via a new `onnx.Conv` -> `hip.causal_conv_with_state` pattern in OnnxToHip. Placed there rather than in the ONNX loop so the existing canonicalizer absorbs the bracketing `Transpose[0,2,1]` pair, which is where most of the win comes from — not the convolution itself.
- **The patch embeds** by relaxing `PatchEmbedConvToGemm` from "kernel == input spatial and output spatial == 1" to "stride == kernel", then re-tightening it to require more than one element in the patch. A 1x1 kernel satisfies `stride == kernel` trivially but is not a patch embed: the split reshape is an identity and the two transposes degenerate into an NCHW <-> NHWC round trip around a contraction that was already a GEMM over C. DETR's ResNet-50 backbone has 33 such convs, and converting them took it from 94 transpose calls to 161.

Everything else `ConvConversion` accepts goes to `hip_conv`: an implicit GEMM over a never-materialised im2col matrix on the `hip_pool` per-axis ABI, fp32 accumulation, arbitrary stride/dilation/group/asymmetric pads, bias fused (removing MIOpen's second `miopenOpTensor` pass). It dispatches between WMMA tiles for fp16, scalar register-tiled ones for fp32/bf16, a dedicated depthwise kernel, and a direct kernel where M is too small to fill a tile row. Launch config comes from `multiProcessorCount` at runtime.

### The ladder, and the fault it hid

Shapes with few output channels cleared no tile and fell to `conv_direct`, which is not a small-problem kernel but a no-reuse one — a thread per output element, re-decoding the contraction index and re-reading the input on every tap. Three changes take work off it:

- A **16-row WMMA rung** accepts tiles whose rows are mostly padding, because wasted matrix-core work is nearly free and the decode is not. ESRGAN's 64->3 output conv at 1000x1000 goes 31.9 ms -> 4.3 ms.
- The **ladder runs twice**, the second pass dropping the requirement that a tile cover the device: half an idle device running a kernel with reuse beats a full one running a kernel without it. The SSD detection heads show it — a 1024-channel 10x10 map into 126 channels gathers 227 MB direct against roughly 12 MB tiled.
- **Depthwise gets its own kernel**, since one output channel per group clears no tile and the contraction is shared by the whole block.

The K-tail fix in the same commit is a real out-of-bounds read, not a tuning change: the staging loops walk a whole BK-deep slice whether or not the contraction has that many taps left, so `makeKSlice`'s past-the-end sentinel sat on the live path for every `K % BK != 0` shape. It set `c0 = in0i`, but `decodeOrigin` gives `org.x0 = o0*s0 - p0`, so on leading padded rows `org.x0 + c0` landed back inside the input and the gather read before the tensor. ESRGAN's head conv is the one corpus shape with `K = 27`; it faulted. Adding `pad_begin` restores the invariant.

## 2. ConvTranspose

The MIOpen transpose path is replaced by a gather-form kernel: each output element solves the inverse equation to find which input taps can reach it, so there are no atomics and no scatter. fp32 accumulation, ONNX `[C, M/group, kH, kW]` weight layout, bias fused — MIOpen could not fuse it here, and `miopenConvolutionForwardBias` was observed to double the deconvolution cost, so it had been running a separate `miopenOpTensor` add over the whole output. There is no op state, no solution cache and no workspace.

**It is a plain one-thread-per-output loop with no tiling.** That is deliberate for correctness on small models and it costs real performance — see the regression below.

## 3. Causal convolution

`kernel_size > 8` was the only live MIOpen route left in `causal_conv_with_state.cpp`; every other branch was dead on the `ndim` check, the activation mapping or the `element_size` validation. It now goes to dynamic-K kernels that take the width at runtime instead of by template unroll:

- **Channels-first prefill** uses dynamic LDS for both the window and the taps.
- **Channels-last prefill** replaces the fixed per-lane register window with a per-lane LDS ring of K floats, indexed `(i - t0) % K`.

The channels-last fold bound in `HipCausalConvCanonicalize` widens from 8 to 128 to match. It stops well short of the kernel's real ceiling, which is set by how much LDS a block can hold and is not knowable at compile time.

Removing the fallback also deleted the descriptor cache, `CausalConvKey`/`CausalConvKeyHash`, and the Find/Forward/bias/Activation chain — about 250 lines. The op state is now empty.

## 4. hip_transpose canonicalisation

Extent-1 axes drop, and axes that stay adjacent under the permutation merge. That is what lets the patch-embed output permute be recognised as the plain 2-D transpose it is and use the tiled kernel already in the file, instead of the generic per-element gather doing `rank` 64-bit divisions per element. With reciprocal division added to the generic path too, the patch-embed permutes go **0.80 ms -> 0.30 ms**.

## Performance

**8050S (32 CU) — not the 8060S the plan calls for.**

Two different baselines appear below and they are not interchangeable.

### A. Encoders, whole branch vs `main`

Wall-clock median, 30 iterations:

| encoder | MIOpen | this | |
|---|---|---|---|
| gemma4 E2B audio | 142.1 ms | 124.9 ms | **1.14x** (-17.2 ms) |
| gemma4 E4B audio | 132.3 ms | 125.0 ms | **1.06x** (-7.3 ms) |
| gemma3-4b vision | 1068.1 ms | 1067.1 ms | flat |

The audio win is mostly transposes, not convolution: 87 calls / 17.0 ms -> 63 calls / 8.8 ms on E2B. The 24 absorbed calls are exactly the 12 depthwise convs' bracketing pairs. Conv itself is 2.2 ms -> 1.0 ms. Vision is flat because its conv path is ~3% of the model, but that path improves 1.49x.

### B. Model zoo, ConvTranspose + causal-conv commit only

Baseline here is the branch **with** the forward-Conv work already in, isolating the last change. 39 model/shape combinations on both builds:

| | |
|---|---|
| running on both builds | 39 |
| failing identically on both (pre-existing) | 9 |
| **broken only by this build** | **0** |
| regressions beyond -3% | 1 |
| numeric deltas above 2% | 0 |

Everything except SAM2.1 lands within ±2%, including 12 conv-free LLM graphs run as a link/regression smoke pass, 7 Whisper encoders, and the Qwen decoder steady state.

### The regressions

**SAM2.1 decoder fp16: 122.5 -> 81.2 QPS, -33.7%.** This is the ConvTranspose kernel. Its 2 nodes go from 0.2 ms to 2.7 ms per inference, which accounts for most of the 4.3 ms end-to-end loss. The cause is the untiled gather described above, and the fix is to tile it — not done here.

**DETR fp16 remains slower than `main`.** The 1x1 guard recovered most of it, and the K-tail/ladder work took it from +36.7% to +28.2%, but it has not been closed.

**Whisper front end.** `whisper l0` and `whisper l1` go 1.494 ms -> 2.822 ms per encoder pass on the per-shape microbenchmark, though the end-to-end encoders measure within ±2% in table B. Models MIOpen served with Winograd on 3x3 stride-1 stay slower, because an im2col implicit GEMM does roughly 4x the arithmetic. Two attempts at the gather — LDS halo staging and a taller wave tile — both measured as net losses and were dropped; with these working sets inside the 32 MB MALL the redundant reads were already cache-served, so the gap is algorithmic and needs a fused Winograd path.

### Where hip_conv's time goes

Ablating whisper_l1 by compiling out one piece at a time:

| component | share |
|---|---|
| im2col gather | 51% |
| LDS staging + fragment loads | 33% |
| matrix cores | 16% |

The kernel is not short of arithmetic — it moves **236 MB to consume 21 MB of unique data**, at ~176 GB/s, close to what this part delivers. The gather cannot be made faster, only smaller. The 11x amplification is three things: re-reading the im2col matrix once per tile row (10x), duplicating each input element across the taps (3x), and stride-2 coalescing waste (2x). A 256-row WMMA tile halved the first. Closing the rest means staging an **input tile** rather than a K-slice, which for 2D needs the block to tile the output spatially rather than linearly — a different kernel, and **not done here**.

## Correctness

Every one of the 39 combinations is either bit-identical to the MIOpen build or inside its own run-to-run noise floor, measured by relative L2 against a same-build repeat run. The 5 Qwen decoders (24-48 `CausalConvWithState` nodes each) are bit-identical across both prefill and decode.

SAM2.1 fp16 is worth calling out because its raw L2 looks alarming until normalised: **0.0047%** new-vs-baseline against a **0.2734%** same-build noise floor. The model is nondeterministic run to run by roughly 58x the size of the A/B difference.

## Test plan

- [x] LIT suite passes, matching the pre-existing baseline. Updated to assert `wrap_conv` / `wrap_conv_transpose` and `CHECK-NOT` the MIOpen symbols, plus new regressions for the causal and patch-embed rewrites.
- [x] `test_whisper.py::_assert_no_silent_fallback` asserts `[REAL] wrap_conv` and the absence of `[REAL] wrap_miopenConvolutionForward`, so a lowering regression to MIOpen fails rather than passing quietly.
- [x] `test_causal_conv_with_state.py` extended with wide-kernel cases (k=9, 12, 16, 31) covering the dynamic-K path.
- [x] 39 model/shape combinations on both builds: 0 broken by this build, 0 numeric deltas above 2%.
- [x] Full numeric suite run: the pre-existing failures (`matmul_nbits`, `qmoe`, `one_hot`, `topk`) reproduce identically on a pre-change build.
- [x] pre-commit clean.
- [ ] **`hip_conv`, `conv_transpose` and the patch-embed rewrite have no in-tree numeric tests.** Files covering all of them were written and passing against the ORT CPU reference during development, but are not included in this PR. Reviewers should treat the kernels as covered by model-level A/B only.
- [ ] **Re-measure the whole branch against `main`.** Tables A and B have different baselines; no single run covers the branch end to end.
- [ ] **Re-run on the 8060S.** Everything above is 8050S.
- [ ] **Tile the ConvTranspose gather** to recover the SAM2.1 regression.
- [ ] Close the DETR gap against `main`.


BoarQing:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/423/

BoarQing:
Acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/477
Perf: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F462&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F476&curr_job=&curr=
Procyon: http://xsjvitisaijenkins:8080/job/daily_hip_ep_onnx/241/
Vision: http://xsjvitisaijenkins:8080/job/daily_hip_ep_onnx/242/

amd-mingw (approved):
LLM/VLM perf/acc ok, most procyon models seems ok, some vision models have gap, approve it and with follow-up PR to continue optimize perf.

PR #836, opened 2026-08-26
feat(hipsr): add the gather op, its ONNX conversion and its lowering

zpye:
## Summary

Adds `hipsr.gather` with its `onnx.Gather` conversion, shape-region recipe and HipsrToLLVM lowering, so an axis-indexed selection runs ONNX -> hipsr -> memref -> LLVM.

## Why

`onnx.Gather` appears twice in `Pipelines/embedding.mlir`, and the two uses are different operations wearing the same name.

Device data is a token-row lookup in the `[vocab, hidden]` table. That is a library call, so it becomes the dialect operation and lowers to `wrap_gather`.

Host data reads one dimension out of an `onnx.Shape` result, which already lives in `#hipsr.mem<host>`. Copying it to the device to launch a kernel over two integers is not worth it, so it becomes a `hipsr.compute` of scalar reads, and the chain Shape -> Gather -> Unsqueeze stays on the host.

## Notes for reviewers

`axis` is normalized during conversion, so the dialect operation only ever sees a non-negative one. The result type comes from the operands rather than from the type ONNX declared, and `inferGatherResultShape` states that shape rule once for both the verifier and the conversion.

The two forms the conversion turns down, an unranked operand and a host index known only at runtime, have no negative test on purpose: they are unsupported rather than invalid, so pinning those diagnostics would make implementing them later look like a regression.

Cursor (Claude Opus 5) assisted with the operation, the conversion, the shape region, the lowering and the tests. I reviewed the result and validated it with the LIT suite.


wcy123 (approved):
LGTM.

PR #837, opened 2026-08-26
feat(hipsr): add the slice op, its ONNX conversion and its lowering

zpye:
Adds `hipsr.slice`, the strided-window operation, with the `onnx.Slice`
conversion, its shape-region recipe and the HipsrToLLVM lowering, so
ONNX -> hipsr -> memref -> LLVM works end to end. The embedding graph needs it
to trim the flattened image features down to the image-token count.

Each window slot (`starts`, `ends`, `axes`, `steps`) has two forms, as
`hipsr.expand`'s shape does: an optional host operand and an optional
`DenseI64ArrayAttr`, exactly one per slot. Entries the compiler can read become
the attribute, entries the graph computes stay the operand, and ONNX's defaults
for an omitted `axes` or `steps` are written in, so nothing downstream carries
the default rule a second time. Entries keep their ONNX meaning: a negative
axis or bound counts back from the end, and a bound may fall outside the axis.

The forms decide the placeholder. A window entirely in attributes lets a
`normal` placeholder work the output shape out from the data's shape, while an
operand has to be read at run time, which takes a `barrier` placeholder holding
the data and that operand for its shape region to read.

The lowering calls `wrap_slice`, passing both shapes as host arrays and each
slot as one host pointer: a stack array for an attribute, the buffer itself for
an operand.

The runtime is unchanged. It copies the window in with an explicit
device-to-host copy, which suits the `hip` dialect's device window but not a
host one, and nothing calls it with a host window yet, so that change lands
with the pipeline that reaches the runtime and can test it.

Cursor (Claude Opus 5) assisted with the operation, the conversion, the shape
region, the lowering and the tests. I reviewed the result and understand it.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Made-with: Cursor


wcy123 (dismissed):
LGTM. Clean implementation following the `hipsr.gather` template with proper verification, shape region population, and lowering.

amd-mingw (approved):

PR #844, opened 2026-08-26
feat(hipsr): add the scatter_nd op, its ONNX conversion and its lowering

zpye:
Adds `hipsr.scatter_nd`, the indexed-slice overwrite operation, with the
`onnx.ScatterND` conversion, the shape-region recipe and the HipsrToLLVM
lowering, so ONNX -> hipsr -> memref -> LLVM works end to end for it. Writing
image features into token embeddings at the positions NonZero found is the last
operation in the embedding graph. Part of the hipsr operation series, after the
transpose, gather and slice ops.

The op takes `ctx`, `data`, `indices` and `updates` as device operands, and an
`init`. It copies `data` and overwrites the slice each row of `indices`
addresses with the matching slice of `updates`. The output keeps `data`'s
shape, so the placeholder needs only `data` and every extent stays symbolic, as
the embedding graph needs.

Overwriting is the only mode. ONNX ScatterND also reduces a duplicate index
into its destination through `reduction`, a read-modify-write rather than an
overwrite, so the conversion rejects those four modes instead of carrying an
attribute nothing can lower.

`wrap_scatter_nd` is the same 16-parameter function the hip dialect calls. Its
sixth pointer is a device count that trims padded index rows; this op has no
such operand, so it goes over as null and the runtime writes every row.

Stacked on #837 (slice), which is stacked on #836 (gather), so only the top
commit belongs to this PR. Retarget it to `main` once the two below it merge.

`Pipelines/embedding.mlir` stays unsupported for reasons outside this op: its
embedding table is an external constant whose data file the test tree does not
carry, `onnx.Equal` types the mask `ui8` where the graph's `onnx.Expand` reads
`i1`, and `onnx.NonZero` has no hipsr conversion. That graph's scatter itself
converts, takes a shape region and reaches the call.

Cursor (Claude Opus 5) assisted with the operation, the conversion, the shape
region, the lowering and the tests. I reviewed the result and understand it.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Made-with: Cursor


wcy123 (dismissed):
LGTM.

amd-mingw (approved):

PR #854, opened 2026-08-27
feat(hipsr): add the copy_d2h op, its bufferization and its lowering

zpye:
## Summary

Adds `hipsr.copy_d2h`, a destination-passing op that makes a device-to-host
transfer explicit in the IR, with its bufferization and its LLVM lowering to a
new `wrap_copy_d2h` runtime symbol.

No callers yet. This is the prerequisite for rewiring `hipsr.nonzero` to
produce a device-side count that the host can read.

## Why

Neither `memref.copy` nor `bufferization.materialize_in_destination` carries a
stream, so neither can express a transfer ordered against GPU work. Without an
op, reading a GPU-computed value on the host is a bare load, which reads stale
memory wherever the pool is real device memory.

## Notes for reviewers

- `wrap_copy_d2h` has no numeric or E2E coverage, because nothing emits the op
  yet. It gets its first real exercise with the `hipsr.nonzero` follow-up.
- It uses `hipMemcpyDefault` rather than an explicit `hipMemcpyDeviceToHost`,
  matching `hipdnn_ep_readback_i32` and `hipdnn_ep_readback_scalar`: an
  explicit D2H fails with `invalid argument` when the source is host-accessible
  memory, and `#hipsr.mem<device>` is a compile-time label that does not pin
  the HIP allocation kind.
- The copy does not synchronize, so a host read of the destination needs a
  separate stream synchronize.
- `AllShapesMatch` compares extents exactly, so a `?` source cannot fill a
  static destination.

AI assistance: substantial. Drafted with Cursor using Claude Opus 5; I reviewed
the result and validated it with the LIT suite and `pre-commit run --all-files`.


wcy123 (commented):
LGTM

wcy123 (approved):

PR #855, opened 2026-08-27
feat(hipsr): add the nonzero op and its ONNX conversion

zpye:
## Summary

Adds `hipsr.nonzero` and the `onnx.NonZero` conversion, so a graph whose element count is only known after a kernel runs can be lowered into the hipsr dialect.

Stacked on #854. Merge that first; this branch targets `feat/hipsr-copy-d2h` and uses the `hipsr.copy_d2h` it introduces.

## Why

Every hipsr op so far has a shape derivable from its operand shapes, so a normal placeholder can size the destination before anything runs. NonZero cannot: the number of non-zero positions depends on the data. The destination is therefore sized twice, once pessimistically so the kernel has somewhere to write, and once for real after the count is known.

That is why the op has two destinations instead of one dynamically sized result. `indices` is allocated for the worst case, where every element is non-zero, and `count` records how many of its columns were filled.

## What

`hipsr.nonzero` is the first hipsr op with two destinations, so `getDpsInitsMutable` spans both adjacent operands instead of forwarding a single accessor. Both are device-resident because the search runs on the device.

`onnx.NonZero` becomes a search, a `hipsr.copy_d2h` bringing the count to the host, and a `hipsr.compute` that narrows the result:

```mlir
%0:2 = hipsr.placeholder(%ctx) ins(%mask)   // worst case: 3 x num_elements, and [1]
%1:2 = hipsr.nonzero(%ctx) ins(%mask) outs(%0#0, %0#1)
%2   = hipsr.placeholder(%ctx) ins(%0#1)    // host count, forwards the source shape
%3   = hipsr.copy_d2h(%ctx) ins(%1#1) outs(%2)
%4   = hipsr.placeholder(%ctx) ins(%2)      // barrier: reads the count to get the width
%5   = hipsr.compute(%ctx) ins(%1#0) outs(%4)   // tensor.extract_slice to that width
```

Only a host read resolves the real column count, so the compute's placeholder is a barrier and this conversion fills its shape region from the copy. The count never becomes a compute operand; the body recovers it from its own destination with `tensor.dim`, which the barrier region has already resolved.

The search placeholder is filled by a recipe registered with `hipsr-populate-shape-region`, yielding the indices shape (input rank by `shape.num_elements`) and the constant count shape `[1]`.

A new `createCopyD2H` helper in `OnnxToHipsrUtils.h` builds a readback in one call: the host destination, its shape region, and the copy. A copy only changes memory space, so the destination forwards the source shape rather than restating what the op's `AllShapesMatch<["src", "init"]>` already pins.

The verifier requires `i64` elements, rank-2 indices with a static row count equal to the input rank, and a static single-element count. The conversion rejects an unranked, scalar, float, or host input, a result that is not a rank-2 `i64` with one row per input axis, and a static column count.

## Notes for reviewers

Not in this change, and both needed before a model can run NonZero: a bufferization model, and a `HipsrToLLVM` lowering with its runtime symbol.

I checked this against `test/lit/Dialect/Hipsr/Pipelines/embedding.mlir`, the motivating graph, which is currently `UNSUPPORTED`. Its NonZero converts and clears the whole `--hipsr-pipeline`, which splits it into two pool domains at the host readback as intended. The file still cannot be enabled, for three reasons outside this change:

- the external-data `onnx.Constant` for the embedding table does not legalize;
- `onnx.Expand` rejects the mask because the converted operand is `ui8` while its declared result is `i1`;
- `hipsr-partition-pool-domains` leaves a destination placeholder in the producing domain while moving the op it initializes into the next one, tripping `requires each result to initialize exactly one hipsr operation`. This reproduces with a runtime-window `onnx.Slice` feeding an `onnx.Shape` and no NonZero present, so it is a pre-existing partitioner bug.

Its `CHECK` lines will also need rewriting whenever it is enabled, since they still expect `onnx.*` ops in output that is now `hipsr.*`.

Prepared with AI assistance (Claude Opus 5 via Cursor), as disclosed in the commit trailers; I reviewed the result and validated it with the full LIT suite.


wcy123 (approved):
LGTM

PR #862, opened 2026-08-27
refactor(conv): drop the dead op-state slot plumbing from the convolution ops

BoarQing:
## Summary

Removes the op-state slot plumbing from `hip.conv`, `hip.conv_transpose` and `hip.causal_conv_with_state`, drops the dead per-session `conv_scratch` buffer and its two exported accessors, and corrects the stale MIOpen documentation those three ops still carried. 19 files, +163 / −382 lines. The only ABI change is the removal of `hipdnn_ep_state_get_conv_scratch` and `hipdnn_ep_state_ensure_conv_scratch`, which had no callers in the tree.

Developer-visible effect: a conv-only graph no longer emits a `hipdnn.num_op_state_slots` attribute at all, so `--generate-op-state-init` becomes a no-op and the per-session empty allocations disappear. The generated dialect docs and `docs/supported-operations.md` stop claiming these ops run on MIOpen.

## Related issue or design

Follow-up cleanup to #834 (`c20892d0`), which took MIOpen out of every convolution path. See [docs/design/op-state-slots-design.md](docs/design/op-state-slots-design.md) for the slot mechanism.

## Why

#834 moved all three convolution ops onto in-tree HIP kernels, but the op-state slot chain they had inherited survived it. That chain existed only to carry the MIOpen descriptor/solution caches. With MIOpen gone the ops are stateless, so every link in the chain was allocating an empty struct per op instance and doing nothing else:

```mermaid
flowchart LR
  TD["HipOps.td: OpStateOpInterface + generateOpStateInit"] --> Pass["--assign-op-state-slots stamps hip.op_state_slot"]
  Pass --> Low["Conv / ConvTranspose / CausalConvWithState lowering: threads an i32"]
  Pass --> Gen["--generate-op-state-init emits a construct call"]
  Gen --> Ctor["hipdnn_ep_op_state_construct_*: allocates an empty struct"]
  Low --> Wrap["wrap_*: ignores the argument"]
```

Dropping `OpStateOpInterface` is the load-bearing change that unlocks the rest: while an op declares the interface the compiler unconditionally emits a construct call, so the symbol has to exist.

The second category is documentation that is simply wrong now. Op summaries and descriptions feed the generated dialect docs, so "2D convolution operation using MIOpen" was the most visible incorrect statement left in the tree.

## What

Slot removal, in dependency order:

- **[HipOps.td](include/hip/Dialect/IR/HipOps.td)** — drop `OpStateOpInterface` and the two `generateOpStateInit` bodies from `Hip_ConvOp`, `Hip_ConvTransposeOp` and `Hip_CausalConvWithStateOp`. `AttrSizedOperandSegments` and the causal conv's `hasCanonicalizer` are kept.
- **Lowering** — stop threading `i32 op_state_slot` in [ConvLowering](lib/Conversion/HipToLLVM/ConvLowering.cpp) (32→31 args), [ConvTransposeLowering](lib/Conversion/HipToLLVM/ConvTransposeLowering.cpp) (27→26) and [CausalConvWithStateLowering](lib/Conversion/HipToLLVM/CausalConvWithStateLowering.cpp) (16→15). `getOpStateSlotValue` stays in `HipToLLVMUtils.h` — six other lowerings still use it.
- **Real runtime** — delete the empty `ConvState` / `CausalConvState` structs, both `hipdnn_ep_op_state_construct_*` functions, and the slot parameter from the three wrappers. Two now-unused `op_state.h` includes go with them.
- **Mock runtime** — delete the two construct stubs and the matching slot parameters.
- **Declarations** — drop the parameter from the three declarations in `hipdnn_ep_runtime.h` and delete the stale "accepted for ABI stability" and `CausalConvState` notes there and in `runtime_state_internal.h`.

Documentation:

- `HipOps.td` section header `===== MIOpen DNN Operations =====` → `===== Convolution Operations =====` (it covers exactly Conv and ConvTranspose), plus the Conv/ConvTranspose summaries and descriptions.
- Comments in [ConvConversion.cpp](lib/Conversion/OnnxToHip/ConvConversion.cpp) and [CausalConvWithStateConversion.cpp](lib/Conversion/OnnxToHip/CausalConvWithStateConversion.cpp), the three runtime file headers, and a dead pointer to `real/miopen.cpp` (a file that no longer exists).
- `docs/supported-operations.md` originally carried three wrong rows here too; #858 landed the same correction upstream first, so the rebase onto `ddb349e1` dropped that hunk as redundant.
- `causal_conv_with_state` removed from the slot-backed op list in the op-state design doc. That line already claimed `conv` is not slot-backed, which was wrong before this PR and is correct after it.

Every other MIOpen reference is deliberately left alone: `miopen.graph`, `miopen.add` and `miopen.softmax` still genuinely use it, and the MIOpen numbers in `conv_kernel.hip` are benchmark measurements that justify the WMMA path, not claims about what the code does.

## Test plan

Built Release/gfx1151 and ran, on this branch and again on a clean rebuild of `main` for comparison:

| Suite | Result |
|---|---|
| LIT (`ctest -R MorphizenMLIRLitTests`) | 440 passed, 13 unsupported, **0 failed** of 453 |
| `test_conv.py`, `test_conv1d.py`, `test_causal_conv_with_state.py`, `test_transpose.py` | 71 passed |
| Full numeric suite | 577 passed, 21 failed — **failure set byte-identical to `main`** |

The 21 numeric failures are pre-existing on `main` (GQA, QMoE, reduce, topk) and are untouched by this change; I diffed the two full-suite runs and the sets match exactly. The `E2E_Compile_*` ctest failures in my local tree are an environment issue (`lld-link: undefined symbol: hipMemcpy` in models such as `sqrt` that contain no conv) and reproduce identically on a clean `main` build.

Payoff verified directly rather than inferred:

- Running `--assign-op-state-slots --generate-op-state-init` over each of the three conv LIT inputs now produces no `hipdnn.num_op_state_slots`, no `hip.op_state_slot` and no construct call.
- A mixed `hip.conv` + `hip.matmul` module renumbers correctly: matmul takes slot 0 (it was slot 1) and the module reports one slot instead of two.

LIT updates: `test_conv.mlir` loses one `i32` from the `wrap_conv` signature check; `test_causal_conv_with_state.mlir` drops `--assign-op-state-slots` from its RUN line and the trailing slot operand from all seven call checks. `test_convtranspose.mlir` and `assign_slots.mlir` needed no change (the former only asserts the call name, the latter uses `hip.matmul`).

Re-verified after the rebase onto `ddb349e1`: the tree builds clean (`hipgpu.dll`, `custom_kernels_gfx1151.dll`, mock runtime and all tools), and LIT still passes. The two files git auto-merged, `hipdnn_ep_runtime.h` and `mock_gpu.cpp`, both compile.

## Validation on every model on this machine

Since the payoff argument is "these operands were dead", the assertion worth
making is that GPU output does not move at all. I built both arms from one
source tree with one CMake configuration — `f65249f3` (this PR's merge base)
and `7f64f27f` (this branch) — and compared them on every ONNX graph
available locally. Each comparison records which EP DLL each arm loaded, so the
arms are provably not sharing a build.

A scan of **414 local graphs** splits them by whether this PR can reach them:
**57 carry** `Conv` (1556 nodes), `ConvTranspose` (4) or
`CausalConvWithState` (180, plus 24 depthwise rank-3 `Conv` nodes that lower to
it); **357 carry none**.

### The mechanism, on production IR

Per-pass compiler IR from both arms, on real models rather than synthetic MLIR.
The slot count falls by exactly the number of conv-family ops in the graph, the
conv ops keep lowering while carrying no slot, and the survivors renumber to a
dense `0..N-1`:

| model | conv-family ops | slots, base → branch | construct symbol dropped |
|---|---|---|---|
| whisper-tiny encoder | 2 `hip.conv` | 43 → 41 | `_conv` |
| sam2.1 decoder fp16 | 1 `hip.conv`, 2 `hip.conv_transpose` | 148 → 145 | `_conv` |
| resnet50 fp16 | 53 `hip.conv` | 119 → 66 | `_conv` |
| Qwen3.5-9B text | 24 `hip.causal_conv_with_state` | 473 → 449 | `_causal_conv_with_state` |
| Llama-3.1-8B prefill (control) | none | 321 → 321 | none |

The conv-free control is the useful half of that table: it confirms a graph
without these ops is not perturbed, which is what makes the 357-graph conv-free
partition a control rather than an untested risk.

### End-to-end token equality

Greedy generation, fixed seed, `do_sample=False`, 32 tokens, same prompt, over
all 26 local OGA model directories. Of the 18 that generate on both arms,
**18/18 emitted an identical token sequence** — including
DeepSeek-R1-Distill-Llama-70B, the Gemma-4 26B-A4B and Qwen3.5/3.6 35B-A3B
MoEs, and all five Qwen causal-conv text models. The other 8 fail identically
on both arms for input reasons (7 Whisper dirs want `audio_features`,
gemma3-4b wants rank-3 `image_features`).

Reported rather than hidden: 7 of those 18 generate degenerate text on this
machine (three Gemma-4 12B builds emit one repeated token; Gemma-4 E2B and Phi-4
emit a single token). That is a pre-existing local issue — the degeneracy is
byte-identical on `main`, and none of these models contain a conv op.

### GPU A/B on the conv-bearing graphs

Of the 57 affected graphs, 37 run standalone. 20 are bit-identical with a
`0.0000%` noise floor, including esrgan (351 `Conv` nodes, the most in the
corpus), simplebev (124), yolov5lu (120), bevformer (55), detr (54),
resnet50 v1.5 (53), ssd-resnet34, mb1-ssd, and all five Qwen causal-conv text
graphs in decode. Four of those five are also bit-identical in prefill; the
fifth, Qwen3.6-35B-A3B, is nondeterministic within a single build (see below).

The other 17 are nondeterministic run-to-run, so a single pair of runs cannot
settle them. Re-running each with 4 to 10 repeats per arm shows the
branch-vs-main difference sitting inside the model's own spread: sam2.1 encoder
fp16, the largest, gives cross `0.1685%` against a within-branch `0.1695%`, and
Qwen3.6-35B-A3B's prefill swings `51%` between two runs of the *same* build. In
other words each arm differs from itself by as much as it differs from the other
arm.

The remaining 20 affected graphs do not run standalone and **fail with the same
exit code on both arms** — int8 models and Whisper's dynamic encoder hit the
CPU-fallback-disabled partitioning limit, and the Qwen vision towers crash on
random `image_grid_thw` data. Pre-existing, and verified on `main`, not assumed.

### Conv-free control graphs

The conv-free partition was swept the same way, ordered to cover every model
directory before repeating shape buckets: **221 of the 357 graphs, spanning all
40 model directories**. 163 of those run standalone and **152 are bit-identical**.

The other 11 all share one signature: the branch-vs-main delta equals the
same-build noise floor to every printed digit, so the difference is the graph's
own nondeterminism rather than the arms disagreeing. Six of them were re-run
with 8 repeats per arm to confirm that reading, and each came back with `cross`
exactly equal to `within_new`. Two are worth naming: whisper-medium's decoder
prefill varies by `363%` *within a single build* (it emits NaNs from random
inputs when run standalone), and Mistral's 12200-token prefill turned out to be
exactly identical across all 16 runs — its one-off `1716%` reading in the
first sweep did not reproduce.

### CPU reference

A four-way comparison (branch vs CPU, main vs CPU, branch vs main, noise floor)
on the 30 conv-bearing graphs small enough to run on CPU. `new_vs_cpu` and
`base_vs_cpu` agree to within `0.0513` percentage points everywhere, so the arms
are equally accurate against CPU rather than merely equal to each other. detr
(`106%`), detr-r50 (`97%`) and yolov5lu (`11%`) have large GPU-vs-CPU gaps that
are identical on both arms — pre-existing, and out of scope here.

### Scope of the above

This was all run against merge base `f65249f3`, before the rebase onto
`ddb349e1`. The rebase replayed the identical patch and dropped only the
`docs/supported-operations.md` hunk that #858 had landed upstream, so no
compiled code changed and the results carry over. The four commits picked up
from `main` (#854, #855, #858, #859) add hipsr `nonzero` / `copy_d2h`, a doc
fix and a CI trigger fix; none touch a convolution path.

## Notes for reviewers

- **Slot renumbering.** `AssignOpStateSlots` assigns dense slots in pre-order, so removing three op types shifts indices for other stateful ops in mixed graphs. This is internal and assigned at compile time, so there is no ABI or serialized-artifact concern, and the one slot-numbering LIT test uses `hip.matmul`. Verified with the mixed-graph check above rather than assumed.
- **Deliberately not included.** `wrap_conv_transpose` still accepts four arguments it never reads (`pad_bottom`, `pad_right`, `output_padding_h`, `output_padding_w`). Unlike the slot, these are meaningful ONNX attributes that the caller has already folded into `output_h`/`output_w`, and the file header documents why each is ignored. Removing them is a separate ABI question, so it is out of scope here.

## Checklist

AI assistance: this change was implemented with Cursor (Claude Opus 5) working from a written plan. Validation was run by hand as described in the test plan, including the clean-`main` rebuild used to establish that the 21 numeric failures and the E2E link failures are pre-existing.


BoarQing:
Perf: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/571/
Acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/572/
Procyon: http://xsjvitisaijenkins:8080/job/daily_hip_ep_onnx/272/
Vision: http://xsjvitisaijenkins:8080/job/daily_hip_ep_onnx/273/

PR #863, opened 2026-08-27
fix(ep): handle constant/initializer graph outputs in fusion

Vandana9628:
## Summary
A graph output that is a bare constant initializer — e.g. an `Identity(constant)` that ONNX Runtime constant folding collapses into a producerless, initializer-backed graph output — was mishandled in three places in the EP fusion path, aborting or crashing the compile of any model that exposes a constant directly as a graph output (a common pattern for surfacing quantization `scale` / `zero_point` metadata).

## The three failure points
1. `node_arg_names_to_nodes()` `LOG(FATAL)`'d (`cannot find producer`) on any producerless output. Only a genuine node-output with no producer indicates a node fused into another subgraph; a constant/initializer output is legal and must not abort the fuse.
2. `build_metadata_json()` emitted the constant into the artifact metadata output list, while `fuse_graph()` could not put it in the `meta_def` outputs — `build_output_index_map()` (MlirCustomOp) then failed to match artifact output → meta_def output.
3. Claiming the constant as a fused output built a fused function whose return referenced an SSA value that does not exist inside the region (segfault in `create_func_func`).

## Fix
Never claim a producerless (constant/initializer) graph output. Both readers of `graph.outputs()` in `pass_main.cpp` (`fuse_graph` and `build_metadata_json`) skip outputs whose `find_producer()` is null, keeping the `meta_def`, the compiled artifact, and the custom-op output map consistent. ONNX Runtime serves an initializer-backed graph output directly. The `node_arg_names_to_nodes()` guard now only FATALs for a genuine producerless node-output.

## Tests
- `GraphTest.TryFuseConstantGraphOutput` (morphizen unit test): constructs a graph with a real compute output plus a constant initializer graph output, asserts `graph_try_fuse` succeeds and excludes the constant from the fused outputs.
- `test/python/test_constant_graph_output.py`: compiles a graph with a constant graph output through the EP and asserts session creation succeeds (covers `fuse_graph` + `build_metadata_json`).

## Files
- `backend-mlir-compiler/level-1-pass/src/pass_main.cpp` (2 spots)
- `morphizen/morphizen-core/src/pass.cpp` (1 spot)


fhanuman (approved):
Either Morphizen or ORT seems to be introducing this illegal IR. I will go ahead and approve the changes; however, it would be good to look into the source of the issue.

PR #864, opened 2026-08-28
perf(conv): restore convolution performance after the MIOpen removal (1.09x over 21 models, 1.22x on resnet50)

BoarQing:
## Summary

Restores convolution performance after the MIOpen removal in #834. The 11 regressed models were losing to **three independent defects plus a correctness bug**, not one:

1. **The tile ladder picked worse tiles on bigger devices.** It rejected any tile producing fewer blocks than the device has CUs, so more CUs meant falling further down the ladder — to narrower row blocks, each re-gathering the im2col matrix once more. Adding CUs does not add bandwidth, so scaling a memory decision with the CU count is backwards. resnet50's stage-4 convolutions took `32x256` at 16 CUs and `16x256` at 20, which is 32 passes over the input where 4 would do, and 38.8% of that model's convolution time.
2. **No Winograd for 3x3 stride-1.** MIOpen's kernel name says the algorithm (`..._f2x3_...`). esrgan, whose 351 convolutions are all 3x3 s1, went 19.4 ms to 92.0.
3. **ConvTranspose was never tiled**, on the stated premise that no supported model contains one. sam2.1's decoder has two: 0.4 ms to 33.8.

Selection now reads the shape and ignores the CU count, so monotonicity holds by construction. It lives in a header that compiles without HIP, so the invariant is checkable with no GPU.

Retuning the rungs surfaced the largest single win, which was not the one the plan predicted: **the wave tile, not the block shape, was limiting the matrix kernels.** A 2x2 wave tile holds 32 accumulator VGPRs where 2x4 holds 64, taking `64x256` from 146 VGPR / 9 waves per SIMD to 97 / 12 and spreading the block over 512 threads instead of 256. That outweighs the doubled gather a narrower row block costs, and it wins in every M bucket from 128 up — so the wide rungs now take a *narrower* tile than the gather term predicts.

## Performance

Shape-level, via `conv_microbench.py` over every distinct convolution shape in the guard set. Both arms hash-verified, 120 iterations, same process:

| model | before | after | |
|---|---|---|---|
| resnet50 | 6.70 ms | 5.48 ms | **1.222x** |
| mb1-ssd | 6.78 ms | 5.61 ms | **1.206x** |
| yolov5lu | 20.25 ms | 17.87 ms | **1.133x** |
| bevformer | 9.32 ms | 8.51 ms | **1.093x** |
| esrgan | 206.7 ms | 194.4 ms | **1.063x** |
| detr | 10.59 ms | 10.54 ms | 1.001x |
| simplebev (fp32) | 24.64 ms | 24.61 ms | unchanged |

Whole-model, 21 models through `onnxruntime_perf_test`, average inference time, **1.093x overall**:

| model | before | after | |
|---|---|---|---|
| google-mobilenet_v2 | 2.22 ms | 1.89 ms | **1.177x** |
| nvidia-resnet50v1.5 | 7.43 ms | 6.30 ms | **1.179x** |
| ultralytics-yolov5lu | 29.86 ms | 26.44 ms | **1.129x** |
| swinv2-base-8x2048x3072 | 8402.6 ms | 7512.7 ms | **1.118x** |
| qfgaohao-mb1-ssd | 6.10 ms | 5.63 ms | 1.082x |
| facebook-detr-resnet-50 | 17.13 ms | 15.91 ms | 1.077x |
| blip/encoder | 14.51 ms | 13.53 ms | 1.072x |
| swinv2-tiny-8x521x1024 | 204.98 ms | 191.85 ms | 1.068x |
| esrgan | 301.32 ms | 283.33 ms | 1.063x |
| swin-tiny-224x224 | 3.69 ms | 3.48 ms | 1.061x |
| bevformer-fp16 | 15.56 ms | 14.91 ms | 1.044x |
| sam2.1/decoder | 5.63 ms | 5.45 ms | 1.034x |
| simplebev | 744.1 ms | 725.4 ms | 1.026x |
| blip/decoder | 5.96 ms | 5.87 ms | 1.016x |
| sam2.1/encoder | 55.18 ms | 55.17 ms | 1.000x |
| swinv2-tiny-512x1024 | 32.95 ms | 33.25 ms | 0.991x |
| swinv2-base-2048x3072 | 956.4 ms | 965.2 ms | 0.991x |
| detr | 22.54 ms | 22.77 ms | 0.990x |
| ssd-resnet34-1200 | 44.59 ms | 45.26 ms | 0.985x |
| convnext | 52.59 ms | 57.14 ms | 0.920x — see below |

**convnext is not a regression.** 18 of its 22 convolutions are depthwise 7x7, which this change cannot touch. Its convolution time measures 18.34 ms before and 18.26 ms after, and three repeat runs put the model at 57.96 mean before against 58.61 after — the 8% is harness variance, and the 52.59 reading was the outlier.

`bevformer-tiny-temporal-sim` fails to load on both arms; pre-existing and unrelated.

The ladder now sits **3.3% off the per-shape oracle, against 18.1% before**.

## Two negative results, recorded in the source so they are not retried

- **Narrower register tiles do not help the scalar path.** 64x32 and 32x64 reach 10 waves per SIMD and 32x32 reaches 12, against 64x64's 8, and all three lose on simplebev's real shapes (12.68 ms against 16.03, 16.93, 20.35). Unlike the matrix tiles these were never accumulator-bound — a 4x4 register tile is 16 of the 177 VGPRs. The aggregate only says otherwise if the extractor's `Nout==1` shape-inference fallbacks are left in, and those are not shapes any model runs.
- **`amdgpu_waves_per_eu(10)` is actively harmful.** It delivers what it asks (64x64 to 144 VGPR at 10 waves, no spill reported) and simplebev's convolutions go **24.6 ms to 43.8**. The registers come back by rematerialising the gather's address arithmetic rather than spilling, so the resource report stays clean while the inner loop recomputes indices it used to hold.

## On the fp16 NaN in bevformer

Investigated and **conv is exonerated**. Every tile variant forced on the failing shape produced the identical 31,744 NaN, which rules out per-variant staging, and a CPU reference agrees with the GPU convolution at 0.99999917 cosine similarity on random inputs. The infinities enter upstream of the convolution. Left open rather than claimed fixed.

## Test plan

- [x] `ConvTileSelectUnitTest` — new, GPU-free, now registered with CTest. It was previously possible to add a tile the kernel could not launch; the suite asserts monotonicity in CU count across 8–128, that every choice is instantiable, and that every table entry is reachable by name.
- [x] `conv_tile_check.py` — all 17 tiles agree with `conv_direct` on pseudorandom data, and each is confirmed to have actually dispatched via the debug log rather than silently falling back.
- [x] Shape-level A/B over every distinct convolution shape in the guard set.
- [x] 21-model A/B, including all 9 models that *improved* under #834 (swin, swinv2 x4, blip x2, convnext, sam2.1/encoder) — the ones a conv retune can most easily pay itself out of.
- [ ] CI perf run to confirm the above on clean hardware.

The numeric suite could not be run locally: `register_execution_provider_library` faults before any convolution executes, an ABI mismatch between the pip `onnxruntime` and the deployed EP. `conv_tile_check.py` was written to cover the same ground for this change specifically, but the suite should be run in CI.

## Note on branch composition

The first 10 commits are the perf-harness tooling this PR's measurements rest on (Phase 0 of the plan) and touch only `tools/perf-harness/**`. The 4 conv commits are the shipping change. Happy to split if preferred.

### DIFF
--- a/docs/perf/qwen36-16k-bottleneck.md
+++ b/docs/perf/qwen36-16k-bottleneck.md
@@ -1,0 +1,500 @@
+<!--
+Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+Licensed under the MIT License.
+-->
+# Qwen3.6-35B-A3B 16K prefill: where the time actually goes
+
+SQTT (RGP) profile of a 16K VLM prefill, taken on repaired model weights. This
+supersedes the withdrawn attribution in `qwen36-ttft-attribution.md`, which was
+measured against a half-zeroed `text.onnx.data`.
+
+## Operating point and build
+
+`vlm_benchmark.py` with `dog.jpg` and `prompt_16k.txt`: **18,646 prompt tokens**
+(16,823 text, 1,823 image), `max_length 33024`, `-e AMDGPU`. Build is
+`feat/rgp-capture-fence-v2` with all three binaries redeployed together
+(`hipgpu.dll`, `custom_kernels_gfx1151.dll`, `hip-compiler.dll`).
+
+Uninstrumented TTFT at this point is **19,526 ms** (3 reps, stddev 293 ms)
+against CI's 16,702 ms at 18,755 tokens. The attribution below is all measured at
+that baseline; the v8 prefill kernel has since been optimised.
+
+> **The 17,587 ms post-optimisation figure quoted below is not reproducible and
+> should not be used.** The same binaries and the same `max_length 33024` now
+> measure **14,734 ms**, and 14,455-14,610 ms with a right-sized KV cache. See
+> "Second profiling round" for the current baseline and for why that number was
+> probably instrumented.
+
+Fitting the 4K and 16K points for each environment separates startup from
+throughput:
+
+| | fixed | per token |
+|---|---:|---:|
+| local, as first measured | 634 ms | 1.013 ms |
+| local, warm and verified clean | 940 ms | **0.729 ms** |
+| CI | 1,016 ms | 0.836 ms |
+
+The first local row was fitted on runs that are now believed to be instrumented.
+On the clean re-measurement local starts up slower than CI but is 13% *faster*
+per token, so on these figures the per-token gap to CI has closed and reversed.
+
+## The layer stack
+
+Op call counts per prefill: `qmoe` 40, `matmul_nbits` 391, `causal_conv` 30,
+`linear_attention` 30, `gqa` 10. That is a 40-layer hybrid stack: 30
+linear-attention layers and 10 full-attention layers, every layer with its own
+MoE block.
+
+Each capture window contains exactly one `topk_routing` and one
+`bucket_tokens`, i.e. exactly one MoE block, which confirms one window equals
+one layer and lets per-layer cost be scaled by layer count.
+
+## Per-layer cost
+
+**Full-attention layer — 463.8 ms GPU busy, 943 dispatches, 0.5% idle**
+
+| kernel | count | total ms | % layer | occ | limiter | bound |
+|---|---:|---:|---:|---:|---|---|
+| `gqa_flash_prefill_v8_kernel` | 1 | 386.6 | 83.3 | 25% | LDS | latency/low-occupancy |
+<!-- The v8 row is the pre-optimisation measurement. It has since been cut to
+     203.9 ms/call uninstrumented; see "Optimising the v8 prefill kernel" below,
+     which also explains why its *instrumented* cost barely moved. -->
+
+| `MatMulNBitsFp16GEMM` | 61 | 29.1 | 6.3 | 55% | VGPR | undetermined |
+| `MatMulNBitsWMMA_NoZP` | 189 | 15.1 | 3.3 | 69% | VGPR | undetermined |
+| `bucket_tokens_kernel` | 1 | 8.9 | 1.9 | 100% | - | undetermined |
+
+**Linear-attention layer — 122.2 ms GPU busy, 490 dispatches, 1.4% idle**
+
+| kernel | count | total ms | % layer |
+|---|---:|---:|---:|
+| `la_pf_pass3_wmma` | 37 | 30.8 | 25.2 |
+| `MatMulNBitsFp16GEMM` | 24 | 20.4 | 16.7 |
+| `la_pf_scan` | 37 | 16.0 | 13.1 |
+| `la_pf_pass1_local` | 37 | 10.4 | 8.5 |
+| `T5LayernormFwdContiguous` | 2 | 9.4 | 7.7 |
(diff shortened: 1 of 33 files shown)


BoarQing:
Closing this. Checked on an 8060S (20 CU, gfx1151) with this branch rebased onto `main`, models run through the built EP under `HIPDNN_EP_DEBUG=1` so the kernel each node dispatches to is measured rather than inferred.

**The headline is carried by a model with no convolution.** The reported 1.093x is a sum-of-times ratio over the 21-model table, and 95.3% of the 934 ms saved is `swinv2-base-8x2048x3072` alone. The per-model geometric mean is 1.049x; excluding that one model the other 19 total 1.018x. That model contains exactly one `Conv`, a `stride == kernel` patch embed that `PatchEmbedConvToGemm` diverts to hipBLASLt Gemm. The batch-1 variant of the same graph (identical `Conv` node) issues **0** `wrap_conv` calls and 316 gemm calls. The same holds for swin-tiny, swinv2-base-2048x3072, swinv2-tiny-512x1024 and blip/encoder; blip/decoder has no `Conv` node at all. Six of the 21 rows cannot be attributed to a convolution change.

**Winograd is gated off precisely where the PR says it was needed.** `hip_conv_winograd_profitable` accepts only `blocks <= kResidentBlocksPerUnit * units`, i.e. 60 here. Over the guard set 487 nodes are eligible and 37 pass. esrgan is 351 eligible and **0** taken: its cheapest 3x3 s1 shape produces 256 blocks and the 1000x1000 layers 3969-7938. Measured, esrgan issues zero Winograd dispatches. Its real tally is 276x `wmma 32x256 2x2`, 74x `wmma 64x256 2x2`, 1x `wmma 16x256 1x2`; against `main` the first and last are identical choices, so the whole 1.063x is 74 nodes changing wave tile from 2x4 to 2x2. resnet50 takes all 13 of its eligible nodes, which is consistent with it being the strongest row.

**ConvTranspose tiling reaches 2 nodes**, both in sam2.1/decoder.

**convnext changes 0 of its 22 convolutions** — 18 depthwise, 4 patch-embed-to-Gemm — so its 0.920x is measurement noise rather than a regression, as the description notes.

Across all 21 models: of 1054 `Conv` nodes, 76 never reach `hip_conv` and 48 are depthwise, and the single most frequent change is 134 nodes moving from wave tile 2x4 to 2x2 on a 64x256 block. That occupancy change is the PR's real content and it stands on its own; it does not need the tile-ladder rewrite, the Winograd kernel, or the 10 commits of perf-harness tooling to land.

Separately, `pre-commit` and `build-windows-mock` are red, the numeric suite was not run, and `docs/perf/qwen36-16k-bottleneck.md` is unrelated to convolution.

The wave-tile change is worth resubmitting on its own, measured with a per-model geometric mean and a convolution-time baseline rather than whole-model wall clock.

PR #866, opened 2026-08-28
fix(hipsr): pair a placeholder with its consumer through their inputs

zpye:
﻿## Summary

```mlir
%x = placeloader ; // mising %y
%d1 = hipsr.cast (%ctx) ins(%y) outs(%x)
```

A `hipsr.placeholder` and its consumer now read the same values, the placeholder on the shape-graph side and the consumer on the data side. Matching the two lists keeps the pair in one pool domain, which they need because they share a destination buffer.

## Why

A placeholder that read fewer values than its consumer could land in an earlier pool domain. The pair then straddled a domain boundary and tripped the placeholder's own result-use rule.

Matching the lists makes the shape graph a copy of the data graph, so the domain walk reaches the same depth from either side. `PartitionPoolDomains` needs no pairing rule of its own and is unchanged here.

Values with no placeholder behind them are outside the rule. A block argument or a constant cannot reach a later domain, and a pool-domain region splits a data value and its counterpart into two block arguments that no longer name each other.

## What

- `PlaceholderOp::verify` gains `verifyConsumerTopology`, which checks both directions, and the ODS docs state the rule.
- `getShapeGraphCounterpart` becomes a free function next to the other hipsr operand helpers, so the verifier and `rewirePlaceholderInputs` resolve inputs the same way.
- `getHipsrInputOperands` returns the data operands of a hipsr op.
- ScatterND, the Gather host path, Shape, and the NonZero narrow barrier now list every operand their consumer reads.
- The NonZero narrow compute takes the host count its barrier reads and ignores it in the body, which sizes the slice from the destination the barrier resolved.
- New LIT cases cover both diagnostics.

## Notes for reviewers

Listing values one side ignores has a cost: it can push a pair one domain later than needed, and NonZero now yields one more value out of the earlier domain.

Developed with Cursor using Claude Opus 5, and validated with the LIT suite and the `-hipsr-partition-pool-domains` output.

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrOps.h
+++ b/include/hip/Dialect/Hipsr/IR/HipsrOps.h
@@ -34,6 +34,15 @@ namespace hipsr {
 // hipsr DPS op. return empty range if op is not a hipsr op.
 ::mlir::OperandRange getHipsrDestinationOperands(::mlir::Operation *op);
 
+// return the data operands of a hipsr op, which sit between the context and
+// the destinations. return empty range if op is not a hipsr op.
+::mlir::OperandRange getHipsrInputOperands(::mlir::Operation *op);
+
+// return the shape-graph value holding value's shape: value itself when it is
+// already a legal placeholder input, otherwise the destination its producer
+// writes into. return value unchanged when no destination matches.
+::mlir::Value getShapeGraphCounterpart(::mlir::Value value);
+
 // compute: use is in Outputs
 // dps: use is in Init
 bool isHipsrDestinationOperand(::mlir::OpOperand &use);
--- a/include/hip/Dialect/Hipsr/IR/HipsrPlaceholderOp.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrPlaceholderOp.td
@@ -29,6 +29,12 @@ def Hipsr_PlaceholderOp : Hipsr_Op<
     results, which keeps the shape graph parallel to the data graph.
     `placeholder_type` is `normal` or `barrier`.
 
+    A placeholder and its consumer read the same values, the placeholder on the
+    shape-graph side and the consumer on the data side, even when one of them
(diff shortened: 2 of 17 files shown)


include/hip/Dialect/Hipsr/IR/HipsrOps.h
```diff
@@ -34,6 +34,15 @@ namespace hipsr {
 // hipsr DPS op. return empty range if op is not a hipsr op.
 ::mlir::OperandRange getHipsrDestinationOperands(::mlir::Operation *op);
 
+// return the data operands of a hipsr op, which sit between the context and
+// the destinations. return empty range if op is not a hipsr op.
+::mlir::OperandRange getHipsrInputOperands(::mlir::Operation *op);
+
+// return the shape-graph value holding value's shape: value itself when it is
```
wcy123, line 41: there are two phases to convert ONNX operation to hipsr. e.g.
```
%x = ONNX.add %a, %b
%y = ONNX.cast %x
```
in the first phase, we don't use place holder entirly

```mlir
%init_x= placeholder %a %b
%x = hipsr.add (%ctx) ins(%a, %b), outs (%init_x)
%init_y = placeholder %x
%y = ONNX.cast (%ctx) ins(%x) outs(init_y)
````
in the second phase it is converted to 

```mlir
%init_x= placeholder %init_a %initb
%x = hipsr.add (%ctx) ins(%a, %b), outs (%init_x)
%init_y = placeholder %init_x
%y = ONNX.cast (%ctx) ins(%x) outs(init_y)
````

so given an value like `x` , getShapeGraphCounterpart return the counterpart of `init_x`,

we need to replace the operand `x` of `placehold` with `init_x`, because there is a rule that all operands of a placeoholder must be a result of another placeholder,

so that the placehold itself is a isomorphizen graph of the data computation.

wcy123 (approved):
LGTM

PR #867, opened 2026-08-28
fix(jit): reserve one ordered allocation per JIT'd object on Windows

zz002:
## Summary

Backport of the Windows JIT allocation fix (`xingy16-amd/hip-ep@968e39d`, author Xing) to this release branch. Sessions using the default LLVM_IR artifact format could die at random with `LLVM ERROR: IMAGE_REL_AMD64_ADDR32NB relocation requires an ordered section layout`; the JIT now reserves one contiguous, correctly ordered allocation per JIT'd object, so that abort cannot happen.

## Related issue or design

Cherry-pick of `xingy16-amd/hip-ep@968e39d7168c9fa21bf3db51da3f048b0ceaaa0f`. The corresponding `main` PR is submitted separately by the original author; this PR only lands the fix on `gpuep/releases/gpuep-rel-2610`.

## Why

LLJIT picks its object linking layer by triple (`LLJITBuilderState::prepareForConstruction`): for x86_64 the rule is `UseJITLink = !TT.isOSBinFormatCOFF()`, so Windows gets RuntimeDyld. MSVC-style x86_64 COFF always carries `.pdata`/`.xdata` SEH unwind tables whose entries are image-relative and are fixed up with `IMAGE_REL_AMD64_ADDR32NB`. `RuntimeDyldCOFFX86_64` resolves those against a faked `__ImageBase` -- the lowest section load address, cached on first use -- and calls `report_fatal_error`, taking the host process with it, whenever a target lands below that base or more than 4 GB above it. It documents the precondition it relies on: the memory manager must guarantee `CodeSection < ReadOnlySection < ReadWriteSection`.

LLJIT's default `SectionMemoryManager()` guarantees neither ordering nor proximity -- code, rodata and rwdata each come from an independent `VirtualAlloc` slab. Whether a process survives comes down to where ASLR put those slabs, so failures are intermittent and get likelier the more modules a session JITs and the more fragmented its address space becomes: worst on large LLMs and on hosts that create many sessions in one process.

## What

`makeLLJITBuilder()` installs, on Windows only, an object linking layer creator that builds an `RTDyldObjectLinkingLayer` over `SectionMemoryManager(nullptr, /*ReserveAlloc=*/true)`. RuntimeDyld then pre-computes the section totals, calls `reserveAllocationSpace`, and the manager carves code, then rodata, then rwdata out of one mapped block in that order. LLJIT builds a fresh manager per object, so every RuntimeDyld instance sees a contiguous, correctly ordered image whose base really is its lowest section. Setting a creator bypasses `LLJIT::createObjectLinkingLayer`, so its two COFF flags are repeated in the creator.

The creator is a functor rather than a lambda because `LLJITBuilderState::ObjectLinkingLayerCreator` changed arity: through LLVM 22 it takes `(ExecutionSession &)`, from LLVM 23 it takes `(ExecutionSession &, jitlink::JITLinkMemoryManager &)`, the second argument being of no use to an RTDyld-based layer. Supplying both call operators binds to either `std::function` without a version macro, and two `static_assert`s keep that property from rotting.

Linux is untouched and keeps JITLink. `HIPDNN_EP_JIT_UNRESERVED=1` falls back to the stock LLJIT layer.

## Test plan

- [ ] `build-windows` green on this branch, producing a `gpu-test-package`
- [ ] `onnxruntime_perf_test` on the resulting package with the default LLVM_IR artifact format: session creation and inference complete, no `IMAGE_REL_AMD64_ADDR32NB` abort

Not built locally; this branch's CI is the verification. On the source branch the original author verified the fix against LLVM 23.1.0 with a module whose functions are `uwtable`, which is what makes the backend emit the `.pdata`/`.xdata` carrying these relocations: with a `MemoryMapper` spacing the three section groups 8 GB apart -- legal under `SectionMemoryManager`'s contract, and what an unlucky ASLR draw amounts to -- the stock manager aborts every run and this change links and executes correctly every run.

## Notes for reviewers

The cherry-pick applied cleanly. The only difference in this file between `main` and this branch is the `HIPDNN_HIPBLASLT_DLL` macro in the library-name list, which lies outside every hunk. `clang-format` on this branch rewraps one comment paragraph and one `static_assert` differently from the source commit; the content is identical.

`main` does not carry this commit yet -- its HEAD is the cherry-pick's parent, `ddb349e`.

## Checklist

### DIFF
--- a/backend-mlir-compiler/custom-op-mlir/src/LlvmIrJit.cpp
+++ b/backend-mlir-compiler/custom-op-mlir/src/LlvmIrJit.cpp
@@ -13,11 +13,14 @@
 #include <llvm/ADT/ArrayRef.h>
 #include <llvm/ADT/SmallString.h>
 #include <llvm/Bitcode/BitcodeReader.h>
+#include <llvm/ExecutionEngine/JITLink/JITLinkMemoryManager.h>
 #include <llvm/ExecutionEngine/JITSymbol.h>
 #include <llvm/ExecutionEngine/Orc/AbsoluteSymbols.h>
 #include <llvm/ExecutionEngine/Orc/ExecutionUtils.h>
 #include <llvm/ExecutionEngine/Orc/LLJIT.h>
+#include <llvm/ExecutionEngine/Orc/RTDyldObjectLinkingLayer.h>
 #include <llvm/ExecutionEngine/Orc/ThreadSafeModule.h>
+#include <llvm/ExecutionEngine/SectionMemoryManager.h>
 #include <llvm/IR/LLVMContext.h>
 #include <llvm/IR/Module.h>
 #include <llvm/Support/Error.h>
@@ -44,6 +47,7 @@ extern "C" const std::size_t runtime_bc_data_size;
 #include <mutex>
 #include <new>
 #include <string>
+#include <type_traits>
 #include <typeinfo>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


amd-mingw (approved):

amd-mingw:
perf: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F462%2F&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F515%2F&curr_job=&curr= 
acc: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6a91128fc9f43c781b07c016

PR #870, opened 2026-08-28
fix(hipsr): give hipsr.equal the i1 result onnx.Equal declares

zpye:
## Summary

`hipsr.equal` produced a `ui8` mask while `onnx.Equal` declares an `i1` one. It
now yields `i1`, which removes the type divergence rather than working around
it.

Stacked on #866; the diff is a single commit once that merges.

## Why

`Onnx_EqualOp` declares its result `TensorOf<[I1]>`, yet `EqualConversion`
discarded that and built the mask as `ui8`, and `EqualOp::verify` required it.

Nothing downstream needed the byte: `hipsr.nonzero` accepts any integer,
`hipsr.expand` only requires input and output to agree, `hipsr.cast` has no
verifier, and the equal lowering passes the **operand** type as its `data_type`.

The divergence had one real cost, and it is why the previous revision of this PR
existed: `ExpandConversion` compared its operand against the declared ONNX
result, saw `ui8` against `i1`, and rejected a mask it could otherwise
broadcast. Retyping the producer removes the mismatch at its source, so the
tolerance previously added to Expand is reverted.

## What

`EqualConversion` builds an `i1` mask, `EqualOp::verify` requires `i1`, and the
ODS description says so. The mask stays `i1` from conversion through NonZero.

Two latent assumptions that a mask occupies a whole byte had to be fixed first:

| Site | Was | Now |
|---|---|---|
| `HipsrPoolAllocPass::staticByteFactor` | `bitWidth / 8`, so `i1` reserved **0 bytes** | rounds up, matching `MemoryLowering` |
| `getHipdnnDataType` | no `i1` case, returned `-1` | `i1` shares the unsigned-byte slot with `ui8` |

Three things ride along:

- Every LIT test this touches now checks its full output.
- The pipeline test loses a hand-edit, so the embedding table keeps the
  external-data form the importer actually emits rather than a form it never
  produces. `test/lit/Inputs/make_external_data.py` materializes the sidecar the
  constant names; only its length matters, since the conversion records the path
  and offset and never reads the weights.
- From review, the runtime data type identifiers move to
  `include/hip/datatype_abi.h`, included by both the runtime header and the
  compiler mappings, instead of being repeated as bare integers. No value
  changed.

## Notes for reviewers

`staticByteFactor` deserves the closest look. It is one line, but it is the
difference between a correctly sized mask buffer and a zero-sized one, and the
failure mode is memory corruption at run time rather than a test failure.
`i4`/`u4` also previously sized to 0 here and now size to one byte per element,
which over-allocates against true bit packing; no hipsr pass or test uses those
types today, and 0 was unusable either way.

The `hipsr-to-llvm` tests now pin upstream's memref-descriptor prologue, so a
change to how upstream builds descriptors means regenerating them.

AI assistance: the dialect and pass changes, the LIT assertions and their test
helper, and this description were drafted with Cursor and Claude Opus 5, then
validated by the full LIT suite (454 passed, 12 unsupported, 0 failed) and by
`pre-commit run --files <changed files>`.

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrEqualOp.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrEqualOp.td
@@ -10,8 +10,9 @@ def Hipsr_EqualOp : Hipsr_DpsOp<"equal",
     Elementwise equality with ONNX broadcasting: `init = lhs == rhs`.
 
     Shapes are right-aligned, and each aligned pair must be equal or contain a
-    one. `init` holds `ui8` rather than the operand element type, because the
-    runtime writes one byte per result element.
+    one. `init` holds `i1` rather than the operand element type, because a
+    comparison yields a bool whatever it compared. The runtime stores that bool
+    one byte per element.
   }];
 
   let arguments = (ins
--- a/include/hip/datatype_abi.h
+++ b/include/hip/datatype_abi.h
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+#ifndef HIP_DATATYPE_ABI_H
+#define HIP_DATATYPE_ABI_H
+
+//===----------------------------------------------------------------------===//
+// Backend-Independent Data Type Identifiers
+//===----------------------------------------------------------------------===//
+//
+// Single source of truth for the element-type identifiers the compiler writes
+// into generated runtime calls and the runtime reads back out. They are our own
+// values -- do NOT assume they match cuDNN or any other library's enum. Each
+// backend provides an explicit mapping function to convert these to
+// library-specific types.
+//
+// Both sides include this header, so a value cannot be changed on one side
+// alone. They stay plain #defines because the runtime's C ABI header uses them.
+//
+// To add a new type:
+//   1. Add the #define here
+//   2. Update hipdnn_ep_datatype_size() and hipdnn_ep_datatype_name() in
+//      lib/Runtime/hipdnn_ep_runtime.h
+//   3. Update the compiler mappings from an MLIR element type:
+//   getHipdnnDataType()
+//      in lib/Conversion/HipToLLVM/HipToLLVMUtils.h and in
+//      lib/Dialect/Hipsr/IR/HipsrLLVMLoweringUtils.cpp, and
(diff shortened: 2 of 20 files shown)


wcy123 (dismissed):
LGTM. The type correction and two fix-first changes are sound.

Side note: `getHipdnnDataType` hardcodes magic numbers (0, 4, 7) that duplicate `HIPDNN_EP_DATATYPE_*` constants from `hipdnn_ep_runtime.h`. Both are in the same repo - they should share a header rather than manually sync.

zpye:
Good catch, fixed in f3b86f5d.

The values now live in `include/hip/datatype_abi.h`, included by both
`lib/Runtime/hipdnn_ep_runtime.h` and the compiler mappings, so neither side
can change one alone. That directory is already on every target's include path
and on the runtime bitcode build's `-I` list, so no build plumbing was needed
beyond adding the header to the bitcode `DEPENDS`.

It turned out to be four sites rather than one:

| Site | Was |
|---|---|
| `getHipdnnDataType` in `HipToLLVMUtils.h` | `0`..`8` with the constant named in a trailing comment |
| `getHipdnnDataType` in `HipsrLLVMLoweringUtils.cpp` | `0`..`8`, bare |
| `getHipdnnInputDataType` in `OnnxToHipUtils.h` | subset, commented |
| `GatherBlockQuantizedLowering.cpp` | `dataDtype = 7` forcing packed nibbles to the unsigned-byte slot |

All four use the names now, and the `-1` returned for a type the runtime cannot
name is spelled `HIPDNN_EP_DATATYPE_UNSUPPORTED`.

One thing your note surfaced indirectly: the "to add a new type" checklist in
the runtime header told the reader to update `getHipdnnDataType()` in
`HipToLLVM.cpp`, a file that no longer exists, and mentioned only one of the
three mappings. It now sits next to the values and names all three.

No value changed. The LIT cases that pin a `data_type` argument in the lowered
call are the check on that, and they pass unmodified (454 passed, 12
unsupported), as do the end-to-end tests for the operations that pass one
(`gather_block_quantized`, `cast`, `equal`, `expand`, `nonzero`, `matmulnbits`,
`qmoe`).

I left the neighbouring `kPool*`, `kGlobalPool*` and `TensorOp` constants in
`HipToLLVMUtils.h` alone to keep this to the datatype duplication you flagged.
They carry the same "must match the runtime" comments, so they have the same
problem and would be a natural follow-up.

wcy123 (commented):
LGTM

wcy123 (approved):

PR #871, opened 2026-08-28
fix: correct GQA KV accounting and signless integer attribute reads

amd-genmingz:
## Summary

Two independent defects, both surfaced by gemma-4-E2B, in different
subsystems:

1. `IntegerAttr::getSInt()` on compiler-synthesized attributes aborts the
   host process during ORT session creation on an assertions-enabled MLIR.
2. Group-query attention mis-accounts KV length, layout, and `seqlens_k`
   for attention layers that consume another layer's present cache as
   their K/V, producing zeroed or wrong-window attention output.

`wrap_group_query_attention` grows one parameter (41 -> 42), so **cached
model artifacts are stale and must be invalidated before testing**.

## Related issue or design

None. See "Notes for reviewers" -- the runtime ABI change may warrant one
under CONTRIBUTING's design-discussion rule.

## Why

**Signless attributes.** `getSInt()` asserts the attribute is `si64`. The
MorphiZen importer encodes scalar ONNX INT attributes that way, but
attributes the compiler synthesizes are signless `i64`. Two sites on the
Transpose+MatMul fold path read synthesized attributes with `getSInt()`,
so any graph feeding a permuted `onnx.Transpose` into `onnx.MatMul` died
mid-session-creation with no diagnostic beyond the assertion line.
gemma-4-E2B's SigLIP vision encoder does this in every layer.

**GQA.** An attention layer whose K/V come from another layer's present
cache has four properties the runtime did not expect together: no
`past_key` operand, `Skv != Sq`, an additive mask, and a per-layer
synthesized `seqlens_k`. Each broke a different assumption. Three of the
four were correlations that happened to hold on every previously tested
model rather than structural invariants:

| Signal used | Structural signal | Held for |
|---|---|---|
| query length | key length | layers where `Sq == Skv` |
| `attention_bias == nullptr` | `past_key == nullptr` | Whisper (bidirectional layers carry no mask) |
| `Skv != sq` implies BNSH | key operand rank | Whisper cross-attention |
| pointer equality | the value itself | models where `seqlens_k` is a graph input |

## What

**`fix(onnx)`** -- read the raw `APInt` instead of `getSInt()` in
`TransposeMatMulFold` (`perm`) and `MatMulConversion` (`hipdnn.transA/B`),
matching what `TransposeConversion` already does for the same `perm`.

**`feat(gqa)`** -- four changes:

- **KV length from the key operand.** `OnnxAttentionConversion` sized the
  present buffer from `dim(q3, 1)`. ONNX Attention permits `Skv != Sq`;
  taking the count from `dim(k3, 1)` is correct for both.
- **Bidirectional path keyed on `past_key`.** `bidirectional_no_past` was
  gated on `attention_bias == nullptr`. A layer can carry a mask and still
  own no past buffer, in which case the decode convention derives
  `past_len = total_seq - sq > past_buf_seq == 0` and rejects the layer
  outright, zeroing its output.
- **`kv_layout_bnsh` through the ABI.** The KV populate previously inferred
  source layout from `Skv != sq`. Layout is a compile-time fact carried by
  the key operand's rank; bare pointers cannot reveal it. The old guess was
  harmless at `kv_num_heads == 1` (BSHD and BNSH are byte-identical there)
  and scrambled heads against seq above it. The causal concat/append path
  now rejects a BNSH source loudly instead of silently skipping a
  transpose.
- **`seqlens_k` read strategy by buffer origin.** A new `SeqlensKHome`
  classification replaces one pointer-keyed cache: external graph inputs
  stay cached, host-mapped scratch is read with a plain host load (no D2H,
  no stream drain), and device-pool buffers are read fresh every call
  because the pool recycles addresses across disjoint live ranges. The
  bidirectional no-past path skips the read entirely, since `total_seq` is
  `skv` and `past_len` is 0 by construction.

## Test plan

- `test/lit/Conversion/hip-to-llvm/test_gqa.mlir`: passes. Covers the
  42-argument signature and both layout cases (rank-3 key -> 0, rank-4 key
  -> 1); the emitted constants were inspected to confirm each check binds
  to `kv_layout_bnsh` rather than to `no_causal` or `op_state_slot`.
- [ ] gemma-4-E2B end-to-end with the model cache cleared: decoder logits
      relative L2, and generated text coherence.
- [ ] Regression all tracked models.

## Notes for reviewers

- **Model cache must be invalidated.** The arity change makes previously
  compiled artifacts incompatible; loading a stale one will fail at the
  call boundary.
- **Two unrelated fixes in one PR.** They are separate commits and each is
  independently reviewable; they ship together because both are required to
  run the model that exposed them. Happy to split if preferred.
- **Public runtime ABI change.** CONTRIBUTING asks for a documented design
  discussion before changing compiler/runtime ABIs. This one adds a
  trailing parameter with a compile-time-known value and no behavior change
  for existing callers, but flagging it explicitly.
- **`kv_num_heads == 1` masks the layout bug.** gemma-4-E2B has
  `kv_num_heads == 1`, so BSHD and BNSH are byte-identical and the old
  layout guess was harmless there. E4B has `kv_num_heads == 2` and needs
  this fix.

## AI assistance
Made with Cursor


amd-yanjunz:

- PR: ROCm/hip-ep#871
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/871#issuecomment-5474528230)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/585/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6a953dc5f38a31f5244cfbf5

amd-genmingz:
main branch : http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/592/

PR #874, opened 2026-08-28
perf(conv): stop the fp32 spill, recalibrate the fp16 ladder, stage the input window

BoarQing:
## Summary

Three changes to `conv_kernel.hip`, measured on gfx1151 (Strix Halo, 40 CU / 20 WGP). Each was found by timing every shape the exposed models run at their real runtime extents against every block shape the ladders contain, so each is a direct A/B on the shapes it affects rather than an inference from the geometry.

**The fp32 tile was spilling, not too coarse.** `conv_tiled_kernel`'s contraction loop has a compile-time bound of `BK=16`, so `#pragma unroll` fully unrolled it. Each step needs `TM+TN` live LDS reads that the scheduler hoists to cover `ds_read` latency, which at 8x8 asks for `16*(8+8)+64` values against 256 VGPRs. `conv_tiled_kernel<float,128,128,16,8,8>` compiled to **396 bytes of scratch per lane at 5 waves/SIMD** — every multiply-add in the hottest loop was a private-memory round trip. That is why the widest scalar tile measured up to 2.3x *slower* than 64x64, which reads as "the narrow tile is better" and is really "the wide tile is broken". Capping the unroll at 4 gives 135 VGPRs, no scratch, 10 waves/SIMD. Two and eight are both worse; the numbers are in the comment. With the spill gone the existing rung order was correct and did not change.

**The fp16 coverage tests were off by 2x, in the direction that hurt.** They ask whether a tile makes at least `units` blocks, and `units` came from `deviceUnitCount()`, which reports WGPs — so each test demanded half the grid it was written to demand. The effect was not that tiles were accepted too readily in general, but that the *specialised wide* ones were accepted on shapes far too small for them. For `Cin=512 Cout=512 20x20 k3 s2` the ladder rejected 128x256 (8 blocks), 256x128 (8) and 128x128 (16) and settled on 32x256 at 1.003 ms; the first two measure 0.743 and 0.659. Stating the tests in CUs sheds those shapes, so 128x128 and 16x256 are now ungated to catch them — otherwise they fall to a narrow rung, which is where the 1.3x-1.7x losses were.

**`conv_window_kernel` stages the input rather than the im2col matrix.** The im2col matrix holds each input element once per kernel tap, so a 3x3 reads the input nine times and none of it survives in cache past L2. esrgan's five RRDB convolutions sit at 71% of the memory roofline moving 5.6x-8.6x their irreducible bytes, so only fewer bytes can move them. This kernel gives a block a `TH x TW` rectangle of the output plane, reads the `(TH+k0-1) x (TW+k1-1)` input window those outputs need once per channel slice, and computes every tap out of LDS. Amplification stops being `kvol` and becomes the halo ratio: 1.33x at 8x32 with a 3x3, against 9x. Scoped to fp16, rank 2, unit stride and unit dilation — that is what fixes the window extent at compile time and lets it be an LDS array.

## Measured

Per-shape convolution totals, node-count weighted:

| model | before | after | |
|---|---|---|---|
| simplebev (fp32) | 849.6 ms | 271.0 ms | 3.14x |
| esrgan (fp16) | 301.3 ms | 166.7 ms | 1.81x |
| yolov5lu (fp16) | 26.6 ms | 20.5 ms | 1.30x |

No shape regresses by more than 1.05x. Both fp16 models land within 1.02x of the best tile available for each shape individually.

Whole-model wall clock, median of 20 inferences with 3 dropped as warmup — only the conv-bearing models moved:

| model | before | after | |
|---|---|---|---|
| simplebev | 1169.0 ms | 551.8 ms | 2.12x |
| esrgan | 367.5 ms | 280.9 ms | 1.31x |
| nvidia-resnet50v1.5 | 11.74 ms | 7.51 ms | 1.56x |
| qfgaohao-mb1-ssd | 11.50 ms | 8.18 ms | 1.41x |
| google-mobilenet_v2_1.0_224 | 3.33 ms | 2.71 ms | 1.23x |
| ultralytics-yolov5lu | 33.39 ms | 28.00 ms | 1.19x |
| bevformer-fp16 | 18.12 ms | 16.34 ms | 1.11x |
| mlcommons-ssd-resnet34-1200x1200 | 58.60 ms | 54.13 ms | 1.08x |
| detr | 31.11 ms | 28.82 ms | 1.08x |

resnet50 and mb1-ssd are the useful surprise: neither is in the corpus any of this was tuned on, and both gain more than yolov5lu did.

## Two things to argue with rather than believe

**The zoo is noisy, and the noise is larger than most of this table.** Two sweeps of the *same* build put `microsoft-swinv2-base-2048x3072` at 1566 and 1302 ms — a 20% spread on a model that dispatches no convolution at all. A single-run A/B here can manufacture a 20% regression out of nothing, and nearly did.

**`sam2.1/encoder` is recorded as unexplained, not explained away.** It is 4-5% slower across four runs per build. It is unlikely to be this change: conv is 6 nodes and 1.48 ms of a 74 ms model, and on those exact 6 shapes the new ladder is *faster* (1.342 ms against 1.483, no shape worse), so 3.4 ms is more than twenty times what the conv budget could move either way. The remaining suspect is code layout in `custom_kernels_<arch>.dll`, which also carries the attention kernels that model is mostly made of.

## Still open

The `conv_transpose` gather is untouched. In sam2.1-decoder it is 3.2 ms of a 10 ms model at 0.4% of compute peak, moving 38x the bytes it needs: one thread per output element summing over `Cin` with no LDS staging, so every output pixel re-reads the channel vector its neighbours just read. It wants the same treatment `conv_window_kernel` just gave the forward path.

The remaining fp16 gap is 0.5 ms on yolov5lu, almost entirely on stride-2 3x3 shapes where 256x128 is the best tile and no simple rung threshold separates them from the shapes that want 128x128. The window kernel does not apply — unit stride is what pins its window extent at compile time.

Two limits on the recalibration are worth stating plainly. It was derived from the only two fp16 models in the corpus, so the whisper shapes cited in the existing rung comments were not re-measured; and the replay searched thresholds and single-rung promotions, not the tile set itself. A new tile shape would need a new sweep.

## Test plan

- [x] Conv and ConvTranspose numeric sweep against a CPU reference: **45 passed, 0 failed** — run against this exact build, not a variant of it.
- [x] Every block shape in both ladders checked against `conv_direct` on pseudorandom inputs; worst error 4.71e-04 fp16, 7.79e-07 fp32.
- [x] All 81 shapes in the runtime census re-timed; largest single regression under 1.05x.
- [x] 21-model wall-clock sweep on the same two builds, repeated where a delta looked real.
- [x] Compiles clean — no new warnings.

## Notes for review

The three changes are independent and land in disjoint regions of the file, so they can be split into separate commits if that reads better. The mechanism for each is in the kernel comments; the evidence and the caveats are in this description.

### DIFF
--- a/lib/Runtime/Kernels/hip/conv_kernel.hip
+++ b/lib/Runtime/Kernels/hip/conv_kernel.hip
@@ -192,6 +192,12 @@ struct ConvParams {
   int32_t nBlkM, nBlkN;
   int32_t gmLog2; // log2 of the swizzle group height
   FastDiv rGroupSpan;
+
+  // Window kernel only: its column blocks tile the output plane in two
+  // dimensions rather than along the flattened position axis, so a block index
+  // has to split back into a row and a column of that 2-D grid.
+  int32_t nBlkW;
+  FastDiv rBlkW;
 };
 
 // The input coordinate that kernel tap (0,0,0) of an output position reads.
@@ -413,10 +419,12 @@ __device__ __forceinline__ bool swizzledTile(const ConvParams &pr, int &blkM,
   return blkM < pr.nBlkM;
 }
 
-// Fill in the tile-grid geometry the swizzle needs for one tile shape.
-static inline ConvParams withTileGrid(ConvParams pr, int64_t bm, int64_t bn) {
+// Fill in the swizzle geometry for a grid of nBlkM row blocks by nBlkN column
+// blocks, however the caller arrived at the column count.
+static inline ConvParams withBlockGrid(ConvParams pr, int64_t bm,
+                                       int64_t nBlkN) {
   pr.nBlkM = static_cast<int32_t>((pr.MPerGroup + bm - 1) / bm);
-  pr.nBlkN = static_cast<int32_t>((pr.Nout + bn - 1) / bn);
+  pr.nBlkN = static_cast<int32_t>(nBlkN);
   // Largest power of two that is both <= kGroupM and <= the tile rows.
   int lg = 0;
   while ((1 << (lg + 1)) <= kGroupM && (1 << (lg + 1)) <= pr.nBlkM) ++lg;
@@ -424,6 +432,24 @@ static inline ConvParams withTileGrid(ConvParams pr, int64_t bm, int64_t bn) {
   pr.rGroupSpan =
       makeFastDiv(static_cast<int64_t>(pr.nBlkN) << lg);
   return pr;
+}
+
+// Fill in the tile-grid geometry the swizzle needs for one tile shape.
+static inline ConvParams withTileGrid(ConvParams pr, int64_t bm, int64_t bn) {
+  return withBlockGrid(pr, bm, (pr.Nout + bn - 1) / bn);
+}
+
+// The same, for the window kernel's 2-D output tile. Its column blocks are a
+// th x tw rectangle of the output plane, not a run of bn flattened positions,
+// because the whole point of the kernel is to stage the input those positions
+// read -- and that is a rectangle only if the output tile is one.
+static inline ConvParams withWindowGrid(ConvParams pr, int64_t bm, int64_t th,
+                                        int64_t tw) {
+  const int64_t nh = (pr.out0 + th - 1) / th;
+  const int64_t nw = (pr.out1 + tw - 1) / tw;
+  pr.nBlkW = static_cast<int32_t>(nw);
+  pr.rBlkW = makeFastDiv(nw);
+  return withBlockGrid(pr, bm, nh * nw);
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


BoarQing:
Perf: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/569/ (tie)
Acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/570/
Procyon: http://xsjvitisaijenkins:8080/job/daily_hip_ep_onnx/269/
Vision: http://xsjvitisaijenkins:8080/job/daily_hip_ep_onnx/271/ 50%+ boost on several models.

amd-mingw (approved):

PR #875, opened 2026-08-29
feat(ep): negotiate the ORT API version at plugin load

zz002:
## Summary

`hipgpu.dll` is built against ORT 1.27 headers but asked the host for `GetApi(ORT_API_VERSION)` unconditionally, so loading the EP into an older runtime crashed with `0xC0000005`. It now negotiates the API version at load time: it runs on any runtime down to ORT 1.24, and reports a descriptive `OrtStatus` instead of crashing on anything older.

## Related issue or design

None.

## Why

`GetApi(v)` returns `nullptr` when `v` exceeds what the host supports, and `CreateEpFactories` dereferenced the result on the next line. Since `ORT_API_VERSION` is baked in at compile time, every runtime older than the one we built against was an access violation right after ORT printed `The requested API version [27] is not available`.

The `ort_version_supported` field of each ABI struct had the same problem in a quieter form: stamping the compile-time constant tells ORT that fields which do not exist in the running runtime's view of the struct are valid.

This follows the `NegotiatedOrtApiVersion` design in onnxruntime/onnxruntime-ep-amdgpu@7ca6647 so the two repos stay comparable. The negotiated version is DLL-local, so `hipgpu.dll` must negotiate for itself even when the amdgpu umbrella's hip shim already did: the shim forwards the original `OrtApiBase` and hands our `OrtEpFactory` straight back to ORT.

Two deliberate differences from upstream. First, the failure path builds its `OrtStatus` from a separate fallback `OrtApi` obtained at a version the runtime does accept, because at that point `Ort::InitApi()` has not run and there is no API to call `CreateStatus` on; `LOG(FATAL)` covers the case where the runtime rejects every version down to 1, since returning `nullptr` would read as success. Second, the code lives in `ort-bridge/src` rather than a shared library: upstream shares it across four backend DLLs, while `hipgpu.dll` has a single plugin entry.

## What

- `morphizen/ort-bridge/src/ort-api-version.{hpp,cpp}`: `kMinOrtApiVersion` (24), `NegotiateOrtApi()`, `NegotiatedOrtApiVersion()`, `FallbackOrtApiForStatus()`. `NegotiateOrtApi()` probes `GetApi()` downwards from the version `GetVersionString()` reports, capped at `ORT_API_VERSION`; starting at the host's own version keeps ORT from logging an "API version not available" line per probe. An unparsable version string falls back to probing from `ORT_API_VERSION`.
- `CreateEpFactories` uses the negotiated `OrtApi` and returns `ORT_EP_FAIL` when the runtime is below `kMinOrtApiVersion`. Its `MY_LOG(1)` line now prints the negotiated version alongside the compile-time one.
- Six ABI structs stamp `NegotiatedOrtApiVersion()` instead of `ORT_API_VERSION`: `OrtEpFactory`, `OrtEp`, `OrtAllocator`, `OrtDataTransferImpl`, `OrtNodeFusionOptions`, `OrtNodeComputeInfo`.

`kMinOrtApiVersion` is 24 to match the umbrella shim's floor. The EP's own API surface needs only 1.23, audited against `\since` in `onnxruntime_c_api.h`, so lowering the floor is a one-constant change.

## Test plan

Verified against official ORT release runtimes (`onnxruntime-win-x64-{1.23.2,1.24.2}.zip` plus the pinned 1.27.0) using a minimal `RegisterExecutionProviderLibrary` host built from the 1.23 headers, so one binary exercises all three:

| runtime | before | after |
|---|---|---|
| 1.27.0 | loads | loads, unchanged |
| 1.24.2 | `0xC0000005` | loads, `AMDGPUExecutionProvider` enumerated |
| 1.23.2 | `0xC0000005` | `requires ORT API >= 24, but the host reports 1.23.2`, no crash |

`MORPHIZEN_DEBUG_ORT_EP_API=1` confirms the negotiated version is 24 on the 1.24.2 host and 27 on the 1.27.0 host, and that ORT no longer logs the out-of-range `GetApi` line.

- [x] Windows build green (Ninja, Release, gfx1151).
- [ ] CI build-windows green.

## Notes for reviewers

No unit test: exercising the negotiation needs a fake `OrtApiBase`, and the target lives in `ort-bridge`, which `morphizen-unit-tests` does not link. The three-runtime check above is the coverage.

Upstream's own version of this path has a latent crash worth knowing about when comparing the two: `RETURN_STATUS` expands to `Ort::GetApi().CreateStatus(...)`, but `Ort::InitApi(ort_api)` runs after the `nullptr` check, so under `ORT_API_MANUAL_INIT` the too-old-runtime branch dereferences an uninitialised `Ort::api` on the first call. That is why this PR uses an explicit fallback API instead.


amd-mingw:
perf: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/547/
acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/548/

amd-mingw:
perf：http://xsjnts.xilinx.com/ROCm/compare?base_url=&base_job=daily_hip_ep_genai&base=462&curr_url=&curr_job=daily_hip_ep_genai&curr=547
acc：http://xsjnts.xilinx.com/ROCm/compare?base_url=&base_job=daily_hip_ep_genai&base=463&curr_url=&curr_job=daily_hip_ep_genai&curr=548

amd-mingw (approved):

PR #861, opened 2026-08-27
feat: Completely remove the dependency on MIOpen.

amd-genmingz:
## Summary

- Remove the last two runtime paths that dispatched GPU work through MIOpen: Gemm no longer uses `miopenOpTensor` to broadcast `beta * C`, and `wrap_power` no longer uses `miopenActivationForward`.
- `wrap_gemm` replace miopen broadcast with local implement.
- `wrap_power` still handles Reciprocal `(0, 1, -1)` and Sqrt `(0, 1, 0.5)` with custom HIP kernels. Any other `(alpha, beta, gamma)` logs `unsupported` and returns `-1`.
- MIOpen is gone from every layer: no handle in `RuntimeState`, no `find_package`, no `-lMIOpen` on the model link, no `MIOpen.dll` in the JIT search list, no vendor stubs, no staged DLL. hipBLASLt is the only vendor library left.
- Session init/cleanup drop `miopenCreate` / `miopenSetStream` / `miopenDestroy`. Init codes 7, 8 and cleanup code 11 are retired; the hipBLASLt codes are unchanged.
- `HIPDNN_EP_DISABLE_VENDOR_BLAS` now means "without hipBLASLt" only.
- `hip.miopen.softmax` and `hip_miopen_softmax` keep their names — they run in-tree kernels, so `grep -i miopen` still hits the dialect and lowering. Renaming is a separate change.

## What

- `runtime_state_internal.h`, `hipdnn_ep_runtime_state.cpp`, `hipdnn_ep_runtime.h`: drop the `miopen_handle` field, its lifecycle calls, and `hipdnn_ep_state_get_miopen_handle`. The hipBLASLt destroy paths are now null-guarded because that handle also stays null under `HIPDNN_EP_DISABLE_VENDOR_BLAS`.
- `hipdnn_graph_runtime.cpp`: same field dropped from `RuntimeStateLayout`, which mirrors `RuntimeState` by offset. These two always have to be edited together.
- `real/gemm.cpp`, `real/power.cpp`: delete `broadcastBiasToOutput`, the softfloat/device-copy broadcast helpers, and the POWER activation descriptor cache.
[...]

PR #877, opened 2026-08-29
fix(build): backport the rel-2610 build fixes and unbreak the mock build

zz002:
## Summary

Backports the build fixes that live only on `gpuep/releases/gpuep-rel-2610` to `main`, and fixes the mock-build breakage one of them introduced. `windows-build-mock` is currently red on the release branch; after this PR it is green on `main`.

## Related issue or design

Backport of the commits Artur Wojcik pushed directly to `gpuep/releases/gpuep-rel-2610`. Determined with `git cherry origin/main origin/gpuep/releases/gpuep-rel-2610` (patch-id based): the other release-only commits are already on `main` under different PR numbers (`#867`→`#869`, the CI push-trigger fix→`#859`, `#875`→`#876`).

## Why

These five never went through a PR, so `main` missed them. Two need explanation:

`3b002576b` ("add miopen package separately installed") is **not** cherry-picked. Its purpose was to re-add `find_package(MIOpen ...)` plus MIOpen include injection, but `main` deliberately removed the MIOpen dependency in #861 (`lib/Runtime/CMakeLists.txt` has zero MIOpen references). Re-introducing it would be a regression. Its two uncontroversial `.gitignore` entries (`.idea/`, `.depend/`) are picked up as part of the `f878504e3` hunk instead.

`1faca9923` ("control flow guard and msvc static runtime") turns the unconditional `/guard:cf` and the enforced static CRT into options that default to `OFF`. On `main` that would relax the current posture, which is CFG always on and a `FATAL_ERROR` when `CMAKE_MSVC_RUNTIME_LIBRARY` selects the dynamic runtime. The commit is taken as-is and the last commit defaults both options back to `ON`, so the effective build is unchanged while the new opt-outs, the Clang CFG path (`-Xclang -cfguard`) and `USE_SPECTRE_MITIGATED_LIBRARIES` become available.

## What

Four cherry-picks keeping the original authorship, plus one commit of our own:

- `gpuep: control flow guard and msvc static runtime` — adds the Clang MSVC-compat flags, the Clang CFG path and the BinSkim/Spectre options.
- `do not override the EP binary name if provided in the command line` — re-applied by hand: `cmake/deps.cmake` was restructured on `main` (the `if(BUILD_EP)` block was flattened) so the hunk context no longer matched; the change is identical.
- `auto detect hipBLASLt library name` — conflict resolved by dropping the `MIOpen.dll` entry (removed from `main` in #861) and keeping only the `HIPDNN_HIPBLASLT_DLL` detection.
- `fix the definition name` — the `.gitignore` hunk also brings in the two entries from the skipped `3b002576b`.
- `fix(build): keep CFG/static-CRT defaults and unbreak the mock build` — ours; see below.

The mock fix: `find_package(hipblaslt CONFIG REQUIRED)` in `backend-mlir-compiler/custom-op-mlir/CMakeLists.txt` fails configure on mock builds, which by definition have no ROCm SDK. This is exactly what breaks `windows-build-mock` on the release branch:

```
CMake Error at backend-mlir-compiler/custom-op-mlir/CMakeLists.txt:80 (find_package):
  Could not find a package configuration file provided by "hipblaslt"
```

Changed to a `QUIET` probe with a fallback to the plain library name, matching how `lib/Runtime/CMakeLists.txt` already probes the same package. `LlvmIrJit.cpp` only uses the macro under `#ifdef _WIN32` and mock builds never load it.

## Test plan

- [x] `python build.py --mock ... --skip_tests` configures and builds (this is the case that is red on the release branch).
- [x] `python build.py ... --skip_tests` real build green (Ninja, Release, gfx1151); configure logs `Control Flow Guard enabled`, confirming CFG is still on after defaulting the option back to `ON`.
- [x] `pre-commit run --all-files` clean on first pass.
- [x] Verified no MIOpen references leaked into `lib/Runtime/CMakeLists.txt` or `LlvmIrJit.cpp`.
- [ ] CI build-windows + build-windows-mock green.

## Notes for reviewers

The mock fix applies equally to `gpuep/releases/gpuep-rel-2610`, where `windows-build-mock` is red right now. This PR only targets `main`; a separate PR for the release branch is not included.

Review focus: the two judgement calls are skipping `3b002576b` on account of #861, and defaulting `USE_BINSKIM_COMPLIANT_COMPILE_FLAGS` / `USE_MSVC_STATIC_RUNTIME` to `ON` to preserve main's behaviour. If either should go the other way, say so and I will adjust.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -159,3 +159,8 @@ _local/
 # Whisper test audio (auto-downloaded at test setup, cached locally; README kept)
 test/python/data/whisper/*.wav
 test/python/data/whisper/librispeech/
+
+.idea/
+.depend/
+CMakeUserPresets.json
+build.*/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,7 @@ if(WIN32)
     add_compile_options("$<$<COMPILE_LANGUAGE:C,CXX>:-fms-extensions;-fms-compatibility;-fdelayed-template-parsing>")
   endif()
 
-  option(USE_BINSKIM_COMPLIANT_COMPILE_FLAGS "Prepare HIP EP for BinSkim Binary Analyzer" OFF)
+  option(USE_BINSKIM_COMPLIANT_COMPILE_FLAGS "Prepare HIP EP for BinSkim Binary Analyzer" ON)
   option(USE_SPECTRE_MITIGATED_LIBRARIES "Use Spectre-mitigated libraries" OFF)
 
   if(MSVC AND MSVC_VERSION GREATER_EQUAL 1900)
@@ -59,7 +59,7 @@ if(WIN32)
     endif()
   endif()
 
-  option(USE_MSVC_STATIC_RUNTIME "Link MSVC runtime library statically" OFF)
+  option(USE_MSVC_STATIC_RUNTIME "Link MSVC runtime library statically" ON)
 
   if(NOT USE_MSVC_STATIC_RUNTIME)
     set(MSVC_RUNTIME_SUFFIX "DLL")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,16 +28,45 @@ project(
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
 
-if(MSVC)
-  if(NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY)
-    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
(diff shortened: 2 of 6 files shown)


amd-mingw (approved):

amd-yanjunz:
ROCm GenAI CI finished with result: SUCCESS.

- PR: ROCm/hip-ep#877
- Source comment: [models=qwen38_27b](https://github.com/ROCm/hip-ep/pull/877#issuecomment-5462444544)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: Qwen3.8-27B-fp16-ve-fp16-int4-k_quant-gs128-text-dml-onnx
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/555/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6a92d7cf8bd9db260f07c512

PR #881, opened 2026-08-31
feat(hipsr): run materialize-init-tensors in the pipeline with shape links

zpye:
﻿## Summary

`buildHipsrPipeline` now runs `hipsr-materialize-init-tensors`. A pool domain leaves the pipeline as constants, shape computations, `tensor.empty` allocations, data ops, then one `hipsr.preserve_shape` per allocation.

The pass emits that order into a fresh domain block instead of moving ops into place, so emission order is final order and nothing is repositioned afterwards. Enabling it in the pipeline turned up two cases it could not yet handle.

## Related issue or design

None. #790 touches this pass too, and it needs a rebase rather than just ordering: it patches `replaceAndCleanup` and the `@already_grouped`, `@static_shape`, and `@multi_dynamic` tests, none of which survive this PR.

## Compiler IR

Trimmed: `#hipsr.mem<device>` and operand type lists omitted.

```mlir
// Before
%init = hipsr.placeholder(%ctx) ins(%a, %b) : tensor<?x512xf16> shape_region {
^bb0(%a_shape: !shape.shape, %b_shape: !shape.shape):
  ...
  hipsr.shape_yield %shape : !shape.shape
}
%0 = hipsr.matmul(%ctx) ins(%a, %b) outs(%init) : tensor<?x512xf16>

// After
%s = scf.execute_region -> !shape.shape {
  %a_shape = shape.shape_of %a : tensor<?x256xf16> -> !shape.shape
  %b_shape = shape.shape_of %b : tensor<256x512xf16> -> !shape.shape
  ...
  scf.yield %shape : !shape.shape
}
%c0 = shape.const_size 0
%extent = shape.get_extent %s, %c0        // only dim 0; 512 is in the type
%d0 = shape.size_to_index %extent
%init = tensor.empty(%d0) : tensor<?x512xf16>
%0 = hipsr.matmul(%ctx) ins(%a, %b) outs(%init) : tensor<?x512xf16>
hipsr.preserve_shape %s, %0 : tensor<?x512xf16>
```

**Fix 1, barrier regions.** Their arguments are `ctx` and the input tensors, which is the placeholder's own operand list, so they map straight through.

```mlir
// Before
^bb0(%rctx: !hipsr.context, %rin: tensor<?x1xf32>, %rextents: tensor<2xi64>):
  %in_shape = shape.shape_of %rin : tensor<?x1xf32> -> tensor<2xindex>
  %column = tensor.extract %rextents[%c1] : tensor<2xi64>

// After, nothing built at region entry
%s = scf.execute_region -> !shape.shape {
  %in_shape = shape.shape_of %in : tensor<?x1xf32> -> tensor<2xindex>
  %column = tensor.extract %extents[%c1] : tensor<2xi64>
  ...
}

// New error: the pool hands buffers out only after the last allocation, and
// -hipsr-partition-pool-domains keeps a barrier one domain past its inputs.
%cast_init = hipsr.placeholder(%ctx) ins(%in) <normal> ...
// expected-error {{barrier input must be allocated outside this pool domain}}
%barrier_init = hipsr.placeholder(%ctx) ins(%cast_init) <barrier> ...
```

**Fix 2, constant dominance.** Conversion emits a constant next to the op that consumes it, so a shape region can read one defined below where its computation has to land. Emitting the constants late fails all three `Dialect/Hipsr/Pipelines/` tests with `operand #0 does not dominate this use`.

```mlir
// Before, %weight sits below the front of the block, where computations go
%cast = hipsr.cast(%ctx) ins(%a) outs(%cast_init)
%weight = hipsr.constant {value = dense<...>} : tensor<2x2xf32>
%add_init = hipsr.placeholder(%ctx) ins(%cast_init, %weight) ...

// After. A constant takes no operands, so every constant in the domain comes
// over first, where it dominates every computation. The data op below reads
// that same clone.
%weight = hipsr.constant {value = dense<...>} : tensor<2x2xf32>
%cast_shape = scf.execute_region -> !shape.shape { ... }
%add_shape = scf.execute_region -> !shape.shape {
  %weight_shape = shape.shape_of %weight : tensor<2x2xf32> -> !shape.shape
  ...
}
%cast_init = tensor.empty() : tensor<2x2xf32>
%add_init = tensor.empty() : tensor<2x2xf32>
%cast = hipsr.cast(%ctx) ins(%a) outs(%cast_init)
%add = hipsr.add(%ctx) ins(%cast, %weight) outs(%add_init)
```

The step hoists every constant rather than the ones a shape region reads: a constant has no operands, so the front of the domain dominates every reader either way, and one walk reaches each constant once. A placeholder input can only be a block argument, another placeholder, or a constant, so nothing else needs to come over early.

## Shape links

The pass built a shape per placeholder result, sized an allocation with it, then dropped it, so nothing downstream could tell which shape belonged to which buffer. `-hip-use-output-allocator` is the only reader of these links, and no pass produced one.

A link names the result of the op that fills the buffer rather than the allocation feeding its `outs`, which is why it comes after the data ops. For a destination-passing op the two become one buffer and the choice looks free, but:

- `hipsr.compute` is not destination-passing. Its result may be a view carrying a shape the destination does not: in `OneShotBufferize/compute.mlir`, a `memref<?xf16>` collapse of a `memref<?x256xf16>` destination.
- Naming the allocation puts a use on the empty tensor after its writer consumed it, so One-Shot Bufferize gives the writer a buffer of its own. Linking the inits in that same graph turns two allocations into four and strands both links on buffers nothing fills (measured by bufferizing the test both ways).

`BufferizableOpInterfaceImpl` already had a `ComputeOp`-only `getResultHeldIn` for this mapping. It becomes `getResultForDestination` beside the other destination helpers, takes any hipsr op, and serves both callers.

## Tests

LIT: 458 passed, 12 unsupported (pre-existing). Most of the diff is regenerated expectations for `Dialect/Hipsr/Pipelines/` and the rewritten pass test, both now carrying the `hipsr.preserve_shape` lines; `embedding.mlir` covers a placeholder whose two results each get one.

Four positive cases remain, each owning a boundary of its own:

| Case | Boundary |
|---|---|
| `@interleaved` | one dynamic extent, plus the shape one placeholder reads from another |
| `@barrier_domain` | two dynamic extents, plus the barrier argument layout |
| `@constant_between_placeholders` | no dynamic extent, plus constant dominance |
| `@no_placeholder` | domain left alone |

Every case spells its whole function out with `CHECK-NEXT`, so a leftover placeholder or an extent read for a static dimension breaks the chain. The single-placeholder, `@static_shape`, and `@multi_dynamic` cases hand their boundaries to the four above; `@already_grouped` goes because it pinned the skip that kept a placeholder from being moved ahead of itself, and a rebuild moves nothing. `invalid.mlir` covers both diagnostics the pass emits.

## AI assistance

AI-assisted (Claude Opus 5 via Cursor): the pass rewrite, the shape-link step and its shared helper, the LIT expectations, and this description, validated by the LIT suite and `pre-commit`. I reviewed the result and own the design decisions.

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrOps.h
+++ b/include/hip/Dialect/Hipsr/IR/HipsrOps.h
@@ -49,12 +49,15 @@ bool isHipsrDestinationOperand(::mlir::OpOperand &use);
 
 // Returns the result that aliases this destination operand.
 // DPS pattern: outs()[i] and result[i] occupy the same buffer.
-// hipsr.compute is not DPS but holds its results in the same positions.
-// Returns null if the operand is not a destination.
+// hipsr.compute is not DPS but stores its results in the same positions.
+// Returns a null OpResult if the operand is not a destination.
 //
 // Example: %r0, %r1 = hipsr.foo outs(%init0, %init1)
 //          getResultForDestination(%init0) -> %r0
 //          getResultForDestination(%init1) -> %r1
+//
+// Example: %r = hipsr.compute outs(%d : tensor<?x256xf16>) : tensor<?xf16>
+//          getResultForDestination(%d) -> %r
 ::mlir::OpResult getResultForDestination(::mlir::OpOperand &use);
 
 } // namespace hipsr
--- a/include/hip/Dialect/Hipsr/IR/HipsrOps.h
+++ b/include/hip/Dialect/Hipsr/IR/HipsrOps.h
@@ -47,8 +47,15 @@ namespace hipsr {
 // dps: use is in Init
 bool isHipsrDestinationOperand(::mlir::OpOperand &use);
 
-// return the result held in the destination use names, or null if none.
-::mlir::OpResult getHipsrResultHeldIn(::mlir::OpOperand &use);
+// Returns the result that aliases this destination operand.
+// DPS pattern: outs()[i] and result[i] occupy the same buffer.
+// hipsr.compute is not DPS but holds its results in the same positions.
+// Returns null if the operand is not a destination.
+//
+// Example: %r0, %r1 = hipsr.foo outs(%init0, %init1)
+//          getResultForDestination(%init0) -> %r0
+//          getResultForDestination(%init1) -> %r1
+::mlir::OpResult getResultForDestination(::mlir::OpOperand &use);
 
 } // namespace hipsr
 } // namespace mlir
--- a/include/hip/Dialect/Hipsr/IR/HipsrOps.h
+++ b/include/hip/Dialect/Hipsr/IR/HipsrOps.h
@@ -47,6 +47,9 @@ namespace hipsr {
 // dps: use is in Init
 bool isHipsrDestinationOperand(::mlir::OpOperand &use);
 
+// return the result held in the destination use names, or null if none.
+::mlir::OpResult getHipsrResultHeldIn(::mlir::OpOperand &use);
+
 } // namespace hipsr
 } // namespace mlir
 
--- a/lib/Dialect/Hipsr/IR/HipsrOps.cpp
+++ b/lib/Dialect/Hipsr/IR/HipsrOps.cpp
@@ -61,7 +61,7 @@ bool mlir::hipsr::isHipsrDestinationOperand(OpOperand &use) {
   return index >= begin && index < begin + destinations.size();
 }
 
-OpResult mlir::hipsr::getHipsrResultHeldIn(OpOperand &use) {
+OpResult mlir::hipsr::getResultForDestination(OpOperand &use) {
   if (!isHipsrDestinationOperand(use)) {
     return {};
   }
--- a/lib/Dialect/Hipsr/IR/HipsrOps.cpp
+++ b/lib/Dialect/Hipsr/IR/HipsrOps.cpp
@@ -60,3 +60,18 @@ bool mlir::hipsr::isHipsrDestinationOperand(OpOperand &use) {
   unsigned begin = destinations.getBeginOperandIndex();
   return index >= begin && index < begin + destinations.size();
 }
+
+// A hipsr op holds the result in the destination slot at the same position.
+OpResult mlir::hipsr::getHipsrResultHeldIn(OpOperand &use) {
+  if (!isHipsrDestinationOperand(use)) {
+    return {};
+  }
+  Operation *op = use.getOwner();
+  unsigned slot = use.getOperandNumber() -
+                  getHipsrDestinationOperands(op).getBeginOperandIndex();
+  // Bufferization keeps the destinations and drops the results they held.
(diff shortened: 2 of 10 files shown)


wcy123 (commented):
**Comment doesn't clarify Normal vs Barrier shape region arguments**

The comment at lines 75-78 doesn't clearly distinguish what Normal vs Barrier placeholders receive in their shape regions.

**Barrier shape region arguments:**
- a. ctx (first argument)
- b. `arith.constant` result
- c. `hipsr.constant` result
- d. Block argument of pool domain block
- e. NOT placeholder results (rejected at line 168)

**Normal shape region arguments:**
- a. `shape.shape_of` arith.constant
- b. `shape.shape_of` hipsr.constant
- c. `shape.shape_of` block argument
- d. Shape from other placeholder's execute_region

The comment should state this distinction clearly.

wcy123 (dismissed):
LGTM

include/hip/Dialect/Hipsr/IR/HipsrOps.h
```diff
@@ -47,6 +47,9 @@ ::mlir::Value getShapeGraphCounterpart(::mlir::Value value);
 // dps: use is in Init
 bool isHipsrDestinationOperand(::mlir::OpOperand &use);
 
+// return the result held in the destination use names, or null if none.
+::mlir::OpResult getHipsrResultHeldIn(::mlir::OpOperand &use);
```
wcy123:
```c++
 // Returns the result that aliases this destination operand.
  // DPS pattern: outs()[i] and result[i] occupy the same buffer.
  // Returns null if the operand is not a destination.
  //
  // Example: %r0, %r1 = hipsr.foo outs(%init0, %init1)
  //          getResultForDestination(%init0) -> %r0
  //          getResultForDestination(%init1) -> %r1

::mlir::OpResult getResultForDestination(::mlir::OpOperand &use);
```

zpye: Renamed to `getResultForDestination` and took your comment, with one line added: `hipsr.compute` reaches this helper too and is not destination-passing, so the DPS line alone would read as the whole contract.

```c++
// Returns the result that aliases this destination operand.
// DPS pattern: outs()[i] and result[i] occupy the same buffer.
// hipsr.compute is not DPS but holds its results in the same positions.
// Returns null if the operand is not a destination.
//
// Example: %r0, %r1 = hipsr.foo outs(%init0, %init1)
//          getResultForDestination(%init0) -> %r0
//          getResultForDestination(%init1) -> %r1
::mlir::OpResult getResultForDestination(::mlir::OpOperand &use);
```

In 8ce09b97. Full LIT suite still at 458 passed, 12 unsupported.

wcy123 (approved):
LGTM

PR #882, opened 2026-08-31
fix(gqa): KV-cache-sharing decoder layers (Gemma 4 E2B/E4B) + gfx1170 WMMA port

anilmartha:
## Summary

Two commits that together make Gemma 4 (Gemma-3n family) KV-cache sharing correct on hip-ep. The first makes the sharing layers run at all; the second makes them compute the right thing when there is more than one KV head.

Gemma 4 shares KV caches across depth. In E2B only the first 15 of 35 decoder layers own a cache (`past_key_values.0..14` / `present.0..14`); in E4B it is the first 24 of 42. The remaining layers have no `past_key`, no `past_value` and no `present` outputs at all — they re-read an earlier layer's already-complete cache through a `Transpose`/`Reshape` off that layer's `present.N.key`. They therefore hand `hip.gqa` the **full Skv-length K/V with no cache to append into**, while the query is a single decode token, and they carry an external additive mask with `is_causal=0` (the mask, not the built-in triangle, encodes causality).

### 1. `fix(gqa): support KV-cache-sharing decoder layers`

Structurally the above is the bidirectional no-past shape the runtime already implements for Whisper's encoder self-attn and decoder cross-attn, but two independent assumptions routed it onto the autoregressive decode path. All 20 of E2B's sharing layers failed on every generated token with

```
gqa_forward_hipblaslt: invalid seqlens_k[0]+1=17 (sq=1, past_len=16, present_seq=1, past_buf_seq=0)
wrap_group_query_attention: legacy decomposed pipeline failed (rc=-1)
```

leaving each layer's output zero-filled, so the model emitted grammatical but meaningless text.

- **`OnnxAttentionConversion.cpp` derived the KV token count from the query's seq extent**, on the documented assumption that Q/K/V share the seq axis. True for a cached layer (K/V carry only the new tokens), false for a sharing layer (1 query, full history in K/V). `present` was sized to a single token — the `present_seq=1` above — and since `GqaLowering` reads `seq_len_kv` off `present_key`'s seq extent, the runtime was told the layer had one key. Now taken from `dim(k3, 1)`, matching the convention the sibling `MultiHeadAttention` lowering already uses for Whisper cross-attn.

- **`gqa_forward_hipblaslt` selected the KV convention with `no_causal && (attention_bias == nullptr)`** — it asked whether a *mask* was absent. In previously-exercised models "no mask" coincided with "no past KV"; Gemma's sharing layers have both a mask and no cache. The mask is orthogonal (Step 8b adds it either way), so the gate now keys on `past_key == nullptr`.

### 2. `fix(gqa): stage BSHD KV with a transpose, not a copy`

With the above, E2B was correct but **E4B was silently wrong** — no diagnostic, just output that stayed fluent for a few hundred tokens and then degraded into repetition and repeated LaTeX arrows.

The no-past path stages the full Skv-length K/V into the BNSD `present` cache the GEMMs read, choosing between a straight D2D copy and the transposing append kernel by testing `sq != Skv`. That was another proxy that held only by coincidence: the one BNSD producer (MHA cross-attn) also happened to be the one with `sq != Skv`. A KV-sharing decode is `sq=1` against a full-history `Skv` but arrives as rank-3 BSHD `[B, Skv, G*d]`, so BSHD data took the BNSD copy.

BSHD and BNSD are byte-identical only at `G == 1`. E2B's sharing layers are `kv_num_heads=1`, so the copy was accidentally correct; **E4B's 18 sharing layers are `kv_num_heads=2`**, so the copy interleaved the heads across the sequence, leaving head `g` at position `s` reading token `(g*Skv + s)/G`. Because those layers still read genuine K/V from genuine tokens — just the wrong ones — the model degraded rather than failing. Prefill was unaffected: `sq == Skv` there already selected the transposing append.

Fixed by branching on the real source layout. `GqaLowering` knows the key memref's rank (rank-4 BNSD vs rank-3 BSHD), so it passes a `kv_bnsd` flag and `update_kv_cache` tests it; the BSHD branch calls `hip_gqa_kv_cache_append` with `src_tokens = Skv`, `past_len = 0`, which places each `(s, g)` in its BNSD slot for any `G` at the same memory traffic as the copy.

### Behaviour-change scope

Both fixes are confined to previously-broken configurations.

For (1): Whisper encoder / cross-attn (`no_causal`, no mask, no past) evaluates identically under both expressions; Whisper decoder self-attn and every GQA model are `no_causal=0` and never reach the flag; `is_causal=0` + mask + a past cache is `false` either way. Only `is_causal=0` + mask + **no** past differs, and that was rejected outright before.

For (2): Whisper cross-attn is rank-4, so it still takes the copy; Whisper encoder self-attn is rank-3 with `sq == Skv`, so the append receives `src_tokens == sq` exactly as before. Only a rank-3 BSHD source with `Skv != sq` differs.

The runtime ABI goes 41 → 42 params. `mock_gpu.cpp` is updated for parity and `hip-to-llvm/test_gqa.mlir`'s signature CHECK gains the trailing `i64`. The archived `gqa_back.cpp` and its `back_wrapper_decl.h` bench shim keep their older signature deliberately — a separate, non-built target that was already two params behind.

### Notes for reviewers

The `seqlens_k[0]+1=17` in the original diagnostic was itself inconsistent with the buggy sizing (a sharing layer computed its own `seqlens_k` as 0, so it should have said `1`). It leaked in from the per-`Compute()` cache in `read_seqlens_k_for_dispatch`, which is keyed **only on the device pointer**: each `onnx.Attention` node synthesizes its own `seqlens_k` scratch buffer, those alias within a forward pass, and a sharing layer read a cached layer's value. `HIPDNN_EP_GQA_CACHE_SEQLENS=0` silenced the message while answers stayed wrong, which is what isolated the two root causes.

That aliasing is moot after these fixes — all layers now compute the same `seqlens_k`, and the sharing layers ignore it on the no-past path — but pointer-keying is only sound when every attention node reads one shared `seqlens_k` graph input (true for ORT-GenAI GQA models, false for the `onnx.Attention` lowering). Left as-is rather than traded against its documented ~30-45 ms/token win; flagged for whoever hits a model with genuinely differing per-layer cache lengths.

Worth noting that both root causes in commit 2 and commit 1 are the same class of defect: a correlated signal (`attention_bias == nullptr`, `sq != Skv`) standing in for the real fact (`past_key == nullptr`, source layout). Both are now plumbed explicitly rather than inferred.

### 3. `fix(kernels): port linear-attention and MHA WMMA sites to the gfx12 encoding`

Independent of the two GQA fixes above, but found while chasing a Qwen3.6-35B-A3B failure on the same hardware, so it rides along here.

`hip_arch_compat.h` probes WMMA capability against **both** spellings of the f16 16x16x16 intrinsic, because gfx1170 sets `__GFX11__` yet exposes only `__builtin_amdgcn_wmma_f32_16x16x16_f16_w32_gfx12`, not the plain `..._w32` that gfx1151 and earlier RDNA3 provide. `gemm_wmma_kernel.hip`, `gqa_kernel.hip` and `matmul_nbits_kernel.hip` were moved onto that abstraction; **`linear_attention_kernel.hip` and `multi_head_attention_kernel.hip` were missed** and still guarded on the gfx11 spelling alone, falling back to `__builtin_trap()`.

So on gfx1170 every WMMA call site in both files compiled to `s_trap`. A compile probe confirms the asymmetry directly: at `--offload-arch=gfx1170` the gfx11 builtin is MISSING and the `_gfx12` builtin is AVAILABLE. Any linear-attention prefill therefore faulted the device with `unspecified launch failure` (HIP 719), poisoning the HIP context so every later call — including `hipMalloc` — failed; only the scalar decode path (`S == 1`, no WMMA) worked. Qwen3.6-35B-A3B died in its **first** decoder layer, at the gated-DeltaNet block, never reaching the MoE.

Both files now use `HIPDNN_WMMA_F32_16X16X16_F16`, `HIPDNN_WMMA_FRAG_ELEMS` (16 on gfx11, 8 on gfx12 where K splits across lane halves instead of replicating), `hipdnn_wmma_k_off(pair)` on every fragment load, and `hipdnn_wmma_acc_row(pair, e)` for the accumulator rows. The trap fallback now keys on `!HIPDNN_HAS_WMMA`, so it still fires on CDNA/wave64 and nowhere else. `multi_head_attention_kernel.hip` also needed `mha_load_dtile()` split in two: the plain 16-wide D-tile copy that stages V into LDS keeps a fixed-width `half16_raw_t`, while the new `mha_load_frag()` returns an arch-width fragment at `dt*16 + k_off`.

`gqa_kernel_back.hip` keeps its gfx11-only guard deliberately — archived variant, absent from CMakeLists, never built.

## Test plan

- [x] `test_onnx_attention.mlir` passes, including a new KV-cache-sharing section pinning `present`'s seq extent to `dim(K,1)`. It is the only lit test exercising native `onnx.Attention`, and all pre-existing sections still pass.
- [x] `hip-to-llvm/test_gqa.mlir` passes with the updated 42-param signature CHECK.
- [x] Real gfx1170, default seqlens cache **enabled**, both models via the AMDGPU EP: `gemma-4-E2B-it-fp16-int4-kquant` answers `The capital of France is **Paris**.` and stays coherent across a 700-token generation; `gemma-4-E4B-it-fp16-int4-kquant` stays coherent across a 900-token generation on the prompt that previously collapsed, including the markdown table that had degenerated into arrow spam.
- [x] E2B also covers the `G == 1` path now taking the append rather than the copy.
- [x] Real gfx1170, WMMA port (commit 3). These compare against an ORT CPU reference, so they cover numerical correctness of the gfx12 fragment layout and not merely the absence of a trap:
  - `test_linear_attention.py` 9/9 — `test_la_chunk_opt_shape[128]` device-faulted before and passes now (`[1]` passed before too: scalar decode, no WMMA).
  - `test_multi_head_attention.py` 7/7 — previously 1 pass then a device fault that killed the process.
  - The 90 Whisper attention tests stay at 90/90; they lower to `hip.gqa` and never exercised either trap.
  - 15 further numeric files pass with no failures.
- [x] Full CI green on commit 1 (`build_real`, `build_linux`, `build_mock`, `gpu_test`, `pre-commit`); re-running for commits 2 and 3.

### Known-failing, pre-existing, out of scope

Verified identical on the pre-change build, so none of these are introduced here:

- `test_gqa.py` fails 14 of 22 both before and after — padded-chunk prefill where `seqlens_k` implies `total_seq < sq`, and a decode needing `total_seq > present_seq`. Both are rejected inside `gqa_forward_fused`, which this PR does not touch.
- `test_conv1d.py` hangs in a MIOpen fp32 convolution on gfx1170 (`CK grouped conv library not found for device gfx1170`).
- `test_matmul_nbits.py` hangs at `lm_head_gpt_oss_shape[15]`.


amd-yanjunz:

- PR: ROCm/hip-ep#882
- Source comment: [node=xconucstrhalo19](https://github.com/ROCm/hip-ep/pull/882#issuecomment-5477547888)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xconucstrhalo19
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/593/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6a958818bf0bd6c7b64cfc1c
- Performance compare: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fids%3D6a9491498bd9db260f07c81b%26collection%3DHIP-EP-Dynamic&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=593

amd-yanjunz:

- PR: ROCm/hip-ep#882
- Source comment: [node=xconucstrhalo19](https://github.com/ROCm/hip-ep/pull/882#issuecomment-5487534282)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xconucstrhalo19
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/607/

amd-yanjunz:

- PR: ROCm/hip-ep#882
- Source comment: [node=xconucstrhalo19](https://github.com/ROCm/hip-ep/pull/882#issuecomment-5526038653)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xconucstrhalo19
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/698/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6a998d2d9527aa1ad9c7af4d
- Compare with latest main branch results: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fids%3D6a986d26b38f5467b26504ac%26collection%3DHIP-EP-Dynamic&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=698

amd-mingw (approved):
LGTM

PR #885, opened 2026-08-31
mlir-imp: preserve unsigned identity for inline dense integer constants (fix QDQ sign-flip)

kirankumarmatam-amd:
## Problem
onnxElementTypeToMlirDenseElementType flattened every signed/unsigned integer to signless, so inline ONNX unsigned constants (UINT8, UINT16, ...) were imported as signless integers of the same width. The HipToLLVM QDQ dtype classifier needs to distinguish an unsigned type from the signed type of the same width; collapsing them makes the two indistinguishable, corrupting DequantizeLinear output.

## Fix
Preserve the unsigned identity of all inline dense unsigned integers (ui8, ui16, ...) so they keep their unsigned type through lowering, matching how external unsigned initializers are already imported (via the offset/size path that bypasses DenseElementsAttr).

## Test plan
- Added unit tests, test/unit/test_dtype_mapping.cpp.
- Re-validated the enabled models, all passing.

### DIFF
--- a/morphizen/mlir-imp/src/mlir-constants.cpp
+++ b/morphizen/mlir-imp/src/mlir-constants.cpp
@@ -75,16 +75,17 @@ mlir::Type onnxElementTypeToMlirDenseElementType(int element_type,
                                                  mlir::OpBuilder &builder) {
   mlir::Type elem = onnxElementTypeToMlirElementType(element_type, builder);
   if (auto intTy = mlir::dyn_cast<mlir::IntegerType>(elem)) {
-    // Preserve unsigned 16-bit (ONNX UINT16) on inline dense constants so they
+    // Preserve unsigned integers (ONNX UINTx) on inline dense constants so they
     // retain their unsigned identity through lowering, matching how EXTERNAL
-    // UINT16 initializers are already imported (ui16, via the offset/size path
-    // that bypasses DenseElementsAttr). A downstream QDQ dtype classifier must
-    // distinguish UINT16 from signed INT16: collapsing UINT16 to signless i16
-    // here makes the two indistinguishable, so inline unsigned constants with
-    // values >= 32768 get read as INT16 and sign-flip. LLVM 22 MLIR accepts
-    // unsigned DenseElementsAttr, and ui16 memrefs already flow end-to-end for
-    // external constants, so keeping ui16 here is consistent and safe.
-    if (intTy.getWidth() == 16 && intTy.isUnsigned())
+    // unsigned initializers are already imported (via the offset/size path that
+    // bypasses DenseElementsAttr). A downstream QDQ dtype classifier must
+    // distinguish an unsigned type from the signed type of the same width:
+    // collapsing e.g. UINT16 to signless i16 makes the two indistinguishable,
+    // so inline unsigned constants with values >= 2^(width-1) get read as
+    // signed and sign-flip. LLVM 22 MLIR accepts unsigned DenseElementsAttr,
+    // and unsigned memrefs already flow end-to-end for external constants, so
+    // keeping the unsigned type here is consistent and safe.
+    if (intTy.isUnsigned())
       return elem;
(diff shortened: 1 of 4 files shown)


morphizen/mlir-imp/src/mlir-constants.cpp
```diff
@@ -75,6 +75,19 @@ mlir::Type onnxElementTypeToMlirDenseElementType(int element_type,
...
+    // UINT16 initializers are already imported (ui16, via the offset/size path
+    // that bypasses DenseElementsAttr). The HipToLLVM QDQ dtype classifier must
+    // distinguish UINT16 (uint16) from signed INT16 (int16, added for the Quark
+    // ResNet50-INT8 bias): collapsing UINT16 to signless i16 here makes it
+    // indistinguishable from INT16, so small inline quantized constants (e.g.
+    // google_bert's per-layer attention sqrt(d_k) scaling scalar) get read as
+    // INT16 and sign-flip on values >= 32768. LLVM 22 MLIR accepts unsigned
+    // DenseElementsAttr, and ui16 memrefs already flow end-to-end for external
+    // constants, so keeping ui16 here is consistent and safe.
+    if (intTy.getWidth() == 16 && intTy.isUnsigned())
```
fhanuman: Is there any reason why we want to check specifically only for `16` bit-width, maybe just check for `isUnsigned` here?

morphizen/mlir-imp/src/mlir-constants.cpp
```diff
@@ -75,6 +75,19 @@ mlir::Type onnxElementTypeToMlirDenseElementType(int element_type,
                                                  mlir::OpBuilder &builder) {
   mlir::Type elem = onnxElementTypeToMlirElementType(element_type, builder);
   if (auto intTy = mlir::dyn_cast<mlir::IntegerType>(elem)) {
+    // Preserve unsigned 16-bit (ONNX UINT16) on inline dense constants so they
+    // retain their unsigned identity through lowering, matching how EXTERNAL
+    // UINT16 initializers are already imported (ui16, via the offset/size path
+    // that bypasses DenseElementsAttr). The HipToLLVM QDQ dtype classifier must
+    // distinguish UINT16 (uint16) from signed INT16 (int16, added for the Quark
```
fhanuman: Please avoid adding models details in the generic logic.

test/unit/test_dtype_mapping.cpp
```diff
@@ -0,0 +1,122 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
```
fhanuman, line 3: probably LIT test is sufficient here?

morphizen/mlir-imp/src/mlir-constants.cpp
kirankumarmatam-amd: Fixed all the comments removing model names.

morphizen/mlir-imp/src/mlir-constants.cpp
kirankumarmatam-amd: I have only tested this with uint16 change for all the current models we have enabled. With just isUnsigned(), I need to re-validate all the models.

test/unit/test_dtype_mapping.cpp
kirankumarmatam-amd: hip-mlir-opt has no ONNX-protobuf front end and only consumes textual MLIR, so the ONNX-dtype -> MLIR-type mapping is not reachable with a LIT test.

morphizen/mlir-imp/src/mlir-constants.cpp
kirankumarmatam-amd: Re-validated all the enabled models with just isUnsigned() after dropping getWidth()==16, and all the models are passing.

fhanuman (approved):

PR #886, opened 2026-08-31
Support packed INT4/UINT4 weights

Vandana9628:
## Summary
Models with **4-bit (INT4/UINT4) weights** currently fail to compile. ONNX INT4/UINT4 has no MLIR element type, so 4-bit initializers are imported as an **i8/ui8 tensor of the LOGICAL element count** while the backing buffer holds only `ceil(numel/2)` **packed bytes**. Verified on a real model: the qkv_proj weight arrives as `tensor<9216x3072x1x1xi8>` .

**Three changes across constant handling:**

1. **`hip.constant` verifier**: It also accepts `size == ceil(numElements/2)` for an 8-bit-integer result. Every other size is still a genuine mismatch.

2. **`hip-externalize-constants`** (`lib/Dialect/Transforms/ExternalizeConstants.cpp`) derived the storage/read size from the logical element count, so it would size the buffer and read the file at twice the packed length.

3. **`convert-onnx-to-hip`** (`lib/Conversion/OnnxToHip/OnnxToHip.cpp`) now stamps a `packed_int4` unit attribute on the `DequantizeLinear` ops that consume such a constant (detected by `size == ceil(numel/2)` on an 8-bit operand), so their downstream lowering nibble-unpacks instead of reading the packed buffer as full-width i8 (an out-of-bounds device read). Covers both inline and external constant paths; splat inline data is skipped.

## Test plan
- [x] LIT: `test/lit/Conversion/onnx-to-hip/test_constant_packed_int4.mlir`.
- [x] Mock build (`build.py --mock`) is clean.

### DIFF
--- a/lib/Conversion/OnnxToHip/OnnxToHip.cpp
+++ b/lib/Conversion/OnnxToHip/OnnxToHip.cpp
@@ -40,6 +40,16 @@ namespace {
 
 constexpr llvm::StringLiteral kOrtMemoryAddressLocation = "*/_ORT_MEM_ADDR_/*";
 
+/// Classification of an 8-bit constant's backing byte size against its element
+/// count, returned by markPackedInt4Consumers so the caller can diagnose a
+/// genuine mismatch.
+enum class PackedInt4Result {
+  NotApplicable, ///< Not an 8-bit static constant, or byte size unknown.
+  Unpacked,      ///< Full-width storage (bytes == numElements): plain int8.
+  Packed,        ///< Half-width storage (bytes == ceil(numElements/2)): int4.
+  SizeMismatch   ///< 8-bit storage that is neither full nor packed: malformed.
+};
+
 /// Mark the DequantizeLinear consumers of a 4-bit-packed constant.
 ///
 /// ONNX INT4/UINT4 is imported as i8/ui8 at the logical element count, but its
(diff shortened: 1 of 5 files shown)


lib/Dialect/IR/HipDialect.cpp
```diff
@@ -200,7 +200,11 @@ LogicalResult ConstantOp::verify() {
   int64_t size = getSizeAttr().getInt();
   if (size <= 0)
     return emitOpError("external source `size` must be positive");
-  if (size != expectedBytes)
+  // ONNX INT4/UINT4 imports as i8/ui8 at the logical element count, but its
```
fhanuman: HIP dialect is supposed to be front-end agnostic, so please avoid adding front-end specific comments.

lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -40,6 +40,36 @@ namespace {
...
+      mlir::dyn_cast<mlir::IntegerType>(tensorType.getElementType());
+  if (!elemIntTy || elemIntTy.getWidth() != 8 || !tensorType.hasStaticShape())
+    return;
+  int64_t numel = tensorType.getNumElements();
+  if (numel <= 0 || packedBytes <= 0 || packedBytes * 2 != numel)
+    return;
+  mlir::OpBuilder builder(constOp->getContext());
+  for (mlir::Operation *user : constOp->getResult(0).getUsers()) {
+    llvm::StringRef opName = user->getName().getStringRef();
+    bool isDq = opName == "onnx.DequantizeLinear";
```
fhanuman, line 94: Curious if there is generic way to address this instead of looking for `onnx.DequantizeLinear`

lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -40,6 +40,36 @@ namespace {
...
+/// is that true backing byte count; only 8-bit storage can halve this way.
+static void markPackedInt4Consumers(mlir::Operation *constOp,
+                                    mlir::RankedTensorType tensorType,
+                                    int64_t packedBytes) {
+  auto elemIntTy =
+      mlir::dyn_cast<mlir::IntegerType>(tensorType.getElementType());
+  if (!elemIntTy || elemIntTy.getWidth() != 8 || !tensorType.hasStaticShape())
+    return;
+  int64_t numel = tensorType.getNumElements();
+  if (numel <= 0 || packedBytes <= 0 || packedBytes * 2 != numel)
```
fhanuman: If the intention is to throw appropriate error, we can enhance the call-site of `markPackedInt4Consumers`

lib/Conversion/OnnxToHip/OnnxToHip.cpp
```diff
@@ -40,6 +40,36 @@ namespace {
...
+/// backing size is known) to make their lowering nibble-unpack. `packedBytes`
+/// is that true backing byte count; only 8-bit storage can halve this way.
+static void markPackedInt4Consumers(mlir::Operation *constOp,
+                                    mlir::RankedTensorType tensorType,
+                                    int64_t packedBytes) {
+  auto elemIntTy =
+      mlir::dyn_cast<mlir::IntegerType>(tensorType.getElementType());
+  if (!elemIntTy || elemIntTy.getWidth() != 8 || !tensorType.hasStaticShape())
+    return;
+  int64_t numel = tensorType.getNumElements();
```
fhanuman, line 80: What is the expectation here for odd number of elements?

lib/Dialect/IR/HipDialect.cpp
Vandana9628: Reworded and made it generic.

lib/Conversion/OnnxToHip/OnnxToHip.cpp
Vandana9628: DequantizeLinear is the only consumer that unpacks. A generic "mark all users" would tag ops that never unpack. Since ONNX guarantees `zero_point.dtype == x.dtype`, tagging the dequant covers both its packed operands.

lib/Conversion/OnnxToHip/OnnxToHip.cpp
Vandana9628: Fixed the marker to use the same ceil relation and added an odd count.

lib/Conversion/OnnxToHip/OnnxToHip.cpp
Vandana9628: Good point. `markPackedInt4Consumers` now returns a `PackedInt4Result`, and the call site in `lowerOnnxConstants` emits an error when the 8-bit source size is neither the full element count nor the packed count. Added `test_constant_packed_int4_invalid.mlir` to cover it.

Vandana9628:
@fhanuman addressed all the comments, this is ready for review.

fhanuman (approved):

PR #890, opened 2026-09-01
feat(qdq): add QuantizeLinear and DequantizeLinear support

amd-genmingz:
## Summary

Adds end-to-end support for ONNX `QuantizeLinear` and `DequantizeLinear`, so graphs containing explicit Q/DQ nodes no longer partition those nodes back to CPU. The change spans all layers: HIP dialect ops, ONNX-to-HIP conversion, HIP-to-LLVM lowering, the runtime wrapper ABI, and new custom HIP kernels.

Supported today: storage types `int8`/`uint8`/`int16`/`uint16`/`int4`/`uint4`, float types `fp32`/`fp16`, and all three ONNX quantization granularities (per-tensor, per-axis, blocked). 

## Why

Q/DQ nodes appear in every quantized ONNX model produced by the standard quantization tooling. Without them, an error will be thrown when the model contains unfused QDQs.

Two ABI choices are worth calling out because they are subtractive and may look like oversights:

- **`output_dtype` is not carried past conversion.** The MLIR importer maps ONNX 8/16-bit storage types one-to-one (`UINT8 -> ui8`, `INT8 -> i8`, `UINT16 -> ui16`, `INT16 -> i16`), so the quantized element type already encodes signedness and width and passing the ONNX attribute as well would invite the two disagreeing. INT4 / UINT4 break that mapping: both import as `i8` / `ui8` at the logical element count, so width travels as `packed_int4` and becomes `out_bits` / `in_bits` at the wrapper; `out_dtype` then supplies only signedness. A `float8` target would still need its own channel.
- **`saturate` reaches `wrap_quantize_linear` but not the kernel.** Per the ONNX spec it only affects `float8` targets, and the range clamp for every integer target supported here (including packed int4) is unconditional, so the kernel has nothing to branch on. It is kept at the wrapper boundary so a future `float8` path does not have to re-thread it through the dialect and lowering.

## What

**Dialect** — `hip.quantize_linear` (`axis`, `block_size`, `precision`, `saturate`) and `hip.dequantize_linear` (`axis`, `block_size`), both DPS ops with an optional `zero_point` operand, auto-reify/auto-infer shape inference, and `HipDstBufferizableModel` registration.

**Conversion** — `QdqConversion.cpp` fills in the ONNX defaults for attributes the source op omits (`axis=1`, `block_size=0`, `precision=0`, `saturate=1`) and drops `output_dtype`.

**Lowering** — `QdqLowering.cpp` emits `wrap_quantize_linear` (16 arguments) and `wrap_dequantize_linear` (14), materializing the input and scale shape arrays as `i64` stack arrays. Element types are validated against an explicit whitelist so an unsupported type fails legalization instead of being silently remapped.

**Runtime** — new `real/qdq.cpp` wrappers plus `mock_gpu.cpp` stubs. Shape, axis, and block-coverage validation happens host-side in `prepare_qdq`, before any launch.

**Kernels** — new `qdq_kernel.hip`. Granularity is resolved once on the host, so the per-tensor path costs no integer division on device; the blocked path folds the same coordinate/`block_size` index as `hip_gather_block_quantized`. Numerics: quantize follows the spec order (round the quotient to nearest-even, add the zero point, then saturate) and clamps in fp32 before narrowing, so an out-of-range or NaN quotient cannot reach an undefined float-to-int conversion; the division is kept as a division rather than a multiply by a reciprocal, which would add a second rounding and shift tie cases by one LSB. Dequantize keeps the zero-point subtraction in `int32`. A null `zero_point` pointer is treated as zero.

**Drive-by fix** — `getHipdnnDataType` gained `UINT16`, and the unsigned check had to be ordered *before* the signless one: `isInteger(16)` also matches `ui16`, so the obvious ordering silently maps every `ui16` tensor to `INT16`. Any future unsigned type needs the same care.

## Compiler IR

<details>
<summary>Before / after (per-axis Q/DQ pair)</summary>

```mlir
// Before
%q = "onnx.QuantizeLinear"(%x, %scale, %zp) {axis = 1 : si64, block_size = 0 : si64, saturate = 1 : si64}
    : (tensor<1x64x32xf32>, tensor<64xf32>, tensor<64xi8>) -> tensor<1x64x32xi8>
%y = "onnx.DequantizeLinear"(%q, %scale, %zp) {axis = 1 : si64, block_size = 0 : si64}
    : (tensor<1x64x32xi8>, tensor<64xf32>, tensor<64xi8>) -> tensor<1x64x32xf32>
```

```mlir
// After
%qinit = tensor.empty() : tensor<1x64x32xi8>
%q = hip.quantize_linear(%ctx) ins(%x, %scale : tensor<1x64x32xf32>, tensor<64xf32>)
     zero_point(%zp : tensor<64xi8>) outs(%qinit : tensor<1x64x32xi8>)
     {axis = 1 : i64, block_size = 0 : i64, precision = 0 : i64, saturate = 1 : i64} : tensor<1x64x32xi8>
%dqinit = tensor.empty() : tensor<1x64x32xf32>
%y = hip.dequantize_linear(%ctx) ins(%q, %scale : tensor<1x64x32xi8>, tensor<64xf32>)
     zero_point(%zp : tensor<64xi8>) outs(%dqinit : tensor<1x64x32xf32>)
     {axis = 1 : i64, block_size = 0 : i64} : tensor<1x64x32xf32>
```

</details>

## Test plan

- [x] `pre-commit run --all-files`
- [x] model with q/dq operator functional testing

## Notes for reviewers

- **`precision` is validated, not implemented.** The kernel computes in fp32 and accepts only TensorProto `0` (unset) and `1` (FLOAT), rejecting anything else rather than silently ignoring it. Note the TensorProto numbering differs from `HIPDNN_EP_DATATYPE`, which is an easy thing to conflate when extending this. An fp16 intermediate is structurally unusable for 16-bit storage anyway, since fp16 cannot exactly represent integers above 2048.
- The conversion LIT test deliberately uses `precision = 16`, a value the runtime rejects. Conversion does not validate attributes, so this is only proving propagation; validation lives at the runtime layer.
- Blocked quantization permits a partial trailing block but requires every non-axis scale dimension to match the input exactly.

## AI assistance

Substantial AI assistance (Cursor, Claude Opus 5) was used for the kernel implementation, the wrapper layer, and the LIT tests. I reviewed the result, own the design decisions above, and validation status is as stated in the test plan.


amd-genmingz:
need #886

amd-mingw (approved):

PR #891, opened 2026-09-01
feat(onnx): add IsInf end-to-end

pragyam32:
Wire onnx.IsInf through hip.isinf, unary elementwise HIP-to-LLVM lowering, mock/real runtime, and the unary HIP kernel.

### DIFF
--- a/lib/Runtime/Kernels/hip/elementwise_unary_kernel.hip
+++ b/lib/Runtime/Kernels/hip/elementwise_unary_kernel.hip
@@ -674,6 +674,12 @@ extern "C" int hip_elementwise_isinf(void *stream, const void *input,
                                      int detect_positive) {
   if (num_elements <= 0)
     return 0;
+  if (!detect_negative && !detect_positive) {
+    fprintf(stderr,
+            "[custom_kernels] hip_elementwise_isinf: at least one of "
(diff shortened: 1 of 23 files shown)


lib/Runtime/Kernels/hip/elementwise_unary_kernel.hip
```diff
@@ -609,6 +609,126 @@ extern "C" int hip_elementwise_floor(void *stream, const void *input,
...
+  int64_t idx =
+      static_cast<int64_t>(blockIdx.x) * blockDim.x + threadIdx.x;
+  if (idx < num_elements) {
+    output[idx] = isinf_to_bool(input[idx], detect_negative, detect_positive);
+  }
+}
+
+extern "C" int hip_elementwise_isinf(void *stream, const void *input,
+                                     void *output, int64_t num_elements,
+                                     int hip_dtype, int detect_negative,
```
asparkhi, line 673: Shall we add a check that one of the flags is set?

asparkhi (changes requested):

lib/Runtime/Kernels/hip/elementwise_unary_kernel.hip
pragyam32: Added the check for flags

PR #893, opened 2026-09-01
fix(onnx): fold ConstantOfShape to scalar fill for Min consumers

pragyam32:
When ConstantOfShape feeds onnx.Min with a dynamic result type, avoid tensor.splat bufferization that emits host memref.store loops into GPU pool memory. Reuse the rank-0 linalg.fill path already used for Where.


asparkhi (approved):
Thanks Pragya! LGTM.

PR #847, opened 2026-08-26
Feat/pack high rank elementwise

nvunnam57128:

PR #899, opened 2026-09-02
fix(onnx-to-hip): pack high-rank onnx.Clip and onnx.Relu to 4-D

nvunnam57128:
## Summary

`PackBroadcastTo4D` did not cover `onnx.Clip` or `onnx.Relu`, so a high-rank Clip reached `HipToLLVM` at full rank and failed to compile. `bevformer-tiny-temporal-sim` now compiles and runs end-to-end; before this change it aborted with `failed to legalize operation 'hip.max' that was explicitly marked illegal`.

## Related issue or design

Follow-up to #847, which introduced the pass for the binary elementwise ops. None otherwise.

## Why

`ElementwiseLowering` rejects rank > 4 because of the 4-D shape-passing ABI, and `PackBroadcastTo4D` exists to keep high-rank elementwise ops under that ceiling. `onnx.Clip` and `onnx.Relu` are not broadcasts, so they were not in the pass's op list, but both expand into `hip.max`/`hip.min` during conversion and therefore hit the same ABI limit. In `bevformer-tiny-temporal-sim`, `Clip_1113` operates on `tensor<4x1x6x2500x1xf32>`; it is the only rank-5 Clip out of 93 in the model.

Adding `"onnx.Clip"` to the existing op list was not sufficient. Clip is ternary, and `PackBroadcastPattern` opened with `op->getNumOperands() != 2`, so Clip would have been registered and then rejected on every match.

## What

- Generalize `PackBroadcastPattern` with a `numDataOperands` count. The leading operands carry the broadcast and are packed; trailing operands are forwarded unchanged.
- Bound the total operand count with `maxOperands`. It exceeds `numDataOperands` only for `onnx.Clip` (1..3); the nine binary ops match at exactly 2 and `onnx.Relu` at exactly 1, which keeps variadic `onnx.Max`/`onnx.Min` out of the pattern.
- Whitelist the pass-through operand types: an operand must be `none` or a rank-0 tensor. A blacklist would let an unranked bound through, since it has no rank to check.
- Skip the rewrite when an op's optional operands are all absent. A Clip with neither bound is an identity, and packing it did no useful work.
- Change `packBroadcast` to take an `ArrayRef<Value>` and run the per-axis broadcast check and the group-safety check across all operands instead of a fixed `lhs`/`rhs` pair.

`onnx.Relu` is included because `ReluConversion` also expands to `hip.max`, so a rank-5 Relu fails identically. bevformer's 96 Relus are all rank ≤ 4, so it is not triggering today, but it is the same exposure and the generalized pattern covers it.

Packing is exact for these two ops: both are elementwise and shape-preserving, `expand_shape` reuses the `collapse_shape` reassociation, and max/min do not accumulate, so no floating-point reassociation is introduced. Unlike the binary ops, the group-safety analysis can never reject them — with one shaped operand whose shape equals the result, every candidate group trivially satisfies `product == outProduct`.

## Test plan

- Full lit suite: 469 passed, 12 unsupported, 0 failures. The refactor touches the shared path for all nine binary ops, so the existing `test_max.mlir` / `test_add.mlir` coverage passing is the signal that the binary behavior is unchanged.
- `test_clip.mlir`: rank-5 Clip with scalar bounds (bevformer's shape), rank-5 Relu, the two-operand spelling with the upper bound omitted, the upper-bound-only spelling, the rank-1 bound left unpacked, and the bounds-less identity left unpacked.
- `test_max.mlir`: rank-5 variadic Max left unpacked, checking both chained `hip.max` ops so a dropped third input fails the test.
- `test_clip_invalid.mlir`: `--verify-diagnostics` for an unranked bound.
- `hip-onnx-runner -m bevformer-tiny-temporal-sim/model.onnx`: session created in ~8 s, inference completed, 3 output tensors, exit 0. Previously failed to compile.
- Numerics vs CPU reference on bevformer with identical seeded inputs: relative L2 of 3.13e-03, 6.61e-04, 6.71e-04 across the three outputs, consistent with ordinary fp32 GPU drift across the attention and matmul stack.
- Corner-case matrix run by hand over 15 shapes and operand spellings — rank 5 through rank 8, zero-sized dimension, all-ones shape, `i1` result via `onnx.Less`, f16, rank-mismatched broadcast, dynamic shape, rank-4 no-op, interleaved broadcast, variadic Max. Every case matched the intended pack/skip decision and the module verified. Rank-8, zero-dim and rank-5 Relu were additionally taken through `--hipdnn-pipeline` to confirm the packed form legalizes.
- `git clang-format --diff HEAD` and a full-file `clang-format` run: clean.

## Compiler IR

<details>
<summary>Before / after, rank-5 Clip</summary>

```mlir
// Before: stays rank-5, rejected by ElementwiseLowering
%1 = hip.max(%ctx) ins(%arg1, %arg2 : tensor<4x1x6x2500x1xf32>, tensor<f32>)
                   outs(%0 : tensor<4x1x6x2500x1xf32>) : tensor<4x1x6x2500x1xf32>
```

```mlir
// After
%collapsed = tensor.collapse_shape %arg1 [[0], [1], [2], [3, 4]]
    : tensor<4x1x6x2500x1xf32> into tensor<4x1x6x2500xf32>
%1 = hip.max(%ctx) ins(%collapsed, %arg2 : tensor<4x1x6x2500xf32>, tensor<f32>)
                   outs(%0 : tensor<4x1x6x2500xf32>) : tensor<4x1x6x2500xf32>
%expanded = tensor.expand_shape %1 [[0], [1], [2], [3, 4]]
    output_shape [4, 1, 6, 2500, 1] : tensor<4x1x6x2500xf32> into tensor<4x1x6x2500x1xf32>
```

</details>

## Notes for reviewers

Three known limitations, all pinned by tests rather than left implicit:

1. A bound that is a rank-1 single-element tensor rather than a rank-0 scalar is not packed, so a rank-5 Clip in that spelling still fails. Such a bound is semantically a scalar and would broadcast correctly against the packed result, so this could be loosened to accept all-ones shapes. Left strict because the ONNX spec specifies scalar bounds and no model exercises the other form. Pinned by `clip_5d_rank1_bound`.

2. Binary packing remains best-effort. An interleaved broadcast such as `tensor<2x3x4x5x6xf32>` against `tensor<1x3x1x5x1xf32>` has no safe 4-group split, since every adjacent pair mixes a broadcast axis with a present one, so the op reaches the ABI limit and fails. Unchanged from #847; bevformer contains no such case.

3. Variadic `onnx.Max`/`onnx.Min` with three or more inputs are still left unpacked, as the file header states. Pinned by `max_5d_variadic`.

The identity-Clip skip is worth calling out as a non-change in emitted IR: the conversion already folded the collapse/expand pair back out, so what the guard removes is a pointless rewrite and a misleading `NumBroadcastsPacked` increment, not a visible defect. Likewise the unranked-bound whitelist — `hip.max` rejects an unranked operand either way, so that was an invalid rewrite on the path to the same diagnostic rather than a silent miscompile.

## Checklist

## AI assistance

Diagnosis, the pattern generalization, the lit tests, the corner-case audit, and this description were prepared with Cursor (Claude Opus 5). Validated by the full lit suite, the corner-case matrix, an end-to-end bevformer compile and run, and the numeric comparison against the CPU reference recorded in the test plan.

### DIFF
--- a/lib/Conversion/OnnxToHip/PackBroadcastTo4D.cpp
+++ b/lib/Conversion/OnnxToHip/PackBroadcastTo4D.cpp
@@ -261,11 +261,15 @@ struct PackBroadcastPattern : public RewritePattern {
       return failure();
 
     // A pass-through operand that carried shape would silently disagree with
-    // the packed result, so only scalars and `none` may ride along.
+    // the packed result, so only scalars and `none` may ride along. This is a
+    // whitelist on purpose: an unranked operand carries no rank to check, so
+    // rejecting merely the ranked-and-non-scalar case would let it through.
     OperandRange trailing = op->getOperands().drop_front(numDataOperands);
     for (Value extra : trailing) {
+      if (isa<NoneType>(extra.getType()))
+        continue;
       auto extraType = dyn_cast<RankedTensorType>(extra.getType());
-      if (extraType && extraType.getRank() != 0)
+      if (!extraType || extraType.getRank() != 0)
         return failure();
     }
 
--- a/lib/Conversion/OnnxToHip/PackBroadcastTo4D.cpp
+++ b/lib/Conversion/OnnxToHip/PackBroadcastTo4D.cpp
@@ -16,7 +16,8 @@
 // but both expand into hip.max/hip.min during conversion and so hit the same
 // rank-4 ceiling in the elementwise ABI; their broadcast lives entirely on the
 // first operand, and Clip's optional min/max bounds are scalars that are
-// forwarded unchanged.
+// forwarded unchanged. A Clip supplying neither bound is an identity and is
+// left alone.
 //
 // Before:
 //   %r = "onnx.Add"(%a, %b) : (...) -> tensor<6x2500x8x1x2x4x2xf32>
@@ -265,19 +266,20 @@ struct PackBroadcastPattern : public RewritePattern {
     // whitelist on purpose: an unranked operand carries no rank to check, so
     // rejecting merely the ranked-and-non-scalar case would let it through.
     OperandRange trailing = op->getOperands().drop_front(numDataOperands);
+    bool anyTrailingPresent = false;
     for (Value extra : trailing) {
       if (isa<NoneType>(extra.getType()))
         continue;
       auto extraType = dyn_cast<RankedTensorType>(extra.getType());
       if (!extraType || extraType.getRank() != 0)
         return failure();
+      anyTrailingPresent = true;
     }
 
     // An op whose optional operands are all absent is an identity: onnx.Clip
     // with neither bound does nothing, so packing it would churn reshapes and
     // inflate the statistic for no gain.
-    auto isAbsent = [](Value v) { return isa<NoneType>(v.getType()); };
-    if (maxOperands > numDataOperands && llvm::all_of(trailing, isAbsent))
+    if (maxOperands > numDataOperands && !anyTrailingPresent)
       return failure();
 
     SmallVector<Value> dataOperands(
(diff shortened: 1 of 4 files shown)


asparkhi (approved):
Thanks Naveen - LGTM.

PR #900, opened 2026-09-02
feat(hipsr): retype shape values so the upstream shape passes can lower them

zpye:
## Summary

Adds `hipsr-convert-shape-to-extent`. The pass retypes `!shape.shape` to
`tensor<Nxindex>` and `!shape.size` to `index`, so the upstream shape passes can
lower the shape math that `hipsr-materialize-init-tensors` emits.

## Why

`--convert-shape-to-std` only matches extent tensors and `index`, so it changed
nothing on our shape-typed IR. Retyping turns it into a real lowering, and it
lets One-Shot Bufferize run on the shape path.

## What

- The pass inlines the `scf.execute_region` and `shape.assuming` regions first,
  so the conversion sees straight-line code.
- It retypes the ops upstream can lower, and lowers the four it cannot:
  `shape.from_extents`, `shape.concat`, `shape.size_to_index`, and
  `shape.const_size`.
- It narrows the extent tensor when the operands give the rank.
  `shape.const_shape` and `shape.from_extents` need this, because MLIR type
  inference and `tensor.from_elements` fix those result types.
- `hipsr.preserve_shape` now accepts all three forms of its shape operand:
  `!shape.shape`, an extent tensor, then an extent memref after bufferization.
  A verifier compares the extent count against the data rank.
- The pipeline runs `-remove-shape-constraints` first, because inlining a
  `shape.assuming` needs the passing witness that pass creates.

```mlir
// Before
%s = scf.execute_region -> !shape.shape {
  %0 = shape.shape_of %a : tensor<?x4xf16> -> !shape.shape
  scf.yield %0 : !shape.shape
}
%c0 = shape.const_size 0
%d = shape.get_extent %s, %c0 : !shape.shape, !shape.size -> !shape.size

// After
%s = shape.shape_of %a : tensor<?x4xf16> -> tensor<2xindex>
%d = shape.get_extent %s, %c0 : tensor<2xindex>, index -> index
```

## Test plan

- LIT suite: 473 passed. New `ConvertShapeToExtent/convert-shape-to-extent.mlir`
  and `invalid.mlir`, plus regenerated pipeline, bufferize, and
  `preserve_shape` expectations.
- Checked by hand that no shape op survives `--convert-shape-to-std` on
  `sample_static.mlir` and `sample_dynamic.mlir`, and that One-Shot Bufferize
  needs no `allow-unknown-ops` for the shape path.
- `pre-commit run --all-files` is clean.

## Notes for reviewers

- `docs/pipeline_pass_menu.md` still needs the two new pipeline entries. I left
  it out of this push; happy to add it before this leaves draft.
- Nothing in CI covers the chain past this pass, so a later change that
  reintroduces a shape op upstream cannot lower would not be caught here.
- `hipsr.constant`, `hipsr.matmul`, and `hipsr.expand` still fail full
  bufferization. That is unchanged by this PR.

## AI assistance

AI-assisted (Claude Opus 5 through Cursor): the pass, the ODS and bufferization
changes, the LIT tests, and this description. I validated the result with the
LIT suite and `pre-commit`, reviewed every line, and own the design decisions.

### DIFF
--- a/lib/Dialect/Hipsr/Transforms/ConvertShapeToExtentPass.cpp
+++ b/lib/Dialect/Hipsr/Transforms/ConvertShapeToExtentPass.cpp
@@ -1,0 +1,276 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+//===- ConvertShapeToExtentPass.cpp - Shape types to extent tensors -------===//
+//
+// Retypes shape values so the upstream shape passes can lower them.
+// !shape.shape becomes tensor<Nxindex>, and !shape.size becomes index.
+//
+// --convert-shape-to-std only matches those new types, so it does nothing
+// until this pass runs. This pass lowers four ops itself, because upstream has
+// no pattern for them: shape.from_extents, shape.concat, shape.size_to_index,
+// and shape.const_size.
+//
+//===----------------------------------------------------------------------===//
+
+#include "hip/Dialect/Hipsr/Transforms/Passes.h"
+
+#include "hip/Dialect/Hipsr/IR/HipsrDialect.h"
+#include "hip/Dialect/Hipsr/IR/HipsrOps.h"
+
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/Shape/IR/Shape.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/SmallVectorExtras.h"
+#include "llvm/ADT/TypeSwitch.h"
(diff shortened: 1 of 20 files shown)


lib/Dialect/Hipsr/Transforms/ConvertShapeToExtentPass.cpp
```diff
@@ -0,0 +1,276 @@
...
+#define GEN_PASS_DEF_CONVERTSHAPETOEXTENTPASS
+#include "hip/Dialect/Hipsr/Transforms/Passes.h.inc"
+
+namespace {
+
+//===----------------------------------------------------------------------===//
+// Rank inference
+//===----------------------------------------------------------------------===//
+
+// Returns how many extents each shape result of `op` holds, or kDynamic when
```
wcy123, line 50: // Returns the extent count for each shape result of `op`, or kDynamic when
  // the rank cannot be inferred from the already-converted `operands`.
  //
  // Unlisted operations get tensor<?xindex>. Example: shape.split_at lowers to
  // tensor.extract_slice, which always returns tensor<?xindex> even when the
  // rank is statically known.

wcy123 (commented):
Nit: comment clarity at https://github.com/ROCm/hip-ep/blob/5330b5ba7ffb61a2ece1ac45a6909c6f356fcdd1/lib/Dialect/Hipsr/Transforms/ConvertShapeToExtentPass.cpp#L50-L56

Suggested rewording:
```cpp
// Returns the extent count for each shape result of `op`, or kDynamic when
// the rank cannot be inferred from the already-converted `operands`.
//
// Unlisted operations get tensor<?xindex>. Example: shape.split_at lowers to
// tensor.extract_slice, which always returns tensor<?xindex> even when the
// rank is statically known.
```

wcy123 (approved):

PR #907, opened 2026-09-03
feat(qdq): fuse DQ/DQ->Add->Q into hip.qadd with PDLL patterns

amd-genmingz:
## Summary

This PR is developed based on PR #767.

PR #767 introduced the PDLL mechanism, which allows us to compile .pdl files into fusion patterns and rewrite rules.

This PR leverages that mechanism to implement the transformation from onnx.DequantizeLinear + onnx.Add + onnx.QuantizeLinear to hip.qadd, and also adds lowering support for hip.qadd.

## Related issue or design

Stacked on #767 (open draft, branch `feat/pdll-qdq-matmul-fusion`, tip `e523460d`).
That commit is an ancestor of this branch, so the diff against `main` includes #767's
commits. See Notes for reviewers for how to sequence the two.

## Why

**Runtime.** The QDQ chain otherwise executes as three ops with two full f32 round
trips through GPU memory. The fused op reads and writes only the quantized width and
never materializes the f32 intermediates.

**Maintainability.** Every other fusion in the tree is a hand-written C++
`RewritePattern`. #767 argued PDLL keeps the match declarative and colocated with the
pattern it describes; this PR is the evidence, since the entire Add fusion is 38 lines
of PDLL and three shared native constraints, and the pass itself does not grow.

### Carried over from #767

The `mlir-pdll` discovery and build step, the PDL dialect registration in `Passes.td`
and `hip-mlir-opt`, the module-relative artifact lookup in `ConvertOnnxToHipPass`
(`dll_path()`, needed because the pass lives in a static library that links into
either the tool or the EP shared library), and the `GetContextArg` /
`ExtractScaleValue` native constraints.

### Extended from #767

- **Aggregate pattern entry point.** `HipFusionPatterns.pdll` `#include`s each pattern
  file, so `mlir-pdll` runs once and emits one PDL module. That means one
  `RewritePatternSet` and one fixpoint, so patterns can trigger each other. Each
  pattern file now carries only its `Pattern` definitions; the native constraints are
  declared once in the aggregate, since PDLL has no include guards.
- **Installable artifact.** #767 passed the build-tree path in as a compile-time
  define, which only resolves from the build directory. The compiled module is now
  installed to `bin/` and located next to the loaded binary, so it works from an
  install tree and inside the shipped EP. A missing artifact is a no-op, and a build
  without `mlir-pdll` warns and disables fusion.
- **`ExtractZeropointValue`.** #767 ignored zero points entirely. The new constraint
  extracts them, with the signedness subtlety below.
- **Zero-point signedness.** The importer gives inline constants signless storage, so
  reading signedness off the zero-point may get wrong number. 
  `ExtractZeropointValue` takes signedness from the Q/DQ op's
  quantized tensor, the only operand that still carries the ONNX type.
- **Dependency tracking.** Every `.pdll` is globbed with `CONFIGURE_DEPENDS`, so
  editing an included pattern rebuilds the module.

### New in this PR

- **Op.** `hip.qadd` builds on `Hip_DpsOp_Broadcast`, so multidirectional broadcast
  shape inference comes for free, and registers the bufferization DPS model.
- **Pattern.** `QDQAddFusion.pdll` matches the DQ/DQ -> Add -> Q chain and rewrites it
  to `hip.qadd` with all three scales and zero points attached.
- **Lowering.** `QAddLowering.cpp` marshals per-operand shape arrays and the
  quantization parameters into `wrap_qelementwise`. It is templated on a `kind`
  selector so `qsub`/`qmul`/`qdiv` can reuse it.
- **Runtime and kernel.** `hip_qelementwise` computes
  `saturate(round(((A - z_a) * s_a + (B - z_b) * s_b) / s_out) + z_out)`
  with per-operand broadcast strides.
- The QMatMul changes have been temporarily removed because they are not 
   yet complete and would introduce errors. A new PR will continue the work 
   to support QMatMul.

## Test plan

- [x] Lit test for test conversion
- [x] functional testing:Functional testing was performed using a quantized official model to verify that qadd can be successfully fused.
- [x] Accuracy testing: The accuracy of qadd was evaluated using test cases covering int8, uint8, int16, and uint16, with sufficiently large computational workloads.

The computed results achieved a rel_l2norm < 5e-6. The observed error is caused by the introduction of M_a and M_b during the mathematical derivation, which introduces a small amount of precision loss.

## Notes for reviewers

- **Sequencing with #767.** #767 is still an open draft and has not merged, so this
  diff contains its commits. This PR also removes everything #767 uniquely added —
  `hip.qmatmul`, `QDQMatMulFusion.pdll`, and `test_qmatmul.mlir` — because a
  quantized MatMul with no lowering or runtime is dead weight in the dialect, and
  because that fusion warrants its own PR. Landing this and closing #767 as superseded
  is the cleanest option; happy to instead rebase onto a merged #767 if maintainers
  prefer to keep its history.
- **Reading the diff.** The PDLL foundation is commits `d1abd797..e523460d` (= #767);
  everything after is the QADD work.
- **New build-time dependency.** `mlir-pdll`. A build without it warns and skips
  fusion rather than failing, so a misconfigured build loses the fusion silently.
  Inherited from #767's design.
- **Limits.** Per-tensor scales only: the constraint requires a splat, so per-channel
  QDQ does not match and falls through to the unfused path. Both operands must share
  one quantized type; mixed precision is not expressible in the current runtime
  signature.
- **Size.** 23 files and roughly 1,030 changed lines, so this will pick up `large-pr`;
  about 200 of those lines are #767's plumbing, reshaped. The remainder does not split
  usefully — the PDLL plumbing has no pattern to run, and the op has no producer until
  the pattern exists — so splitting would land dead code. Happy to peel the runtime
  kernel out if reviewers prefer.

## AI assistance


wcy123 (approved):
LGTM

PR #908, opened 2026-09-03
feat(hipsr): bufferize the hipsr pipeline with One-Shot Bufferize

zpye:
## Summary

The hipsr pipeline now ends in memrefs instead of tensors. `buildHipsrPipeline`
continues past `hipsr-convert-shape-to-extent` with four passes:

- `shape-to-shape-lowering` rewrites `shape.num_elements` as a `shape.reduce`,
  which `convert-shape-to-std` cannot lower on its own.
- `convert-shape-to-std` clears the remaining shape ops.
- `one-shot-bufferize` converts tensors to memrefs.
- `convert-linalg-to-loops` lowers the `linalg.map` that `shape.broadcast`
  leaves behind.

## Why

#900 retyped the shape math so the upstream shape passes could lower it. That
was the last thing blocking One-Shot Bufferize on this pipeline, and the rest of
the lowering needs buffers.

## What

`one-shot-bufferize` runs with `bufferize-function-boundaries`, an identity
layout map, and `use-encoding-for-memory-space`, so
`tensor<4xf16, #hipsr.mem<device>>` becomes `memref<4xf16, #hipsr.mem<device>>`.

New bufferization models:

- `hipsr.constant` points into the constants blob, which the runtime owns, so it
  bufferizes like upstream `arith.constant`: an identity-layout memref that
  nobody may write, with the blob attributes unchanged.
- `matmul`, `expand`, `add`, and `nonzero` get the shared destination-passing
  model.

`hipsr.compute` now reuses an input buffer for an output only when the body
never reads the matching destination argument, because that reuse also changes
the argument's type.

New per-op tests live under
`test/lit/Dialect/Hipsr/Transforms/OneShotBufferize/`. Regenerating the three
pipeline tests against the bufferized output is most of the diff.

Written with AI assistance. I reviewed the result and validated it with the LIT
suite.

### DIFF
--- a/include/hip/InitAllPasses.h
+++ b/include/hip/InitAllPasses.h
@@ -192,7 +192,6 @@ inline void registerAllPasses() {
     mlir::registerReconcileUnrealizedCastsPass();
     mlir::memref::registerResolveShapedTypeResultDimsPass();
     mlir::registerConvertLinalgToLoopsPass();
-    // Upstream passes that lower what hipsr-convert-shape-to-extent retypes.
     mlir::registerRemoveShapeConstraintsPass();
     mlir::registerShapeToShapeLoweringPass();
     mlir::registerConvertShapeToStandardPass();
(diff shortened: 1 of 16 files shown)


wcy123 (commented):
Single-statement if with early return needs curly braces per LLVM standard:

https://github.com/ROCm/hip-ep/blob/54998fcbc82f1764bad96c97d640af8d4b2b4a16/lib/Dialect/Hipsr/Transforms/BufferizableOpInterfaceImpl.cpp#L278-L282

wcy123 (approved):

PR #914, opened 2026-09-07
perf(kernels): halve softmax traffic and vectorise the 4-D broadcast add

BoarQing:
## Summary

Two independent memory-traffic fixes to the custom kernels that dominate
windowed-attention vision models. Both were found by measuring achieved
bandwidth against the compulsory traffic each operation must move, and both
close most of the remaining gap.

| operation | before | after | change |
|---|---:|---:|---|
| row-wise softmax | 254 ms | 56 ms | stop copying the input; wave-per-row kernel |
| 4-D broadcast add (attention bias) | 345 ms | 186 ms | vectorised fast path |

Measured on swinv2-base at 2048x3072, fp16, gfx1151, summed over all
invocations in one warm inference.

## What

**Softmax — remove the copy.** `hip_miopen_softmax` copied the entire input
tensor over the output with `hipMemcpyAsync`, then ran an in-place kernel over
the result. The copy moved exactly as many bytes as the kernel itself, so the
operation paid twice its compulsory traffic; on the dominant shape it moved
402 MB and cost about as much as the softmax.

All six row-wise kernels now take separate source and destination pointers.
Each either holds its row in registers or reads strictly ahead of what it
writes, so accepting a distinct destination is a signature change rather than a
restructure, and passing one buffer as both still behaves as before. The
previous in-place entry points remain as wrappers that do exactly that, so
external callers are unaffected. `softmax_row_2d_inplace_kernel` becomes
`softmax_row_2d_general_kernel`, since it now serves both cases.

This commit also carries the wave-per-row path, which keeps a row in registers
so it is read once and written once instead of read three times and written
twice, and a fix for a race in the multi-wave LDS reduction.

**Elementwise — widen the access.** The general 4-D broadcast kernel assigns
one element per thread, so it issues two-byte accesses and recomputes the index
with four runtime integer divisions and four modulos per element. On the
bias-add shapes the narrow access, not the arithmetic, dominated: removing the
divisions alone left the kernel well short of copy speed, while widening the
access closed the gap. Shapes whose broadcast pattern folds into a contiguous
inner run now take a vectorised path where each thread moves a wide aligned
chunk; the guard and the term folding are computed host-side, and shapes that
do not qualify use the general kernel unchanged.

## Test plan

- **Numeric (GPU vs ORT CPU), gfx1151:** 62/62 pass — 24 softmax cases covering
  the wave path, the strided fallback and chained-softmax write coverage, and
  38 elementwise cases covering both the fast path and the general kernel.
  These were run against this exact build; they are not part of this PR (see
  below).
- **fp32 softmax end to end:** L2 difference 1.3e-05 against ORT CPU over 23M
  elements, all values finite.
- **Compiler suites:** `MorphizenMLIRLitTests`, `OutputAllocatorUnitTest` and
  `StaticPlugins` pass.
- **CI on the previous head of this branch** (identical kernel sources, plus
  the tests): Linux build, mock build, real build, GPU test and pre-commit all
  green.
- **Capture evidence:** an RGP capture re-fenced on the dominant softmax shape
  now opens on the softmax kernel at dispatch 0 with no preceding gap and
  contains no `__amd_rocclr_copyBuffer`, where the previous capture contained
  exactly one immediately before the kernel.

## Notes for reviewers

- **No tests are included in this PR, by request.** The numeric cases and the
  standalone A/B microbenchmarks that back these two changes exist and pass,
  and the results above come from running them, but they are held back for a
  separate change. Note that `test/numeric/` is not currently executed by any
  CI workflow, so this does not change what CI verifies — but it does mean the
  new softmax kernels and the new elementwise fast path land without committed
  coverage a reviewer can run. I would recommend landing that follow-up
  alongside or shortly after this PR.
- **The two commits are independent** and touch disjoint kernels. They are
  presented together because they were measured together on the same model; if
  reviewing them separately is preferable, the PR splits cleanly at the commit
  boundary.
- **Aliasing.** The kernels tolerate source and destination being the same
  buffer, but the EP's pool planner assigns distinct offsets to overlapping
  lifetimes, so a graph-generated softmax cannot actually alias. The tolerance
  exists for the in-place wrappers and is exercised by the microbenchmark
  rather than by an ONNX-level test.

## AI assistance disclosure

Substantial AI assistance, disclosed per `CONTRIBUTING.md`. An AI tool (Claude
Opus 5 via Cursor) implemented these changes and wrote this description. The
work was validated by execution rather than review alone: the numeric results,
the build results and the capture evidence quoted above are from actual runs on
gfx1151 hardware, and the per-operation timings come from back-to-back A/B runs
with only the change under test differing between builds. The contributor
remains the author and is accountable for the code and the claims.

### DIFF
--- a/lib/Runtime/Kernels/hip/elementwise_binary_kernel.hip
+++ b/lib/Runtime/Kernels/hip/elementwise_binary_kernel.hip
@@ -879,6 +879,19 @@ __device__ __forceinline__ void ew_store(__half *p, uint32_t i, float v) {
   p[i] = __float2half(v);
 }
 
+__device__ __forceinline__ float ew_to_float(float v) { return v; }
+__device__ __forceinline__ float ew_to_float(__half v) {
+  return __half2float(v);
+}
+
+template <typename T> __device__ __forceinline__ T ew_cast(float v);
+template <> __device__ __forceinline__ float ew_cast<float>(float v) {
+  return v;
+}
+template <> __device__ __forceinline__ __half ew_cast<__half>(float v) {
+  return __float2half(v);
+}
+
 // op: 0=add, 1=mul, 2=min, 3=max
 __device__ __forceinline__ float ew_apply(int op, float a, float b) {
   switch (op) {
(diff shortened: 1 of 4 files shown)


amd-yanjunz:

- PR: ROCm/hip-ep#914
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/914#issuecomment-5580367639)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/795/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6a9fc4ed29d3ad702c0f7533
- Compare with latest main branch results: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fids%3D6a9efb7a7174cc4cb7c7ba82%26collection%3DHIP-EP-Dynamic&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=795

BoarQing:
vision baseline: http://xsjvitisaijenkins:8080/job/daily_hip_ep_onnx/327/
vision for PR: http://xsjvitisaijenkins:8080/job/daily_hip_ep_onnx/324/
Procyon: http://xsjvitisaijenkins:8080/job/daily_hip_ep_onnx/328/

image

amd-mingw (approved):

PR #915, opened 2026-09-07
feat(hipsr): run hipsr-use-output-allocator in the pipeline

zpye:
`hipsr-use-output-allocator` was complete but never scheduled, so every graph
output stayed a plain `memref.alloc`. This adds it to the HIPSR pipeline after
`convert-linalg-to-loops`, which places it before the pool allocator (outputs
must stay runtime-owned) and before preserve-shape removal (it reads those ops
to size dynamic outputs).

Scheduling it exposed a bug. By that point the graph is bufferized, so a
preserved shape is an index memref, and the pass read extents with
`shape.get_extent`, which does not accept one. Given this input:

```mlir
#dev = #hipsr.mem<device>

func.func @graph(%ctx: !hipsr.context, %shape: memref<2xindex>, %m: index,
                 %in: memref<?x4xf32, #dev>) -> memref<?x4xf16, #dev> {
  %out = memref.alloc(%m) : memref<?x4xf16, #dev>
  hipsr.cast(%ctx) ins(%in : memref<?x4xf32, #dev>)
                   outs(%out : memref<?x4xf16, #dev>)
  hipsr.preserve_shape %shape, %out : memref<2xindex>, memref<?x4xf16, #dev>
  return %out : memref<?x4xf16, #dev>
}
```

the pass used to fail, which is why `sample_dynamic` and `embedding` stopped
verifying:

```
error: 'shape.get_extent' op operand #0 must be shape or extent tensor,
       but got 'memref<2xindex>'
```

It now reads extents with `memref.load`, and `hipsr.cast` writes straight into
the buffer the allocator returns:

```mlir
func.func @graph(%arg0: !hipsr.context, %arg1: memref<2xindex>, %arg2: index,
                 %arg3: memref<?x4xf32, #dev>) -> memref<?x4xf16, #dev> {
  %0 = memref.load %arg1[%c0] : memref<2xindex>
  %1 = hipsr.alloc_output(%arg0, %0) {out_idx = 0 : i64} : memref<?x4xf16, #dev>
  hipsr.cast(%arg0) ins(%arg3 : memref<?x4xf32, #dev>)
                    outs(%1 : memref<?x4xf16, #dev>)
  hipsr.preserve_shape %arg1, %1 : memref<2xindex>, memref<?x4xf16, #dev>
  return %1 : memref<?x4xf16, #dev>
}
```

Running after bufferization is now the documented contract, and a clear
diagnostic rejects a shape the pass cannot read.

That output also shows the second fix. The pass used to materialize every
dimension of both the external and the internal type, so the static `4` became
an `arith.constant` nothing used, and the internal extents became `memref.load`s
that have a read effect and survive DCE. It now builds only the extents it
uses, and the internal ones only when the output needs a `reinterpret_cast`
view.

The pass's own tests separated cases with `// ---` but never passed
`-split-input-file`, so every `CHECK-LABEL` matched against a single module.
That is fixed, with whole-function checks and a new `invalid.mlir` covering the
shapes the pass rejects.

Validated with the HIPSR LIT suite: 107 tests pass, including the three
regenerated pipeline goldens.

AI assistance: the implementation, test updates, and this description were
prepared with Cursor and Claude, reviewed by me, and validated by the tests
above.

### DIFF
--- a/include/hip/Dialect/Hipsr/Transforms/Passes.td
+++ b/include/hip/Dialect/Hipsr/Transforms/Passes.td
@@ -249,10 +249,14 @@ def HipsrUseOutputAllocatorPass
   let summary =
       "Rewrite returned memref.alloc outputs to hipsr.alloc_output";
   let description = [{
-    Replaces each graph-output `memref.alloc` with `hipsr.alloc_output`, reusing
-    the alloc's dynamic-size operands and setting `out_idx` to its position in
-    `func.return`. Output buffers then come from the EP output allocator and are
-    owned by the EP/runtime instead of being deallocated or pooled.
+    Replaces each graph-output `memref.alloc` with `hipsr.alloc_output`, taking
+    dynamic sizes from the allocation's `hipsr.preserve_shape` and setting
+    `out_idx` to its position in `func.return`. Output buffers then come from
+    the EP output allocator and are owned by the EP/runtime instead of being
+    deallocated or pooled.
+
+    Runs after bufferization, since a shape it reads must be an index memref,
+    and before `hipsr-pool-alloc` and `hipsr-remove-preserve-shape`.
 
     An allocation counts as a graph output when `func.return` returns it either
     directly or through view-like aliases. The view operations are left in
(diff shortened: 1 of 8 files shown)


wcy123 (approved):
LGTM

PR #916, opened 2026-09-07
feat(qdq): add QDQ Mul fusion and hip.qmul quantized elementwise op

amd-genmingz:
## Motivation

Quantized models emit `DequantizeLinear -> Mul -> QuantizeLinear` chains.
Without fusion, each chain runs as three separate ops with float round-trips.
`hip.qadd` already covers the additive case, so Mul only needs the same path
with a different scale folding.

## Details

- Add `Hip_QMulOp` (`hip.qmul`) as a broadcasting DPS op carrying the QDQ
  scales/zero points, with `getDpsInitsMutable`/`getEffects` and a bufferizable
  interface model.
- Add `QDQMulFusion.pdll`, a pure-PDLL pattern that matches the QDQ Mul chain
  and rewrites it into `hip.qmul`, reusing the existing native constraints.
- Rename `QAddLowering.cpp` to `QElementwiseLowering.cpp` and make the shared
  lowering template kind-aware. For Mul the three scales collapse into a single
  coefficient (`M_a = s_a * s_b / s_out`, `M_b` unused), so the existing
  `wrap_qelementwise` ABI is reused unchanged.
- Extend the HIP kernel with `HIP_QELEMENTWISE_MUL`: the rescale now happens
  after the integer differences are combined, which keeps Add numerically
  identical while letting Mul apply one fused rescale.
- Update the runtime header and kernel docs to state the per-kind meaning of
  the coefficient slots.
- If the fused pdl mlir cannot be found, an error will be thrown.

###  Why update qdq_fusion_pass.hpp
`QDQAdd` and `QDQMul` are structurally identical and rooted on the same op
(`onnx.QuantizeLinear`). `-convert-pdl-to-pdl-interp` merges same-root patterns
into a single matcher tree and uniques every predicate the patterns share, so a
shared predicate is evaluated once, in the block where the two branches still
coincide.
That is correct for a yes/no predicate but not for a native constraint that
*returns a value*. `GetContextArg`, `ExtractScaleValue` and
`ExtractZeropointValue` were declared as result-producing `Constraint`s and
called from the match section. After uniquing, their results are defined in the
block owned by the merged node, while each pattern keeps its own
`pdl_interp.record_match` on its own branch. One branch's `record_match` then
operands a value defined in a sibling block that does not dominate it, which is
exactly what the verifier reports — hence the note pointing at "op in the same
region".

## Test Plan

- [x] `test/lit/Conversion/onnx-to-hip/test_qmul.mlir`: QDQ chain fuses into
      `hip.qmul`, rank-1 rhs stays unpadded, and no ONNX/constant ops are left
      behind.
- [x] `test/lit/Conversion/hip-to-llvm/test_qmul.mlir`: `hip.qmul` lowers to
      `wrap_qelementwise` with the folded coefficient.
- [x] Existing `hip.qadd` LIT tests still pass (kernel refactor must not change
      Add numerics).
- [x] Numeric check of quantized Mul against CPU reference on GPU, including a
      broadcast shape.

## AI assistance

### DIFF
--- a/.github/workflows/windows-build-real.yml
+++ b/.github/workflows/windows-build-real.yml
@@ -157,6 +157,9 @@ jobs:
           cp "${{ runner.workspace }}/install/bin/"*.dll staging/bin/ 2>/dev/null || true
           cp "${{ runner.workspace }}/install/bin/"*.exe staging/bin/ 2>/dev/null || true
 
+          # cp fusion pattern file
+          cp "${{ runner.workspace }}/install/bin/HipFusionPatterns.pdl.mlir" staging/bin/
+
           cp "${{ runner.workspace }}/amdgpu-artifacts/bin/"* staging/bin/
 
           # Smoke: every per-arch kernel DLL must be present in staging/bin/
--- a/include/hip/Dialect/IR/HipBufferize.h
+++ b/include/hip/Dialect/IR/HipBufferize.h
@@ -210,6 +210,7 @@ registerHipBufferizableOpInterfaceModels(DialectRegistry &registry) {
     SizeOp::attachInterface<HipDstBufferizableModel<SizeOp>>(*ctx);
     LoopOp::attachInterface<HipDstBufferizableModel<LoopOp>>(*ctx);
     QAddOp::attachInterface<HipDstBufferizableModel<QAddOp>>(*ctx);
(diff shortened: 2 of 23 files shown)


wcy123 (approved):
LGTM.

PR #921, opened 2026-09-08
feat(qdq): fuse DQ+MatMul+Q into a new hip.qmatmul op

amd-genmingz:
## Summary

Adds `hip.qmatmul` and a PDL pattern that collapses
`DequantizeLinear, DequantizeLinear -> MatMul -> QuantizeLinear` into it,
executed by a new integer-accumulating HIP kernel. Covers A8W8 and A16W8 with
per-tensor scales.

## Related issue or design

None. Follows the existing QDQ Add/Mul fusion structure in
`lib/Conversion/OnnxToHip/pdl/`.

## Why

A QDQ MatMul chain runs three kernels and materializes an f32 intermediate the
size of the product. Accumulating in integers removes that buffer and is also
more accurate than the f32 reduction it replaces: the accumulation is exact, and
only the final rescale rounds.

## What

- **Dialect**: `hip.qmatmul` with per-tensor scale/zero-point attributes and
  `transA`/`transB`. Verification and shape reification share
  `reifyMatmulLikeShape` with `hip.matmul` instead of duplicating it.
- **Fusion**: `QDQMatMulFusion.pdll` requires splat scales, readable zero
  points, static result shapes, an 8- or 16-bit A and Y, and an 8-bit B. Adds
  native helpers `IsEightBitQuantized`, `IsEightOrSixteenBitQuantized`,
  `HasStaticShapedResults`, and a generic `ExtractAttrInt64`.
- **Lowering**: folds `M = A_scale * B_scale / Y_scale` at compile time and
  computes `b_batch_stride` — `K*N` for a per-batch weight, 0 for a broadcast
  one — with a runtime path for unknown leading dims.
- **Runtime/kernel**: `wrap_qmatmul` over a tiled kernel (64x64x16, dot4
  fragments) plus a GEMV path for `M == 1`. A 16-bit A is byte-split into hi/lo
  planes so the 8-bit dot instructions still apply; the zero-point corrections
  ride the same dot pipeline and are applied in the epilogue.
- **Portability**: templated `hipdnn_dot4<ASigned, BSigned>` covers all four
  sign combinations — `v_dot4_i32_iu8` on RDNA3/4, scalar fallback for
  mixed signs on CDNA.
- **Build**: the compiled `HipFusionPatterns.pdl.mlir` is staged per config so
  it tracks source edits, and an empty pattern module is now a no-op rather than
  an abort.

## Test plan

- [x] `test/lit/Conversion/onnx-to-hip/test_qmatmul.mlir` — the chain fuses into
      one `hip.qmatmul` carrying the folded attributes.
- [x] `test/lit/Conversion/hip-to-llvm/test_qmatmul.mlir` — argument order and
      sources for the A16W8 + `transB` + rank-3 per-batch case.
- [x] End-to-end accuracy on gfx1151 vs the ORT CPU EP, 12 fixtures covering the
      A8W8/A16W8 sign combinations, both kernel paths, per-batch and broadcast
      weights, non-tile-aligned extents, and omitted zero points. HIP matches an
      exact int64 numpy golden **bit-for-bit in all 12**. Each graph is only the
      fusion pattern and runs under `HIPDNN_EP_STRICT=1` with CPU fallback
      disabled, so a compile failure would fail session creation; every log
      shows exactly one `wrap_qmatmul` and no leftover Q/DQ runtime calls.
- [x] Accumulator bound: K = 32768, 99.2% of the runtime's uint8 x uint8 limit,
      with `|corrected|` peaking at 76% of the int32 range — still exact.
- [x] Verify the functionality using a real model; the logs show that `wrap_qmatmul` 
       is being called.

## Notes for reviewers

- `transA`/`transB` are plumbed through but are always 0 from an ONNX graph
  today: `TransposeMatMulFold` stamps them inside the pre-lowering loop, while
  this pattern runs once at module level before it. They are in place for when
  that ordering changes. Consequently the kernel's transposed paths have LIT
  coverage only, not runtime coverage.
- B must be 8-bit. A16W16 would need a wider accumulator and a second kernel.
- The int32 accumulator bounds K (33025 for uint8 x uint8, 131071 for
  int8 x int8); the runtime rejects a larger K rather than silently overflowing.
- On CDNA a mixed-sign pair — the common uint8 activation x int8 weight — takes
  the scalar fallback in `hipdnn_dot4`. Only RDNA3/4 was measured.
- At 22 files / ~1.5k lines this will trip the `large-pr` label. It is a
  vertical slice (dialect, fusion, lowering, runtime, kernel) committed layer by
  layer; splitting it horizontally would leave an op with no lowering, or a
  kernel nothing reaches.

## AI Assistance

The code was developed collaboratively with Cursor, and all code has been
reviewed.

### DIFF
--- a/include/hip/Dialect/IR/HipBufferize.h
+++ b/include/hip/Dialect/IR/HipBufferize.h
@@ -210,6 +210,7 @@ registerHipBufferizableOpInterfaceModels(DialectRegistry &registry) {
     SizeOp::attachInterface<HipDstBufferizableModel<SizeOp>>(*ctx);
     LoopOp::attachInterface<HipDstBufferizableModel<LoopOp>>(*ctx);
     QAddOp::attachInterface<HipDstBufferizableModel<QAddOp>>(*ctx);
+    QMatMulOp::attachInterface<HipDstBufferizableModel<QMatMulOp>>(*ctx);
     // hip.if is a DPS control-flow op (getDpsInitsMutable, results alias
     // o_init) just like hip.loop. Without this model one-shot-bufferize aborts
     // with "op was not bufferized: hip.if" for any graph containing onnx.If,
--- a/include/hip/Dialect/IR/HipOps.td
+++ b/include/hip/Dialect/IR/HipOps.td
@@ -4914,15 +4914,23 @@ def Hip_QMatMulOp : Hip_DpsOp<"qmatmul", /*traits=*/[],
     M = A_scale * B_scale / Y_scale   (folded at lowering time)
     ```
 
-    Shape contract matches `hip.matmul` without the transpose attributes:
+    Shape contract matches `hip.matmul`:
       A: [..., M, K], B: [..., K, N], Y: [broadcast(A.batch, B.batch), M, N].
+
+    Optional `transA` / `transB` (0/1) apply a last-dimension swap on the
+    corresponding operand before multiply, with the same meaning as on
+    `hip.matmul`: `transA` reads A as [..., K, M] and `transB` reads B as
+    [..., N, K]. They carry the compile-time fold of
+    `Transpose(perm=[..,r,r-2])` into MatMul, so a quantized attention
+    `Q @ K^T` still fuses. When unset they default to 0.
 
     Quantization is per-tensor, so every scale and zero point is an attribute.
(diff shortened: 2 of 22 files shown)


wcy123 (approved):
LGTM.

PR #923, opened 2026-09-08
ci(windows): build OGA from the AMDGPU fork branch to exercise #719 on Gemma-4

BoarQing:
## What

Stacked on top of #719 (`perf/gqa-fuse-bias-softmax`) so CI builds and tests that branch's GQA change against an OGA that can load Gemma-4. This PR adds exactly one commit; the rest of the diff is #719 and should be reviewed there.

The change is the `windows-deps.yml` half of #599, re-applied rather than cherry-picked so this branch stays one reviewable commit:

- OGA is checked out from `amd-genmingz/onnxruntime-genai @ test_0_3` instead of `microsoft/onnxruntime-genai @ v0.14.0`. The fork branch carries the AMDGPU MorphiZen integration that `OGA_PR_PATCHES: "2194"` used to supply **plus sliding_window support, which upstream v0.14.0 does not have**. `OGA_PR_PATCHES` is emptied accordingly.
- The OGA cache key is rev'd `mm2` -> `mm3` and now contains the repo and ref rather than a version string, so the old version-keyed artifact set cannot be silently reused.
- `requests` is pip-installed before `build.py`, whose `tools/python/util` dependency resolver imports it at module scope even though `--ort_home` means nothing is downloaded.

Authored by @amd-genmingz in #599. Only the Windows deps workflow is touched; `linux-build.yml` still pins the upstream `OGA_VERSION`, matching #599.

## Why

#719 is a Gemma-4 optimisation and its numbers were taken locally, so it is worth putting it through the CI GPU suite. `windows-build.yml` chains `prebuild` (this workflow) -> `build_real` -> `gpu_test` on the StrixHalo runner and has no draft gate, so opening this as a draft is enough to run the full chain.

Separately, this is the first chance to see the fork OGA on CI hardware. Gemma-4 currently does not run through genai locally on any commit tested: `vlm_benchmark.py` aborts in the EP's output-allocator callback before reaching a single GQA dispatch, because with `past_present_share_buffer: true` genai binds `present` to the same fixed-capacity buffer as `past` while the EP reports `past_buf + new`:

```
output allocator callback failed for out_idx 1: ... output 'present.0.key'
has shape {1,8,1024,256} but the computed output shape for this run is {1,8,3264,256}
```

That reproduces on `main`, `e2a8f041`, `8b86c388`, `e1735368` and `bf15cac1`, so it is pre-existing and unrelated to #719. Whether the fork's sliding_window support changes that behaviour is the open question.

## What this does and does not prove

**Read this before treating a green run as validation.** There is no Gemma-4 test in CI -- `.github/` contains no reference to the model at all. So a green run here shows:

- the fork OGA branch builds on the Windows CI toolchain, and
- the existing GPU suite still passes with it, and with #719's kernel change underneath.

It does **not** show that Gemma-4 loads or that the `present.0.key` mismatch is resolved, because nothing in the suite exercises that path. Confirming the blocker needs a Gemma-4 case added to the GPU test, which is deliberately not in this PR.

## Notes for reviewers

- This targets `perf/gqa-fuse-bias-softmax`, not `main`, so the diff stays reviewable. It should not be merged before #719; if #719 lands first, this reduces to the single `windows-deps.yml` commit and can be retargeted at `main`.
- The cache-key bump means the first run pays a full OGA build rather than a cache hit, so expect the `prebuild` job to be slow once.
- #599 also mentions an onnxruntime-extensions bump to PR 1091 in its description, which is **not** in its diff and therefore not here either.

### DIFF
--- a/.github/workflows/windows-deps.yml
+++ b/.github/workflows/windows-deps.yml
@@ -30,8 +30,14 @@ env:
   LLVM_TAG: llvmorg-22.1.0
   ONNXRUNTIME_VERSION: "1.27.0"
   ONNXRUNTIME_PR_PATCHES: ""
-  OGA_VERSION: "0.14.0"
-  OGA_PR_PATCHES: "2194"
+  # Fork branch carrying the AMDGPU MorphiZen integration (former PR 2194) plus
+  # sliding_window support, which upstream v0.14.0 does not have.
+  OGA_REPO: "amd-genmingz/onnxruntime-genai"
+  OGA_REF: "test_0_3"
+  # Optional space-separated microsoft/onnxruntime-genai PR numbers applied on
+  # top of the fork checkout via pull/<n>.patch + git am. Empty because the fork
+  # branch already carries the integration that 2194 used to supply.
+  OGA_PR_PATCHES: ""
   ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: "true"
   AMDGPU_EP_REF: "05eaa2a938f71e98b49bd1be3fb03720167be5cc"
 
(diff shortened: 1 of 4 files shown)


amd-yanjunz:

- PR: ROCm/hip-ep#923
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/923#issuecomment-5582861628)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/806/

PR #924, opened 2026-09-08
feat(hipsr): let a compute body write the destination it is given

zpye:
## Summary

A `hipsr.compute` that produced new data built a fresh tensor and yielded that,
leaving the destination it was given unwritten. The result and the init were
then two different buffers, so a reader of the init saw whatever the buffer held
before the body ran. Shape and gather bodies now insert into the destination and
yield what they wrote, so the result and the init name one buffer.

That only holds once the destination is writable. A pool domain and a compute
scope ops the pipeline already placed rather than call them, so an entry
argument names its operand's buffer and is writable exactly when that buffer is.
Reporting entry arguments as read-only made the whole alias set of any buffer
reaching one non-writable, which forced a copy wherever that buffer was written.
A compute input stays read-only when its body writes it, so an input the body
clobbers is still copied, and a function argument is unaffected.

Allocations drop from 18 to 17 on `sample_static`, 19 to 17 on `sample_dynamic`,
and 50 to 42 on `embedding`. Every `memref.copy` left in the goldens
materializes a shape-graph concat into its index buffer; none moves data.

The pipeline goldens now check every output line with a `CHECK-NEXT` chain, so
an added allocation or copy fails the test instead of slipping between two
matches.

## Related issue or design

None.

---

Written with Cursor (Claude Opus 5) and validated with the LIT suite. I reviewed
and understand the result.

### DIFF
--- a/lib/Conversion/OnnxToHipsr/GatherConversion.cpp
+++ b/lib/Conversion/OnnxToHipsr/GatherConversion.cpp
@@ -50,7 +50,7 @@
 //   ^bb0(%body_ctx, %in, %dest):
 //     %p = arith.constant 0 : index
 //     %e = tensor.extract %in[%p]
-//     %out = tensor.from_elements %e
+//     %out = tensor.insert %e into %dest[]
 //     hipsr.compute_yield %out
 //   } : tensor<i64, #hipsr.mem<host>>
 //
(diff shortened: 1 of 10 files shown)


wcy123 (approved):
LGTM

PR #925, opened 2026-09-08
fix(onnx-to-hip): use in-tree mlir-pdll target at configure time, fix multi-config generator dest path

pschlan-amd:
## Summary

find_program cannot see mlir-pdll until the FetchContent LLVM build produces the binary, which disabled PDL fusion on from-source trees. ConvertOnnxToHipPass requires the compiled patterns, so a missing tool is now a configure error.

The same pattern is already used for lib/Runtime, for example.

Also fix an issue where the `HipFusionPatterns.pdl.mlir` file is placed in the wrong path when using a multi-config generator. (Not seen in CI since it uses Ninja.)

## Related issue or design

None

## Test plan

Local build was failing before (since a recent commit caused `python/` to error out during build when mlir-pdll is n/a). With the change, it passes.

## Checklist

Debugged and resolved with assistance of Cursor. The code changes have been reviewed and understood.


pschlan-amd:
Closed in favor of https://github.com/ROCm/hip-ep/pull/935

PR #926, opened 2026-09-08
Convert hip.add to tosa.add

nvunnam57128:
Adds the first conversion pattern to `convert-hip-to-tosa`, which shipped as a stub with an empty `RewritePatternSet`. `hip.add` now lowers 1-1 to `tosa.add` inside `rock.kernel` functions so rocMLIR can absorb it into a fused kernel.

Based on `feat/rocmlirtriton_fusion` at d7d8705 ("Use conversion driver instead of greedy driver"), so the pattern is an `OpConversionPattern` reading operands off the adaptor.

### Heads-up: two fixes were needed before any pattern could fire

Neither is `hip.add`-specific — both block every future op converter too, so flagging them explicitly for review.

**1. The TOSA dialect was never loaded.** `MLIRTosaDialect` is linked into the `HipToTosa` target, but that only supplies C++ symbols; the dialect was absent from `hip-mlir-opt`'s registry, so creating a `tosa` op failed with ``Dialect `tosa' not found``. Marking the dialect legal on the `ConversionTarget` does not load it either. Fixed by overriding `getDependentDialects` in the pass class, which keeps the change local to `HipToTosa.cpp` — the TableGen base emits that method as an empty virtual, so no `Passes.td` or `InitAllPasses.h` edit is required.

**2. `applyFullConversion` was rejecting `func.func`.** The target marked `hip` illegal and `tosa` legal, but `applyFullConversion` treats an op with *no registered legality* as illegal. That includes the root `func.func` and its terminator, so the pass failed on every `rock.kernel` function with `failed to legalize operation 'func.func'` — independent of any patterns. Fixed with `markUnknownOpDynamicallyLegal`, since only the hip dialect is being converted away here.

### Broadcasting

Operands are mapped only when they already carry the result's rank. TOSA broadcasts size-1 dimensions only, while hip uses rank-extending ONNX/NumPy broadcasting, so a lower-rank operand would yield invalid TOSA.

This limits real-world coverage more than it might appear. In a ResNet-50 dump, ReLU lowers to `hip.max` against a rank-0 `tensor<f16>`, and `hip.add` bias forms show the same shape. Handling those needs a `tosa.reshape` to pad the rank with 1s first, the way `MultiBroadcastConverter` does in rocMLIR's `MIGraphXToTosa.cpp`. Deliberately left out of this PR to keep it reviewable.

### Testing

- `test_add.mlir` — same-rank operands, size-1 broadcast, and a function without `rock.kernel` (pass early-returns).
- `add-invalid.mlir` — rank-extending broadcast, rank-0 scalar operand, and dynamic shapes. These need a separate `--verify-diagnostics` file because full conversion turns a rejected op into a pass failure that emits no output for `FileCheck`.

Full lit suite: 478 passing. `Conversion/onnx-to-hip/test_qadd.mlir` fails both before and after this change; it never runs `convert-hip-to-tosa`.

### Also included

`.cursor/skills/add-hip-to-tosa-op/SKILL.md`, documenting the 21 verified 1-1 op pairs, the safe decompositions, and the traps found while wiring this up — the logical-vs-bitwise `RockTosaToElementwise` gap, `tosa.mul`'s shift operand, and the per-op result-type syntax difference in `HipOps.td`. Happy to split this into a separate PR if preferred.

### DIFF
--- a/lib/Conversion/HipToTosa/HipToTosa.cpp
+++ b/lib/Conversion/HipToTosa/HipToTosa.cpp
@@ -1,3 +1,8 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
 #include "hip/Dialect/IR/HipDialect.h"
 #include "hip/Dialect/Transforms/Passes.h"
 
--- a/lib/Conversion/HipToTosa/HipToTosa.cpp
+++ b/lib/Conversion/HipToTosa/HipToTosa.cpp
@@ -16,6 +16,51 @@ namespace mlir::hip {
 
 namespace {
 
+// TOSA broadcasts size-1 dimensions only and requires every operand to carry
+// the result's rank, so hip's ONNX/NumPy rank-extending broadcast does not
+// always survive a 1-1 mapping.
+bool isTosaCompatibleOperand(Value operand, RankedTensorType resultType) {
+  auto operandType = dyn_cast<RankedTensorType>(operand.getType());
+  if (!operandType || !operandType.hasStaticShape())
+    return false;
+  if (operandType.getElementType() != resultType.getElementType())
+    return false;
+  if (operandType.getRank() != resultType.getRank())
+    return false;
+
+  ArrayRef<int64_t> shape = operandType.getShape();
+  ArrayRef<int64_t> resultShape = resultType.getShape();
+  for (int64_t i = 0, e = resultType.getRank(); i < e; ++i)
+    if (shape[i] != resultShape[i] && shape[i] != 1)
+      return false;
+  return true;
+}
+
+// The hip context and the DPS `outs` buffer are both dropped: the result type
+// already encodes the destination.
+struct AddConverter final : public OpConversionPattern<hip::AddOp> {
+  using OpConversionPattern<hip::AddOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(hip::AddOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // Memref mode (post-bufferization) has no SSA result to replace.
+    if (op.getNumResults() != 1)
+      return rewriter.notifyMatchFailure(op, "expected tensor mode");
+
+    auto resultType = dyn_cast<RankedTensorType>(op.getResult(0).getType());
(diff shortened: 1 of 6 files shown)


lib/Conversion/HipToTosa/HipToTosa.cpp
```diff
@@ -27,8 +84,13 @@ class HipToTosaPass : public impl::ConvertHipToTosaPassBase<HipToTosaPass> {
     ConversionTarget conversion(*ctx);
     conversion.addIllegalDialect<HipDialect>();
     conversion.addLegalDialect<tosa::TosaDialect>();
+    // applyFullConversion treats an op with no registered legality as illegal,
+    // which would reject the enclosing func.func and its terminator. Only the
+    // hip dialect is being converted away here, so everything else is legal.
+    conversion.markUnknownOpDynamicallyLegal([](Operation *) { return true; });
```
ilango100: No need of this, I've added func dialect as legal. Please rebase.

lib/Conversion/HipToTosa/HipToTosa.cpp
```diff
@@ -16,7 +21,59 @@ namespace mlir::hip {
...
+  }
+};
+
 class HipToTosaPass : public impl::ConvertHipToTosaPassBase<HipToTosaPass> {
+  // TOSA is linked in but absent from hip-mlir-opt's dialect registry, so the
+  // pass declares it here to get it loaded before any tosa op is created.
+  // Marking the dialect legal on the ConversionTarget does not load it.
+  void getDependentDialects(DialectRegistry &registry) const override {
+    registry.insert<tosa::TosaDialect>();
+  }
```
ilango100: This is also added in passes.td, rebase.

ilango100 (changes requested):

lib/Conversion/HipToTosa/HipToTosa.cpp
```diff
@@ -29,6 +79,7 @@ class HipToTosaPass : public impl::ConvertHipToTosaPassBase<HipToTosaPass> {
     conversion.addLegalDialect<tosa::TosaDialect, func::FuncDialect>();
```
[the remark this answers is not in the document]
ilango100: We'll be marking ub.poison as legal op and then add canonicalizer pass to remove the dead ub.poison op.

PR #933, opened 2026-09-08
perf(gqa): narrow the decomposed prefill's key range per query chunk, -50.5% TTFT at 16K

BoarQing:
Rebased onto current `main`. Single commit; `lib/Runtime/real/gqa.cpp`, `lib/Runtime/Kernels/hip/gqa_kernel.hip` and `lib/Runtime/Kernels/include/hip_custom_kernels.h`.

This now carries what was #945 as well. The two were the same idea arriving in two steps — narrow each query chunk's key range, then get the upper end of that range right — and separating them left the first half asserting a bound the second half had to go back and fix. #945 is closed in favour of this.

## What was wrong

`kv_span` is computed once per op, before the chunk loop, from `past_len`:

```cpp
const int64_t lo = past_len - local_window_size + 1;
if (lo > 0) { kv_lo = lo; kv_span = total_seq - kv_lo; }
```

`past_len` is the absolute position of the **first** query row of the call, so at a fresh prefill it is 0, `lo` goes negative and the branch never fires. Window narrowing has only ever engaged at decode. #795 recovers Gemma-4's 1024-token window at compile time from the additive mask, but that recovery could not reach prefill through this path.

## What this does

Derives each chunk's key range from its own query range. For rows `[q0, q0+c)` the lower bound is `past_len + q0 - W + 1`, from the recovered window.

The **upper** bound is the part worth reading carefully, because the obvious answer is wrong.

For an op that declares `is_causal`, it is the chunk's own last row: `past_len + q0 + c`. `hip_gqa_causal_mask_f32` masks exactly the entries being dropped, so this is provably the same arithmetic.

For a bidirectional op it is read out of the additive mask, by a new kernel `hip_gqa_bias_key_extent` that scans the region above each chunk's last row and returns the highest key position any row in the chunk attends. An entry whose additive bias is at or below -65504 — the fp16 lower bound, the not-attended sentinel `softmax_f32_to_out_kernel` already documents — contributes exactly zero to the softmax, so excluding it is bit-exact. Anything less negative counts as attended, which only narrows less. The scan floors at the diagonal, so it can never be looser than the causal answer.

**A recovered window does not license the causal bound.** This is the trap, and an earlier revision of this PR fell into it. `AttentionWindowFold` matches `And(q >= k, q - k < W)`, whose first conjunct is the causal predicate — but it matches that as one *disjunct* of the keep-condition, not as the whole of it. Gemma-4's windowed mask is `And(pad, Or(win, seg))`, and the fold accepts it because the same-image-block leg `seg` cannot reach further **back** than the window: 254 soft tokens per block against a window of 1024. That argument bounds the range below and says nothing about above. `seg` is bidirectional, so it reaches one image block **above** the diagonal at a bias of 0, and clamping there deletes attended keys outright — a changed function, not a reassociation. So `no_causal` takes its upper bound from the mask whether or not a window was recovered, and only the attribute buys the diagonal.

Two things this had to get right mechanically:

- **`strideA` becomes explicit.** Once a chunk reads a sub-range, the dense `m*k` default is shorter than the `kv_span*d` the expand actually packed, and every head after the first would mis-stride silently.
- **Descriptors are resolved before the workspace is sized.** That region is allocated once and the loop cannot grow it. The shape-keyed cache collapses the 26 chunks to four entries, so this costs hash lookups rather than heuristic queries.

## Measurements

Gemma-4 26B-A4B at a 16,384-token prompt with an image, on gfx1151. TTFT interleaved and then order-reversed, both arms built from the same tree minutes apart, each with its own `TEMP` so neither sees the other's autotune cache:

| arm | block 1 | block 2 | mean p50 |
|---|---|---|---|
| `main` (`de258777`) | 35,353.1 ms | 35,291.3 ms | 35,322.2 ms |
| this branch | 17,512.7 ms | 17,787.8 ms | 17,650.2 ms |

**−17,672.0 ms, −50.03%.**

This was run twice, as two independent A/Bs several builds apart. The other gave 35,327.3 → 17,502.6 ms, −50.46%. The two `main` arms agree to 0.01% of each other, which is a useful check on the harness: whatever run-to-run spread the candidate arm shows is not drift in the baseline. For reference, the per-chunk narrowing alone measured −42.67% against the `main` of the time; the mask-derived bound on the global layers is the rest.

Per layer at that operating point, from the runtime's own per-op accounting:

| layers | shape | key extent per chunk | score elements | vs unnarrowed |
|---|---|---|---|---|
| 25 local | d=256, W=1024 | `[640, 1663]` | 26,012,784 | **10.14x** |
| 5 global | d=512, no window | `[640, 16241]` | 137,034,081 | **1.93x** |

The global layers' 137,034,081 is exactly the causal triangle summed over the chunking, `Σ c·(q0+c)` — that is, the mask states a bound that happens to coincide with causality on this input, and the scan recovers it rather than assuming it.

From an SQTT capture of the same operating point, per **local** layer:

| component | base | this branch | speedup | utilisation |
|---|---|---|---|---|
| score GEMM | 404.6 ms | 39.9 ms | 10.14x | 21% of a memory floor |
| softmax | 231.8 ms | 18.5 ms | 12.56x | 70% of a bandwidth floor |
| value GEMM | 106.4 ms | 10.5 ms | 10.12x | 48% of a memory floor |
| **GQA total** | **753.3 ms** | **78.9 ms** | **9.55x** | |

The measured 10.14x on the score GEMM lands exactly on the 10.14x analytic work reduction, so the narrowing captures what is available to it.

Those utilisation figures use a 59.4 TFLOP/s fp16 ceiling for gfx1151 (40 CU × 2.9 GHz × 512 FLOP/clk); the same model gives 122.9 TFLOP/s for a 7900 XTX against AMD's published 122.8. An earlier revision of this PR quoted 22% and 82% *of a compute floor* from a 24.6 TFLOP/s figure that was low by 2.4x. At the correct ceiling the compute floor drops below the 256 GB/s memory floor for both GEMMs, so they are memory-bound, and the value GEMM is not near its floor. Worth knowing when deciding what to look at next; it does not touch the measured times or the speedups, neither of which has a hardware ceiling in it.

GQA falls from 73.1% to well under half of GPU busy time, so attention is no longer the dominant prefill cost.

## Correctness

Recovering the key extent per chunk from the score GEMM's wavefront counts in an SQTT trace, the measured extents match the closed form on **all 52 captured chunks** — 640, then 1280, then 1663 steady, with a 1264 tail. That is the narrowing rule reproduced in hardware, not inferred from timings.

The mask-derived upper bound is bit-exact by construction, as argued above. The window-derived lower bound is exact but **not bitwise**: the value GEMM's `k` and the softmax's reduction length both shrink, so the same non-masked terms are summed in a different order and hipBLASLt may pick a different algorithm. Bitwise equality against the un-narrowed path is the wrong expectation. To separate a plumbing bug from that reassociation, development carried a switch that held every chunk at the op-level key range while keeping the per-chunk plumbing live — descriptor-per-chunk resolution, the explicit `strideA`, the per-chunk operand offsets. In that mode the output is bitwise identical to base, which is what isolates the two. For scale on the remaining difference, changing `sq_chunk` from 640 to 1280 on base, with no narrowing at all, diverges the generated text earlier than this change does.

On the diagonal-bound trap specifically, since "it measured fine" is exactly what made it survive: at the default score budget the 254-token image block sits at the head of the prompt and `sq_chunk` is 640, so no block straddles a chunk boundary and the two bounds agree to the element — 26,012,784 either way. Forcing `sq_chunk=127` puts a boundary inside the block; the bounds then separate by 16,891 entries per head and the 64-token greedy continuation diverges, in its description of the image. `sq_chunk` falls out of the score budget, so it shrinks as context grows: past roughly 44K tokens it is below 254 and even a leading block straddles. The mask-derived bound is what makes all of these agree.

`test/numeric/tests/test_gqa.py` shows no new failures, including under forced chunking (`HIPDNN_EP_GQA_SCORE_BUDGET_MB=1`); the 14 pre-existing failures are identical on both arms and the 8 passing tests report identical `max_diff`.

## Cost

The mask scan is one kernel over half the mask plus one D2H of a few ints, per chunked op that is `no_causal` — 30 layers on Gemma-4, roughly 250 MB read apiece. The mask is shared across layers of a kind, so a cache keyed on the bias pointer and the call's geometry would collapse those 30 scans to 2; that is left out here because the scan is already small against the score GEMMs it bounds. Any failure in it leaves the bound unset and the op scores its full key range, as before.

## Scope

`use_no_expand` is excluded from chunking entirely, so this applies to the expand flavour only. The chunk loop also runs once unless the score pair exceeds the 1 GiB budget, and with a single chunk `q0 + c == sq` and narrowing buys nothing — so shorter prompts see no change until `sq_chunk` is decoupled from the memory budget. Whisper's encoder self-attention and decoder cross-attention carry no additive mask to read a bound from and keep their full range.

### DIFF
--- a/lib/Runtime/Kernels/hip/gqa_kernel.hip
+++ b/lib/Runtime/Kernels/hip/gqa_kernel.hip
@@ -1162,6 +1162,123 @@ extern "C" int hip_gqa_add_attention_bias_f32(
   GQA_RETURN_LAUNCH_STATUS();
 }
 
+// Highest key position any query row in a chunk can attend, read off the
+// additive mask instead of assumed.
+//
+// The decomposed prefill tiles the query range into chunks and gives each its
+// own key extent. For a causal op that extent is the chunk's own last row,
+// which is exact and needs no mask inspection. A BIDIRECTIONAL op
+// (onnx.Attention is_causal=0) has no such bound in its attributes, so every
+// chunk scores every key -- and on Gemma-4's 5 global layers that is the
+// largest single block of wasted work in a 16K prefill, because their mask is
+// causal apart from a same-image-block leg reaching one image block above the
+// diagonal. AttentionWindowFold cannot supply the bound: the block extent is a
+// property of the input, not of the mask's structure.
+//
+// The mask itself already carries it, exactly, for one streaming read. An entry
+// whose additive bias is at or below -65504 -- the fp16 lower bound, and the
+// sentinel HF exports write for "not attended" -- cannot survive the softmax:
+// relative to any score the GEMM ahead of it can produce, exp() of it
+// underflows to exactly zero. That is the same property the mask relies on to
+// mask at all, and softmax_f32_to_out_kernel above already documents this
+// value for the same reason. So dropping those key positions from a chunk's
+// range is bit-exact rather than an approximation, and anything less negative
+// than the sentinel is treated as attended, which only narrows less.
+//
+// Only the region ABOVE the chunk's last row is scanned. Keys at or below it
+// are kept whatever the mask says, so the answer is max(diagonal, highest
+// attended key above it) and the work is half the mask rather than all of it.
+// The max is reduced over every bias batch/head plane as well as every row in
+// the chunk, so a broadcast plane and a per-head mask both yield a bound valid
+// for all of them.
+#define GQA_BIAS_EXTENT_THREADS 256
+
+// Additive bias at or below this cannot survive the softmax. Not a tolerance:
+// it is the fp16 lower bound, which is the sentinel itself.
+#define GQA_BIAS_MASKED_MAX (-65504.0f)
+
+template <typename TBias>
+__global__ void bias_key_extent_kernel(
+    const TBias* __restrict__ bias, int* __restrict__ out,
+    int bias_batch, int bias_heads, int bias_sq, int bias_total_seq,
+    int sq, int sq_chunk, int past_len) {
+  const int chunk = blockIdx.x;
+  const int q0 = chunk * sq_chunk;
+  if (q0 >= sq)
+    return;
+  const int c = min(sq_chunk, sq - q0);
+
+  // Absolute key position of the chunk's last query row. Keys at or below it
+  // are attended by the causal leg and are never dropped, so it is the floor.
+  const int diag = past_len + q0 + c - 1;
+  int best = diag;
+
+  const int scan_lo = diag + 1;
+  if (scan_lo < bias_total_seq) {
+    const size_t plane = static_cast<size_t>(bias_sq) * bias_total_seq;
+    const int planes = bias_batch * bias_heads;
+    for (int p = 0; p < planes; ++p) {
+      for (int r = 0; r < c; ++r) {
+        const size_t row_base = static_cast<size_t>(p) * plane +
+                                static_cast<size_t>(q0 + r) * bias_total_seq;
+        // `k > best` first: once this thread has seen a high hit, the lower
+        // columns cannot raise the answer and the load is skipped.
+        for (int k = scan_lo + threadIdx.x; k < bias_total_seq;
+             k += blockDim.x) {
+          if (k > best &&
+              static_cast<float>(bias[row_base + k]) > GQA_BIAS_MASKED_MAX)
+            best = k;
+        }
+      }
+    }
+  }
+
+  __shared__ int red[GQA_BIAS_EXTENT_THREADS];
+  red[threadIdx.x] = best;
+  __syncthreads();
+  for (int s = blockDim.x / 2; s > 0; s >>= 1) {
+    if (threadIdx.x < s)
+      red[threadIdx.x] = max(red[threadIdx.x], red[threadIdx.x + s]);
+    __syncthreads();
+  }
+  if (threadIdx.x == 0)
+    out[chunk] = red[0];
+}
+
+extern "C" int hip_gqa_bias_key_extent(
+    void* stream_ptr, const void* bias, void* out_chunk_hi,
+    int bias_batch, int bias_heads, int bias_sq, int bias_total_seq,
+    int sq, int sq_chunk, int past_len, int num_chunks,
+    int bias_element_size_bytes) {
+  if (!bias || !out_chunk_hi || num_chunks <= 0 || sq_chunk <= 0 ||
+      bias_total_seq <= 0 || bias_sq <= 0)
+    return -1;
+  hipStream_t stream = static_cast<hipStream_t>(stream_ptr);
+  if (bias_batch <= 0)
+    bias_batch = 1;
+  if (bias_heads <= 0)
+    bias_heads = 1;
+  (void)hipGetLastError();
+  if (bias_element_size_bytes == 2) {
+    bias_key_extent_kernel<__half>
+        <<<dim3(num_chunks), GQA_BIAS_EXTENT_THREADS, 0, stream>>>(
+            static_cast<const __half*>(bias), static_cast<int*>(out_chunk_hi),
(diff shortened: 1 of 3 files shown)


BoarQing:
Gemma Perf: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F832%2F&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F833%2F&curr_job=&curr=
Gemma acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1773/

amd-yanjunz:

- PR: ROCm/hip-ep#933
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/933#issuecomment-5596266647)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/842/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6aa122953388831dc060f5c7
- Compare with latest main branch results: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fids%3D6aa04cedfdbe2b87391b4c31%26collection%3DHIP-EP-Dynamic&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=842

amd-mingw:
acc failed for gemma-4-26B
image

amd-mingw (changes requested):
need fix gemma-4-26B accuracy issue

BoarQing:
Gemma Perf: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F891%2F&base_job=daily_hip_ep_oga&base=891&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F885&curr_job=daily_hip_ep_oga&curr=885

Gemma Acc: http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1775/

amd-yanjunz:

- PR: ROCm/hip-ep#933
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/933#issuecomment-5601713168)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/858/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6aa193d686c7afa3c42f883d
- Compare with latest main branch results: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fids%3D6aa04cedfdbe2b87391b4c31%26collection%3DHIP-EP-Dynamic&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=858

amd-mingw (approved):

PR #935, opened 2026-09-08
fix(onnx-to-hip): minor build fixes

pschlan-amd:
## Summary

find_program cannot see mlir-pdll until the FetchContent LLVM build produces the binary, which disabled PDL fusion on from-source trees.

The same pattern is already used for lib/Runtime, for example.

Also disable DIA to prevent issues with ATL dependency.

## Related issue or design

None

## Test plan

Local build was failing before (since a recent commit caused `python/` to error out during build when mlir-pdll is n/a). With the change, it passes.

## Checklist

Debugged and resolved with assistance of Cursor. The code changes have been reviewed and understood.

### DIFF
--- a/lib/Conversion/OnnxToHip/CMakeLists.txt
+++ b/lib/Conversion/OnnxToHip/CMakeLists.txt
@@ -6,23 +6,36 @@
 # ============================================================================
 # PDLL Pattern Compilation
 # ============================================================================
-# Locate mlir-pdll. LLVM_TOOLS_BINARY_DIR is exported by find_package(MLIR) and
-# covers both the prebuilt SDK and an installed LLVM; the _deps path covers an
-# LLVM built inside this project's FetchContent tree. Default search paths stay
-# enabled so a tool on PATH is still usable.
-find_program(MLIR_PDLL_EXE mlir-pdll
-    HINTS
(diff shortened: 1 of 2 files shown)


lib/Conversion/OnnxToHip/CMakeLists.txt
```diff
@@ -47,7 +78,23 @@ if(MLIR_PDLL_EXE)
         "Compiled PDL fusion patterns that must ship beside the EP library")
     message(STATUS "OnnxToHip: PDLL fusion enabled — patterns: ${PDL_OUTPUT}")
 else()
-    message(WARNING "OnnxToHip: mlir-pdll not found — PDLL fusion disabled")
+    message(FATAL_ERROR
```
amd-bqiao: expect to be FATAL?

amd-bqiao (dismissed):
lgtm

lib/Conversion/OnnxToHip/CMakeLists.txt
pschlan-amd: Idea was yes, since it's now required in https://github.com/ROCm/hip-ep/blob/00e4fcbd2ea5c56dd4f365f0dd91fc2a2b819b5b/python/CMakeLists.txt#L53 - however I now see python build is optional, so I'll revert this to warning.

amd-bqiao (approved):

PR #937, opened 2026-09-08
fix(rms-norm): derive reduction width from axis, not scale

AMDmoore:
## Summary

RMSNorm used `scale`'s element count as the reduction width. That is wrong when the scale spans several rows (grouped gain, e.g. scale `[G, D]` against input `[N, G, D]` with `axis = -1`): the kernel reduced over `G*D` instead of `D` and applied the wrong gains.

Lowering now computes `norm_num_elements` as the product of the input dims from `axis` on, and the kernel indexes scale by `row % scale_rows`. Ordinary shared-gain RMSNorm (`scale_rows == 1`) is unchanged.

## Related issue or design

None.

## Why

ONNX reduces over the axes that `axis` selects and only then multiplies by scale. Scale may cover several rows; it must not widen the reduction. Runtime still does not consume `axis` — that folding happens in HipToLLVM, which is the only place that sees both the attribute and the input shape.

## What

- `RmsNormOpLowering` emits `norm_num_elements` from `axis`.
- `wrap_rms_norm` / `hip_rms_norm` take that width plus `scale_rows`.
- Reject a scale whose length is not a multiple of the normalized width (interior broadcast such as `[G, 1]` vs `[N, G, D]`).
- LIT `test_rms_norm.mlir` updated for the 11-parameter ABI.

## Test plan

- [x] Windows gfx1151: `build-hip-ep.bat` Release configure + build + install from this branch.
- [x] `pre-commit run --files` on the 7 changed files (lintrunner reformatted `NormLowering.cpp`).
- [ ] CI LIT (`test_rms_norm.mlir`).

## Notes for reviewers

`axis` remains on the runtime ABI but is unused (`(void)axis`); the new `norm_num_elements` argument is the folded result.

## Checklist

- [x] User-facing or design documentation was updated when needed (kernel/runtime comments).

Cursor (Grok 4.6) assisted with the port onto `main`, formatting, and this PR. The RMSNorm grouped-scale behavior was authored in the source tree and reviewed before submit.


amd-mingw (approved):
LGTM

PR #943, opened 2026-09-09
fix(hipsr): keep barrier placeholders on the data values

zpye:
## Summary

A barrier placeholder's region reads the *contents* of its inputs, not just their shapes. The conversion post-process rewired every placeholder input to its shape-graph counterpart, which pointed barriers at a destination buffer rather than at the value written into it. Barriers now keep the data values.

## Why

`hipsr.producer` and `hipsr.consumer` below stand in for any hipsr op; types and regions are elided.

```mlir
// A producer is handed a destination and yields the value written into it.
%dest  = hipsr.placeholder(%ctx) ins(%src) {normal}
%value = hipsr.producer(%ctx) ins(%src) outs(%dest)

// Before: the barrier named %dest, while its own consumer named %value.
%init  = hipsr.placeholder(%ctx) ins(%dest) {barrier}
%out   = hipsr.consumer(%ctx) ins(%value) outs(%init)

// After: both name %value.
%init  = hipsr.placeholder(%ctx) ins(%value) {barrier}
%out   = hipsr.consumer(%ctx) ins(%value) outs(%init)
```

One-Shot Bufferize runs the producer in place on `%src`, so nothing ever writes `%dest`. The barrier's region read it anyway and computed a shape out of uninitialized memory.

## What

- `OnnxToHipsr.cpp`: skip barriers when rewiring placeholder inputs into the shape graph.
- `HipsrPlaceholderOp.cpp`: a barrier may name data results, and its inputs must equal its consumer's operands. This also rejects a permuted list, which the region would have read positionally.
- `MaterializeInitTensorsPass.cpp`: still reject a barrier input allocated inside the same pool domain, now tested by block rather than by defining operation.
- Goldens regenerated. Most of the diff is `Pipelines/embedding.mlir` and `PartitionPoolDomains/materialization.mlir`; the substantive change is about 25 lines.

LIT: 480 passed, 12 unsupported. Prepared with Cursor and Claude Opus 5, and reviewed; commit trailers record the assistance.

### DIFF
--- a/include/hip/Dialect/Hipsr/IR/HipsrPlaceholderOp.td
+++ b/include/hip/Dialect/Hipsr/IR/HipsrPlaceholderOp.td
@@ -24,15 +24,16 @@ def Hipsr_PlaceholderOp : Hipsr_Op<
       body writes into, so the placeholder carries that destination's shape,
       which may differ from the result held there.
 
-    Inputs feed the shape graph and must be block arguments or results of
-    `hipsr.placeholder`, `arith.constant`, or `hipsr.constant` — never data
-    results, which keeps the shape graph parallel to the data graph.
-    `placeholder_type` is `normal` or `barrier`.
+    A `normal` placeholder's inputs feed the shape graph and must be block
+    arguments or results of `hipsr.placeholder`, `arith.constant`, or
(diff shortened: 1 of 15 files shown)


wcy123 (approved):
LGTM.

PR #945, opened 2026-09-09
Draft

BoarQing:


BoarQing:
Folded into #933 and closed in favour of it.

The two were one idea arriving in two steps -- narrow each query chunk's key range, then get the upper end of that range right -- and keeping them apart left #933 asserting a bound this PR had to come back and correct. Worse, the split hid a real bug between them: #933 took the diagonal as an upper bound on any op with a recovered window, on the grounds that AttentionWindowFold's \And(q >= k, q - k < W)\ carries a proof of causality. It does not. The fold matches that as one disjunct of \And(pad, Or(win, seg))\, and the same-image-block leg is bidirectional, so it reaches an image block above the diagonal at a bias of 0. This PR wired \hip_gqa_bias_key_extent\ only to the bidirectional-without-window layers, so the 25 windowed ones kept the bad bound.

#933 now carries the kernel from here plus that fix, as a single commit rebased onto current main, and measures -50.46% TTFT at 16K against main de258777 (35,327.3 -> 17,502.6 ms, interleaved and order-reversed).

PR #947, opened 2026-09-09
feat(runtime): add OP_PROFILE to the qmatmul, qadd, and qmul wrappers

amd-genmingz:
## Summary

Adds an `OP_PROFILE` scope to the three quantized-fusion runtime wrappers that
were missing one, so `HIPDNN_EP_PERF=1` reports `qmatmul`, `qadd`, and `qmul` as
their own rows in the `[PERF]` table and in the Chrome trace. No behavior change
when profiling is off.

## Related issue or design

- Design: `docs/design/per-op-profiling.md`
- Ops instrumented here: #921 (`hip.qmatmul`), #916 (`hip.qmul`), #907 (`hip.qadd`)

## Why

Per-op GPU time is derived by differencing consecutive markers on the in-order
stream (`op_profile_resolve_and_print`: `gpuMs = cumMs - tlPrevMs`). An op
without a marker is therefore not merely unreported — its GPU time is absorbed
into the next instrumented op's duration. On QDQ models the quantized
MatMul/Add/Mul sit on the hot path, so their cost was being silently attributed
to neighbouring ops, which is misleading exactly where the recent fusion work
needs measurement.

## What

- `wrap_qmatmul`: label `qmatmul`, shape tag `m=..,n=..,k=..,b=..:<a>/<b>-><y>`.
  The dtypes are part of the tag because A16W8 and A8W8 take different kernel
  paths and should not share a row.
- `wrap_qelementwise`: `hip.qadd` and `hip.qmul` share this single wrapper, so a
  fixed label would merge them into one row. The label is derived from the
  `kind` argument, the same way `global_pool.cpp` labels by mode. Shape tag is
  `<dtype>:<out dims joined by x>`, bounded by the buffer size and null-safe on
  `out_shape`.
- `hipdnn_ep_runtime.h`: new `hipdnn_ep_qelementwise_kind_name()` placed next to
  the existing `hipdnn_ep_tensor_op_name()`, matching the pool / global-pool
  naming helpers.

No CMake change is required: `op_profile.h` is already in the bitcode `DEPENDS`
list, and both `.cpp` files are already registered in `compile_to_bitcode` and
`RUNTIME_BC_MODULES`.

## AI Assistance

### DIFF
--- a/lib/Runtime/hipdnn_ep_runtime.h
+++ b/lib/Runtime/hipdnn_ep_runtime.h
@@ -51,6 +51,17 @@ static inline const char *hipdnn_ep_tensor_op_name(int64_t op) {
   }
 }
 
+static inline const char *hipdnn_ep_qelementwise_kind_name(int64_t kind) {
+  switch (kind) {
+  case HIPDNN_EP_QELEMENTWISE_ADD:
+    return "qadd";
+  case HIPDNN_EP_QELEMENTWISE_MUL:
+    return "qmul";
+  default:
+    return "qelementwise_unknown";
+  }
+}
+
 static inline int64_t hipdnn_ep_datatype_size(int64_t data_type) {
   switch (data_type) {
   case HIPDNN_EP_DATATYPE_FLOAT:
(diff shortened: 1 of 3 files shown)


amd-mingw (approved):
LGTM

PR #948, opened 2026-09-09
feat(hipsr): run hipsr-pool-alloc so each domain uses one buffer pool

zpye:
## Summary

This turns on `hipsr-pool-alloc` in the HIPSR pipeline, right after `hipsr-use-output-allocator`.

Every temporary buffer in a pool domain used to get its own `memref.alloc`. Now they are all views into one pool that the runtime hands out. All three pipeline goldens end with zero device allocations, so the graph no longer asks the device allocator for one buffer per temporary on every inference.

The pass already existed. This change makes it work on real pipeline output and switches it on.

## Before / after

Two temporaries in one domain, in the order the pipeline writes them: every allocation first, then the compute ops.

```mlir
hipsr.pool_domain(%ctx, %in : !hipsr.context, memref<4x1024xf16, #hipsr.mem<device>>) {
^bb0(%dctx: !hipsr.context, %din: memref<4x1024xf16, #hipsr.mem<device>>):
  %a = memref.alloc() : memref<4x1024xf16, #hipsr.mem<device>>
  %b = memref.alloc() : memref<4x1024xf16, #hipsr.mem<device>>
  hipsr.add(%dctx) ins(%din, %din) outs(%a)
  hipsr.add(%dctx) ins(%a, %a) outs(%b)
  hipsr.pool_domain_yield
} {domain_id = 0 : i64}
```

They become one pool and two views into it:

```mlir
hipsr.pool_domain(%arg0, %arg1 : !hipsr.context, memref<4x1024xf16, #hipsr.mem<device>>) {
^bb0(%arg2: !hipsr.context, %arg3: memref<4x1024xf16, #hipsr.mem<device>>):
  // ... each size rounded up to the 256-byte pool alignment ...
  %6 = arith.addi %2, %5 : index
  %7 = hipsr.get_pool(%arg2, %6) {domain_id = 0 : i64} : memref<?xi8, #hipsr.mem<device>>
  %view   = memref.view %7[%c0][] : memref<?xi8, #hipsr.mem<device>> to memref<4x1024xf16, #hipsr.mem<device>>
  %view_3 = memref.view %7[%2][]  : memref<?xi8, #hipsr.mem<device>> to memref<4x1024xf16, #hipsr.mem<device>>
  hipsr.add(%arg2) ins(%arg3, %arg3) outs(%view)
  hipsr.add(%arg2) ins(%view, %view) outs(%view_3)
} {domain_id = 0 : i64}
```

Two buffers that are never live at the same time share an offset, so the pool is as big as the peak rather than the sum.

## Why

Hand-written test input has three properties that real pipeline output does not have. Each one used to make the pass fail.

**Not every buffer is device memory.** The pool lives on the device, so a `#hipsr.mem<host>` shape buffer cannot be carved out of it. It keeps its own allocation.

**The next domain can read a pooled buffer.** The partitioner splits a domain at a barrier and passes each domain's results to the next one, so this is normal. In `sample_static` the cast output crosses the split:

```mlir
%0:3 = hipsr.pool_domain(...) {                                   // domain 0
  %alloc_9 = memref.alloc() : memref<2x1xf32, #hipsr.mem<device>>
  hipsr.cast(%arg3) ins(%alloc_8) outs(%alloc_9)
  hipsr.pool_domain_yield %alloc_9, %4, %2
} -> memref<2x1xf32, #hipsr.mem<device>>, ... {domain_id = 0 : i64}
%1 = hipsr.pool_domain(%arg0, %0#0, %0#1, %0#2 : ...)             // domain 1 reads it
```

`hipsr.pool_domain` implements `RegionBranchOpInterface`, so `BufferViewFlowAnalysis` reports the domain result `%0#0` as an alias of `%alloc_9`. `%0#0` is defined outside the body, which means all of its users are outside the body too. That is not the allocation escaping, so the pass drops `%0#0` from the alias list. It only drops an alias defined by the domain op itself. The `hipsr.pool_domain_yield` inside the body is still a normal use, and that is what keeps the grouper from reusing the yielded buffer's offset.

**A domain can have nothing to pool.** It may only compute shapes on the host, or write straight into a graph output. That no longer fails.

The pass is stricter in one place. If a device allocation is still sitting at the top level of a domain body after the rewrite, the pass now fails. Leaving it there would quietly bring back the per-inference allocation this pass exists to remove.

## Notes for reviewers

- `hipsr-pool-alloc` has to run after `hipsr-use-output-allocator`, so graph outputs stay owned by the runtime and never land in the pool.
- Every view is created before any allocation is erased, so an erase cannot invalidate the builder's insertion point.
- `hipsr-materialize-init-tensors` now reads all the dynamic sizes first and creates the `tensor.empty` ops afterwards. It used to emit each init tensor as a unit, which left the second allocation's size reads after the first allocation. Bufferization turned that into a `memref.load` between two `memref.alloc`s, and the pool can no longer read those sizes at the point where it is taken. The new order also drops one duplicate `shape.get_extent` per allocation.
- The leftover-allocation check only looks at the top level of a domain body. The pass does not pool nested regions, and failing on them would turn a missing feature into a whole-model CPU fallback.
- No `hipdnn.pool_size`/`buffer_offsets`/`buffer_count` is emitted. Each `hipsr.get_pool` carries its own `domain_id` and a size computed in the IR. That is the runtime-managed path `generate-interface` already falls back to, and a dynamic size has no fixed offset to report.
- `PoolAlloc/basic.mlir` has a new case, `yielded_buffer_read_by_next_domain`, for the cross-domain read. Drop the alias filter and it fails with `allocation escapes IsolatedFromAbove domain`. The three pipeline goldens cover it too.

Full LIT suite: 484 passed, 12 unsupported (all already unsupported on `main`), 0 failed.

Prepared with AI assistance (Cursor, Claude), reviewed and validated locally.

### DIFF
--- a/lib/Dialect/Hipsr/Pipelines/Pipelines.cpp
+++ b/lib/Dialect/Hipsr/Pipelines/Pipelines.cpp
@@ -28,6 +28,7 @@
 //   --one-shot-bufferize
 //   --convert-linalg-to-loops
 //   --hipsr-use-output-allocator
+//   --hipsr-pool-alloc
 void mlir::hipsr::buildHipsrPipeline(OpPassManager &pm,
                                      const HipsrPipelineOptions & /*options*/) {
   pm.addPass(createAddContextArgPass());
@@ -58,6 +59,7 @@ void mlir::hipsr::buildHipsrPipeline(OpPassManager &pm,
   pm.addNestedPass<func::FuncOp>(createConvertLinalgToLoopsPass());
 
   pm.addNestedPass<func::FuncOp>(createHipsrUseOutputAllocatorPass());
+  pm.addPass(createHipsrPoolAllocPass());
 }
 
 void mlir::hipsr::registerHipsrPipelines() {
--- a/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
+++ b/lib/Dialect/Hipsr/Transforms/HipsrPoolAllocPass.cpp
@@ -7,6 +7,7 @@
 
 #include "hip/Dialect/Hipsr/IR/HipsrDialect.h"
 #include "hip/Dialect/Hipsr/IR/HipsrOps.h"
+#include "hip/Dialect/Hipsr/IR/HipsrTypes.h"
 
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Bufferization/Transforms/BufferViewFlowAnalysis.h"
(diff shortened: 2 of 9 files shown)


wcy123 (dismissed):
LGTM

wcy123 (approved):

PR #950, opened 2026-09-09
Convert the shape-manipulating hip ops to TOSA

nvunnam57128:
## What

Converts the shape-manipulating ops to TOSA so rocMLIR can absorb them into a fused kernel: **Transpose, Reshape, Squeeze, Unsqueeze, Flatten and Slice**.

rocMLIR does not execute these — it folds them into the neighbouring conv or gemm as a layout change. `TosaToRock.cpp`'s `TransposeRewritePattern` rewrites a transpose after a conv into an `output_layout` attribute and one after a matmul into `transpose_c`, while `mergeTransposeWithGemmLikeOp` pushes a transpose on the input side into `input_layout` / `filter_layout`. Left in the HIP dialect, a transpose instead lowers to `llvm.call @wrap_transpose`, a standalone kernel and a full extra pass over memory.

## How the six ops route here

Only Transpose arrives as a hip op. `OnnxToHip` deliberately decomposes the others into builtin tensor metadata ops, because they move no data and a DPS hip op would force an allocation and a copy:

| ONNX op | What reaches this pass |
| --- | --- |
| Transpose | `hip.transpose` |
| Reshape, Squeeze, Unsqueeze, Flatten | `tensor.collapse_shape` / `tensor.expand_shape` |
| Slice (constant bounds, positive step) | `tensor.extract_slice` |
| Slice (negative step, runtime bounds) | `hip.slice` — not claimed |

So the six ops are covered by three patterns: `TransposeConverter`, a `ReshapeConverter` templated over collapse/expand, and `ExtractSliceConverter`. `hip.slice` exists precisely for the cases `tosa.slice` cannot express — it has no stride operand and no reverse traversal — so it is left for the runtime.

## Legality: borrowed vs owned ops

The hip ops stay unconditionally illegal, matching the 18 already converted: one this pass claims must convert or the pass fails.

The tensor ops are different. They are borrowed, not owned — a kernel can legitimately hold a strided or rank-reducing extract, or a dynamically shaped reshape, and rocMLIR consumes those directly. Marking them illegal outright would turn previously-fine kernels into hard errors, so they are illegal only where a pattern will succeed:

```cpp
conversion.addDynamicallyLegalOp<tensor::ExtractSliceOp>(
    [](tensor::ExtractSliceOp op) { return !isTosaExpressibleSlice(op); });
```

## Testing

New `test_shape.mlir` (227 lines) covers the conversions, the passthrough forms, and the rejection. It includes a case with the shape `hip-fuse-rocmlir` actually produces — a `ub.poison` context, a `tensor.empty` per outs buffer and a `hip.conv` anchor — and checks all three survive while the shape ops convert around them. The existing tests hand-write kernels taking their buffers as block arguments, which is the blind spot that let the earlier `applyFullConversion` regression reach a real model.

Full lit suite: 479 passing. The one failure, `test_qadd.mlir`, runs only `--convert-onnx-to-hip`, which this branch does not touch; it was failing before this work.

## Note for reviewers

rocMLIR's pipeline opens with `--tosa-to-tensor`, which converts `tosa.reshape` back into a collapse/expand pair and `tosa.slice` back into `tensor.extract_slice`. For the reshape family that makes the conversion a round trip, since hip-ep already emits the form `tosa-to-rock` consumes. Happy to drop `ReshapeConverter` and keep only Transpose if that is the preference — the ops would then be covered by leaving them alone.

Separately, `FuseROCMlir`'s `isaPointwiseOp` list contains only elementwise ops, so none of these are outlined into a `rock.kernel` yet and the patterns will not fire on a real model until that changes. Follow-up.

### DIFF
--- a/lib/Conversion/HipToLLVM/RocMlirLowering.cpp
+++ b/lib/Conversion/HipToLLVM/RocMlirLowering.cpp
@@ -61,22 +61,21 @@ struct RocMlirOpLowering : public ConvertOpToLLVMPattern<RocMlirOp> {
     // ELF carries embedded NULs, so keep the exact byte length (createGlobal
     // string stores the StringRef's full range, no NUL-termination scan).
     StringRef callee = op.getKernel();
-    Value kernelBinary = globalString(module, rewriter, loc,
-                                      (callee + "_binary").str(),
-                                      op.getKernelBinary());
+    Value kernelBinary =
+        globalString(module, rewriter, loc, (callee + "_binary").str(),
+                     op.getKernelBinary());
     // NUL-terminate the function name so a C `char*` consumer can read it.
     std::string funcNameStr = callee.str();
     funcNameStr.push_back('\0');
-    Value funcName =
-        globalString(module, rewriter, loc, (callee + "_name").str(),
-                     funcNameStr);
+    Value funcName = globalString(module, rewriter, loc,
+                                  (callee + "_name").str(), funcNameStr);
 
     // Operand data pointers, in kernel-arg order: inputs first, then output.
     SmallVector<Value> argPtrs;
     for (Value in : adaptor.getInputs())
       argPtrs.push_back(extractContiguousMemRefPtr(in, rewriter, loc));
-    argPtrs.push_back(extractContiguousMemRefPtr(adaptor.getOutput(), rewriter,
-                                                 loc));
+    argPtrs.push_back(
+        extractContiguousMemRefPtr(adaptor.getOutput(), rewriter, loc));
 
     // alloc(kernargs, size): stack buffer of N pointer slots.
     int64_t numArgs = static_cast<int64_t>(argPtrs.size());
@@ -88,8 +87,8 @@ struct RocMlirOpLowering : public ConvertOpToLLVMPattern<RocMlirOp> {
     for (int64_t i : llvm::seq<int64_t>(0, numArgs)) {
       Value idx = LLVM::ConstantOp::create(rewriter, loc, i32Type,
                                            rewriter.getI32IntegerAttr(i));
-      Value slot = LLVM::GEPOp::create(rewriter, loc, ptrType, ptrType,
-                                       kernargs, idx);
+      Value slot =
+          LLVM::GEPOp::create(rewriter, loc, ptrType, ptrType, kernargs, idx);
       LLVM::StoreOp::create(rewriter, loc, argPtrs[i], slot);
     }
 
--- a/lib/Conversion/HipToTosa/CMakeLists.txt
+++ b/lib/Conversion/HipToTosa/CMakeLists.txt
@@ -17,5 +17,6 @@ add_dependencies(HipToTosa
 target_link_libraries(HipToTosa PUBLIC
 	HipDialectIR
 	MLIRPass
+	MLIRTensorDialect
 	MLIRTosaDialect
 )
--- a/lib/Conversion/HipToTosa/HipToTosa.cpp
+++ b/lib/Conversion/HipToTosa/HipToTosa.cpp
@@ -11,6 +11,7 @@
 #include <mlir/Dialect/Tosa/IR/TosaOps.h>
 #include <mlir/Dialect/Tosa/Utils/ConversionUtils.h>
 #include <mlir/Dialect/UB/IR/UBOps.h>
+#include <mlir/Dialect/Utils/StaticValueUtils.h>
 #include <mlir/IR/BuiltinTypes.h>
 #include <mlir/IR/MLIRContext.h>
 #include <mlir/IR/PatternMatch.h>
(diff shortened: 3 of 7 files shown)


nvunnam57128:
Rebased onto the current `feat/rocmlirtriton_fusion`. The conflicts were both with @ilangoR's `b7634844` ("Fix HipToTosa for matmul.mlir") and both were textual — adjacent edits, no disagreement about behaviour:

1. **The const-shape helper.** `b7634844` generalised `reshapeTo3D` into `reshapeTo(input, shape, rewriter)`; this branch had extracted the const-shape half into `createConstShape` because `ExtractSliceConverter` needs a bare `!tosa.shape` value with no reshape attached. Resolved by keeping both, with `reshapeTo` calling `createConstShape`, so there is one place that builds a `tosa.const_shape`.
2. **The conversion target.** `b7634844` added `addLegalOp<ub::PoisonOp, tensor::EmptyOp>()`; this branch added `TransposeOp` to the illegal list plus the three dynamic-legality rules. Both kept.

While resolving (2) I reworded the comment above `addLegalOp`. It justified the line as needed because "a full conversion still requires every remaining op to be legal", but this pass uses `applyPartialConversion`, so that is not why the line is needed — it is needed because `tensor.empty` is left dead once the DPS `outs` operand is dropped. The line stays either way; only the reasoning changed.

Verified after the rebase: `hip-mlir-opt` builds clean, and the lit suite is unchanged at 479 passing.

---

**Two pre-existing failures on this branch, neither from this PR.** Both `pre-commit` and the Windows builds are `pull_request`-triggered, and `b7634844`..`7e8eeaf2` were pushed straight to `feat/rocmlirtriton_fusion` rather than through a PR, so this branch is the first thing to run either check against them.

**`pre-commit` — fixed here, in its own commit.** `lintrunner` runs over all files and found 55 `clang-format` violations across four files this PR otherwise has no reason to touch:

| File | Violations |
| --- | --- |
| `tools/hip-rocmlir-compiler/hip-rocmlir-compiler.cpp` | 40 |
| `lib/Conversion/HipToLLVM/RocMlirLowering.cpp` | 10 |
| `lib/Runtime/hipdnn_ep_runtime.h` | 4 |
| `lib/Conversion/HipToTosa/HipToTosa.cpp` | 1 |

The last one is `MatMulConverter`'s `rewriter.replaceOp`, which came in with `b7634844` and is preserved verbatim through the rebase. The fix is the output of the pinned `clang-format 16.0.1` and nothing more — line rewrapping and comment reflowing, verified to leave the token stream of every file identical. It is a separate commit, so it can be dropped if you would rather format these yourself.

**Windows builds — left alone, needs a decision.** `build_mock` and `build_real` both fail to compile `hip-rocmlir-compiler`:

```
tools\hip-rocmlir-compiler\hip-rocmlir-compiler.cpp(57):
fatal error C1083: Cannot open include file: 'dlfcn.h'
```

The tool is POSIX-only by design — as its own `CMakeLists.txt` notes, it reaches the rocMLIR pipeline by `dlopen`/`dlsym`-ing `librockCompiler.so` rather than link-time coupling to it — but `tools/CMakeLists.txt` adds it on every platform. Guarding that one `add_subdirectory` with `if(NOT WIN32)` is enough; the `hipdnn_ep_harden_comgr_safe` loop in the root `CMakeLists.txt` already skips absent targets via `if(TARGET ...)`. I have not touched it, since it is your tool and you may prefer a `LoadLibrary` shim over dropping it from the Windows build. Happy to push either.

`build-linux` passes.

nvunnam57128:
Pushed one more commit: `hip.expand` (ONNX `Expand`) now converts too.

It is a broadcast rather than a shape manipulation, so it sits slightly outside what the title says — happy to retitle, split it out, or drop the commit if you would rather keep this PR to the original six.

**Why it belongs with the rest.** Left alone, `hip.expand` lowers to a `wrap_expand` kernel that materialises the whole result — the same separate-kernel cost the other ops here exist to avoid.

**How it converts.** TOSA has no broadcast op, so there is nothing to map onto directly. rocMLIR spells one as a multiply against a tensor of ones at the result shape, leaning on the implicit broadcasting of TOSA's elementwise ops, and `MIGraphXToTosa` lowers `migraphx.multibroadcast` — the op ONNX `Expand` maps to on that path — exactly this way:

```
%0 = "tosa.const"() <{values = dense<1.000000e+00> : tensor<64x768x2304xf16>}> : () -> tensor<64x768x2304xf16>
%1 = "tosa.const"() <{values = dense<0> : tensor<1xi8>}> : () -> tensor<1xi8>
%2 = tosa.mul %x, %0, %1 : (tensor<1x768x2304xf16>, tensor<64x768x2304xf16>, tensor<1xi8>) -> tensor<64x768x2304xf16>
```

That is the same constant and the same operand order as the `CHECK` lines rocMLIR pins in its own `mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir`, down to the f16 element type.

**The multiply does not survive.** `TosaToRock`'s `mulBroadcast` matches this idiom, recognises the ones through `isConstantOne`, drops the multiply and rewrites the broadcast into a `rock.transform` — a coordinate remap rather than a copy. I traced that path by hand rather than assuming it, since no `rocmlir-opt` binary is built here: `mulBroadcast` inspects `getInput2()` for the constant and returns `getInput1()` as the real value, which is why the ones have to be the second operand; `isConstantOne` reaches `isSpecificValueAttribute`, whose `SplatElementsAttr` branch ends at `isExactlyValue(1.0)`, satisfied by a splat `dense<1.000000e+00>`. The integer case lands in the `IntegerAttr` branch, which accepts `1.0` because it floors to an exact `int64_t`.

**Rank extension** reuses what `BinaryConverter` already does: `EqualizeRanks` prepends 1s before the multiply, since TOSA broadcasts size-1 dimensions only once both operands carry the result's rank. The `shape` operand is never read — ONNX shape inference has already resolved the broadcast into the result type, which is also the only form available here, as a shape computed on the device is not visible to the compiler.

**One deliberate asymmetry.** `hip.expand` is marked *conditionally* illegal, unlike every other hip op this pass claims. `OnnxToHip` supports a dynamically shaped `Expand` on purpose, reading the extents back from the device with a stream sync, so that form is one the compiler deliberately produces and has to stay a `wrap_expand`. An unconditional `addIllegalOp` would turn it into a hard failure of the pass.

**Testing.** New `test_expand.mlir`, 8 cases: the same-rank and rank-extending broadcasts, rank-0 and integer sources, a broadcast feeding a matmul anchor in the shape `hip-fuse-rocmlir` produces, both dynamic forms passing through, and the non-kernel no-op. I checked it is not vacuous — mutating the expected ones constant from `1.0` to `7.0` makes it fail. Full lit suite is now 480 passing; the one failure remains `test_qadd.mlir`, which runs only `--convert-onnx-to-hip` and was failing before this branch.

The `FuseROCMlir` caveat from the description applies here too, and a little more sharply: `isaPointwiseOp` only absorbs an anchor's *consumers*, while an expand typically *feeds* the anchor, so making this fire on a real model needs producer-side fusion rather than just another entry in that list.

PR #951, opened 2026-09-09
feat(qdq): fuse DQ+Conv1x1+Q into a new hip.qconv op

amd-mounikk:
## Summary

- Adds `hip.qconv`, a fused quantized 1×1 convolution, plus the PDLL pattern, HipToLLVM lowering, runtime wrapper, and W4A16 GPU kernel.
- Matches `DequantizeLinear(activation) + DequantizeLinear(weights) -> Conv -> QuantizeLinear` and rewrites it to a single op.
- Unfused path today: dequantize activations to f32, expand 4-bit weights to f32, run a float conv, requantize — three full-precision temporaries per layer that nothing downstream reads.
- Weight expansion is the expensive part: a `[8192, 2048]` weight is 8 MiB packed (4-bit) and 64 MiB as f32, so most time is spent moving bytes that exist only to be multiplied and discarded.
- Fused path never materialises those temporaries: weights stay packed into the register, nibbles unpack in-register, accumulation is integer, quantized output comes from the epilogue.
- Geometry that makes this possible: 1×1 kernel, unit stride, unit dilation, no padding is one dot product per output position down the channel axis — a GEMM with `M = out_channels`, `K = in_channels`, `N = spatial_size`. No im2col, no per-axis geometry in the runtime.
- Developer-visible effect: matching subgraphs lower to one `wrap_qconv` call instead of five ops, and a `qconv` row appears in the `[PERF]` table.
- On `PSU_MF_LORA_128` all 196 convolutions fuse; end-to-end latency drops from 730 ms to 678–694 ms (gfx1151, Release, `hip-onnx-runner`).
- Non-matching convolutions keep the existing path. Change is additive: 19 files, +1,639/−0.

## Related issue or design

- None.

## Why

- Activation QDQ is per-tensor, so input/output scales and zero-points are scalars and fold into attributes.
- Host further folds them into `input_scale / output_scale` and `1 / output_scale` so the kernel never divides.
- Weight scales and zero-points are per output channel and cannot fold; on a real model they arrive as external constants whose values are not readable at fusion time, so they stay operands.
- Packing must be inferred, not read from the type: ONNX INT4/UINT4 imports as i8/ui8 at the logical element count, and this pass runs before `lowerOnnxConstants` stamps `packed_int4`.
- `IsPackedInt4PerChannelWeight` therefore uses the constant’s backing byte count (`ceil(numel/2)`), covering both inline dense values and the external location/offset/size form used by multi-megabyte weights.

## What

- `hip.qconv` in `HipOps.td` as a DPS op, plus `getDpsInitsMutable`/`getEffects` and bufferization registration.
- `QDQConvFusion.pdll` with three native constraints: `IsUint16Quantized`, `IsFusableQConvGeometry`, `IsPackedInt4PerChannelWeight`.
- `QConvLowering.cpp` → 19-parameter `wrap_qconv`; validates rank-4 NCHW, geometry, and element types.
- `wrap_qconv` in `real/qconv.cpp` plus the `mock_gpu.cpp` stub.
- `qconv_kernel.hip`: tiled path for `spatial_size > 8`, GEMV path otherwise.
- Lit tests for fusion and lowering.
- Scope (enforced by constraints): UINT16 per-tensor activations; per-output-channel packed INT4/UINT4 weights (`weight_axis` 0); 1×1 / unit stride / unit dilation / no padding / group 1.

## Test plan

- `pre-commit run --all-files --hook-stage manual` — all 9 hooks pass (also green on this PR).
- `ninja check-hip-mlir-lit` — 486 passed, 12 unsupported, 0 failed; both new `test_qconv.mlir` files execute (not skipped).
- Standalone harness vs independent CPU reference — 14 shapes: both dispatch paths, signed/unsigned weights, 4-bit and 8-bit storage, with/without bias, ragged extents, batch > 1. All bit-exact.
- Single-op ONNX vs ORT CPU (inputs identical between runs, `qconv` present in each EP log):
  - 2048 → 1024, P=1 (GEMV): 1,024 elems, 0 differing
  - 2048 → 1024, P=128 (tiled): 131,072 elems, 0 differing
  - 2048 → 8192, P=128: 1,048,576 elems, 8 differing, max 1 LSB
  - 8192 → 2048, P=128: 262,144 elems, 3 differing, max 1 LSB
- Nothing differs by more than 1 LSB (0.0015% of UINT16 range). Reference accumulates in f32; kernel accumulates integers and scales once, so a `.5` boundary can round either way.

## Performance

- `PSU_MF_LORA_128` end-to-end: 730 ms → 678–694 ms.
- Hardware / method: gfx1151, Release, real runtime, `hip-onnx-runner` on shipped inputs, vs the same build with the pattern not firing.

## Checklist


amd-mingw (approved):
Non-blocking: 1×1 scope makes sense for W4A16 LLM. Is there a tracked model/benchmark where 3×3 QDQ Conv is a hotspot on the unfused path? If not, I'd treat general conv QDQ fusion as a separate follow-up once we have a concrete target (vision W4 backbone vs INT8 CV).

PR #952, opened 2026-09-09
feat: pin LLVM to rocmlirTriton's patched 23.0.0git snapshot

pragyam32:
### RTTI stays on

Matching rocmlirTriton's `LLVM_ENABLE_RTTI=OFF` broke the Linux link with hundreds of `undefined reference to 'typeinfo for mlir::Pass'` / `mlir::ConversionPattern` / `mlir::Dialect`. hip-ep compiles its own libraries with RTTI on ELF, so their vtables reference typeinfo for the MLIR bases they derive from, and an `-fno-rtti` LLVM never emits those symbols. MSVC hid this because hip-ep already builds with `/GR-` on Windows.

`LLVM_ENABLE_RTTI=ON` is restored in `cmake/deps.cmake` and both workflows, matching what `main` already used. Only `test/plugin/sample_plugin` opts out of RTTI on ELF, and no-RTTI objects link against an RTTI LLVM without issue. The point of this PR is sharing rocmlirTriton's LLVM *source* pin; RTTI is a build-flag choice local to hip-ep.

`linux-build.yml` now saves `llvm-install` immediately after the LLVM build rather than relying on `actions/cache`'s post step, which is skipped when the job fails and therefore discarded a three-hour build. This mirrors the restore/save pattern `windows-deps.yml` already uses.

### Merge with main

`lib/Conversion/OnnxToHip/CMakeLists.txt` resolved in favour of `main`, which landed the same in-tree `mlir-pdll` fix this branch carried, in a better form. `cmake/deps.cmake` keeps both sides: `LLVM_ENABLE_DIA_SDK=OFF` from `main` plus `LLVM_ENABLE_TERMINFO=OFF` and `LLVM_ENABLE_ASSERTIONS=ON` from this branch.


amd-yanjunz:

- PR: ROCm/hip-ep#952
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/952#issuecomment-5613225361)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/881/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6aa24ed29a4cc5d9c2dd9c9a
- Compare with latest main branch results: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fids%3D6aa1b5b8c1fd62e7952f88b7%26collection%3DHIP-EP-Dynamic&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=881

PR #903, opened 2026-09-02
feat(matmul_nbits): offline autotune LUT, WMMA/GEMV tuning, and K-padding

ZhanxingPu:
## Summary

Gives the int4 MatMulNBits kernels an offline autotune lookup table (LUT): the
first call for a shape resolves its tile config from a table compiled into the
custom-kernels DLL instead of paying a runtime sweep. Alongside it, a series of
WMMA/GEMV tuning, coverage, and correctness fixes.

Runtime behavior is unchanged by default: the per-process tune cache is still
consulted first, then the LUT, then the sweep. The table only makes the *first*
encounter with a shape free.

## Related issue or design

In-tree design and status: `lib/Runtime/Kernels/hip/autotune/matmul_nbits/`
(`README.md` for the design, `PLAN.md` for the current state and remaining work).

## Why

The tuner benchmarks 29 GEMV or 68 WMMA candidate dispatches on the first call
for each unseen shape, and that cost lands in the middle of inference. The LUT
removes it for shapes measured offline; a log-space nearest-neighbour tier
covers unmeasured shapes without a runtime sweep. The other commits fix WMMA
tile selection and widen the paths that can be tuned at all.

## What

Offline LUT (`autotune/matmul_nbits/`):
- Two-tier lookup keyed on categorical fields (phase, bits, group size,
  zero-point, row stride) matched exactly, then the nearest measured point in
  log space over (M, N, K) -- Tensile-style, distance 0 == exact. FlatBuffers
  table, geometry stored rather than config indices (validated against the live
  tables at load, so a stale row is skipped, not launched).
- `update_lut.py` pipeline: measure -> fit weights -> build -> compile. Winners
  come from the tuner's own debug log; phantom (near-zero/negative) readings are
  dropped and near-ties broken toward the larger warp tile offline.
- Embedded into `custom_kernels_<arch>` via `cmake/xxd.py`; absent-table archs
  fall back to the sweep. `HIPDNN_MATMUL_AUTOTUNE_MODE=online` bypasses the
  table; `HIPDNN_MATMUL_LUT_LOG=1` traces lookups.

Kernel / dispatch:
- Widen the int4 WMMA warp tile and cut K-loop overhead.
- Padrow: route int4 WMMA prefill through a cache-line-padded B on
  aliasing-K shapes (K a multiple of 256), with the row stride a runtime arg.
- K-padding: K%32!=0 int4 prefill (e.g. down_proj with K=4304) now runs the
[...]

PR #953, opened 2026-09-09
test: matmul_nbits K-padding zero-point fix on a gemma4-capable OGA build (#933 + #599)

BoarQing:
**Test PR — not for merge.** Base is `perf/gqa-chunk-kv-span`, so the diff here is #599's workflow change plus the matmul_nbits fix; #933 carries the GQA kernel work and reaches CI through the base.

## What this stacks

| Component | Source | In the diff? |
|---|---|---|
| `perf(gqa)`: per-chunk key-range narrowing | #933, via base branch | no (base) |
| `ci(windows)`: gemma4-capable OGA build | #599, cherry-picked | yes |
| `fix(matmul_nbits)`: K-padding zero-points | new here | yes |

#599's two commits are cherry-picked unmodified with authorship intact; the resulting `.github/workflows/windows-deps.yml` is byte-identical to #599's — both are blob `94b2308c`. It repoints the OGA checkout from `microsoft/onnxruntime-genai@v0.14.0` to `amd-genmingz/onnxruntime-genai@test_0_3`, which carries the AMDGPU MorphiZen integration plus sliding-window support.

## The matmul_nbits fix

The nightly VLM run has been emitting fluent nonsense (`-y own-y own…`) on gemma-4-26B-A4B QMoE at every prefill length. That is not #933 and not #719. Bisected against CI's own `gpu-test-package` artifacts, vision session on the EP:

| build | date | output |
|---|---|---|
| `aaff3f91` | Sep 2 08:01 | correct, TTFT 16,120 ms |
| `1723fd36` | Sep 3 16:31 | correct, TTFT 26,216 ms |
| `13f44fb8` (#903) | Sep 4 08:10 | **garbage**, TTFT 1,595 ms |

`1723fd36` is `13f44fb8`'s direct parent, so it is that one commit — #903, "offline autotune LUT, WMMA/GEMV tuning, and K-padding". It made GPU vision ~20x faster and numerically wrong at the same time.

#903 added a WMMA branch for `K % 32 != 0` int4 prefill that zero-pads K and runs the ordinary kernel. That branch reads `zp_for_fp16_paths`, but the wrapper only built the unpacked fp16 zero-point buffer when `K % 32 == 0` — the exact complement of the new branch's predicate. So every asymmetric `K % 32 != 0` prefill reached it with `zp_for_fp16_paths` still pointing at the nibble-packed `zero_points` and read those bytes as fp16: wrong values, plus a read past the end (packed is `[N,(ngk+1)/2]` bytes, the branch indexes `[N,ngk]` fp16 — 78,336 against 311,040 at N=1152, ngk=135).

SigLIP-So400m's MLP is 1152→4304→1152, and `4304` is the only ragged dimension anywhere in this model: `4304 % 32 == 16`. The `down_proj` (M=2268 N=1152 K=4304, group_size 32) is therefore the single qualifying matmul, and every decoder matmul has `K % 32 == 0`. That is why the decoder stayed correct while the vision encoder garbled, deterministically, on every image.

Two changes:

- Build the fp16 zero-points for any `bits=4, batch==1, M>1` shape rather than only `K % 32 == 0`. This keeps K=4304 on the fast path instead of reverting to the naive kernel it used before #903.
- Gate both WMMA branches on `fp16_zp_ready`, the flag already guarding the col-major GEMV path. Absent the buffer they now decline and fall through to a kernel that reads the uint8 zero-points, rather than computing silently wrong results. The K-padding branch shipped this bug by inferring the buffer's presence from a coincidence between two gates instead of checking.

`mn_kpad_test` covered K=4304 and passed, because it only ever ran symmetric (`zero_points=nullptr`), so `has_zp` was false and the bad pointer was never dereferenced. It now runs every shape in three zero-point modes — symmetric, asym with only the uint8 buffer, asym with both — and adds the shipped M=2268 geometry. It remains a standalone dev tool, not wired into CMake.

## What this run will and will not show

**The matmul fix is directly observable here.** The `Run VLM benchmark` job iterates the persisted `amdgpu-vlm-models` directory with `dog.jpg`, and the publish step captures the generated text into the `Output` column of the `VLM Benchmark Results` comment. That column is where the `-y own-y own…` came from, so it is also where the fix shows up. Local result on gfx1151 with the CI config unchanged — vision session still on the EP with `profile: hip` and `disable_cpu_ep_fallback: 1`:

| prefill tokens | TTFT | output |
|---|---|---|
| 274 | 907 ms | coherent caption |
| 412 | 1,000 ms | coherent caption |
| 2,303 | 2,446 ms | coherent caption |
| 16,399 | 13,822 ms | coherent caption |

Under `HIPDNN_EP_DEBUG=1` the `down_proj` takes the K-pad path 27 times, once per SigLIP layer, with zero naive fallbacks — so the fast path is retained, not traded away for correctness. The vision-on-CPU arm is unchanged at 5,462 ms, confirming `K % 32 == 0` shapes were untouched.

**#933's perf claim is still not corroborated.** The VLM job runs at `--max_length 4096` and the default prompt, so roughly a 274-token prefill. #933's win is a 16K prefill: −50.5% TTFT (35,327.3 → 17,502.6 ms, interleaved and order-reversed against `main`), measured locally on gfx1151. Nothing in CI times a prefill at that length, and the mask-derived upper bound #933 fixes only matters where an image block straddles a query chunk boundary, which needs a long enough context to force small chunks. A green run here says nothing regressed; the numbers in #933 rest on the local measurements recorded there.

Note this corrects the earlier revision of this description, which claimed CI could not load the model at all. That was wrong: `.github` contains no literal `gemma` reference, but the VLM and OGA jobs iterate model directories staged on the GPU runner, which is precisely how the nightly hit this bug.

## Disposition

Nothing here should be merged. The workflow change belongs to #599 and the GQA kernel change to #933. The matmul_nbits fix is the one piece with no home yet — #903 is broken on `main` today, so any model whose intermediate size is not a multiple of 32 will garble on the vision path. Worth splitting into its own PR against `main` once this run confirms it.

Supersedes #941.

### DIFF
--- a/.github/workflows/windows-deps.yml
+++ b/.github/workflows/windows-deps.yml
@@ -587,6 +587,11 @@ jobs:
         if: steps.cache-oga.outputs.cache-hit != 'true'
         shell: bash
         run: |
+          # build.py imports tools/python/util, whose dependency_resolver
+          # imports requests at module scope. --ort_home means nothing is
+          # actually downloaded, but the import still has to resolve.
+          python -m pip install -q requests
+
           python "${{ github.workspace }}/source-oga/build.py" \
             --config Release \
             --cmake_generator Ninja \
--- a/.github/workflows/windows-deps.yml
+++ b/.github/workflows/windows-deps.yml
@@ -30,8 +30,14 @@ env:
   LLVM_TAG: llvmorg-22.1.0
   ONNXRUNTIME_VERSION: "1.27.0"
   ONNXRUNTIME_PR_PATCHES: ""
-  OGA_VERSION: "0.14.0"
-  OGA_PR_PATCHES: "2194"
+  # Fork branch carrying the AMDGPU MorphiZen integration (former PR 2194) plus
+  # sliding_window support, which upstream v0.14.0 does not have.
+  OGA_REPO: "amd-genmingz/onnxruntime-genai"
+  OGA_REF: "test_0_3"
+  # Optional space-separated microsoft/onnxruntime-genai PR numbers applied on
+  # top of the fork checkout via pull/<n>.patch + git am. Empty because the fork
+  # branch already carries the integration that 2194 used to supply.
+  OGA_PR_PATCHES: ""
   ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: "true"
   AMDGPU_EP_REF: "05eaa2a938f71e98b49bd1be3fb03720167be5cc"
   # Space-separated onnxruntime/onnxruntime-ep-amdgpu PR numbers applied on top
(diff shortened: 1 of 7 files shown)


BoarQing:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1774/
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/857/

PR #956, opened 2026-09-10
feat(ep): deliver provider options to runtime subsystems at session init

mingyueliuh:
## Summary

Introduce a generic reference-init configuration channel for passing EP provider options across the C ABI boundary to runtime subsystems.
 GQA autotune mode is the first consumer:

```text
HIPDNN_GQA_AUTOTUNE_MODE  >  gqa_autotune_mode provider option  >  build default
```

Behaviour is unchanged when neither lever is set, and the environment variable
keeps its existing precedence and parsing.

## Why

Runtime subsystems (e.g. GQA autotune) could only be configured via
environment variables. Provider options set through the ONNXRuntime API
had no path into the runtime.

## What

The EP publishes its provider-option map as a C accessor table; init copies the
entries onto `RuntimeState`, which outlives the call:

```text
MlirCustomOp                     builds a C accessor table over its option map
  -> inference_init              generated artifact entry point
  -> hipdnn_ep_state_init_with_fs
                                 copies the entries onto RuntimeState
  -> hip_gqa_autotune_create     receives the gqa_autotune_mode string
  -> chooseMode                  env var > provider option > build default
```

- **`include/hip/init_config_abi.h`** (new) -- the config ABI: an opaque `self`
  plus `provider_option_count` and `provider_option_at`. Pure C, because it
  crosses from the MSVC-built EP into Clang-compiled JIT'd model code. It is
  enumeration-only, and borrowed for the duration of the call.
- **`inference_init(out_state, fs, config)`** -- one pointer wider. The metadata
  blob address and size stay compiled in, so the public arity grows by one, not
  three. `config` may be null, which is how `hip-test` and any other caller
  without an ORT session invokes an artifact.
- **`RuntimeState::provider_options`** (new) -- the copy init makes.
  `hipdnn_ep_runtime_get_provider_option()` answers every later lookup, and
  unknown keys are kept for future consumers.
- **`chooseMode()`** -- weighs the two levers and the default in one place, and
  logs the decision behind the existing `HIPDNN_GQA_LUT_LOG` switch:

  ```text
  [gqa-autotune] mode=lookup (from the default) HIPDNN_GQA_AUTOTUNE_MODE=[unset], gqa_autotune_mode=[bogus]
  ```

A lever set to a value naming neither mode is skipped rather than honoured, so a
typo in the environment variable does not hide the provider option.

## Test plan

Built and installed RelWithDebInfo on Windows / gfx1151.
- The GQA and output-allocator E2E compile/execute tests pass. They run
  `hip-test`, so they cover the new three-argument `inference_init` end to end.
- On GPU through `test_onnx_runner`: a single-node
  `com.microsoft::GroupQueryAttention` decode model (fixed 256-slot KV cache,
  8 Q heads over 2 KV heads, `head_dim` 64), each case from a fresh EPContext
  cache. Every run returned 0.

  | provider option | env var | mode taken | reported source | observed behaviour |
  |---|---|---|---|---|
  | `lookup` | unset | lookup | `gqa_autotune_mode` | one table lookup, one config |
  | `online` | unset | online | `gqa_autotune_mode` | no table lookup, 8 candidates swept |
  | absent | unset | lookup | the default | one table lookup, one config |
  | `lookup` | `online` | online | `HIPDNN_GQA_AUTOTUNE_MODE` | no table lookup, 8 candidates swept |
  | `bogus` | unset | lookup | the default | one table lookup, one config |

## Notes for reviewers

- **Cached artifacts must be invalidated.** The failure mode is quiet: an
  artifact compiled before this declares the four-argument
  `hipdnn_ep_state_init_with_fs`, the JIT resolves that by name against the new
  five-argument definition, and the `config` pointer the runtime then
  dereferences is whatever the call left in that register.
- No backward-compatibility path is included. An earlier revision of this branch
  added a second entry point (`inference_init_v2`) beside the original; it was
  dropped once compatibility was confirmed not to be a requirement, since it
  cost two generated functions, two runtime symbols and a probe in the EP. The
  first five commits still describe that design and the last one replaces it, so
  the final state is best read from the diff rather than commit by commit.
- `parseMode()` returns `bool` through an out-param rather than
  `std::optional`. That translation unit is also compiled by `hipcc` as `-x hip`
  with no `-std=`, where MSVC's STL gates `<optional>` on `_HAS_CXX17`, which
  clang does not define there, so the header contributes nothing.
- `provider_option_at` indexes a `std::map` with `std::next`, O(n) per call and
  O(n^2) over the copy. With a handful of options, once per session, an index is
  not worth its weight.

## Checklist

## AI assistance

Written with Cursor (Claude Opus 5). The design decisions were directed and
reviewed by the author, and validated with the builds, LIT run, unit tests and
on-GPU matrix above.


amd-yanjunz:

- PR: ROCm/hip-ep#956
- Source comment: [models=all node=xconucstrhalo19](https://github.com/ROCm/hip-ep/pull/956#issuecomment-5646677549)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xconucstrhalo19
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1056/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6aa57ef4320ddd91343d11a5
- Compare with latest main branch results: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fids%3D6aa455b1b8c6c4d87e0673e4%26collection%3DHIP-EP-Dynamic&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=1056

amd-mingw:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1057/
http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6aa5760143a08cad3c3d111a

amd-mingw (approved):

PR #961, opened 2026-09-10
perf(allocator): pool large pinned buffers by capacity

BoarQing:
## What

Pool pinned host buffers above 16 MB, keyed by the buffer's **capacity** rather than by a fixed size class. Two files, allocator only.

Requests above 16 MB were allocated at their exact size and released straight back to the driver on every `Free`. That is the right call for a one-off transient, but it is the wrong one for the allocation that actually dominates: a KV tensor under `past_present_share_buffer=false`, which grows by one token's worth every decode step. No two steps ask for the same size, so an exact-size pool could never hit and every step paid a fresh page-pin per tensor.

Both Gemma-4 models set `past_present_share_buffer=false` because they package `max_length=262144`, at which a shared cache would be 96.0 GiB (12B) and 60.0 GiB (26B). They cannot use the shared-buffer path, so this is the path that has to work.

## Mechanism

- **Reuse by capacity.** A request takes the smallest retained buffer that covers it, within `capacity <= size + size/32`. This is what lets one buffer absorb a whole run of growing requests instead of one step each, and the bound stops a small request from consuming a buffer a much larger one will need.
- **Headroom only on observed growth.** Buffers are allocated at exactly the requested size until a request exceeds the largest size seen in its octave. The growth flag latches per octave, because every layer asks for the same size within one decode step and only the first of them sees an increase. Once latched, capacity is `size + size/64` page-rounded — continuing the size-class table's own geometry, whose steps go 1/2, 1/4, 1/16, 1/32 as buffers grow.
- **Outgrown capacities are dropped** as soon as the next one is allocated, so a growth transition costs one set of buffers rather than two.
- **Retention is bounded by peak outstanding bytes**, with LRU eviction — the allocator's own high-water mark, so the bound scales with the context the model is run at instead of with a fixed share of RAM.
- **`hipHostMalloc` failure releases the whole pool and retries once.** Failing while holding reclaimable pinned pages would be worse than not pooling at all.

## Results

gfx1151, 16384-token prompt, 128 generated. Two arm orders interleaved and averaged, each arm with its own `TEMP` so autotune and MLIR caches cannot leak between them. Peak is process peak working set, which on this APU covers both CPU and GPU since pinned pages are shared.

| model | TPS before | TPS after | peak before | peak after |
| --- | --- | --- | --- | --- |
| gemma-4-26B-A4B-it-qmoe | 13.25 | **35.03** (+164%) | 21284 MB | **12901 MB** (−39.4%) |
| gemma-4-12b-it-kquant | 19.14 | 19.17 (+0.2%) | 25641 MB | **20338 MB** (−20.7%) |

The 12B does not speed up and is not expected to: at 16 K it is bound by the 7.29 GB of decoder weights it reads per token, not by allocation. It still gives back 5.2 GB of peak memory.

**At a 24 K context the 26B no longer runs at all on `main`** — the driver refuses a 100872192-byte pin part way through decode and the run dies with `hipHostMalloc(Mapped|Coherent) failed ... out of memory`. Pooled, it completes at 31.52 tok/s and 18344 MB.

### Neutrality

On models that share their KV buffer there is no churn for a pool to remove, so the only thing an allocator change can do is cost memory. Peak memory delta vs `main`:

| model | context | Δ peak |
| --- | --- | --- |
| DeepSeek-R1-Distill-Llama-70B | 16384 | −0.09% |
| Llama-3.1-8B-awq | 16384 | 0.00% |
| Mistral-7B-Instruct-v0.3 | 16384 | +0.09% |
| Qwen3.6-35B-A3B | 16384 | −0.15% |
| Qwen3.8-27B k_quant | 16384 | +0.16% |
| gpt-oss-20b | 4096 | +0.02% |
| qwen3-30b-a3b | 2048 | −0.00% |

TPS on these ranges −0.10% to +1.91%. The 70B swings ±6% between runs, but in both arm orders the arm that ran *first* got 5.25 tok/s and the arm that ran *second* got 4.93 regardless of which arm it was — it pages from disk (119 s TTFT at 5.5 GB resident), so that is position, not the allocator. `gemma3-4b-it` is not covered: it fails identically on both arms with an `image_features` input-rank error, unrelated to this change.

### Mechanism verification

Measured with temporary instrumentation (not in this diff):

| | buffers reused | driver calls | distinct capacities | peak retained |
| --- | --- | --- | --- | --- |
| Mistral-7B @ 16 K (shared KV) | 64 | 64 | 1, at the exact requested size | 2064 MB |
| Gemma-4 26B @ 16 K | 15180 | 182 | 7 | 7240 MB |
| Gemma-4 26B @ 24 K | 15180 | 182 | 7 | 10858 MB |

Mistral never rounds a buffer — the growth flag never latches, so a fixed-shape model pays no waste. The 26B reaches a 98.8% reuse rate, and its 182 driver calls across 60 KV tensors is exactly the 3 per tensor the design predicts: one exact-size probe plus a ping-pong pair. Both numbers are **identical at 24 K**, so the driver-call count does not grow with context, while retention scales exactly 1.5× with the 1.5× context.

### DIFF
--- a/morphizen/ort-bridge/src/morphizen-hip-gpu-allocator.cpp
+++ b/morphizen/ort-bridge/src/morphizen-hip-gpu-allocator.cpp
@@ -75,6 +75,60 @@ int SizeClassIndex(size_t size) noexcept {
   return -1;
 }
 
+constexpr size_t kPageSize = 4096;
+
+// Capacity for a large request that is known to be growing. The 1/64 headroom
+// is the reuse span: a Gemma-4 KV tensor is 64 MiB at a 16 K context and grows
+// 4096 bytes per token, so one capacity covers the next 256 tokens.
+//
+// Continuing the size-class table's geometry, whose steps go 1/2, 1/4, 1/16,
+// 1/32 as buffers grow, because the waste is proportional to the size.
+constexpr size_t LargeCapacity(size_t size) noexcept {
+  const size_t want = size + size / 64;
+  return (want + kPageSize - 1) / kPageSize * kPageSize;
+}
+
+// Checked at compile time because no unit-test target covers this file and
+// returning less than the request would overflow the buffer.
+constexpr bool LargeCapacityHolds() {
+  for (size_t n = 16ull << 20; n <= 16ull << 30; n += n / 8) {
+    const size_t c = LargeCapacity(n);
+    if (c < n + n / 64 || c > n + n / 64 + kPageSize) {
+      return false;
+    }
+  }
+  return true;
+}
+static_assert(LargeCapacityHolds(),
+              "LargeCapacity must cover the request with its growth headroom "
+              "and waste at most one page beyond it");
+static_assert(LargeCapacity(64ull << 20) - (64ull << 20) >= 128 * 4096,
+              "one capacity must span more than a 128-token generation at a "
+              "16 K context");
+
+int OctaveOf(size_t size) noexcept {
+  int octave = 0;
+  while (size > 1) {
+    size >>= 1;
+    ++octave;
+  }
+  return octave;
+}
+
+// Hand a batch of buffers back to the driver with pool_mutex_ released:
+// hipHostFree unpins pages and is as heavyweight as the hipHostMalloc that
+// pinned them, so holding the lock would serialize every other allocation
+// behind it.
+void ReleaseToDriver(const std::vector<void *> &buffers, int device_id) {
+  if (buffers.empty()) {
+    return;
+  }
+  ScopedDevice _(device_id);
+  for (void *p : buffers) {
+    (void)hipHostFree(p);
+  }
+}
+
 int TryGetDeviceId(const OrtMemoryInfo *memory_info) noexcept {
   if (memory_info == nullptr) {
     return -1;
@@ -107,6 +161,62 @@ HipGpuAllocator::HipGpuAllocator(const OrtMemoryInfo *memory_info,
   GetStats = nullptr;
 }
 
+// A capacity the model has just grown past cannot serve this request or any
+// later one, so its idle buffers are dead weight. Bounded to the immediate
+// neighbourhood below `size` so an unrelated, much smaller tensor pooled in
+// the same octave is left alone.
+void HipGpuAllocator::DropOutgrown(size_t size, std::vector<void *> &out) {
+  auto it = large_free_.lower_bound(size - size / 16);
+  while (it != large_free_.end() && it->first < size) {
+    for (void *ptr : it->second.free) {
+      ptr_to_size_.erase(ptr);
+      large_retained_bytes_ -= it->first;
+      out.push_back(ptr);
+    }
+    it = large_free_.erase(it);
+  }
+}
+
+// One buffer from the least-recently-used capacity, skipping `keep` because
+// evicting there would free a buffer of exactly the size being made room for.
+// Buffers nothing can ask for again are never touched, so they are always the
+// oldest and go first.
+bool HipGpuAllocator::EvictLruLarge(size_t keep, std::vector<void *> &out) {
+  auto victim = large_free_.end();
+  for (auto it = large_free_.begin(); it != large_free_.end(); ++it) {
+    if (it->first == keep) {
+      continue;
+    }
+    if (victim == large_free_.end() ||
+        it->second.last_used < victim->second.last_used) {
+      victim = it;
+    }
+  }
+  if (victim == large_free_.end()) {
+    return false;
+  }
+  void *ptr = victim->second.free.back();
+  victim->second.free.pop_back();
+  ptr_to_size_.erase(ptr);
+  large_retained_bytes_ -= victim->first;
+  out.push_back(ptr);
+  if (victim->second.free.empty()) {
+    large_free_.erase(victim);
+  }
+  return true;
+}
+
+void HipGpuAllocator::DrainLarge(std::vector<void *> &out) {
+  for (auto &entry : large_free_) {
+    for (void *ptr : entry.second.free) {
+      ptr_to_size_.erase(ptr);
+      out.push_back(ptr);
+    }
+  }
+  large_free_.clear();
+  large_retained_bytes_ = 0;
+}
+
 void *ORT_API_CALL HipGpuAllocator::AllocImpl(OrtAllocator *this_,
                                               size_t size) {
   if (size == 0) {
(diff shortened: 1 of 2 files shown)


amd-yanjunz:

- PR: ROCm/hip-ep#961
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/961#issuecomment-5616573952)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/896/

BoarQing:
Gemma Perf: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F891%2F&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F893%2F&curr_job=&curr=

BoarQing:
Force-pushed a revision. The first version pooled by rounded capacity and
capped retention, which was enough for the 30-layer decoder it was measured on
and wrong in general. Testing the combined stack (#962) on a second model
showed why.

At 16384 prompt / 128 output, build 893 against build 891, which differ by
exactly this commit:

| | 26B-A4B (30 layers) | 12B (48 layers) |
| --- | ---: | ---: |
| TPS | 18.74 -> 36.03 | 19.31 -> 19.32 |
| per-token std, ms | 36.57 -> 0.30 | 0.74 -> 0.95 |
| peak host memory, GB | 12.54 -> 15.28 | 13.05 -> 20.97 |

The 12B gained nothing and paid 7.9 GB. Cause: a capacity is only reusable
while something keeps asking for it, and a monotonically growing KV tensor
eventually outgrows one and never asks again, stranding a full set of buffers
at the capacity it left behind. Nothing released them. The 12B has 96 large KV
tensors per step against the 26B's 60, so its stranded 5.0 GiB plus its 5.9 GiB
live set exceeded the cap: the pool held memory it could not reuse and kept
making the driver calls it exists to remove. The 26B stayed under the cap and
hid the flaw entirely.

Worse, the stranding is not a rare alignment. A KV tensor's bytes per token is
a power of two and benchmark prompt lengths are too, so a long-context run
starts exactly on an octave edge, where the rounding step doubles, and is
guaranteed to cross it.

Three changes, all of them generic rather than shaped to these models:

- `LargeCapacity` rounds up the request plus a 1/64 growth headroom, so a
  request landing just under a power of two takes the class above it and a
  whole generation of growth shares one capacity. Worst-case waste goes from
  1/8 to 1/8 + 1/64.
- Reaching the retention cap now evicts the least-recently-used capacity
  instead of refusing to retain the incoming buffer, so the buffers that
  survive are the ones still in rotation. The cap is no longer a cliff, and the
  one-shot warning now fires only when eviction cannot make room, which is the
  genuinely bad state.
- Capacities untouched for two decode steps' worth of allocator traffic are
  released. The threshold is derived from the allocator's own traffic (a step
  is about two large operations per live buffer), not configured.

The invariant checks were model-specific too -- three hand-picked KV tensor
sizes -- and are now a constexpr sweep over every octave the large pool can
reach, asserting that capacity covers the request, that waste stays bounded,
that the headroom survives rounding, and that capacity is monotonic. One worked
example is kept as a regression check for the exact alignment that caused this.

The pool also logs one summary line at teardown, because none of the above is
visible from throughput and diagnosing it previously meant patching
instrumentation into a build. It sits at INFO normally and is raised to WARNING
when the pool has given up, so the counters land next to the warning that says
why.

Local re-measurement on the 26B at 16K, 128 tokens: 27.85 ms/token mean, 27.87
p50, 28.18 p90, std 0.28, with the pool reporting

```
large-buffer pool: 15240 reuses, 122 driver allocations, 4 capacity classes
(2 live at teardown), 0 cap evictions, 2 stale trims,
peak retained 7920 MB of a 8146 MB cap
```

A 99.2 percent hit rate, and two of the four capacities released by the trim
rather than held for the session. Short context is unchanged at 19.39 ms/token.

Expectation for the 12B is honest rather than optimistic: the stranded set goes
away, but one live KV set is what the pool *is*, so its peak memory will land
between the 13.05 GB baseline and the 20.97 GB this PR previously cost. Its
throughput may not move much either -- at roughly 141 GB/s for 7.29 GB of
decoder weights per token it is close to bandwidth-bound, and unlike the 26B it
never showed the stall pattern this removes. CI on #962 will show both.

BoarQing:
Gemma Perf: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F946%2F&base_job=daily_hip_ep_oga&base=946&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F961%2F&curr_job=daily_hip_ep_genai&curr=961

amd-yanjunz:

- PR: ROCm/hip-ep#961
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/961#issuecomment-5628869407)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/966/

amd-yanjunz:

- PR: ROCm/hip-ep#961
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/961#issuecomment-5633731638)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1030/

amd-yanjunz:

- PR: ROCm/hip-ep#961
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/961#issuecomment-5643360630)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1048/

amd-yanjunz:

- PR: ROCm/hip-ep#961
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/961#issuecomment-5644046806)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1050/

BoarQing:
Gemma Perf: 
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F1044%2F&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F1052%2F&curr_job=&curr=

Gemma Accuracy:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1780/

General Perf & Acc:
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F1037&base_job=&base=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F1053&curr_job=&curr=

http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1053/

amd-mingw (approved):
LGTM

PR #969, opened 2026-09-10
Initial rocMLIR/rocMLIRTriton integration

ilango100:
## Summary

Adds an end-to-end path to compile a fused ONNX GEMM (+ elementwise) subgraph through rocMLIR/Triton into a launchable GPU kernel embedded in the model's LLVM bitcode.

New pieces:
- A **HIP→TOSA conversion** (`convert-hip-to-tosa`) and a **`hip-fuse-rocmlir`** fusion pass that outlines a fused GEMM into a `rock.kernel` function behind a new **`hip.rocmlir`** dispatch op.
- A new **`hip-rocmlir-compiler`** tool that drives the ONNX→HIP head, hands the TOSA kernel to `librockCompiler.so` (rocMLIR high-level + backend pipelines) via `dlopen`, embeds the compiled GPU binary back into `hip.rocmlir`, runs the ONNX→HIP tail, lowers to LLVM, and emits OS-portable bitcode (`-o <out.bc>`).
- A **HIP→LLVM lowering** for `hip.rocmlir` plus a **`wrap_rocmlir` runtime wrapper** that loads and launches the embedded kernel with `hipModule*`, including op-profiling support.

Developer-visible: `hip-rocmlir-compiler model.onnx.mlir -o model.bc` produces a `.bc` that `hip-test`/the EP can run, with the GEMM offloaded to a rocMLIR-generated kernel.

## Related issue or design

None (feature series; see per-commit history on `feat/rocmlirtriton_fusion`).

## Why

rocMLIR/Triton generates high-performance fused GEMM kernels, but hip-ep had no way to route ops to it. `librockCompiler.so` statically embeds its own LLVM/MLIR copy, so it cannot share an `MLIRContext` with the tool (dialect/TypeID collisions). The tool therefore crosses the boundary only through serialized IR text and the `.so`'s exported MLIR C-API, keeping the two LLVM copies isolated. The compiled kernel is embedded inline in `hip.rocmlir` so the resulting `.bc` stays self-contained and consumable by the existing JIT/native loaders.

## What

- **Dialect/ops**: `hip.rocmlir` op (`kernel` symbol, ctx, variadic inputs, DPS output, `kernel_binary`/`grid_size`/`block_size` attrs); DPS bufferization model registered.
- **Passes**: `hip-fuse-rocmlir` (matches Conv/Gemm + pointwise, sets `rock.kernel`/`rock.arch`), `convert-hip-to-tosa` (add/mul/unary/binary + matmul, collapsing batched A into M for TOSA's equal-batch matmul), `rocmlir-pipeline`. `buildOnnxToHipPipelineTail` was made public so tools can reuse the exact tail ordering.
- **`hip-rocmlir-compiler`**: dlopen/dlsym of the rocMLIR C-API; perfConfig search-space enumeration + affix (or `--perf-config <str>`); backend knobs read from the chosen perfConfig; `--dump-high-level` to stop after the high-level pipeline and write rock MLIR; `-o` bitcode emission mirroring `hip-compiler`. `ROCK_COMPILER_SO`/`ROCK_ARCH` env overrides.
- **HIP→LLVM**: `RocMlirLowering.cpp` emits the kernel binary + name as globals, stages operand pointers into a kernargs buffer, and calls `wrap_rocmlir(state, binary, name, block, grid, kernargs, size)`.
- **Runtime**: `wrap_rocmlir` (real) caches `hipModuleLoadData`/`hipModuleGetFunction` per binary and launches via `hipModuleLaunchKernel` with the packed kernargs "extra" config; clears/checks `hipGetLastError` around the launch; `OP_PROFILE("rocmlir", ...)` scope. Mock wrapper is an empty stub.

## Test plan

- `cmake --build build` — compiler, tools, and runtime bitcode build clean.
- LIT: added `test/lit/Conversion/hip-to-tosa/{test_unary,test_binary,test_matmul}.mlir`; verified `convert-hip-to-tosa` output.
- `hip-rocmlir-compiler matmul.mlir -o out.bc` → valid LLVM bitcode (`file`, `llvm-dis`), contains `inference_init/compute/cleanup`, `@rocMlir0_binary`, and the `wrap_rocmlir` call with correct kernargs (`grid=704512, block=64, size=24`); `hip-inspect` reads metadata/constants.
- `--dump-high-level` writes `rock.gemm`/`rock.transform` MLIR; `--perf-config` affixes a supplied config (invalid configs correctly rejected by the backend).
- Verified the embedded `runtime.bc` in `hip-test` carries the instrumented `wrap_rocmlir` (disassembled).

## Notes for reviewers

- **Boundary contract**: the tool and `librockCompiler.so` each link their own LLVM/MLIR; only serialized IR crosses. `dlopen` uses `RTLD_LOCAL` so the `.so`'s symbols can't interpose the tool's MLIR. Do not pass live MLIR objects across.
- **Ordering**: `hip-use-output-allocator` must stay before pooling; the tool keeps the module in hip form after `fuse-rocmlir`, compiles a clone, then embeds + runs the standard tail.
- **`wrap_rocmlir` uses the standard `hipModuleLaunchKernel`** (grid = blocks, block = threads/block), matching rocMLIR's `mlirGetKernelAttrs` = `{block, grid, cluster}`.
- Runtime/model-cache invariant: rebuild runtime bitcode and invalidate cached models when validating.

## Checklist


asparkhi (approved):

amd-mingw:
Hi @ilango100 ,
Could you share more context on this PR?
(1)Background / motivation
(2)Problems it aims to solve
(3)Future roadmap
(4)New dependencies (build vs runtime)
(5)Expected impact on release package size
Thanks! cc @asparkhi

asparkhi:
Hi @amd-mingw 

Background/Motivation: RMT is used by MIGraphX. It is a kernel generator that can produce Triton kernels on-the-fly. This saves us from the need to use HIP custom kernels. Its performance is expected to be better than HIP custom kernels.
 
Future roadmap: Presently it supports 4 major ops: conv, matmul, attention, GEMM. In the future, it is expected to  be the default path in HIP-EP.
 
New dependencies: Triton-windows (LLVM23)

Another PR is pipeline to move HIP-EP to this version of LLVM. @amd-mounikk  can share the PR.

Release size: There is no change in the package size due to this PR as the rocmlirTriton's shared library is accessed with dlopen in this PR. We expect the hipgpu.dll size to grow by at least 40MB when this library starts getting built along with hipgpu.dll.

asparkhi:
This is how the final integration of RMT into HIP-EP will look like:
RMTIntoHIPEP

amd-mingw:
@asparkhi  “ We expect the hipgpu.dll size to grow by at least 40MB when this library starts getting built along with hipgpu.dll.”
----the size increase is the main concern, in the 11D release, reduce size is the top-1 task, and we do a lot of work/experiment to reduce the size, include hipgpu.dll. you should consider it when merge this solution.

asparkhi:
@amd-mingw  we have a follow up PR which will build RMT under an environment variable only. So, it will be an experimental flow until proven.

PR #957, opened 2026-09-10
feat(matmul_nbits): support fp32 A/output via device-side fp16 cast for u2/u3/u4

ZhanxingPu:
## Summary

MatMulNBits kernels (u2/u3/u4) previously required fp16 activations. This PR
adds transparent fp32 input/output support by inserting device-side cast
buffers around the existing fp16 dispatch path.

### Changes

**`matmul_nbits_kernel.hip`**
- When `element_size_bytes == 4` and `bits ∈ {2, 3, 4}`, allocate scratch
  buffers and launch a vectorised `float2→half2` cast kernel before the
  existing fp16 dispatch, then cast `C` back to fp32 afterwards.
- Scratch buffers are grown/reused from existing allocations; no extra
  `hipMalloc` per call.
- fp16 fast-path is unchanged. `bits=8` still rejects non-fp16.

**`test/example/gemm_fp16u2,u3,u4`**
- Add `--fp32` flag / `test_fp32` target; validates fp32 output against the
  dequant-to-fp16 reference.
- u3 example adds a stub `resolve()` so the kernel links without FlatBuffers.

## Test plan

- [ ] `make test_fp32` passes in gemm_fp16u2, gemm_fp16u3, gemm_fp16u4 examples
- [ ] Existing `make test` (fp16 path) unaffected
- [ ] Checked that `bits=8` fp32 input is still rejected

PR #971, opened 2026-09-10
feat(matmul_nbits): support fp32 A/output for bits=8 via the shared fp16 cast path

ZhanxingPu:
## Summary
Extends fp32 A / fp32 output support to `bits=8` MatMulNBits, reusing the
device-side fp16 cast path introduced for u2/u3/u4 in #957. int8 MatMulNBits
now accepts `element_size_bytes==4` (fp32) inputs/outputs, previously only
fp16 was allowed for bits=8.

## Why
#957 added fp32 A/C for bits=2/3/4 by casting A→fp16, running the unchanged
fp16 compute path, and casting the result back to fp32, but its dispatch was
gated `element_size_bytes==4 && bits != 8`, explicitly excluding int8. So
bits=8 with fp32 activations fell through to the "only fp16 supported" error.
Nothing in the cast path is bits-specific, so int8 only needed to be let in.

Overwriting the activation buffer in place (to avoid the fp16 scratch) was
considered and rejected: in the target model 55% of MatMulNBits activations
are shared by multiple consumers (QKV projections share the input LayerNorm
output; gate/up projections share the pre-feedforward LayerNorm output), so an
in-place fp32→fp16 cast would corrupt the peer ops reading the same tensor.

## What
- `matmul_nbits_kernel.hip`: drop `&& bits != 8` from the fp32 cast-wrapper
  gate (`element_size_bytes == 4`). fp32 bits=8 now casts A→fp16, calls the
  unchanged fp16 int8 dispatch (WMMA/GEMV/naive), and casts the fp16 result
  back to fp32. Reuses the existing `g_fp32cast_a/c` scratch, so the int8
  kernels stay fp16-only — no element-type template, no code-size growth.
- Add `gemm_fp16i8/`: standalone HIP-SDK test (Makefile, data generator,
  no-LUT autotune stub, GPU-vs-CPU driver with a `--fp32` mode).

## Test plan
On gfx1151:
- **i8**: fp16 (regression) and fp32, prefill (WMMA) and decode (GEMV),
  with and without zero_points — all pass.
- **u2/u3/u4 regression**: the change is a no-op for bits 2/3/4 (`bits != 8`
  is always true there, so the gate is unchanged); spot-checked u4 fp16/fp32
  to confirm no regression.

## AI assistance
Investigation, the one-line gate change, and the test program were developed
with AI assistance (Cursor) and reviewed by the author; commit carries the
`Co-Authored-By` / `Made-with` trailers per CONTRIBUTING.md.

### DIFF
--- a/lib/Runtime/Kernels/hip/matmul_nbits_kernel.hip
+++ b/lib/Runtime/Kernels/hip/matmul_nbits_kernel.hip
@@ -6193,9 +6193,13 @@ extern "C" void hip_matmul_nbits_dequant_b_fp16(
 }
 
 /* -----------------------------------------------------------------------
- * fp32 A / fp32 output cast patch (element_size_bytes==4, bits != 8):
+ * fp32 A / fp32 output cast patch (element_size_bytes==4, all bits incl. 8):
  * device-side, vectorized float<->fp16 conversion so the existing fp16
- * compute path (WMMA/GEMV/naive below) runs completely unchanged.
+ * compute path (WMMA/GEMV/naive below, and the bits=8 dispatch) runs
+ * completely unchanged. Trading a persistent (cached, grow-on-demand) fp16
+ * scratch pair for zero per-kernel code duplication -- fp32 is a cold input
+ * type, so this keeps the binary small instead of templating every hot
+ * int8/int4 kernel on the element type.
  * ----------------------------------------------------------------------- */
 
 __global__ void cast_fp32_to_fp16_kernel(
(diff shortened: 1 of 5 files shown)


amd-yanjunz:

- PR: ROCm/hip-ep#971
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/971#issuecomment-5618478286)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/907/
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6aa307b25412e2e397397b78
- Compare with latest main branch results: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fids%3D6aa1b5b8c1fd62e7952f88b7%26collection%3DHIP-EP-Dynamic&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=907

amd-mingw (approved):

PR #977, opened 2026-09-11
refactor(hipsr): inline region ops after pool allocation

zpye:
## Summary
- Flatten `hipsr.pool_domain` and `hipsr.compute` after pool allocation so later module passes see one function block.
- Keep pool planning inside each domain; inlining only removes the region boundary afterward.
- Update pipeline goldens so they no longer expect `hipsr.compute` / `hipsr.pool_domain` after `--hipsr-pipeline`.


wcy123 (approved):
LGTM. Simple 68-line pass that flattens `hipsr.pool_domain` and `hipsr.compute` regions after pool allocation is complete. Uses PostOrder walk to inline from inside out.

PR #982, opened 2026-09-11
feat(hipsr): free pipeline-owned host buffers

zpye:
## Summary
- Run `--buffer-deallocation-pipeline` and `--optimize-allocation-liveness` after `hipsr-inline-regions` so host shape buffers are freed after their last use.
- Memory effects:
  - Hipsr DPS ops report a memref read on each input and a memref write on each destination. Tensor operands report none, so unused tensor copies are dead.
  - `hipsr.alloc_output` reports `Allocate` on its result so a returned output stays owned and dealloc does not clone it.
  - `hipsr.get_pool` stamps `bufferization.manual_deallocation` so `--ownership-based-buffer-deallocation` does not free the pool.


wcy123 (approved):
LGTM.

PR #987, opened 2026-09-11
feat(hip-to-tosa): convert hip.gemm to tosa.matmul + tosa.add

nvunnam57128:
`hip.gemm` is a registered fusion anchor in `FuseROCMlir`, but it had no TOSA
converter and no entry in `HipToTosa`'s illegal-op list. Since the pass uses
`applyPartialConversion`, an op that is neither listed as illegal nor matched by
a pattern is simply treated as legal and passed through without a diagnostic.

So `FuseROCMlir` outlined the gemm into a `rock.kernel`, `convert-hip-to-tosa`
converted everything around it and left the gemm alone, and rocMLIR was then
handed a hip dialect op it has never seen. It died with `0xC0000409`
(`STATUS_STACK_BUFFER_OVERRUN`) — no error message, no source location. Any
model containing an ONNX Gemm hit this.

The unconverted op reaching the kernel boundary looked like:

```mlir
func.func @rocMlir0(...) attributes {rock.arch = "gfx1151", rock.kernel} {
  %2 = hip.gemm(%0) ins(%arg0, %arg1, %arg2 : ...) outs(%1 : ...)
  %3 = tosa.sigmoid %2
  return %3
}
```

The sigmoid converted; the gemm did not.

## What this does

Expands `hip.gemm` the way ONNX defines it — `Y = alpha * A' * B' + beta * C`:

- optional `transA` / `transB` become `tosa.transpose`
- `tosa.matmul` is batched, so a batch of one is wrapped around the rank-2
  operands and stripped afterwards
- `alpha` and `beta` become splat `tosa.mul`s, elided entirely when they are
  1.0, which is the common case
- `C` is left-padded with ones to the result's rank, because TOSA expands only
  size-1 dimensions while ONNX broadcasts unidirectionally

The same kernel now converts to:

```mlir
%7  = tosa.matmul %5, %6, %2, %2 : (tensor<1x32x64xf32>, tensor<1x64x128xf32>, ...) -> tensor<1x32x128xf32>
%8  = tosa.reshape %7, %1 : (tensor<1x32x128xf32>, !tosa.shape<2>) -> tensor<32x128xf32>
%9  = tosa.reshape %arg2, %0 : (tensor<128xf32>, !tosa.shape<2>) -> tensor<1x128xf32>
%10 = tosa.add %8, %9 : (tensor<32x128xf32>, tensor<1x128xf32>) -> tensor<32x128xf32>
%11 = tosa.sigmoid %10 : (tensor<32x128xf32>) -> tensor<32x128xf32>
```

Expanding into separate ops costs nothing downstream: MIGraphX's ONNX parser
desugars `Gemm` into exactly the same dot plus broadcast add, and on gfx1151
rocMLIR fuses that chain back into a single kernel (`mlir_dot_add_sigmoid`).

Integer gemm is rejected rather than converted. `alpha` and `beta` are f32
attributes applied as elementwise multiplies in the result type, so an integer
result would round them away silently; failing legalization is the honest
outcome.

## Validation

- New `test/lit/Conversion/hip-to-tosa/test_gemm.mlir` covers the basic form,
  omitted `C`, `transA`, `transB`, non-unit alpha/beta, a full-rank `C`, a
  column-vector `C` broadcasting along N, the non-kernel no-op, and rejections
  for dynamic shapes, integer element types, and a non-broadcastable `C`.
- Full lit suite passes (the one failure, `onnx-to-hip/test_qadd.mlir`, is
  pre-existing on this base and unrelated — its RUN line never touches
  `HipToTosa`).
- End to end through `hip-rocmlir-compiler` against `librockCompiler`, the same
  model that previously crashed now compiles to a 5288-byte GPU binary
  (grid 16, block 64).

### DIFF
--- a/lib/Conversion/HipToTosa/HipToTosa.cpp
+++ b/lib/Conversion/HipToTosa/HipToTosa.cpp
@@ -167,6 +167,123 @@ Value createSplatFloat(ConversionPatternRewriter &rewriter, Location loc,
       DenseElementsAttr::get(type, rewriter.getFloatAttr(elemType, ap)));
 }
 
+// Swap the two dimensions of a rank-2 tensor.
+static Value transpose2D(Value input, ConversionPatternRewriter &rewriter,
+                         Location loc) {
+  auto type = cast<RankedTensorType>(input.getType());
+  ArrayRef<int64_t> shape = type.getShape();
+  return tosa::TransposeOp::create(rewriter, loc,
+                                   type.clone({shape[1], shape[0]}), input,
+                                   rewriter.getDenseI32ArrayAttr({1, 0}));
+}
+
+// hip.gemm is ONNX Gemm: Y = alpha * A' * B' + beta * C, where A and B are
+// optionally transposed and C broadcasts to [M, N]. TOSA has no fused
+// equivalent, so this spells the whole thing out. MIGraphX's ONNX parser
+// desugars Gemm the same way -- into a dot plus a broadcast add -- and rocMLIR
+// fuses the resulting chain back into one kernel, so the expansion costs
+// nothing downstream.
+struct GemmConverter final : public OpConversionPattern<hip::GemmOp> {
+  using OpConversionPattern<hip::GemmOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(hip::GemmOp op, OpAdaptor adaptor,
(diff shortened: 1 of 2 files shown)


nvunnam57128:
Superseded: reopening from a branch on ROCm/hip-ep rather than a fork, matching the convention used by the other hip2tosa branches.

PR #988, opened 2026-09-11
feat(hip-to-tosa): convert hip.gemm to tosa.matmul + tosa.add

nvunnam57128:
﻿`hip.gemm` is a registered fusion anchor in `FuseROCMlir`, but it had no TOSA
converter and no entry in `HipToTosa`'s illegal-op list. Since the pass uses
`applyPartialConversion`, an op that is neither listed as illegal nor matched by
a pattern is simply treated as legal and passed through without a diagnostic.

So `FuseROCMlir` outlined the gemm into a `rock.kernel`, `convert-hip-to-tosa`
converted everything around it and left the gemm alone, and rocMLIR was then
handed a hip dialect op it has never seen. It died with `0xC0000409`
(`STATUS_STACK_BUFFER_OVERRUN`) ΓÇö no error message, no source location. Any
model containing an ONNX Gemm hit this.

The unconverted op reaching the kernel boundary looked like:

```mlir
func.func @rocMlir0(...) attributes {rock.arch = "gfx1151", rock.kernel} {
  %2 = hip.gemm(%0) ins(%arg0, %arg1, %arg2 : ...) outs(%1 : ...)
  %3 = tosa.sigmoid %2
  return %3
}
```

The sigmoid converted; the gemm did not.

## What this does

Expands `hip.gemm` the way ONNX defines it ΓÇö `Y = alpha * A' * B' + beta * C`:

- optional `transA` / `transB` become `tosa.transpose`
- `tosa.matmul` is batched, so a batch of one is wrapped around the rank-2
  operands and stripped afterwards
- `alpha` and `beta` become splat `tosa.mul`s, elided entirely when they are
  1.0, which is the common case
- `C` is left-padded with ones to the result's rank, because TOSA expands only
  size-1 dimensions while ONNX broadcasts unidirectionally

The same kernel now converts to:

```mlir
%7  = tosa.matmul %5, %6, %2, %2 : (tensor<1x32x64xf32>, tensor<1x64x128xf32>, ...) -> tensor<1x32x128xf32>
%8  = tosa.reshape %7, %1 : (tensor<1x32x128xf32>, !tosa.shape<2>) -> tensor<32x128xf32>
%9  = tosa.reshape %arg2, %0 : (tensor<128xf32>, !tosa.shape<2>) -> tensor<1x128xf32>
%10 = tosa.add %8, %9 : (tensor<32x128xf32>, tensor<1x128xf32>) -> tensor<32x128xf32>
%11 = tosa.sigmoid %10 : (tensor<32x128xf32>) -> tensor<32x128xf32>
```

Expanding into separate ops costs nothing downstream: MIGraphX's ONNX parser
desugars `Gemm` into exactly the same dot plus broadcast add, and on gfx1151
rocMLIR fuses that chain back into a single kernel (`mlir_dot_add_sigmoid`).

Integer gemm is rejected rather than converted. `alpha` and `beta` are f32
attributes applied as elementwise multiplies in the result type, so an integer
result would round them away silently; failing legalization is the honest
outcome.

## Validation

- New `test/lit/Conversion/hip-to-tosa/test_gemm.mlir` covers the basic form,
  omitted `C`, `transA`, `transB`, non-unit alpha/beta, a full-rank `C`, a
  column-vector `C` broadcasting along N, the non-kernel no-op, and rejections
  for dynamic shapes, integer element types, and a non-broadcastable `C`.
- Full lit suite passes (the one failure, `onnx-to-hip/test_qadd.mlir`, is
  pre-existing on this base and unrelated ΓÇö its RUN line never touches
  `HipToTosa`).
- End to end through `hip-rocmlir-compiler` against `librockCompiler`, the same
  model that previously crashed now compiles to a 5288-byte GPU binary
  (grid 16, block 64).

## Also here: hip-rocmlir-compiler on Windows

Folded in because CI's `build_mock` and `build_real` jobs run on `windows-2022`
and this was breaking them independently of the change above.

`tools/hip-rocmlir-compiler` reaches `librockCompiler`'s pipeline symbols
through the POSIX loader, and included `<dlfcn.h>` unconditionally. MSVC has no
such header, so the tool could not compile on Windows at all. The include is now
guarded, with `dlopen` / `dlsym` / `dlerror` stand-ins over `LoadLibraryA`,
`GetProcAddress` and `FormatMessageA`, so the loading logic itself stays
platform-independent.

`windows.h` is included after the LLVM/MLIR headers and behind
`WIN32_LEAN_AND_MEAN` and `NOMINMAX` so its macro surface cannot rewrite them.
`RTLD_LOCAL` is a no-op: a DLL never contributes its exports to a process-global
namespace, so the symbol isolation it buys on ELF â€” keeping the library's own
MLIR from interposing this executable's copy â€” is already how the Win32 loader
behaves.

The same commit adds `--dump-hip` and `--dump-tosa`, which write the module at
the two points where the pipeline hands work across a stage boundary. They were
what made the silent pass-through above visible in the first place. Both are
diagnostics on the way to `-o`, so failing to write one is reported without
failing the compile. The TOSA dump is taken before the non-kernel funcs are
dropped, so it matches what `hip-mlir-opt --convert-hip-to-tosa` prints rather
than showing only the kernels that cross into the `.so`.

### DIFF
--- a/lib/Conversion/HipToTosa/HipToTosa.cpp
+++ b/lib/Conversion/HipToTosa/HipToTosa.cpp
@@ -355,6 +355,24 @@ Value createSplatFloat(ConversionPatternRewriter &rewriter, Location loc,
       DenseElementsAttr::get(type, rewriter.getFloatAttr(elemType, ap)));
 }
 
+// Multiply by a scalar that ONNX carries as an f32 attribute. hipBLASLt scales
+// in f32 even when the data is f16 or bf16 (scaleType = HIP_R_32F while
+// dataType is HIP_R_16F), so for those types widen to f32 around the multiply
+// rather than rounding the scalar down to the data type and losing it there.
+static Value scaleByF32(Value value, float scale, RankedTensorType type,
+                        ConversionPatternRewriter &rewriter, Location loc) {
+  bool widen = !type.getElementType().isF32();
+  RankedTensorType mulType = widen ? type.clone(rewriter.getF32Type()) : type;
+  if (widen)
+    value = tosa::CastOp::create(rewriter, loc, mulType, value);
+  value = tosa::MulOp::create(rewriter, loc, mulType, value,
+                              createSplatFloat(rewriter, loc, mulType, scale),
+                              createZeroMulShift(rewriter, loc));
+  if (widen)
+    value = tosa::CastOp::create(rewriter, loc, type, value);
+  return value;
+}
+
 // hip.gemm is ONNX Gemm: Y = alpha * A' * B' + beta * C, where A and B are
 // optionally transposed and C broadcasts to [M, N]. TOSA has no fused
 // equivalent, so this spells the whole thing out. MIGraphX's ONNX parser
@@ -386,9 +404,12 @@ struct GemmConverter final : public OpConversionPattern<hip::GemmOp> {
       return rewriter.notifyMatchFailure(
           op, "A, B, and result element types must match");
     // alpha and beta are f32 attributes folded in as elementwise multiplies in
-    // the result type, so an integer gemm would silently round them away.
-    if (!isa<FloatType>(elementType))
-      return rewriter.notifyMatchFailure(op, "only float gemm is supported");
+    // the result type, so an integer gemm would silently round them away. f64
+    // is excluded separately: the hipBLASLt path accepts it, but TOSA has no
+    // f64 tensor type, and failing legalization beats emitting invalid TOSA.
+    if (!elementType.isF32() && !elementType.isF16() && !elementType.isBF16())
+      return rewriter.notifyMatchFailure(
+          op, "only f32, f16, and bf16 gemm are supported");
 
     Location loc = op.getLoc();
     bool transA = op.getTransA() != 0;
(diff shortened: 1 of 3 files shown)


nvunnam57128:
Thanks — all five points were valid and are addressed in ce2f3cb8.

**f64 legalization.** Correct, and this was the one producing invalid IR. `isa<FloatType>` admits f64, and the runtime really does accept it (`kTypeFloat64` → `HIP_R_64F` with `HIPBLAS_COMPUTE_64F` in `lib/Runtime/real/gemm.cpp`), so an f64 gemm built a `tosa.matmul` chain that TOSA cannot represent. The check is now `isF32() || isF16() || isBF16()`, with a `@gemm_f64` rejection case.

**f16/bf16 alpha precision.** Also correct: `scaleType` stays `HIP_R_32F` while `dataType` is `HIP_R_16F`, so scaling happens in f32. A new `scaleByF32` helper widens around the multiply for the narrow types. f32 gemm is byte-identical to before — no casts are emitted — and both scalars are still elided entirely at 1.0, so this only affects non-unit alpha/beta on half precision. Covered by `@gemm_f16_alpha`:

```mlir
%8  = tosa.reshape %6, %7 : (tensor<1x4x16xf16>, !tosa.shape<2>) -> tensor<4x16xf16>
%9  = tosa.cast %8 : (tensor<4x16xf16>) -> tensor<4x16xf32>
%12 = tosa.mul %9, %11, %10 : (tensor<4x16xf32>, tensor<4x16xf32>, tensor<1xi8>) -> tensor<4x16xf32>
%13 = tosa.cast %12 : (tensor<4x16xf32>) -> tensor<4x16xf16>
```

**Windows library resolution.** Good catch, and I can confirm it from use: every invocation during development needed `ROCK_COMPILER_SO` set explicitly. I fixed the loader calls but left `resolveSoPath()` returning `build/librockCompiler.so`. There is now a `defaultSoPath()` that returns `build/rockCompiler.dll` under `_WIN32`; the help text prints it rather than hard-coding the POSIX name, and the missing-symbol diagnostic no longer names the `.so`.

**Dump write errors.** Right, only the open error was checked. `dumpModule` now flushes and tests `has_error()` before reporting success, and warns with the underlying message otherwise.

**`--dump-tosa` comment.** Accurate criticism of my own comment — `tpm` canonicalizes before the dump, so it is not what `hip-mlir-opt --convert-hip-to-tosa` prints. The comment now says so explicitly.

Re-validated: hip-to-tosa lit suite 10/10, full suite 486 passed with the one pre-existing unrelated `onnx-to-hip/test_qadd.mlir` failure, and end to end through `librockCompiler` the gemm still compiles to the same 5288-byte binary (grid 16, block 64).

PR #842, opened 2026-08-26
feat(kernels): ship one gfx11-generic library instead of per-arch kernel DLLs

zz002:
## Summary

CI now builds a single `gfx11-generic` kernel library instead of one `custom_kernels_<arch>` per gfx115x target, so the package ships one kernel binary that runs on the whole gfx11 family.

## Related issue or design

None.

## Why

`gfx11-generic` is an LLVM generic target: one code object valid on gfx1100-1153. Shipping per-arch slices multiplies the kernel binary by the number of archs we cover, while every slice holds the same kernels. On a gfx1151 machine the generic build is not slower than the native one (see Performance), so those slices buy nothing at package level.

## What

- `windows-build.yml` / `linux-build.yml`: `HIP_ARCHITECTURES` becomes `gfx11-generic`, and the Linux packaging smoke check derives the kernel `.so` name from the same variable.
- `LlvmIrJit.cpp`: the kernel-library loader tries `custom_kernels_<arch>` first and falls back to the generic library of that arch family, so a package may ship the generic library alone or add per-arch libraries that override it. `genericTargetFor` maps arch to generic target using the AMDGPU Generic Processors table in LLVM AMDGPUUsage, because the arch digits do not imply the family (gfx1030 belongs to gfx10-3-generic, and no gfx10-generic exists).
- `cmake/deps.cmake`: `gfx11-generic` passes the Windows TheRock arch check, since it needs a toolchain that can emit it rather than a bundle slice. The Linux branch no longer derives the tarball name from `HIP_ARCHITECTURES`.
- `cmake/deps.txt`: `therock_linux` pins the full archive basename, matching how `therock_windows` is already spelled.

## Test plan

- [x] Local Windows build (TheRock 7.11, Ninja): `custom_kernels_gfx11-generic.dll` (12.93 MB) and `hipgpu.dll` link clean.
- [x] Local A/B on gfx1151, same EP build with only the kernel library swapped: no regression (numbers below).
- [ ] `build-windows` green, package carries a single `custom_kernels_gfx11-generic.dll`.
- [ ] `build-linux` green, package carries `libcustom_kernels_gfx11-generic.so`.
- [ ] `gpu-test` green on gfx1151, EP resolving kernels through the generic library.

## Performance

Qwen3.5-35B-A3B vision + text, gfx1151, OGA `model_benchmark`, 3 iterations.

| Metric | gfx1151 kernels | gfx11-generic kernels |
|---|---:|---:|
[...]

PR #826, opened 2026-08-25
fix(gqa): update LUT with autotune choice on 7.14

ZhanxingPu:
## Summary

Autotune logs collected from a second gfx1151 machine across five
production models show that NW4_BKV64_MT1 is consistently faster than
the NW2_BKV32_MT1 / NW4_BKV32_MT1 entries that ship in the current LUT,
across all seq-length buckets tested.

Models verified: DeepSeek-R1-70B, Llama-3.1-8B, Mistral-7B,
Phi-4-14B, Qwen2.5-14B (all use d=128, hpg ∈ {4, 5, 8}).

## Changes

- **gfx1151.json / gfx1151.fb** — 400 PrefillV7 rows updated to
  `NW4_BKV64_MT1`:
  - ExactHeadGroup: `H64_G8`, `H32_G8`, `H40_G10`, `H40_G8`
    (all sq-length buckets, covers d=128 at hpg=8/10)
  - HeadGroup / Geometry fuzzy tiers: hpg=8, hpg=4, hpg=5
  - .fb recompiled with flatc v25.12.19

- **CMakeLists.txt** — revert `HIPDNN_EP_GQA_AUTOTUNE_ONLINE_DEFAULT`
  back to `OFF`; the option was flipped to ON during data collection and
  should not ship.

- **gqa_autotune_sweep.cpp** — sweep tool quality-of-life:
  suppress per-iteration prints during warmup passes; extend default
  timing budget (target_ms 40→1000, max_iters 200→2000) for more
  stable measurements on short-latency shapes.

## Test plan

- [ ] Rebuild `hip-ep` on gfx1151; confirm LUT is embedded in runtime.bc
- [ ] Run prefill benchmark with `HIPDNN_EP_PERF=1` on DeepSeek-R1-70B
  and Llama-3.1-8B; verify latency at representative seq lengths
  (sq=512, 1024, 4096) matches or improves on prior results
- [ ] Confirm `HIPDNN_EP_GQA_AUTOTUNE_ONLINE_DEFAULT` defaults to lookup
  mode (no online GPU sweep at session create)

PR #992, opened 2026-09-12
build(gqa-lut): sweep prefill across the KV range, and let one phase be rebuilt alone

BoarQing:
Sweep-grid and rebuild tooling only. The retuned table that used to sit on top of
this branch is reverted in `4cedb546`, so the net diff here is `update_lut.py`
alone; measurements on the shipped `gfx11-generic` package show the table would
have been a net regression, and the harness that produced it is not currently a
faithful proxy for what ships.

## What is left here

`update_lut.py` gains two things, both independent of which target is swept:

- the prefill sweep grid covers the KV range instead of a single corner, so a
  geometry like `sq=2048, skv=2048` is measured rather than extrapolated from
  `skv=65536`
- a single phase can be rebuilt without re-sweeping the others

## Why the table commit was dropped

CI moved the kernel build target to `gfx11-generic` in #842, a day before the
table on this branch was measured on a `gfx1151`-targeted build. The optimal
prefill config is not the same on the two, so the table was tuned for a binary
that is no longer the one shipping.

Reading the in-kernel autotuner out of the shipped `gfx11-generic` package
(`HIPDNN_GQA_AUTOTUNE_LOG` for the launched config, `HIPDNN_GQA_AUTOTUNE_MODE=online`
with `HIPDNN_PREFILL_TUNE_DEBUG` for every candidate's cost), then scoring both
tables against those measured timings:

| table | summed prefill time over all measured geometries |
| --- | --- |
| a perfect table | 57,648 ms |
| `main` | 66,325 ms (+8,676) |
| this branch's table | 68,142 ms (+10,493) |

So the table was 1,817 ms *worse* than `main` across the measured set. It is a
mix of two opposite effects:

- **d=128 `PrefillV7`, where it loses.** On `gfx11-generic`, `main` already
  launches the measured optimum at every d=128 geometry tested — 4 models, both
  lengths, 36 geometries, zero headroom. The dropped table moved them to
  `bkv=32`, which is right on `gfx1151` and wrong here: Mistral 16K 155.4 →
  178.1 ms/call, and DeepSeek-70B 16K 309.5 → 355.8 ms/call, which at 80 layers
  alone is +3,696 ms of TTFT.
- **d=256 `PrefillV8`, where it wins.** `main` has a genuine coverage hole here:
  its 58 V8 points sit only at `skv=128` and `skv=65536`, with nothing on the
  `sq == skv` diagonal, so production geometries resolve at distance 5.4–6.3 and
  land on the *slowest* of the five candidates. The expanded grid fixes that —
  Qwen3.8-27B 16K goes 899.5 → 681.3 ms/call. That is what the grid change in
  this PR is for, and it is worth having.

The d=128 loss is larger than the d=256 win, so shipping the table as measured
was not viable. Regenerating it correctly is worth doing: the hole it closes is
11.8% of TTFT on Qwen3.8-27B and 16.4% on Qwen3.6-35B at 16K.

## Why the grid change alone does not finish the job

Pointing the sweep at the shipping target is necessary but not sufficient. Built
with `--offload-arch=gfx11-generic`, the standalone sweep binary still disagrees
with the shipped DLL on the same shape:

| `d=128 sq=skv=16384`, H32/G8 | standalone sweep | shipped `gfx11-generic` DLL |
| --- | --- | --- |
| `NW4 BKV32 MT1` | 241 ms (its pick) | 178 ms |
| `NW4 BKV64 MT1` | 509 ms | 155 ms (actual optimum) |

`BKV=32` differs by about a third, which is plausible harness overhead, but
`BKV=64` differs by 3.3x and flips the ranking. That is not explained by the
offload arch, by `-O2` vs `-O3` (the DLL builds kernels at `-O2`, the sweep
Makefile hardcodes `-O3`; both were measured), or by `max_seq` (16384, 32768 and
131072 all land within 1%). Until that divergence is understood, a table built
from this harness cannot be trusted for `gfx11-generic`, whatever grid it sweeps.

The in-kernel autotuner readout used for the numbers above does not have this
problem, because it runs inside the binary being shipped. That is already how
`matmul_nbits` is documented to source its winners; the GQA pipeline uses a
separate sweep binary instead, and that structural difference is what this
measurement exposed.

Whatever produces the next table, `gpu_arch` must stay stamped `gfx1151`. Both
loaders compare that field to the runtime device via `currentGpuArch()`, which
reports `gfx1151` regardless of build target, so stamping `gfx11-generic` would
make the runtime discard the table and silently fall back to heuristics.


BoarQing:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1072/
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1073/
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1074/
http://xsjvitisaijenkins:8080/job/daily_hip_ep_oga/1782/

BoarQing:
Closing: the premise was wrong and the change is a measured regression.

## What this PR assumed

That d=128 prefill was resolving to the wrong config. The evidence was the
resolver distance -- Mistral-7B answers `sq=skv=2048` from `d=5.099` where
decode resolves at `d=0.042` -- plus a measured 2.3-2.6x throughput collapse
with prompt length on all three d=128 models.

## Why that was wrong

The distance is real but harmless. `main`'s table answers every production
coordinate from its `sq=4096 skv=65536` point, and 23 of its 35 `PrefillV7`
points carry the same config, so a far neighbour still lands on the correct
answer. A cosmetically poor resolution is not an incorrect one.

The throughput collapse was an artifact of the rig I measured on: ROCm 7.11.0
with `HIP_ARCHITECTURES=gfx1151`. Production builds ROCm 7.14.0 with
`gfx11-generic`. `BKV` trades LDS, hence occupancy, against per-tile
amortization, and that crossover moves between the two toolchains.

## What the shipped binary actually wants

Online autotune sweeping all 12 V7 candidates inside the CI-built
`custom_kernels_gfx11-generic.dll`:

| coordinate | best | time | `BKV=32`, same NW |
| --- | --- | --- | --- |
| `sq=skv=2048` | `NW=4 BKV=64 MT=1` | 3.704 ms | 4.100 ms (+10.7%) |
| `sq=skv=16384` | `NW=4 BKV=64 MT=1` | 164.46 ms | 227.07 ms (+38.1%) |

`main` already resolves to `nw=4 bkv=64 mt=1` at 2048, 8192 and 16384. There is
no headroom. This PR moved 454 of its 460 `PrefillV7` points onto `bkv=32`.

## Measured cost

Both arms are CI-built packages, run interleaved and order-reversed on one
gfx1151 host. Baseline is `main@90c2b457`, this branch's own base.

| model | length | `main` | this PR | delta | fwd/rev spread |
| --- | --- | --- | --- | --- | --- |
| Mistral-7B | 8192 | 6077 ms | 6257 ms | +2.9% | 0.15% |
| Llama-3.1-8B | 8192 | 6341 ms | 6520 ms | +2.8% | 0.10% |
| Mistral-7B | 16384 | 14490 ms | 15235 ms | +5.1% | 0.06% |
| Llama-3.1-8B | 16384 | 15146 ms | 15938 ms | +5.2% | 0.48% |

Consistent with the Jenkins fleet run on #993, which measured +6.0% on
Mistral-7B and +6.6% on Llama-3.1-8B at 16K.

## The reason this got as far as CI

`lut/gfx1151.json` carries a `rocm_version` field for exactly this hazard, and
it is inert at both ends:

- `scripts/update_lut.py` sets `ROCM_VERSION = 70151803` as a hardcoded
  constant, so a table measured on 7.11 is stamped identically to one measured
  on 7.14.
- `gqa_autotune.cpp` never reads the field. It validates `schema_version`,
  `kernel_abi` and `gpu_arch`; `rocm_version` has no reader.

So a table measured against the wrong SDK loads silently and is trusted.
Worth fixing independently of this PR -- stamp the SDK version actually used
and have the loader reject a mismatch. #826 ("update LUT with autotune choice
on 7.14") already had to rewrite 800 lines of this table for a version bump,
so the failure mode has precedent.

## Not carried forward

`4f5f8b3e` (wider prefill KV ladder, `build --phase`) is dropped too. Its
rationale was that chunked prefill pins `seq_q` at the chunk size, which is why
it trims `sq >= 2048 and skv > 4 * sq`. On the stack measured here prefill is
issued as a single call with `sq == seq_kv` at both 2048 and 16384, so that
premise is unverified as well.

PR #962, opened 2026-09-10
[DO NOT MERGE] test: PR 961 capacity-keyed large-buffer pool on PR 599's OGA build

BoarQing:
[DO NOT MERGE] CI-only. Runs PR #961's rewritten large-buffer pool on PR #599's OGA build, which is what actually exercises the Gemma-4 models the pool targets. Not for merge; #961 is the change under review.

- `761edb30`, `54f5ea29` -- PR #599, build OGA from the AMDGPU fork
- `2bda665b` -- PR #961, pool large pinned buffers by capacity

Rebuilt on current `main`. The allocator commit is byte-identical to #961's head.

PR #993, opened 2026-09-12
[DO NOT MERGE] test: PR 994 autotune retune on PR 599's gemma4-capable OGA build

BoarQing:
Benchmark vehicle for [#994](https://github.com/ROCm/hip-ep/pull/994). Not for
merge.

| commit | source | why |
| --- | --- | --- |
| `2719575c` | #994 | the DLL-bound matmul_nbits + GQA prefill retune under test |
| `df4e9d9f`, `5408d20d` | #599 | build OGA from the AMDGPU fork, so gemma4-unified models load |
| `3530f144` | `6c7e7ac5`, replayed | untrimmed TheRock dist, so gemma-4's vision pooler GEMM does not silently fail |

PR 961's allocator commit is deliberately excluded, so this measures the LUT
change alone.

## The deps fix

`f8532c0a` pins `therock-dist-windows-gfx11-all-7.14.0`, which **is not
published** -- that asset was deleted from the v0.3.0 release, so the prebuild's
*Fetch MIGraphX + TheRock SDKs* step 404s. Worth flagging on
[#962](https://github.com/ROCm/hip-ep/pull/962): it carries the same commit and
only goes green because its prebuild restores from cache, so it fails the same
way on a miss.

Two of the three untrimmed candidates are dead ends:

| asset | outcome |
| --- | --- |
| `gfx11-all-7.14.0` | 404, deleted from the release |
| `gfx115X-all-7.14.0` | downloads, but ships no `bin/hipconfig`, so `find_package(HIP)` fails with `HIPCONFIG_EXEC-NOTFOUND` |
| `gfx1151-7.14.0` | full SDK, untrimmed, **builds green** |

`3530f144` pins the last of those (same change as `6c7e7ac5`). Verified locally
as well as in CI: it carries `bin/hipcc`, `bin/hipconfig` and 276 gfx1151
hipblaslt/Tensile files, against 18 of 100 per arch in the trimmed dist.
Dropping the other gfx11 members costs nothing -- `cmake/therock.cmake`
FATAL_ERRORs on any Windows `HIP_ARCHITECTURES` outside gfx1150-gfx1153 plus
`gfx11-generic`, the GPU test node is gfx1151, and the dist still carries
`oclc_isa_version_11-generic.bc` so `HIP_ARCHITECTURES=gfx11-generic` builds
unchanged.


BoarQing:
Closing with [#992](https://github.com/ROCm/hip-ep/pull/992). This PR existed only to benchmark that table on top of [#599](https://github.com/ROCm/hip-ep/pull/599)'s gemma4-capable OGA build, and it did its job: the Jenkins fleet run measured +6.0% TTFT on Mistral-7B and +6.6% on Llama-3.1-8B at 16K, i.e. the table was a regression. Root cause and the measured read-out from the shipped binary are written up in #992.

PR #999, opened 2026-09-14
feat(hipsr): add the externalize-constants pass

zpye:
`hipsr-externalize-constants` already stamps each constant and writes `constants.bin`. `--generate-interface` needs the layout on the module, so the pass now also writes `hip.constants_file`, `hipdnn.constant_sizes`, and `hipdnn.constant_offsets`. It runs last in `--hipsr-pipeline`.

AI assistance was used on this change.


wcy123 (approved):
LGTM

PR #1000, opened 2026-09-14
feat(qmatmul): support per-column and int4 weight quantization, and tune the M==1 decode path

amd-genmingz:
## Summary

Extends `hip.qmatmul` from per-tensor-only 8-bit weights to per-output-column
scales/zero points and 4-bit packed weights, and rewrites the M==1 GEMV decode
kernel around in-block split-K with an autotuned geometry.

## Related issue or design

No tracking issue yet. Design notes for both parts live outside the repo; the
op-level contract is documented inline in `HipOps.td` and
`hipdnn_ep_runtime.h`, which are the intended sources of truth after merge.
Happy to open an issue and link it before review if maintainers prefer.

## Why

- Some model quantize weights per output column, often at 4 bits.
  Without that form the QDQ subgraph never fuses, so the matmul falls back and
  the whole quantized path is lost.
- `qmatmul` dominated GPU time in decode (199.1 ms, 95.1% of total). The kernel
  was latency-bound: one DRAM round trip per k-step with only ~256 B in flight,
  so it was nowhere near bandwidth limits.

## What

**Per-column / int4 weights**

- `hip.qmatmul` gains optional `B_scales` / `B_zero_points` operands,
  `B_quant_axis`, and a `packed_int4` unit attribute; `B_scale` /
  `B_zero_point` become optional so the two forms are mutually exclusive.
  A verifier enforces that, plus operand rank, element type, and length.
- Two new PDLL patterns (`QDQMatMulPerColumnW4`, `QDQMatMulPerColumnW8`) at
  benefit 9, so per-tensor still wins the tie where both match.
- Kernel ABI: `hip_qmatmul` / `wrap_qmatmul` take `B_scales`,
  `B_zero_points`, `b_bits`, and `AY_ratio`. For per-column 4-bit B, folding
  the zero point into nibble widening drops the `rowA` and `K*A_zp*B_zp`
  correction terms entirely.

**GEMV decode path** (three commits, reviewable in order)

- Vectorize over column groups (`kGemvCols`).
- Split K inside the block over `threadIdx.y` and merge slices through LDS;
  this is what cut the serial k-chain and took decode 199.1 ms → 67.6 ms.
- Templatize the K-tile and merge via LDS atomics instead of per-slice
  buffers, so LDS no longer scales with split count and finer tiles can
  expose more slices. Geometry is picked by a 12-entry config table tuned
  on-device once per shape and cached, with a safe single-slice fallback.

**Cleanup / infra**

- `hipsr-externalize-constants` no longer stamps `hip.constants_file` /
  `hipdnn.constant_{sizes,offsets}` and is removed from `buildHipsrPipeline`;
  `hip-externalize-constants` owns storage policy and artifact I/O.

## Test plan

- [ ] CI Accuracy and Performance Verification

## Notes for reviewers

- 25 files / ~2.1k changed lines crosses the `large-pr` thresholds. The weight
  quantization work and the GEMV tuning are independent and could be split;
  they are together here because both land on `qmatmul_kernel.hip` and would
  otherwise conflict. Happy to split if preferred.
- The 67.6 ms decode number is measured through the split-K commit. The
  K-tile templating on top of it is implemented but **not yet measured** —
  worth confirming before we claim the remaining headroom.
- The templated geometry produces 384 kernel instantiations (12 configs × 32
  dtype triples). Compile time cost is accepted deliberately; flag it if that
  is too steep.
- M=128 prefill is untouched by this PR (188.8 ms, unchanged). It runs the
  tiled path and sits at ~9% of dot4 peak, so it is still open ground.
- `hipsr-externalize-constants` is now a pure IR transform with no caller in
  `lib/` or `tools/`. Intentional, but worth a second opinion on whether it
  should be deleted instead of kept.
## AI Assistance


amd-genmingz:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1161
image

amd-mingw (approved):
with a follow-up PR switch autotune to LUT solution

PR #1004, opened 2026-09-14
fix(runtime): record every PERF phase event before querying it

pschlan-amd:
## Summary

Fixes the `[PERF]` phase breakdown, which reported `H2D 0.00 ms | Compute 0.00 ms | D2H 0.00 ms` on every inference, and removes the misleading `[custom_kernels] hip_gather launch failed: invalid resource handle` messages that came with it. On a Gemma decode run the two are the same bug: 1085 gather "launch failures" across 1086 inferences, while gather actually runs 3-4 times per inference and never failed.

## Why

`g_perf.h2d_end` lost its only `hipEventRecord` when `prepare_output` was removed for the output-allocator ABI. Under that ABI the generated graph calls only `prepare_input` and `stream_sync` (see `GenerateInterface.cpp`), so `finalize_output` never runs either and `d2h_start` was unrecorded as well. All three `hipEventElapsedTime` calls therefore failed every inference and left the phase floats at their initialized zero.

`hipEventElapsedTime` on an unrecorded event returns `hipErrorInvalidResourceHandle`, and HIP holds that error in the thread-local last-error slot until something reads it. The four query sites discarded the return code with `(void)`, so the first kernel that checked after its launch picked the error up as its own. That is the embedding gather, the first op in the graph, which is why exactly one message appeared per inference and why only gather ever complained.

Beyond the noise, this made every custom kernel's post-launch error check untrustworthy: a genuine launch failure would be attributed to whichever kernel read the slot first, and a stale error could mask it.

## What

- Record `h2d_end` at the end of every `prepare_input` (last one wins, the same approach `d2h_end` already used, since the generated graph offers no end-of-marshalling hook).
- Close any window the callbacks left open in `stream_sync` before recording `d2h_end`, so no span can ever reference an unrecorded event. With runtime-owned outputs this makes Compute span the whole kernel sequence and D2H degenerate, which is the truth when nothing is copied back.
- Gate `d2h_start` on an explicit flag instead of `d2h_count == 0`, which never leaves zero when every output takes the alias fast path.
- Route both query sites through `perf_resolve_spans`, which reports a failed span once instead of discarding it, reports any HIP error already pending at the inference boundary rather than overwriting it, and clears the last-error slot so profiling cannot leak error state into kernels.
- Drop the duplicate `inference_num++` in `stream_sync` that made the one-line flush and the block header disagree by two (`[PERF] #18` vs `[PERF] inference #20` for the same inference).
- Clear the last-error slot before the launch in `hip_gather`, per the "clear before launch, return `hipGetLastError()` after" invariant in `CLAUDE.md`, so its check reports only its own launch.
- Add `hipGetLastError` to the mock runtime, which the shared tensor TU now needs.

## Test plan

- Compile-checked all four affected paths on Windows: real-runtime bitcode, mock-runtime bitcode of the same TU, `mock_gpu.cpp`, and the `gather_kernel.hip` object for gfx1151. Clean under `-Wall` apart from pre-existing third-party header warnings.
- `pre-commit run --files <changed files>` passes.
- Reran the Gemma gqa-verify scenario with `HIPDNN_EP_PERF=1` on gfx1151 after invalidating the model cache: no `hip_gather launch failed` lines, no `[PERF] WARNING` lines, and the `Compute:` figure is now plausible and consistent with the `gpu=` wall figure instead of 0.00 ms.

## Notes for reviewers

Two related issues are deliberately left out of this change:

- Only `elementwise_binary_kernel.hip` and now `gather_kernel.hip` clear the last-error slot before launching. The remaining custom kernels should follow, but that is a large mechanical sweep across ~56 files and is pure hardening now that this error source is gone.
- `g_perf` is a single process-global while a two-session run (e.g. OGA) interleaves inferences from both sessions, so the phase counters and events are still shared across sessions. The per-op `[PERF]` table is unaffected because `op_profile` state is already per-session.

AI assistance: the root-cause analysis, the implementation, and this description were prepared with Cursor (Claude Opus 5) and reviewed by me. Validation is as described in the test plan, including the GPU run on gfx1151.


pschlan-amd:
Closing in favor of https://github.com/ROCm/hip-ep/pull/1018 since PRs from a fork break the "gpu_test" action.

PR #1013, opened 2026-09-15
feat(hipsr): lower nonzero, alloc_output, and preserve_shape to LLVM

zpye:
Lower `hipsr.nonzero`, `hipsr.alloc_output`, and `hipsr.preserve_shape` to LLVM.

- nonzero calls the search kernel (count is `i32`; positions stay `i64`)
- alloc_output asks the EP output allocator for a buffer
- preserve_shape is metadata, so the conversion erases it with a plain rewrite (a ConvertOpToLLVM adaptor would need LLVM-convertible operand types)

First of two stacked PRs. Follow-up: #1014.

AI-assisted; reviewed.

PR #1014, opened 2026-09-15
feat(hipsr): add LLVM lowering to --hipsr-pipeline

zpye:
Fold LLVM lowering into `--hipsr-pipeline` so one flag goes from ONNX to LLVM.

Adds `expand-strided-metadata`, `lower-affine`, `convert-scf-to-cf`, `reconcile-unrealized-casts`, and `convert-to-llvm`. Pipeline LIT now checks the full LLVM IR.

`lower-affine` and `reconcile-unrealized-casts` are needed for embedding collapse/expand stride products.

Stacked on #1013. Retarget to `main` after that merges.

AI-assisted; reviewed.


wcy123 (approved):
LGTM.

Extends `--hipsr-pipeline` to go all the way to LLVM IR. [Adds 5 passes](https://github.com/ROCm/hip-ep/blob/03a0294f44d55cf5a4447334ba470e1b6409831e/lib/Dialect/Hipsr/Pipelines/Pipelines.cpp#L92-L100):

1. `expand-strided-metadata` (collapse_shape/expand_shape have no LLVM pattern)
2. `lower-affine` (expand-strided-metadata emits affine.apply)
3. `convert-scf-to-cf` (convert-linalg-to-loops leaves scf.for/scf.if)
4. `reconcile-unrealized-casts` (drop leftover conversion casts)
5. `convert-to-llvm`

Test updates rebaseline 3 pipeline tests to full LLVM IR (embedding.mlir, sample_dynamic.mlir, sample_static.mlir).

PR #1017, opened 2026-09-15
feat(qgemm): fuse QDQ Gemm into hip.qgemm with per-channel and 4-bit weights

amd-genmingz:
## Summary

Adds `hip.qgemm`, the integer-domain fusion of DequantizeLinear x2 (or x3) ->
ONNX Gemm -> QuantizeLinear, covering an optional bias, per-output-channel
weight quantization, and 4-bit packed weights. Includes the PDLL fusion
patterns, the HIP-to-LLVM lowering, the `hip_qgemm` kernel ABI with real and
mock runtime glue, and a dedicated M==1 decode kernel.

## Related issue or design

No tracking issue. Follows the QDQ fusion design already used by
`QDQMatMulFusion.pdll` and `QConvLowering.cpp`. Op semantics are documented
in `HipOps.td`, the runtime ABI in `hipdnn_ep_runtime.h`.

## Why

Gemm was the last common quantized layer with no fused path, so a quantized
FC layer dequantized to f32, ran an f32 Gemm, and requantized.

## What
**Op and dialect**
- `hip.qgemm` with `AttrSizedOperandSegments`: optional `B_scales` /
  `B_zero_points` in a `b_per_channel(...)` operand group, optional `C`
  broadcast to [M, N], and `alpha` / `beta` / `transA` / `transB`.
- `B_bits` is 4 or 8. At 4, `B` and `B_zero_points` keep their logical shape
  and 8-bit element type while each byte holds two values, low nibble first --
  the same ONNX convention already used by `hip.qconv`.
- Verifier enforces operand ranks, C broadcastability, 8-bit B storage, the
  all-or-nothing per-channel pair, and one scale/zero point per N.
  `reifyResultShapes` follows the same 2D rule as `hip.gemm`.
**Fusion**
- Four PDLL patterns: `QDQGemm` / `QDQGemmNoBias` at benefit 10, and the
  per-channel variants at benefit 9, so a per-axis weight whose scales happen
  to be splat still takes the per-tensor path where the coefficient folds to a
  scalar instead of costing a load per output.
- Separate patterns because a PDLL operand list matches an exact arity and a
  rewrite cannot branch; each writes its own operand segment sizes.
- New native constraints/rewrites: `IsEightSixteenOrThirtyTwoBitQuantized`,
  `IsPerChannelWeight`, `ExtractQuantBits`, `ExtractAttrF32`.
- Fix: ONNX si64 attributes are normalized to signless i64 before they reach
  hip ops, which also affects the plain `hip.gemm` conversion.
**Lowering and runtime**
- Lowering folds `M_ab = alpha*s_a*s_b/s_y` and `M_c = beta*s_c/s_y`, so the
  kernel never divides. A per-channel `B_scales` is the one factor that cannot
  fold and stays a device array applied in the epilogue.
- Zero points are expanded into `rowA` / `colB` correction terms so the inner
  loop runs on raw operands; each term is only accumulated when its zero point
  can be non-zero, so symmetric weights cost nothing. Accumulators are int64
  because a 16-bit activation against a far-end 4-bit zero point overflows
  int32 after two terms.
- `hip_qgemm` / `hip_qgemm_workspace_bytes` added to the kernel ABI, with
  `wrap_qgemm` in both the real and mock runtimes.
**M==1 decode kernel**
The 16x16 reference kernel collapses at M==1: N/16 blocks of 16 live threads
each walk all of K one byte at a time. The specialized path gives a thread
`kCols` consecutive columns so B is read as whole dwords, stages A once per
k-tile in LDS, and splits K twice -- `kSplits` slices side by side inside a
block merging through LDS atomics, plus a grid-level split into scratch with a
requantizing second pass for a narrow N that cannot fill the device on column
windows alone. Geometry is a per-shape property, so it is templated and
measured on device once per shape (12-entry table, `HIPDNN_QGEMM_GEMV_AUTOTUNE`,
safe single-slice fallback).

## Test plan

- [x]  `onnx-to-hip/test_qgemm.mlir` — 4 cases: per-tensor A8W8 with int8 bias
      and non-default `alpha`/`beta`/`transB`; per-tensor bias-free `ui8`;
      per-channel A16W4 with int32 bias; per-channel A8W8 bias-free
- [x] `hip-to-llvm/test_qgemm.mlir` — 2 cases: static per-tensor with folded
      `M_ab`/`M_c`; dynamic-M per-channel 4-bit, checking the full call ABI
- [x] accuracy test with CPUEP

Category | Count | Result
-- | -- | --
Per-channel w4 (backbone, INT4 weight + INT4 zp) | 196 | 196/196 PASS
Per-tensor w8 (LoRA, B as graph input) | 392 | 392/392 PASS
Exact match (L2 = 0) | 495 | —
Non-zero L2 (within budget) | 93 | max L2 ≈ 2.24, max abs = 1

- [x] performance test by CI

## Notes for reviewers

- `qgemm_kernel.hip` is a correctness-first reference implementation, not a
  tuned one. No `test/numeric/` coverage yet, so the last two test-plan boxes
  are open; happy to split the kernel into a follow-up if reviewers prefer to
  land the compiler path first.
- The two `.github/workflows/` files come from PR 534 and will drop out of this
  diff once that lands and this branch rebases.

## AI Assistance

The code was developed collaboratively with Cursor, and all code has been
reviewed.


amd-mingw (dismissed):
with a follow-up PR switch autotune to LUT solution

amd-mingw (dismissed):

amd-mingw (approved):

PR #1023, opened 2026-09-16
feat(ep): embed the PDL fusion patterns in the EP library

amd-genmingz:
## Summary

The compiled PDL fusion patterns are now linked into the EP library as a byte
array instead of being shipped as a loose `HipFusionPatterns.pdl.mlir` next to
it. `ConvertOnnxToHipPass` reads them from memory, so the pass no longer has to
locate its own binary at runtime, and no packaging path has to carry the file.

## Related issue or design

None. No design document described the previous file-based distribution, so
nothing in `docs/` needed updating.

## Why

Shipping the patterns as a data file made a correct build silently
non-functional. Every consumer — the install tree, the wheel, and the CI
staging step — had to remember to copy one extra file beside the EP, and a
missed copy only surfaced as a failed model compile at inference time.

The previous lookup also required `dll_path()`, a `GetModuleHandleEx` /
`dladdr` helper whose result depends on the final link target, plus a
`windows.h` include in an otherwise portable compiler source file.

## What

Build pipeline in `lib/Conversion/OnnxToHip/CMakeLists.txt`:

1. `mlir-pdll -x mlir` produces the PDL module as before.
2. `mlir-opt --strip-debuginfo --emit-bytecode` condenses it: 41645 -> 3480
   bytes. `mlir-pdll` hardcodes `enableDebugInfo()`, so two thirds of its
   output was 234 location aliases each repeating an absolute source path;
   stripping them also keeps the build directory out of the shipped library.
   Bytecode then folds the remaining names into a string table.
3. `cmake/xxd.py` emits the array and an appended accessor pair.

`mlir-opt` is optional. Without it the textual module is embedded unchanged,
which the parser accepts because it dispatches on the bytecode magic bytes.
Python is now required in this directory: `find_package(Python3)` only
populates the scope it is called from, and this directory is configured before
`lib/Runtime`.

The generated translation unit keeps the array `static` and exports only

```c++
extern "C" const unsigned char *hip_pdl_fused_pattern_data(void);
extern "C" size_t hip_pdl_fused_pattern_size(void);


wcy123 (approved):
LGTM.

PR #1024, opened 2026-09-16
perf(gqa): chunk a windowed prefill so the key narrowing can engage

BoarQing:
Chunks a windowed prefill so the key narrowing that already exists can actually run.

## The bug

`gqa_forward_hipblaslt` recovers a per-chunk key bound from the sliding window, but it is keyed off `q0`, so it narrows nothing when there is only one chunk. Chunking is decided purely by a byte budget, and at 2283 tokens on a 16-head model the score pair is ~500 MB against a 1 GiB default. So a windowed prefill takes the single-chunk path, scores the full key range, and discards the window. The narrowing code was written and correctly guarded; nothing ever entered it.

## The change

Chunk on the window itself, independently of size, gated on the same `chunk_narrow_ok` the narrowing needs plus a check that there is more to drop than the chunk already covers.

## Chunk height is measured, not assumed

gemma-4-12b at 2283 tokens, 1024 window, against an unchunked 2748 ms:

| rows | TTFT | |
|---|---|---|
| 1024 | 2663 ms | -3.1% |
| 512 | 2516 ms | -8.4% |
| **256** | **2441 ms** | **-11.2%** |
| 128 | 2457 ms | -10.6% |

A peak rather than a plateau: shorter chunks score fewer keys per row but issue more passes and hand the GEMMs a smaller `n`.

## Validation

DLL-swap A/B, interleaved, three rounds:

| point | base | cand | delta |
|---|---|---|---|
| 12B @2K | 2663 ms | 2467 ms | **-7.34%**, CI [-209, -182] |
| 12B @2K reversed | 2659 ms | 2474 ms | **-6.93%** |
| 12B @128 | 506 ms | 505 ms | -0.09%, spans zero (gated off) |
| 12B @16K | 21022 ms | 20990 ms | -0.15%, spans zero |

Gated off at 128 by design (392 < 1024+256). No regression at 16K, where the byte budget already chunked.

## Caveat

The 2K gain reads 7% here and 11% in the earlier env-var sweep. The candidate lands at 2441-2474 ms in both runs; the baseline is what moved between sessions. Only within-invocation deltas are quoted.

## Not included

Bit-exactness is argued in the existing comments (an additive bias at or below -65504 contributes exactly zero to the softmax, so dropping those keys is exact) but is not verified numerically in this PR. Worth a logits diff before merge.

---

## Correction: this is NOT bit-exact

I originally flagged bit-exactness as argued-but-unverified. Verified now, and it does not hold.

**Control** (same DLL, two runs, 64 greedy tokens on the 2283-token multimodal prompt): **byte-identical output.** The runtime is deterministic run to run, so the comparison below is attributable to this change.

**Test** (baseline vs candidate, same prompt): output matches for ~45 tokens, then diverges at character 226 of 302:

```
base: ...the high computational and memory demands of generative AI (like ChatGPT-4...
cand: ...the high computational and memory costs   of generative AI (like ChatGPT-4...
```

Both outputs remain coherent and on topic; it is a single near-tie flipping under greedy argmax and then propagating.

**Why.** The narrowing itself is exact - it only drops keys whose additive bias is at or below -65504, which contribute exactly zero to the softmax. But chunking also changes the score GEMM's `n` from 2283 to 256, so hipBLASLt's heuristic selects a different kernel, and a different tile accumulates in a different order. The arithmetic is mathematically equivalent, not bitwise identical.

This is the same class of numerical change as any kernel retune or tile change, and by that standard it is acceptable. But it should be reviewed as such rather than as a no-op, and the PR previously implied otherwise.

---

## Corrected numbers (earlier validation was contaminated)

The first validation ran with a leaked `HIPDNN_EP_GQA_SCORE_BUDGET_MB=224` in the benchmarking shell, left over from a previous sweep. `ab_interleaved.ps1` clears only the env keys that some arm in the *current* manifest declares; these arms declare only `dll`, so the union is empty and the stale variable passed into both arms. At 2K it forced both to the budget's chunking; at 16K it forced both to 149-row chunks, which is why 16K looked like a no-op.

Re-measured with the environment explicitly cleared:

| point | base | cand | delta |
|---|---|---|---|
| 12B @2K | 3151 ms | 2873 ms | **-8.80%**, CI [-292, -263] |
| 12B @16K | 23328 ms | 22272 ms | **-4.53%**, CI [-1084, -1028] |
| 12B @128 | - | - | gated off (392 < 1024+256) |

Both are better than first reported (-7.34% and "no change"). **16K is a real win, not a no-op** - 256-row chunks beat the byte budget's 640 by 4.5%.

Control: a build with `kWindowChunkRows = 640` measures +0.05% against pre-change (interval spans zero), confirming that when the constant does not bind, behaviour is identical to the old path.

## Chunk height, re-swept clean

At 2K, with the environment clean: 192 -> 2860 ms, 256 -> 2873, 320 -> 2885, 384 -> 2904. At 16K: 256 -> 22272, 384 -> 22726, 640 -> 23339. Smaller is better at both lengths, and the curve is shallow near the optimum. 256 sits close to best at 2K and is clearly best of those tested at 16K. 192 is marginally better at 2K (-0.45%) but was not tested at 16K.


BoarQing:
Rebased and retargeted. Base is now `perf/allocator-pool-large-reland` (#1068) rather than `main`.

## Why not main

This branch's merge-base was `86a4cc3e`, which **is #961** — the large pinned-buffer pool. #961 was reverted by #1056, so rebasing onto today's `main` would have moved this work onto a tree where the allocator pooling is gone, which is not the environment any number in the description above was measured in.

#1068 relands #961 with the defect that caused the revert fixed, so basing on it both contains all of current `main` and restores the allocator behaviour this branch was developed against. It should land after #1068.

This matters for the 16K figures specifically. The Jenkins `daily_hip_ep_genai` log for the 26B at `in 16384` without the pool reads 9.51 TPS average against a 12.53 p50, a 37 ms per-token standard deviation and a 257 ms worst token — the decode transient dominates, and any GQA change measured against that baseline is measuring the allocator instead.

## Rebase details

Clean — no conflicts, despite six files overlapping with what `main` changed since the merge-base (`matmul_nbits_kernel.hip`, `hip_custom_kernels.h`, `hipdnn_ep_runtime.h`, `hipdnn_ep_runtime_state.cpp`, `hipdnn_ep_runtime_tensor.cpp`, `runtime_state_internal.h`). All twelve commits replayed. I diffed the pre- and post-rebase branch diffs against their respective bases: the only differences are blob hashes, hunk offsets, and one context line `main` renamed nearby, so the content is unchanged. Release build on gfx1151 is clean.

`pr1024-prerebase` is tagged locally at `59eb7517` if the previous head is ever needed.

## One new commit: `style: satisfy pre-commit on the perf-harness files`

`pre-commit` did not pass on this branch, before the rebase as well as after, so this is a pre-existing gap rather than rebase fallout. Three parts, none behavioural:

- clang-format reflows in `debug_log.h`, `op_profile.cpp`, `real/gqa.cpp`, and ruff-format across `tools/perf-harness`. Wrapping only.
- `flag_parity.py` drops a dead `inputs` local (ruff F841). It was set to `None` on both paths of the tokenizer probe and never read. Done by hand because ruff will not rewrite a tuple unpacking.
- A blank line after the licence block in each of the seven perf-harness `.ps1` files.

That last one is worth flagging properly, because the hook's own "fix" is destructive. `.ps1` maps to the `script` type in `LICENSES/settings.json`, whose `lineCommentStartPattern` is `^\s*#`. With `-f`, `licenseheaders` therefore treats the entire contiguous `#` run at the top of the file as the licence header it owns, and replaces it. On `ab_interleaved.ps1` that deleted 54 lines of measurement guidance; across the seven files, 140 lines. Running the hook and accepting its output silently destroys the documentation.

Every `.ps1` already in the tree escapes this by having a non-comment line directly after the licence (`param(` in the `.cursor/skills` scripts), and `.pre-commit-config.yaml` escapes it with exactly the blank line used here. So the separator is the existing convention rather than a new one, and the documentation is preserved intact.

## Measurement, in progress

Now re-measuring on the new base with CI's own case matrix from `daily_hip_ep_genai` — `vlm_benchmark.py`, `--max_tokens 128 --num_iterations 3 --warmup_iterations 1`, at `in` 19 / 128 / 2048 / 16384 with `--max_length` 8192 and 33024 respectively — across the three Gemma models available locally.

Three arms, because three of this branch's four optimisations are default-off env flags (`HIPDNN_EP_GQA_FUSE_APPEND`, `HIPDNN_EP_QMOE_ROUTER_GEMV`, `HIPDNN_EP_MATMUL_NBITS_WIDE_K`) and only the window chunking is unconditional: #1068 alone, this branch with flags unset, and this branch with all three on. The middle arm is what CI would see if this merged today.

Numbers to follow. The existing figures in the description were measured against a `main` that still had #961 and with a contaminated environment in the first round, so treat them as superseded until the new table lands.

BoarQing:
## Gemma performance on the new base, measured with CI's own case matrix

Rebased onto #1068 and re-measured. Headline: **the default-on part of this PR is a real prefill win at long context and nothing at short context, exactly as its gating predicts. The three default-off flags add no prefill gain and cost 2-3% of decode throughput.**

### Method

The case matrix is the Jenkins `daily_hip_ep_genai` job's, argument for argument: `vlm_benchmark.py` with `--image_path dog.jpg`, `--max_tokens 128 --num_iterations 3 --warmup_iterations 1`, at `in` 19 / 128 / 2048 / 16384 with `--max_length` 8192, 8192, 8192, 33024. Same machine, same ORT, same `amdhip64`, same model files; only `hipgpu.dll` and its paired `custom_kernels_gfx1151.dll` differ between arms.

Three arms:

| arm | contents |
|---|---|
| `base` | #1068 (`main` + #961 + the eviction fix) |
| `cand` | + this PR, flags unset — what CI would see if it merged today |
| `candflags` | same binary as `cand`, with `HIPDNN_EP_GQA_FUSE_APPEND=1 HIPDNN_EP_QMOE_ROUTER_GEMV=1 HIPDNN_EP_MATMUL_NBITS_WIDE_K=1` |

The third arm exists because three of this PR's four optimisations are default-off `env_enabled` gates; only the GQA window chunking is unconditional. A plain DLL-swap A/B would have measured only the chunking and reported the rest as "no effect".

Each arm gets its own `TEMP`, the profiler variables are cleared, the flag keys are written for every arm (removed where unwanted) so they cannot leak between arms, and a run is discarded unless the driver reports `Device: AMDGPU`. Every invocation was repeated with the arm order reversed; a delta that changes sign under reversal is position, not the change.

### TTFT, cand vs base, both orders

| model | case | forward | reverse | mean | verdict |
|---|---|---:|---:|---:|---|
| 26B-A4B-qmoe | in19 | +1.88% | −1.39% | +0.24% | sign flips — noise |
| 26B-A4B-qmoe | in128 | −0.05% | −0.46% | −0.25% | negligible |
| 26B-A4B-qmoe | **in2048** | **−8.76%** | **−8.72%** | **−8.74%** | consistent |
| 26B-A4B-qmoe | **in16384** | **−3.27%** | **−5.64%** | **−4.45%** | consistent |
| 12b-kquant | in19 | −0.55% | +0.00% | −0.28% | sign flips — noise |
| 12b-kquant | in128 | −3.80% | +0.00% | −1.90% | sign flips — noise |
| 12b-kquant | **in2048** | **−9.64%** | **−9.22%** | **−9.43%** | consistent |
| 12b-kquant | **in16384** | **−4.33%** | **−3.63%** | **−3.98%** | consistent |
| gemma3-4b | in19 | −0.22% | +1.24% | +0.51% | sign flips — noise |
| gemma3-4b | in128 | +0.13% | −0.67% | −0.27% | negligible |
| gemma3-4b | **in2048** | **−4.34%** | **−5.29%** | **−4.81%** | consistent |
| gemma3-4b | **in16384** | **−5.60%** | **−4.47%** | **−5.04%** | consistent |

The pattern is the change's own gating made visible. `in19` and `in128` are 274 and 392 prompt tokens, below the `window + kWindowChunkRows` threshold, so chunking never engages and those four rows flip sign between orders — which is what a null result should look like. Every case that clears the threshold improves, in both orders, on all three models.

Prefill throughput moves with it: +9.6% (26B @2K), +10.4% (12B @2K), +5.6% (gemma3-4b @2K), and +3.8 to +6.4% at 16K.

`gemma3-4b` was not previously covered. It benefits too, −4.8% at 2K and −5.0% at 16K, so the change is not specific to the 12B it was developed on.

### Decode throughput

`cand` does not move TPS: within ±1.4% everywhere, sign inconsistent. That is the expected result — this is a prefill change, and decode does not enter the chunked path.

`candflags` is consistently **worse** on TPS, on every model and every length:

| model | in19 | in128 | in2048 | in16384 |
|---|---:|---:|---:|---:|
| 26B-A4B-qmoe | −2.3 / −2.9% | −2.7 / −2.0% | −3.8 / −3.1% | −0.7 / −0.4% |
| 12b-kquant | −2.7 / −2.3% | −3.0 / −2.4% | −3.4 / −2.1% | −1.8 / −1.3% |
| gemma3-4b | −2.6 / −2.5% | −2.9 / −2.7% | −3.0 / −2.8% | −2.6 / +0.4% |

(forward / reverse.) Twenty-two of twenty-four measurements are negative, so this is not noise. And `candflags` shows no TTFT advantage over `cand` — the prefill gain is entirely the chunking. On this evidence the three flags are not ready to be defaulted on, and `HIPDNN_EP_QMOE_ROUTER_GEMV` in particular deserves a second look, since the router GEMV was the one expected to help decode.

One exception worth investigating rather than claiming: at 26B `in19` only, `candflags` peak memory drops to 1198-1201 MB against base's 1859-2054 MB, a 35-42% reduction that appears at no other length. I have not explained it and would not quote it as a result yet.

### Incidental corroboration for #1068

The same CI case run on `base` reproduces the allocator win independently, against the `daily_hip_ep_genai` log for the identical 26B `in 16384` case on a build without the pool:

| | CI log, no pool | `base` here, #1068 |
|---|---:|---:|
| TPS | 9.51 | **34.25-34.95** |
| ms/token, avg | 105.13 | **28.6-29.2** |
| ms/token, std | 37.20 | **0.46-0.55** |
| ms/token, max | 257.83 | ~30 |

Different harness, same case definition. The 37 ms standard deviation collapsing to 0.5 ms is the re-pinning transient going away, and 34.8 TPS lands on #961's original 35.03 claim.

### Caveats

**Arms were block-sequential within an invocation, not interleaved per case.** The loop is arm-outer, so `base`'s in2048 and `cand`'s in2048 are separated by other runs. That is weaker than `ab_interleaved.ps1`'s per-case pairing, and it is a consequence of the harness gap below. Order reversal is the control, and it works — the four null cases flip sign while the eight real ones do not, and the 2K deltas reproduce to within 0.5 percentage points across two independent invocations (−8.76 vs −8.72, −9.64 vs −9.22). But paired per-case statistics would be stronger, and I would want them before treating the 16K figures (which spread −3.3 to −5.6%) as precise.

**Absolute numbers are not comparable to CI.** TTFT at `in128` on the 26B is 830 ms here against CI's 2015 ms on the same 392-token workload — different machine and a much newer build. Only within-invocation deltas are quoted.

**16K needed the commit ceiling raised.** A 16K Gemma-4 run adds ~42 GB of commit on top of this box's ~30 GB resting charge against a 74.9 GB limit, so my guard rail had to go to 74 GB. Physical memory stayed above 31 GB free throughout, so the documented hard-reboot hazard was never in play. The first 16K attempt aborted on the guard for all three arms.

### Harness gap this exposed

`bench_tps.ps1` drives `model_benchmark.exe`, which cannot load a multimodal export, and `ab_interleaved.ps1` rejects `-Driver vlm` under `-Metric tps`. So this PR's own harness cannot measure decode throughput on the Gemma-4 models this PR targets, which is why the measurement above uses a separate CI-parity runner and a weaker pairing. Extending `bench_tps.ps1` to accept `-Driver vlm` and relaxing that guard would close it. I did not do it here — it is a harness change and belongs in its own PR rather than inside a rebase.

### Suggested next step

Split this PR. The window chunking is measured, consistent across three models and two arm orders, and defaults on: it is ready to review on its own. The three flagged optimisations currently cost decode throughput and buy nothing measurable, so they need either a fix or a justification, and they will hold up the part that works.

BoarQing:
## Scoped down to the runtime changes; the profiling work stays local

Force-pushed. This PR is now **10 files, +727/−21, all under `lib/Runtime`**, down from 40 files and +5553. Everything under `tools/` and every profiling hook has been removed and kept on a local branch.

### Why: the profiling commits were breaking `build_mock`

```
FAILED: [code=1] lib/Runtime/runtime_op_profile.bc
D:/a/hip-ep/hip-ep/lib/Runtime/op_profile.cpp:14:10: fatal error: 'hip/hip_runtime.h' file not found
```

`profile: decode-step RGP capture and a per-token memory-floor model` added `#include <hip/hip_runtime.h>` and a `hipDeviceSynchronize()` call to `op_profile.cpp`. That file is compiled into `runtime_op_profile.bc`, and the mock configuration (`build.py --mock`, no GPU/HIP/TheRock) has no HIP headers, so it cannot build. `build_mock` was the only failing check; `build_real`, `build-linux`, `gpu_test` and `pre-commit` all passed.

`op_profile.cpp` and `op_profile.h` were touched by that one profiling commit and by nothing else, so removing the profiling commits restores both files to the base exactly — `git diff` against `perf/allocator-pool-large-reland` for them is empty, and that base passes `build_mock`.

### What is still here

The four optimisations and their gates, nothing else:

| commit | what |
|---|---|
| `perf(gqa)` fuse the decode K rope and KV cache appends into one launch | `HIPDNN_EP_GQA_FUSE_APPEND`, default off |
| `perf(qmoe)` purpose-built narrow-N fp16 GEMV for the MoE router | `HIPDNN_EP_QMOE_ROUTER_GEMV`, default off |
| `perf(matmul_nbits)` widen TILE_N for tall-K decode GEMV shapes | `HIPDNN_EP_MATMUL_NBITS_WIDE_K`, default off |
| `perf(gqa)` chunk a windowed prefill so the key narrowing can engage | unconditional |
| `style` clang-format the new gqa and debug_log code | wrapping only |

I checked before splitting that none of the perf commits reference `rgp_capture_fence`, `op_profile` or `RGP_FENCE`, so the two halves are genuinely independent rather than merely separable on paper.

### What moved out

`tools/perf-harness/**` (README, 13 analysis scripts, 8 bench scripts, 2 capture scripts, `common.ps1`), `tools/rgp_parser` occupancy changes, and the `op_profile.*` plus `hipdnn_ep_runtime_tensor.cpp` profiling hooks. Held locally on `local/1024-with-profiling`, tagged `pr1024-with-profiling`.

Worth being explicit that this is a loss, not just a tidy-up: the measurements in my previous comment were produced with that harness, and it is not in the tree for anyone to reproduce them with. If the profiling work is wanted upstream later it needs the mock build dealt with first — guarding the HIP include and the `hipDeviceSynchronize()` behind the same mock switch the rest of the runtime uses would be the obvious approach, but that is a separate change and I have not done it.

### Measurements still stand

The numbers in the previous comment came from the unconditional window chunking, which is unchanged here. Restating the part that matters, TTFT versus base with the arm order reversed between two independent invocations:

| model | in2048 | in16384 |
|---|---:|---:|
| gemma-4-26B-A4B-qmoe | −8.76% / −8.72% | −3.27% / −5.64% |
| gemma-4-12b-kquant | −9.64% / −9.22% | −4.33% / −3.63% |
| gemma3-4b | −4.34% / −5.29% | −5.60% / −4.47% |

`in19` and `in128` flip sign between orders, which is the chunking correctly gating off below `window + 256` tokens.

The three default-off flags still cost 2-3% of decode throughput in 22 of 24 measurements with no prefill gain over the chunking alone, so my recommendation from the previous comment is unchanged: the window chunking is ready to review on its own, and the three flagged optimisations need either a fix or a justification before they should hold it up.

### Checks

`pre-commit run --from-ref perf/allocator-pool-large-reland --to-ref HEAD` passes. Release build on gfx1151 is clean. I did not reproduce `build_mock` locally — CI is the check for that, and the failing translation unit is now byte-identical to a base that passes it.

BoarQing:
## Profiled the three flags individually. One of them was the whole regression, and it is now mostly fixed.

My earlier comment measured all three flags on together and reported 2-3% of decode lost with no prefill gain. That could not say which flag was responsible, so this round turns them on one at a time. All arms share **one binary**, so nothing here can be a build difference.

### Which flag costs the decode

Decode ms/token against the same binary with all flags off. Three models, `in128` and `in2048`, both arm orders, so 12 points per flag.

| flag | mean | slower at | range |
|---|---:|---:|---|
| `HIPDNN_EP_GQA_FUSE_APPEND` | +0.31% | 10/12 | −0.48% .. +0.92% |
| `HIPDNN_EP_QMOE_ROUTER_GEMV` | +0.35% | 9/12 | −1.12% .. +2.24% |
| **`HIPDNN_EP_MATMUL_NBITS_WIDE_K`** | **+2.98%** | **12/12** | **+1.36% .. +3.80%** |

`fuse` and `router` are inside noise with signs that flip between orders. `widek` is slower at every single point in both orders on all three models, and it accounts for essentially the whole 2-3%.

Worth noting for `router`: its gate is not MoE-specific (`M == 1 && elem_size == 2 && N <= 256 && K >= 512`), and on the 12B a debug run logs no declines at all, which is consistent with it never matching — the 12B's heavy GEMVs are int4 `matmul_nbits`, not the fp16 the gate requires. So on non-MoE models it is a genuine no-op rather than a neutral win.

### Why WIDE_K regressed, and the fix

Replaying the promotion rule over `autotune/matmul_nbits/lut/gfx1151.json` shows what it touches. At `K > N` it promotes 34 of 139 dp4a points, and three of them are Gemma's:

| shape | what it is | ratio K/N | promotion |
|---|---|---:|---|
| 2048 x 2816 | k/v projection | 1.38 | `(32,1) → (64,8)` |
| 2816 x 4096 | o projection | 1.45 | `(64,1) → (128,8)` |
| 2816 x 8192 | MLP down | 2.91 | `(128,1) → (256,8)` |

The first two are the problem, and the mechanism is the block-count term rather than the ratio. `kWideKMinBlocks = 256` with `tile_n = 8` admits any N >= 2048, so Gemma's k/v lands exactly on the threshold — where Qwen3-30B's k/v had been *excluded* for leaving only 128 blocks after widening. The term that reads like the safety valve is what let them in. At 2.9 and 5.8 MB of weights these are bandwidth-bound, and the 1.78x the promotion is predicated on is an issue-rate ceiling that assumes the opposite.

So the scope limit is now `K >= 2N` instead of `K > N`: the saving is a fixed fraction of the integer-dot cost, so the shape has to be tall enough for that cost to dominate the weight stream, not merely exceed N. That drops the LUT count from 34 to 22 and removes both problem shapes. Qwen3-30B's o-projection sits exactly on the new boundary (N=2048, K=4096) and still promotes; all three of the rule's own Qwen tripwires still pass.

### It keeps the gain on the model it was built for

The promotion was designed and validated on Qwen3-30B, so tightening it needed a regression check there. 128/128, both arm orders:

| arm | round 1 | round 2 | mean TPS | vs off |
|---|---:|---:|---:|---:|
| off | 66.85 | 66.33 | 66.59 | — |
| `K > N` | 69.35 | 68.00 | 68.68 | **+3.1%** |
| `K >= 2N` | 68.76 | 68.41 | 68.59 | **+3.0%** |

The optimisation is real — 3% on Qwen — and the tightening keeps essentially all of it.

### And most of the Gemma cost goes away

| | mean | slower at | worst |
|---|---:|---:|---:|
| `K > N` | +2.98% | 12/12 | +3.80% |
| `K >= 2N` | +0.67% | 10/12 | +1.70% |

With the flag off, the two binaries are within ±0.9% with mixed signs, which is the expected no-op: the rule early-returns when the flag is disabled.

### Three things I am not claiming

**The residual is real and I have not explained it.** The 12B is still +1.1% to +1.7% in all four of its points. What still promotes there is the MLP down-projection at ratio 2.91 — decisively tall by the new test, and still not paying. So the ratio is a better discriminator than `K > N` but not the right one, and narrowing it further needs per-shape evidence rather than a third threshold.

**The exploration hook is broken.** `HIPDNN_EP_MATMUL_NBITS_WIDE_K_CFG` exists to pick the TILE_N target on the wall clock, which is exactly the question here. Forcing a config index inflates commit charge by ~20 GB on the 12B and ~16 GB on the 26B, costs 25% of decode, and aborts the 26B on my memory guard rail. That is a separate defect and it is why the target was not chosen by sweeping. Only config 17 `(256,8)`, closest to what the rule picks anyway, behaves normally.

**Enabling the flag changes numerics.** On the 12B over 32 greedy steps, tokens are identical but worst logit cosine is 0.99468 (mean 0.99915) — against an A/A control of exactly **1.00000000**, so the runtime is bit-deterministic and the divergence is the promotion reordering the dp4a accumulation. That fails the harness's own parity threshold. Same class as the chunking this PR already documents, but it means defaulting this flag on would need an accuracy check wider than three models.

**No RGP capture.** I dropped it. The causal chain is carried by the LUT replay plus the Qwen control, and the hook that would have supported a capture-based sweep is the broken one above. The "bandwidth-bound rather than issue-bound" reading is therefore an inference from the shapes and the measurements, not a counter measurement.

### Where this leaves the PR

The window chunking remains the part that pays and defaults on. Of the three flags: `widek` now helps Qwen by 3% and costs Gemma 0.67%, which justifies keeping it default-off; `fuse` and `router` are measurably neutral, and `router` provably does not fire on non-MoE models at all. If you want the PR smaller, those two are the candidates to drop, since neither has a measured benefit on any model tested here.

BoarQing:
## Swept the chunk height; 256 was not the optimum

The height was set from one measurement — gemma-4-12b at 2283 tokens, where 128 rows read 0.7% worse than 256 and the curve looked like a peak. I swept it properly and that result does not hold. `kWindowChunkRows` is now `kScoreChunkAlign` (128) rather than twice it.

### Method

Eight heights (0, 128, 192, 256, 320, 384, 512, 640) as eight arms **inside one invocation on one binary**, differing only in a local env override of the constant. That removes the last way a DLL-swap A/B can mislead: the arms cannot differ in anything but the number, because there is only one binary. Three models, `in2048` and `in16384`, both arm orders. 96 runs, no failures.

The override was local only and is reverted; the committed change is the constant.

### Result

TTFT against the unchunked path, averaged over both orders:

| height | 26B 2K | 26B 16K | 12b 2K | 12b 16K | g3-4b 2K | g3-4b 16K |
|---|---|---|---|---|---|---|
| **128** | **−8.8%** | **−5.9%** | **−10.2%** | **−6.1%** | **−4.9%** | **−4.5%** |
| 192 | −8.5% | −4.8% | −8.9% | −4.9% | −4.4% | −4.0% |
| 256 | −8.2% | −4.7% | −9.2% | −4.6% | −4.7% | −4.3% |
| 320 | −7.4% | −2.5% | −8.5% | −2.2% | −4.6% | −4.2% |
| 384 | −6.7% | −2.6% | −7.8% | −2.8% | −4.6% | −4.1% |
| 512 | −5.6% | 0.0% | −6.4% | −0.6% | −4.5% | −3.4% |
| 640 | −5.1% | +0.7% | −5.8% | +0.9% | −4.0% | −2.5% |

128 is the argmin in **11 of 12** model/length/order cells and wins on the mean in all 12. Restricted to the tile-aligned heights (128/256/384/512/640) the curve is monotone in all six columns, so shorter chunks scoring fewer keys per row beats the extra passes and the smaller `n`, at least down to the quantum.

The two heights that are *not* multiples of `kScoreChunkAlign`, 192 and 320, sit off that trend in four of the six columns. That is the same effect from the other side, and it is why 128 rather than something smaller: below one quantum the chunk stops being tile-friendly at all.

### Two things the sweep settled beyond the height

**The 16K gain comes from going below the byte budget, not from chunking.** `w640` was included as a control because that is roughly what the budget picks unaided at 16K. It measures within 1% of not chunking at all (22,473 vs 22,424 ms on the 12B). So at 2K this change enables chunking, and at 16K it shrinks a chunk the budget had already chosen — two different mechanisms, which is why the gain differs between the two lengths.

**Re-measured on this branch's own base**, not only where the sweep ran, at 2K over two reversed orders:

| model | 256 | 128 | delta | sign holds in both orders |
|---|---|---|---|---|
| gemma-4-26B | 1982 ms | 1970 ms | −0.59% | yes |
| gemma-4-12b | 2670 ms | 2644 ms | −0.97% | yes |
| gemma3-4b | 2328 ms | 2319 ms | −0.40% | no |

### Where this is soft

The margins between adjacent heights are 0.2-1.6%, near this harness's noise floor for TTFT, and I would not claim any single pair on its own. What carries the choice is the consistency — 12 of 12 cells, monotone in every column, and the direction reproducing on a second base.

gemma3-4b is the honest exception. Its curve spans 0.9% end to end, the 128-vs-256 sign flips between orders, and it gains 0.16-0.17%. The height simply does not matter for that model; the gain is specific to the two Gemma-4 models.

The 16K sweep needed the #1068 base, because with #961 reverted the 12B at 16K exhausts this box's commit limit and every arm aborts. That is not a property of the chunking, and the 2K numbers plus the re-measurement above are on this branch's own base.

BoarQing:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1585/

amd-yanjunz:

- PR: ROCm/hip-ep#1024
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/1024#issuecomment-5759839983)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1587/
- Main baseline: not usable, the baseline uses a different NODE
- Package build run: https://github.com/ROCm/hip-ep/actions/runs/35585190155
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6ab136c62225a21cf8e4c828
- Compare with latest daily baseline (fallback): http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fids%3D6ab0217915d05d850a51da84%26collection%3DHIP-EP-Dynamic&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=1587

amd-mingw (approved):

PR #1027, opened 2026-09-16
build: share LLVM 23 without the AMDGPU backend

amd-mounikk:
## Summary
- Shares rocmlirTriton's LLVM 23 with hip-ep (same stack as the green shared-LLVM work), but builds that LLVM with `LLVM_TARGETS_TO_BUILD=X86` only.
- hip-ep host codegen never selects AMDGPU; leaving it in the shared LLVM made `lld`'s `InitializeAllTargets*` pull ~20 MiB of AMDGPU into `hipgpu.dll`. Local build dropped the DLL from 116.4 MiB to 86.13 MiB.
- Busts the rocmlirTriton CI cache keys (`-clang-x86` / `linux-rocmlirtriton-x86-`) so Windows/Linux rebuild LLVM without AMDGPU.

## Notes
- rocmlirTriton device compilation still wants AMDGPU. This draft is to measure hip-ep size/CI with a host-only LLVM, not to claim rockCompiler still produces HSACO from this tree.
- Runtime HIP/hipBLASLt (TheRock `amdhip64_7.dll`, `hipblaslt.dll`) is unchanged.

## Test plan
- [ ] pre-commit
- [ ] Linux Build
- [ ] Windows Build (rocmlirTriton cache miss ~2h, then hip-ep)
- [ ] Confirm `hipgpu.dll` has no `LLVMAMDGPU*` / `AMDGPUTargetMachine`
- [ ] Smoke `hip-onnx-runner` on a non-LLM model with ORT 1.27 next to the exe


amd-mounikk:
Superseded by the LLVM 23 bump that drops rocmlirTriton entirely.

This PR tried to reach LLVM 23 through rocmlirTriton's vendored external/llvm-project subtree and then drop AMDGPU from that shared build. That is not possible: rocMLIR is itself an AMDGPU kernel compiler, and its librockCompiler archive names LLVMAMDGPUCodeGen/Desc/AsmParser as literal members, so LLVM_TARGETS_TO_BUILD=X86 fails while assembling rockCompiler.mri:

    ninja: error: 'mlir/tools/rocmlir-lib/LLVMAMDGPUAsmParser', needed by
    'mlir/tools/rocmlir-lib/rockCompiler.mri', missing and no known rule to make it

Building LLVM 23 directly from upstream sidesteps the conflict.

PR #1033, opened 2026-09-16
feat(hip-onnx-runner): add --no-run, --provider-options, --allow-cpu-fallback

mingyueliuh:
﻿## Summary

Adds three flags to `hip-onnx-runner` so a model can be loaded through the EP for its initialization side effects (compilation, MLIR dumps) without running inference:

- `--no-run` creates the session and exits before `Run()`
- `--provider-options key=value` (repeatable or comma-separated) passes arbitrary EP provider options
- `--allow-cpu-fallback` skips `session.disable_cpu_ep_fallback`

## Why

The model-compatibility tooling needs the graph the compiler actually sees, which the MorphiZen init pass can dump. Getting that dump requires two things the runner could not express: selecting an alternate EP configuration (`config_file`) and stopping after session creation.

The third flag came out of testing. An init-only configuration claims no node, so `session.disable_cpu_ep_fallback=1` (previously unconditional) fails session creation *after* the dump was already written. Relying on the non-zero exit and checking for the file would hide real failures, so the strictness is now opt-out instead of unconditional. It stays on by default, since a graph the EP failed to claim should still surface as a session error rather than silently running on CPU.

## What

- `--no-run` skips input generation, inference and output dumps; it warns when `--dump-level` was also given and tears down in the same order as the normal path (session released before the EP library is unloaded).
- Provider options are parsed before the EP library is loaded, so a malformed entry fails fast. Merge order is environment defaults, then `--provider-options`, then `--mlir-dump-dir`: an explicit CLI option beats an ambient env var, while the dedicated dump flag keeps ownership of `dump_dir` (and warns when it overrides a conflicting value). The effective options are printed.
- `--allow-cpu-fallback` prints a notice and leaves `session.disable_cpu_ep_fallback` unset.

## Test plan

| Case | Result |
|---|---|
| `-n --no-run` | session created, no inference, exit 0 |
| EP + `--no-run --provider-options artifact_format=LLVM_IR` | option printed, session compiled (4.6 s), no inference, exit 0 |
| EP + init-only config, no `--allow-cpu-fallback` | session creation fails (expected; this is what motivated the flag) |
| EP + init-only config + `--allow-cpu-fallback` | exit 0, 13 ms (no compilation), dump written |
| Same on a 1.8B ONNX model | exit 0, 2.3 s, 0.42 MB text MLIR dump |

## Checklist


amd-mingw (approved):
LGTM

PR #1041, opened 2026-09-17
feat(hip-to-tosa): lower comparison, logical and sign ops to TOSA

nvunnam57128:
## Summary

Adds HipToTosa conversions for `hip.equal`, `hip.less`, `hip.and`, `hip.or`, `hip.not` and `hip.sign`. Until now a `rock.kernel` containing any of them could not be absorbed into a rocMLIR fused kernel, because the pass had no pattern for them.

Because `OnnxToHip` already decomposes `onnx.Greater` into `hip.less(B, A)` and both or-equal forms into `hip.not(hip.less(..))`, this also reaches ONNX `Greater`, `GreaterOrEqual` and `LessOrEqual` without any new op or pattern for them — for graphs whose BOOL is carried as `i1`. `OnnxToHip` preserves the importer's result type, and where that is `ui8` (ORT's byte-boolean representation, as in `test/lit/Conversion/onnx-to-hip/test_greater.mlir`) the op stays on the existing runtime path rather than converting; see the type gating below.

## Related issue or design

None.

## Why

The comparisons cannot reuse `BinaryConverter`. They produce `i1` while their operands carry the type being compared, and `isTosaCompatibleOperand` checks each operand's element type against the _result's_ — which would reject every well-formed comparison. The existing comment on `BinaryConverter` already called this out; `ComparisonConverter` is the pattern it was pointing at.

TOSA spells the ordered comparisons as `greater` and `greater_equal` only, so `hip.less` becomes `tosa.greater` with its operands swapped. This is the same identity `OnnxToHip` uses in the other direction, so an ONNX `Greater` swapped on the way in is swapped back here and ends up as the `tosa.greater` it started as.

`hip.and`, `hip.or` and `hip.not` are spelled with the `tosa.bitwise_*` ops rather than the `tosa.logical_*` ones. The logical spelling converts cleanly in this pass and then dies downstream: rocMLIR's `RockTosaToElementwise` has no pattern for `tosa.logical_and`/`_or`/`_not` and marks every surviving `tosa` op illegal. Confirmed against `rocmlir-opt --rock-tosa-to-elementwise`, where `tosa.bitwise_and` and `tosa.bitwise_xor` lower to `arith.andi`/`arith.xori`. On `i1` the bitwise and logical operations coincide, and the `i1` gate is what keeps that true.

## What

- `ComparisonConverter<HipOpTy, TosaOpTy, SwapOperands>` — new pattern for `hip.equal` and `hip.less`. Runs the broadcast check against a result-shaped tensor of the _operand_ element type.
- `hip.and`/`hip.or` map to `tosa.bitwise_and`/`tosa.bitwise_or`. `hip.not` becomes `tosa.bitwise_xor` against an all-ones constant, since `x ^ true` is `!x` on `i1`; `tosa.bitwise_not` has no downstream pattern either.
- `SignConverter` — TOSA has no sign op, so `hip.sign` expands to the nested select its definition spells out: `select(x > 0, 1, select(0 > x, -1, 0))`, guarded by a self-compare on floats so NaN is forwarded rather than collapsed to zero. ONNX defines `sign(NaN) = NaN` and `lib/Runtime/real/sign.cpp` propagates it as a deliberate delta from ORT, so without the guard a fused model would disagree with the unfused one. Documented with a `Before:`/`After:` IR example.
- **These ops are claimed by element type and shape, not outright.** They are registered with `addDynamicallyLegalOp`, not `addIllegalOp`. An op marked illegal that this pass cannot rewrite aborts the whole function's conversion and takes the fusible ops around it down too, which is strictly worse than leaving it for the runtime lowering that already handles it. Each predicate therefore claims exactly what its pattern accepts — element types, static shapes, and every tensor operand's broadcastability — so nothing reaches a pattern that then refuses it.
- The type gates are allow-lists. `Tosa_FloatTensor` is `AnyFloat` and `Tosa_Int` is any signless or unsigned integer, so `f64`, `f8`, `f80`, `i4` and `i128` all satisfy the op verifiers while nothing downstream can lower them. Floats are restricted to the `f16`/`bf16`/`f32` set `ConvConverter` and `GemmConverter` already use; integers to the signless widths ONNX produces. Unsigned operands are excluded from the comparisons because TOSA integers are signless, so a `ui8` 255 would compare as `-1`.

## Test plan

New `test/lit/Conversion/hip-to-tosa/test_compare_logical.mlir` covering the op mappings, broadcasting through an `i1` result (both size-1 and rank-extension), the sign expansion on `f32` and `i32` including the NaN guard, the `rock.kernel` guard, and the forms the pass declines: `ui8` booleans and `ui8` comparison results, unsigned and `f8` comparison operands, `f64` and `i4`, non-`i1` operands to the logical ops, `i1` to `sign`, dynamic shapes, un-broadcastable operands, and a static result with a dynamic operand.

The file has **no `expected-error` cases**, which is the invariant rather than an accident: a diagnostic appearing there means a legality predicate and its pattern have drifted apart.

```
ninja hip-mlir-opt          # clean build, Release/Ninja/MSVC
llvm-lit test/lit           # 523 passed, 0 failed, 9 unsupported (532 discovered)
```

No regressions; the pre-existing `hip-to-tosa` tests all still pass.

## Notes for reviewers

`hip.atan`, `hip.round` and `hip.mod` are deliberately **not** converted. None has a TOSA counterpart in the LLVM this tree pins, and rocMLIR's `MIGraphXToTosa` has no precedent for any of them either, so each would be an invented multi-op expansion: `round` is ties-to-even (so `floor(x + 0.5)` is wrong on halfway cases and it needs an odd/even correction), `mod` branches on its `fmod` attribute and the integer case wants floor-division while `tosa.intdiv` truncates toward zero, and `atan` has no closed form in TOSA at all. They are left dynamically legal too, so the pass leaves them alone rather than failing on them.

The `ui8` boolean family is the main coverage gap this leaves. Normalizing importer BOOL results to `i1` would widen the end-to-end path, but that belongs in `OnnxToHip` rather than here, where it would mean inventing a `ui8`↔`i1` bridge around every TOSA comparison.

## Checklist

AI assistance: the conversion patterns, the lit test and this description were drafted with Cursor (Claude Opus 5). Validated by building `hip-mlir-opt` and running the full lit suite as above. The TOSA op availability and the `OnnxToHip` decompositions were verified against `TosaOps.td` and the `OnnxToHip` sources; the downstream `bitwise` vs `logical` gap and the predicate/pattern mismatches were each reproduced directly (`rocmlir-opt --rock-tosa-to-elementwise` for the former, `hip-mlir-opt --convert-hip-to-tosa` on the three aborting forms for the latter) rather than assumed.


asparkhi (approved):

asparkhi:
@copilot resolve the merge conflicts in this pull request

nvunnam57128:
@copilot remove these two commit ids
@nvunnam57128
[fix windows build fallback for prebuilt artifacts](https://github.com/ROCm/hip-ep/pull/1041/commits/e13a7a48163aed9955f246f205546f8d2eb3fdf6) 
[e13a7a4](https://github.com/ROCm/hip-ep/pull/1041/commits/e13a7a48163aed9955f246f205546f8d2eb3fdf6)
@nvunnam57128
[scope windows prebuild fallback artifacts per run](https://github.com/ROCm/hip-ep/pull/1041/commits/8e94fb27964cc408262498b076ecbc0729635f2e) 
[8e94fb2](https://github.com/ROCm/hip-ep/pull/1041/commits/8e94fb27964cc408262498b076ecbc0729635f2e)

asparkhi (approved):

nvunnam57128:
@copilot resolve the merge conflicts in this pull request

nvunnam57128:
@copilot resolve the merge conflicts in this pull request

nvunnam57128:
@copilot resolve the merge conflicts in this pull request

nvunnam57128:
@copilot resolve the merge conflicts in this pull request

PR #1043, opened 2026-09-17
kernels: int8/dp4a MatMulNBits paths, GQA autotune key bucketing, elementwise flat path

penglu12-quark:
Kernel-side work that came out of getting EAGLE-3 and MTP speculative decoding running on gfx1151
(Strix Halo) against Qwen3.5-9B and Qwen3.6-27B int4 targets.

Merged up to `744e7bb0`, so the diff reads against current `main`. Opened from a branch in this repo
rather than a fork so the GPU test job can post its own result tables (a fork PR only gets a
read-only token, and both publish steps fail with `Resource not accessible by integration`).

## What is here

| File | Change |
|---|---|
| `lib/Runtime/Kernels/hip/matmul_nbits_kernel.hip` | int8 GEMV paths (`matmul_nbits_kernel_i8`, `matmul_nbits_gemv_kernel_i8`), a dp4a GEMV kernel, an activation-quantisation kernel, M-blocked GEMV for small M, and per-shape autotune selection across the gemv / dp4a / wmma families |
| `lib/Runtime/Kernels/hip/gqa_kernel.hip` | bucket the KV length by `log2` in the flash-decode autotune key |
| `lib/Runtime/Kernels/hip/elementwise_binary_kernel.hip` | `ew_flat_kernel` fast path for the no-broadcast case, where both operands already match the output shape |
| `lib/Runtime/Kernels/include/hip_custom_kernels.h` | declarations for the new entry points |
| `lib/Runtime/real/matmul_nbits.cpp` | dispatch into the new paths; scratch sized `M*K` instead of `K` |
| `lib/Runtime/debug_log.h` | `HIPDNN_EP_MATMUL_DP4A_MAX_M` lever and dp4a doc comments |
| `scripts/run_spec_decode_pipeline.py`, `docs/spec_decode_pipeline.md` | the benchmarking pipeline these kernels were tuned against |

The last row is independent of the kernels and can move to its own PR if you would rather review it
separately.

## Behaviour changes worth flagging

Two things change what happens by default, not just what is available behind a lever:

- **dp4a eligibility widened from `M == 1` to `M <= 7`.** Tunable via `HIPDNN_EP_MATMUL_DP4A_MAX_M`,
  which defaults to 7. Speculative decoding verifies several draft tokens per step, so the shapes
  that matter are exactly the small-M ones the old `M == 1` guard excluded.
- **`HIPDNN_EP_EW_FLAT` defaults on.** On Qwen3-8B prefill at 200 positions, the 72 multiplies go
  from 6.4 ms to 3.6 ms.

## The GQA autotune key change

This is the one I would most like a view on, because it is a caching-behaviour change that affects
every GQA user, not just speculative decoding.

`flashDecodeKey` previously included the exact KV length `skv`. In a decode loop `skv` grows by one
per token, so every call misses the autotune cache and the tuner re-tunes forever — the cache can
never do its job in the workload it exists for. Bucketing by `log2(skv)` means the key only changes
when the context doubles, so a generation costs a handful of tunes instead of one per token.

For scale on why this became blocking rather than a nice-to-have: in the speculative-decoding
benchmarks these kernels were built for, autotune warming accounted for 66% of wall-clock time on
the 27B target and 82% on the 9B. Those runs already include this change, so I am quoting it as
motivation, not as a measured before/after — I have not run a clean A/B and would want to before
claiming a number. If you prefer, this hunk is small and self-contained and can land on its own
ahead of everything else.

## Conflict resolution in `matmul_nbits_kernel.hip`

Upstream reworked this file substantially since the fork point, and the timing helpers collided:
`#996` added `settleClock()` / `timeConfigPeak()` for gfx1150 LUT generation, which time a config at
its peak with a settled clock, while this branch probes configs against a cold cache by evicting L2
before each timed launch. Both are kept — `timeConfigPeak()` now takes a `void* evict` and calls
`gemvEvict()` before each launch, so an eviction buffer is used when one is supplied and the
upstream peak-timing behaviour is unchanged when it is not.

## Testing

In-tree kernel suites, run against this branch on gfx1151 (Radeon 8060S), group size 128,
K=5120, N=2880, at the M values speculative decoding actually hits:

| Suite | M=1 | M=4 | M=128 |
|---|---|---|---|
| `gemm_fp16u3` (u3) | 2880/2880, max_abs 0.0005 | 11520/11520, max_abs 0.0010 | 368640/368640, max_abs 0.0039 |
| `gemm_fp16u3` (u4) | 2880/2880, max_abs 0.0010 | 11520/11520, max_abs 0.0020 | 368640/368640, max_abs 0.0078 |
| `gemm_fp16i8` | 2880/2880, max_abs 0.0312 | 11520/11520, max_abs 0.0625 | 368640/368640, max_abs 0.1250 |

Every element matches the NumPy reference within the suite tolerance; the worst case in each column
is one fp16 ULP at that output magnitude. The same u3/u4 suite built from `main` at M=128 gives
identical error (0.0039 / 0.0078), so accuracy is unchanged. u4 throughput at M=128 is 0.158 ms
against 0.165 ms for `main` on the same machine.

For the GQA change, `gqa/decode` and `gqa/prefill` have test drivers but no committed `Makefile` —
`.gitignore` line 7 is a bare `Makefile` rule that matches at every level, so only the suites that
were force-added past it build from a clean checkout. I built both from the `hipcc` command lines in
their own READMEs and ran them against this branch and against `main`:

- `gqa/decode`: all 52 cases pass on both, with the same verdicts and the same kernel path chosen.
  relL2 against the CPU reference stays in 2.7e-4 to 2.9e-4.
- `gqa/prefill`: one case, `llama-win-d128` (D128 with a 128-wide window), fails **identically on
  `main`** — the suite expects the kernel to decline the shape and it now returns `rc=0`. That looks
  like a stale expectation left behind when windowed D128 gained support, not something this branch
  causes. Every other case passes with matching verdicts.

On latency, most decode cases land within run-to-run noise of `main`, but a handful came out
substantially faster on this branch on the autotuned path — up to 37x on one small-`skv` case. That
is the shape of the effect the bucketing is meant to have, since `main` re-tunes on cache misses,
but it is one run per case and I would want repeats before quoting any of it as a result.

End-to-end, this was exercised through the EAGLE-3 / MTP speculative decoding harness on gfx1151 —
Qwen3.5-9B and Qwen3.6-27B int4, greedy, across a 480-instance Spec-Bench sample — with the
speculative arms checked for token-identical output against a no-drafter baseline.

## Note on the GPU test job's L2 table

The L2 numbers this job reports (259.905, 25.19, 39.54) are the same on `main`; they are not a
regression from this branch. The metric is `sqrt(sum over all elements of (ep - cpu)^2)` with no
normalisation by element count or reference magnitude, so it scales with tensor size: at the
368,640-element output above, the 0.01 threshold allows an RMS error of 1.65e-5 per element, roughly
470 times smaller than the single fp16 ULP the kernels are already within. Worth treating as a
separate issue against the threshold rather than a gate on this PR.


amd-yanjunz:

- PR: ROCm/hip-ep#1043
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/1043#issuecomment-5711510532)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1324/
- Main baseline commit: 2044d66
- Main baseline build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1306/
- Package build run: https://github.com/ROCm/hip-ep/actions/runs/35197640160
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6aabc4a811bb40542ac18b32
- Compare with main baseline 2044d66: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fdatabase%3Ddevelopment%26collection%3Ddaily_hip_ep_genai%26id%3D6aab96b77b0e9f197d204efa&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=1324

amd-mingw (changes requested):
some perf seems drop a lot from the test result: https://github.com/ROCm/hip-ep/pull/1043#issuecomment-5711529594

amd-yanjunz:

- PR: ROCm/hip-ep#1043
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/1043#issuecomment-5755262531)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1555/
- Main baseline build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1528/
- Package build run: https://github.com/ROCm/hip-ep/actions/runs/35559898127
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6ab0cc292225a21cf8e4c6a2
- Compare with main baseline 8c888c0: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fdatabase%3Ddevelopment%26collection%3Ddaily_hip_ep_genai%26id%3D6ab012a615d05d850a51da72&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=1555

amd-yanjunz:

- PR: ROCm/hip-ep#1043
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/1043#issuecomment-5757240416)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1575/
- Main baseline build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1528/
- Package build run: https://github.com/ROCm/hip-ep/actions/runs/35559898127
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6ab0fc0a87baba1ecde4c6e7
- Compare with main baseline 8c888c0: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fdatabase%3Ddevelopment%26collection%3Ddaily_hip_ep_genai%26id%3D6ab012a615d05d850a51da72&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=1575

penglu12-quark:
> some perf seems drop a lot from the test result: [#1043 (comment)](https://github.com/ROCm/hip-ep/pull/1043#issuecomment-5711529594)

hi @amd-mingw , Please help to review again!

sapareek:
@penglu12-quark  can we check if Auto Tuning is really needed here? 
@ZhanxingPu please review auto tuning part of the kernels. 
Python script can be its own PR - that is better.

penglu12-quark:
> if Auto Tuning is really needed here?

Autotuning needed here: speculative decoding verifies at `M = k+1`, which lands
right on the GEMV/WMMA crossover, so auto tuning help to perform better for the verification pass.

ZhanxingPu:
> @penglu12-quark can we check if Auto Tuning is really needed here? @ZhanxingPu please review auto tuning part of the kernels. Python script can be its own PR - that is better.

I reviewed the autotuning changes against `origin/main`. I think the following issues should be addressed before merging:

  1. The newly added GEMV-M, dp4a-M, and GEMV-M-i8 tuners use `WARMUP=1` followed by one five-launch aggregate measurement. This does not follow the existing
  `origin/main` methodology and can mis-rank candidates due to APU clock drift and candidate order. Please reuse the existing `settleClock()` and
  `timeConfigPeak()` helpers, including the positive-duration guard. Passing `nullptr` for cache eviction is reasonable when comparing candidates in the same
  weight-traffic class.

  2. `preferWmmaTier()` always measures WMMA first and GEMV second. Either callback may also trigger its own config sweep, so the two families can be measured
  under different thermal/cache conditions. Please resolve the child configs first, settle the clock, and compare fixed-config launches using interleaved or
  order-reversed rounds before caching the result.

  3. The new `tier_cache` key omits `has_bias`. GEMV fuses bias, while WMMA uses an additional bias kernel, so biased and bias-free calls may have different
  family winners. Please include `has_bias` in the key.

  4. These new small-M and family decisions have no offline-LUT representation, so normal execution can perform mandatory runtime sweeps for every new shape.
  Please either extend the LUT, or keep these runtime sweeps behind explicit `online` mode and use a validated heuristic in normal lookup mode.

  For GQA, flash decode was already log2-bucketed on the PR base. This PR actually changes the v7/v8 prefill cache keys from exact `skv` to buckets. Please
  update the description and provide lower/interior/upper-point measurements showing that the v7/v8 winner remains stable within each bucket; otherwise the
  cached result depends on which `skv` is encountered first.
  
cc @amd-mingw

penglu12-quark:
Thanks — all five addressed, pushed as `9dd2170c`.

**1. Timing methodology.** `tuneGemvMConfig`, `tuneDp4aMConfig` and `tuneGemvMConfigI8` now call `settleClock()` once before the candidate loop and  `timeConfigPeak()` per candidate, with the same `WARMUP=3 / ITERS=5 /  kFloorIters=15` and positive-duration guard the M==1 tuners use. The `WARMUP=1`  plus single five-launch aggregate is gone. Agreed on `nullptr` eviction, and the  code now states why rather than leaving it implicit: every candidate in a `gemvMCandidates()` list reads the weight the same number of times, so they are  one weight-traffic class and there is nothing for a cold probe to expose. The  M==1 tuners, which do compare across traffic classes, still evict.

**2. `preferWmmaTier()` ordering.** Restructured as you describe: both child callbacks run once first so each family's own config sweep is resolved and cached before anything is timed, then `settleClock()`, then four comparison rounds with the order reversed on alternate rounds (WMMA first on 0 and 2, GEMV first on 1 and 3), each a single isolated launch per family. Median of four per family, rejecting non-positive durations, and only then cached.

**3. `has_bias` in the tier key.** Fixed — `preferWmmaTier()` takes `has_bias`, the key carries a `_b1`/`_b0` suffix, and the dispatch site passes `bias != nullptr`.

**4. Offline-LUT representation.** Took your second option: the sweeps are  behind  IPDNN_MATMUL_AUTOTUNE_MODE=online`,  and lookup mode uses a closed-form  heuristic with zero tuning launches. Tier tuning follows the same gate and  defaults off in lookup mode, where the fixed `kWmmaMinM` cutoff decides;
`HIPDNN_EP_MATMUL_TIER_TUNE=1/0` overrides either way.

The heuristic is fitted to the sweep's own measurements rather than argued from  the hardware model, scored per shape against the sweep's winner and weighted by how many MatMulNBits nodes of that shape the graph runs — a projection it gets wrong runs once per layer while the lm_head runs once per token, and an unweighted average hides exactly that. Ranked by the worst of three measured sets rather than the mean, which is what caught a real bug: the rule keyed
`tile_n` off `M`, but `M` only *selects* `tile_m` (M<=2 -> 2, 3..4 -> 4, 5..7 -> 8), and what shares the register budget with `tile_n` is `tile_m`. While the fit reached only M=4 the two were the same variable; at M=5..7 it asked for `partial[8][4]` and spilled, 2.35x on 5120x6144 uint4. That is not an edge case for MTP — at k=3 the 27B's verify width averages 5.9 and reaches 11.

**5. v7/v8 prefill keys.** You are right on both counts — flash decode was already bucketed on the base, and the actual change here is v7/v8 prefill going from exact `skv` to buckets. Description corrected.

On stability, measured rather than assumed: `gqa_autotune_sweep` times every candidate at three `skv` per bucket (first, middle, last) with `max_seq` pinned to the top of the bucket so the KV stride is constant and `skv` is the only thing moving, at `sq` 2 and 8 since the calls that made the exact-`skv` key miss  are speculative verify passes rather than prompt prefills.

amd-yanjunz:

- PR: ROCm/hip-ep#1043
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/1043#issuecomment-5778285230)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1626/
- Main baseline: pending package run selection

amd-yanjunz:

- PR: ROCm/hip-ep#1043
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/1043#issuecomment-5778514476)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1631/
- Main baseline: pending package run selection

amd-yanjunz:

- PR: ROCm/hip-ep#1043
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/1043#issuecomment-5786050350)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1648/
- Main baseline: pending package run selection

penglu12-quark:
hi @sapareek  @amd-mingw  @ZhanxingPu ,
Please help to review it again! 

cc @ChaoLi-AMD

CaymanYang:
I see all points Zhanxing mentioned have been modified. LGTM. Thanks.

amd-yanjunz:

- PR: ROCm/hip-ep#1043
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/1043#issuecomment-5803960744)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1683/
- Main baseline: pending package run selection

amd-yanjunz:

- PR: ROCm/hip-ep#1043
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/1043#issuecomment-5804752225)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1684/
- Main baseline: pending package run selection

amd-yanjunz:

- PR: ROCm/hip-ep#1043
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/1043#issuecomment-5805866477)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1686/
- Main baseline: pending package run selection

amd-mingw (approved):

amd-yanjunz:

- PR: ROCm/hip-ep#1043
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/1043#issuecomment-5806989242)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1693/
- Main baseline: pending package run selection

PR #1045, opened 2026-09-17
feat(hipsr): compile a model through the hipsr pipeline end to end

zpye:
## Summary
- Opt in to the hipsr lowering path from the compiler driver with `HIPDNN_EP_HIPSR=1`, and add one E2E matmul model that compiles and runs on the GPU under `HIPDNN_EP_STRICT=1`.
- Load the modeled ONNX dialect on that path (`convert-onnx-to-hipsr` matches ODS op classes), and share `registerConvertToLLVMInterfaces` between the driver and `hip-mlir-opt`.
- Stop trailing-padding the constants file (VRAM size is last offset plus size), and keep a cwd `DiskFileSystem` fallback alive until the pass manager finishes so `hip-compiler` can still write constants.


wcy123 (approved):
LGTM.

PR #1046, opened 2026-09-17
feat: add hip-fusion-transform pass and migrate QDQ fusion to the HIP layer

amd-genmingz:
### Summary

Introduces `hip-fusion-transform`, a HIP-to-HIP PDLL rewriting pass that runs after shape inference, and moves every QDQ fusion pattern from the ONNX layer onto it. The ONNX-layer PDL machinery is deleted rather than left dormant.

### Related issue or design

No tracking issue. The new pass, its pipeline position, and its ordering constraint are documented in [`docs/pipeline_pass_menu.md`](docs/pipeline_pass_menu.md); the pass contract lives inline in `Passes.td` and `Pipelines.h`.

### Why

Fusion need to be **fully contained within the HIP dialect so that it no longer depends on the ONNX dialect**. The old patterns were rooted on `onnx.*` ops, which tied Q/DQ fusion — a HIP-level concern about which quantized HIP ops exist — to the shape of the incoming ONNX graph. The migrated patterns are rooted on `hip.*` ops, reaching into `tensor.*` only where the conversion itself emits it, so the HIP dialect now owns its own optimizations and any frontend that lowers into it inherits them.

### What

- **New pass**: `hip-fusion-transform`, inserted in the ONNX-to-HIP pipeline tail directly after `InferShapes`. It must precede `hip-externalize-constants`, because its scale and zero-point constraints read `hip.constant` payloads; running it later makes every fusion silently decline.
- **New CMake module**: `cmake/HipPdllPatterns.cmake` generalizes PDLL → bytecode → embedded byte array, parameterized by entry point, symbol, and target name.
- **Patterns**: eight pattern files plus one aggregate entry point, backed by native helpers in `hip_fusion_transform.hpp`.
- **New mechanism — round-trip elimination** (`QdqRoundTrip.pdll`): replaces the six-op SISO fusion with three patterns that drop matching Q/DQ pairs and re-run the layout op on quantized data, covering the DPS case, the tensor-dialect case, and the degenerate `DQ → Q` pair left by a dissolved `onnx.Identity`.
- **`qlpnormalization`**: `onnx.LpNormalization` is decomposed before this pass runs, so the pattern recognizes the equivalent `hip.rms_norm` via the L2 identity `L2(x) = rms_norm(x, 1/sqrt(N), eps=0)`, comparing the scale bitwise to match the decomposer's rounding.
- **Removals**: `lib/Conversion/OnnxToHip/pdl/` (nine files, including an 839-line helper header), 159 lines of embed logic, and the PDL dialect dependencies. This also removes a latent trap where disabling the PDLL build made `convert-onnx-to-hip` fail outright.
- **Tests relocated**: `test/lit/Conversion/onnx-to-hip/test_q*.mlir` → `test/lit/Transforms/HipFusionTransform/`.
- **Target rename**: the embed target is now `HipFusionTransformPdll`; the wheel and tool references follow it.
- **Temporary, not part of this feature**: the two CI workflow files. See the TODO below.

### Test plan

- [x] 30 lit test functions across 8 files in `test/lit/Transforms/HipFusionTransform/`, covering every migrated op
- [x] CI is returning its focus to model accuracy and performance, and we expect them to remain unchanged.

### Notes for reviewers

- `5773470a3` (sample plugin test carrier fix) also rode along on this branch and could be split out; it does not touch the fusion path.
- `qlpnormalization` now depends on the decomposer's output shape. If the decomposition changes, the pattern silently stops firing rather than failing, which costs the fusion but not correctness.
- Round-trip and layout patterns require single-use results, which is slightly narrower than the ONNX-layer behavior.

### AI Assistance

The code was developed collaboratively with Cursor.


amd-genmingz:
focus model: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1427
all model(with perf): http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1436

amd-mingw (approved):
Thanks! LGTM.

PR #1047, opened 2026-09-17
perf(qgemm): vectorize tiled staging loads and double-buffer the tile LDS

AMDmoore:
## What

Two changes to the tiled quantized GEMM's k-step, both in staging rather than in the dot4 work itself.

**Quad-level bounds guard.** The `kDotK` elements a thread stages are adjacent along the fast axis of `A` and of a `[N, K]` `B`, but the guard that keeps a tile from reading past the end of the matrix was applied per element. That put a branch and an independent 64-bit address on every staged load, so the loads could not coalesce and the 4-bit reads could not share the byte that holds two of their nibbles. Hoisting the guard to the whole quad lets the fast path issue one wide load per operand; a ragged `k` edge still takes the element-wise form and computes the same values.

On the `uint16 x int8` instance this turns `4x global_load_u16` into `1x global_load_b64` for `A`, and `4x global_load_i8` into `1x global_load_b32` for `B`.

**Double-buffered tile LDS.** The tile LDS was single-buffered, so the barrier closing each k step ordered the next tile's global loads behind the current tile's dots and left their latency exposed in front of the LDS store that consumes them. Two alternating buffers remove that barrier. A wave stages a tile only after reading the previous one, and the barrier at the head of each step keeps it from running more than one tile ahead, so the buffer it overwrites was last read a step earlier and every wave has passed that point. One barrier after the loop hands the LDS back to the caller. LDS footprint doubles from a small base and does not change occupancy.

## Relationship to #1044

This branch **supersedes #1044**. The staging rework is written on top of that PR's specialization commit, which is included here as the first commit — `main` has none of `QGemmTileLds`, `kTransA`, or the dispatcher, so the change does not stand alone.

The specialization commit's message has been corrected while cherry-picking. It previously attributed its modest gain to "a per-iteration floor unrelated to instruction issue". That was wrong: the flags are launch-uniform, so the bodies the specialization removes were already being branched over at runtime, and the static instruction count it reduced was never the executed count. There is no separate throughput floor.

## Results

Microbenchmark over the tiled kernel's hot prefill shapes, measured against the specialization commit this branch also carries (interleaved A/B, same process, same clocks):

- geometric mean ~1.16x, best shape ~1.38x, no shape regressed
- the specialization commit was itself ~1.06x over `main`, so the two together are roughly ~1.23x on those shapes

End to end on a LoRA decoder model with an 8k prompt, swapping only the kernel DLL so both runs share one compiled model artifact:

- time to first token ~+7%
- decode throughput flat, as expected — decode runs the GEMV path, not the tiled kernel

## Validation

- **Numerics unchanged.** Output digests over 14 shapes are byte-identical before and after: aligned 4-bit and 8-bit cases, `M`/`N`/`K` ragged in every combination, shapes small enough to trigger split-K, and `M = 1`.
- **No race from the removed barrier.** Six repeat runs of the digest set produce one distinct result.
- **End-to-end text unchanged.** Greedy decode output on the LoRA model is byte-identical before and after (same length, same SHA-256).

## Temporary commits — drop before merge

The six `ci:` commits are cherry-picked from #534 and are byte-identical to that PR's workflow files. They are only here so CI can resolve the OGA dependency on this branch, and should be dropped once #534 lands.

## AI assistance

Developed with Cursor. The AI assisted with the ISA analysis that located the cost (instruction mix of the k-step, load widths, which blocks are branch-guarded), the hardware microbenchmarks used to rule out earlier hypotheses, the kernel edits, and the differential correctness harness. Validated as described above: digest equality across shapes including ragged and split-K cases, repeat runs for race detection, interleaved A/B for performance, and an end-to-end byte-identical output check. Commits carry the `Co-authored-by` trailer.


AMDmoore:
benchmark:
http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6aac0c263b5d33c8ffc18bf6

PR #1051, opened 2026-09-17
perf(qgemm): optimize W4 prefill kernels

donghanbj:
## Summary

- specialize the uint16 x signed-W4 transposed-weight prefill path with packed int4 loads and co-issued A/B memory operations
- add shape-gated 64x64x32 and M=128 tiled kernels for the fixed 128-token PSU prefill graph
- tune split-K occupancy targets for the PSU N=32 and N=1024 projections
- add a standalone correctness/performance microbenchmark covering base and LoRA projection shapes

## Performance

PSU_MF_LORA_GEMM-model-2.1.1_SWA on gfx1151, no ktrace/ROCTx, decode autotune disabled, `-w 1 -r 5`:

- 128-token prefill: 727.77 -> 1034.62 tok/s (+42.2%), TTFT 171.76 -> 120.82 ms
- 2K prefill: 717.66 -> 1017.50 tok/s (+41.8%), TTFT 2.765 -> 1.950 s
- 8K prefill: 702.64 -> 970.50 tok/s (+38.1%), TTFT 11.283 -> 8.169 s

A fixed 100-dispatch RGP window reduced aggregate QGemm GPU time from 6.035 ms to 3.844 ms (-36.3%).

## Validation

- all 14 PSU QGemm microbenchmark shapes pass sampled int64 CPU-reference checks
- ONNX-to-HIP QGemm fusion and HIP-to-LLVM QGemm lit checks pass
- gfx11-generic and gfx1170 kernel libraries build successfully
- 128/2K/8K model benchmarks complete without decode or memory regression


lib/Runtime/Kernels/hip/qgemm_kernel.hip
```diff
@@ -1474,6 +1809,56 @@ int qgemm_gemv_nk_launch(const QGemmArgs &g, const QGemmRequant &requant,
...
+
+template <typename TA, typename TB, typename TY>
+int qgemm_gemv_nk_launch(const QGemmArgs &g, const QGemmRequant &requant,
+                         bool need_row_a, bool need_col_b) {
+  constexpr bool kCanTuneNk =
+      std::is_same_v<TA, uint16_t> && std::is_same_v<TB, int8_t> &&
+      std::is_same_v<TY, uint16_t>;
+  if constexpr (kCanTuneNk) {
+    int config = qgemm_gemv_nk_config_override();
+    if (config == 0 && qgemm_device_cu_count() == 20) {
```
amd-mingw: qgemm_device_cu_count() == 20  hardcode is not good.

lib/Runtime/Kernels/hip/qgemm_kernel.hip
```diff
@@ -1731,8 +2248,48 @@ int qgemm_launch_tiled(const QGemmArgs &g, const QGemmRequant &requant,
...
+      std::is_same_v<TY, uint16_t>;
+  const bool use_w4_specialized =
+      kCanUseW4Specialized && qgemm_tiled_w4_specialized_enabled() &&
+      !g.trans_a && g.trans_b && g.b_bits == 4 &&
+      (reinterpret_cast<uintptr_t>(g.B) % alignof(uint16_t)) == 0 &&
+      (g.K % kDotK) == 0;
 
   (void)hipGetLastError();
+  if constexpr (kCanUseW4Specialized) {
+    // K=32 staging wins for the PSU Up/Gate and O/Down shapes, but regresses
```
amd-mingw: don't show model name

lib/Runtime/Kernels/hip/qgemm_kernel.hip
```diff
@@ -1731,8 +2248,48 @@ int qgemm_launch_tiled(const QGemmArgs &g, const QGemmRequant &requant,
...
+      hipLaunchKernelGGL(
+          qgemm_tiled_k32_w4_kernel,
+          dim3(static_cast<unsigned>(g.N / kTileN),
+               static_cast<unsigned>(g.M / kTileM)),
+          dim3(kTiledBlock), 0, g.stream, a_ptr, g.B, y_ptr, g.N, g.K,
+          requant, need_row_a, need_col_b);
+      return static_cast<int>(hipGetLastError());
+    }
+
+    // Screening showed this geometry wins on the PSU Up/Gate projection
```
amd-mingw: don't show model name

amd-mingw (changes requested):

lib/Runtime/Kernels/hip/qgemm_kernel.hip
donghanbj: Fixed in 342939e.

There were two of these gates, and they needed different treatment.

The one on the single-slice K path was simply unnecessary — a K of one 32-element slice leaves all but one slice of the generic 32-slice block empty whatever the CU count, so the condition on `g.K` alone carries the invariant. Gate removed.

This one does depend on machine width: a block owns `kCols * kWaves` outputs, so a wider group yields fewer blocks. Rather than pinning it to a part, it now takes a tuned config only while the resulting grid still puts at least one block on every scheduling unit. On the part these were measured on that keeps both selections exactly as before; on a wider part the narrow-N choice steps aside instead of under-filling the device.

Verified with a ktrace capture before and after: identical launch counts for every kernel (71,120 / 14,224 / 49,784 / 14,224 GEMV dispatches, 224 tiled), only run-to-run timing noise.

lib/Runtime/Kernels/hip/qgemm_kernel.hip
donghanbj: Fixed in 342939e. Removed from four comments and from one identifier (`kPsuDecodeTypes` -> `kSplitAdapterTypes`), replaced with the shapes the reasoning actually depends on.

PR #1052, opened 2026-09-17
feat(hip-to-tosa): lower hip.softplus by decomposing to TOSA

pragyam32:
TOSA has no softplus op, so a rock.kernel containing hip.softplus could not be absorbed into a rocMLIR fused kernel. Expand to the numerically stable form the hip_softplus kernel uses, max(x, 0) + log(1 + exp(-abs(x))), rather than naive log(1 + exp(x)), which overflows on ordinary f16 values.

## Summary

Adds HipToTosa conversion for ``hip.softplus``. TOSA has no 1-1 mapping, so a ``rock.kernel`` that contains it can now be absorbed into a rocMLIR fused kernel by expanding to TOSA elementwise ops.

## Related issue or design

None
## Why

Naive ``log(1 + exp(x))`` overflows on ordinary f16 values (``exp(20)`` is already Inf). The expansion matches ``hip_softplus``: ``max(x, 0) + log(1 + exp(-abs(x)))``.

## What

- ``SoftplusConverter`` emits ``tosa.abs`` / ``tosa.negate`` / ``tosa.exp`` / ``tosa.add(+1)`` / ``tosa.log`` plus ``tosa.maximum(x, 0)``, then ``tosa.add``.
- ``hip.softplus`` is marked illegal for this pass, same as ``hip.sqrt``.
- LIT coverage in ``test_decompose.mlir``: mapping, outlined ``ub.poison`` kernel, dynamic-shape and integer rejections.

## Test plan

- ``pre-commit run --all-files`` — passed
- ``llvm-lit`` ``Conversion/hip-to-tosa/test_decompose.mlir`` — passed (same converter, earlier ``hip-ep`` build)

## Notes for reviewers
hip.sqrt and the other decompositions deliberately stay in addIllegalOp, so a dynamic-shape or integer form of those — and now of hip.softplus — still fails the pass. The same argument for dynamic legality applies to them, but that is pre-existing behaviour with its own diagnostics at the end of test_decompose.mlir, and changing it is outside this change's scope. It is a reasonable follow-up.

The static-shape and float-element-type restriction in SqrtConverter carries over unchanged — TOSA exp/log are FP-profile ops and need a static ranked tensor — so hip.softplus only reaches TOSA for static float tensors. The difference from before is that a rock.kernel containing that form is now converted rather than failing the pass.

## Checklist


nvunnam57128:
@copilot resolve the merge conflicts in this pull request

PR #1061, opened 2026-09-18
[HipToTosa] Convert hip.top_k

nvunnam57128:
Converts `hip.top_k` in `--convert-hip-to-tosa`. It already had an `OnnxToHip` conversion and a `HipToLLVM` runtime lowering, so only the TOSA path was missing. Independent of #1058; branched from `main`.

## Approach

TOSA has no sort and no top-k. Its only order-aware operations are `tosa.argmax` and `tosa.reduce_max`, so TopK expands into K rounds of *take the maximum, then mask out the position just taken so the next round finds the runner-up*, concatenating the per-round slices at the end.

**The masking is the part worth reviewing.** The obvious form — mask everything equal to the round's maximum — is wrong under ties, because it erases every member of the tie at once. For `[5, 5, 1, 5]` with `K = 3` it produces `[5, 1, -inf]` where ONNX wants `[5, 5, 5]`. I verified that empirically rather than reasoning about it, and it's covered by a regression case in the cross-check script.

The converter instead compares *positions*: an iota along `axis` against the round's `argmax`. Since argmax names exactly one position, each round removes exactly one element, and ties come out right.

**K comes from the result shape, not the `k` operand.** That operand is a runtime tensor, but the values result is K-wide along `axis` by construction, so the shape is the authoritative static source and dynamic shapes decline cleanly.

`K = 1` skips the masking and both concats entirely, collapsing to `reduce_max` + `argmax` + a reshape. That's the common classification shape and worth the special case.

**Smallest-first** would need an argmin, which TOSA does not have, so the input is negated and the same largest-first rounds run, with the values negated back at the end. Negating the minimum of an integer range overflows, so integers keep to largest-first.

## Legality gating

Each round costs a `reduce_max`, an `argmax`, a compare and a select, so K is bounded rather than unrolled without limit. That bound, integer smallest-first, and f64 are gated with `addDynamicallyLegalOp`, following the `isTosaExpressibleExpand` precedent already in this file: declining them is a routine outcome rather than a defect, so those forms stay `hip` ops instead of failing the pass. Shape disagreements are deliberately *not* in the predicate, so they remain hard failures inside the pattern.

## Validation

- Full lit suite: 524 passed, 0 failed.
- New lit file covers `K = 2` (asserting two rounds with exactly one masking step between them, since the last round has nothing left to search), `K = 1` (asserting no select and no concat), negative `axis` with i32 indices, smallest-first (asserting the negate at both ends), and integer input. Three cases assert that a too-wide K, integer smallest-first, and f64 are left as `hip` ops with no `tosa.argmax` emitted. One rejection case checks the diagnostic.
- Because lit only checks which ops appear, the expansion was separately cross-checked against ONNX reference semantics over **5400 randomized cases** spanning ranks 1-3, every axis, every K, both orderings, and both float and integer data — including deliberately tie-heavy inputs drawn from a three-value range, which is the case the naive mask gets wrong. All exact, values and indices.

## Known limitation

`tosa.argmax` has no lowering in rocMLIR. I checked rather than assumed: it is not one of the six `TosaToRock` anchors, not among the 34 `RockTosaToElementwise` patterns, and not data movement, so `tosa-to-tensor` does not take it either — the string `ArgMax` does not appear anywhere in that tree.

This splits the op awkwardly. The **values** half lowers fine (`reduce_max` is an anchor; `equal`, `select` and `concat` are all handled — `tosa-to-tensor` does convert concat). The **indices** half does not lower at any K, and `hip.top_k` is destination-passing with both outputs, so the working half cannot be emitted alone.

Unlike the gather and scatter families, where the blocker is a missing lowering for an inherently irregular op, this one is a single missing primitive over a reduction rocMLIR already performs — `rock.reduce` does max today and would need an index-carrying variant. That looks like the cheapest of the outstanding TOSA gaps to close, if someone wants to pick it up.

`hip.top_k` is also not in `FuseROCMlir`'s pointwise list, so it is not outlined into kernels today regardless.

## AI assistance

Written with Cursor. The converter and tests were authored with AI assistance; I reviewed the result. Validation was not taken on trust: the expansion was checked against an independent ONNX reference over the case count above, the tie-handling claim was confirmed by running the incorrect variant and observing it fail, and the rocMLIR coverage claim was confirmed by searching that tree directly. The commit carries a `Co-authored-by` trailer.


nvunnam57128:
Folded into #1058 as requested, so the whole HipToTosa op batch reviews and lands together. The commit is unchanged apart from being rebased onto #1058's head; review it there.

PR #1064, opened 2026-09-18
feat(hip-to-tosa): lower Shape/Size/Identity/ConstantOfShape residues to TOSA

pragyam32:
These ONNX ops have no 1-1 TOSA mapping except Identity. Fold their OnnxToHip residues (from_elements, splat, dense arith.constant, hip.size) to tosa.const so a rock.kernel can absorb them.
TOSA has no Shape, Size, or ConstantOfShape. Identity is 1-1 (tosa.identity), but there is no hip.identity: convert-onnx-to-hip already forwards the SSA value.

image

hip.size is marked illegal for this pass, same as hip.gather

## Summary

Adds HipToTosa conversion for Shape, Size, Identity, and ConstantOfShape. TOSA has no 1-1 mapping except Identity. A rock.kernel that contains their OnnxToHip residues can now be absorbed by expanding to tosa.const (Identity is already a no-op).
## Related issue or design

None.
## Why

These ops are folded or rewritten before HIP dialect in most graphs. The pieces that survive (tensor.from_elements, dense arith.constant, tensor.splat, hip.size) are not TOSA, so rocMLIR cannot absorb them until they become tosa.const. Identity is already forwarded at OnnxToHip; TOSA’s 1-1 form would be tosa.identity.
## What

- FromElementsConverter / TensorConstConverter / SplatConverter fold static Shape and ConstantOfShape residues to tosa.const.
- SizeConverter folds static hip.size to tosa.const(prod(shape)). Dynamic hip.size stays illegal.
- Identity has no hip op; LIT covers the forwarded return %x kernel.
- LIT coverage in test_shape_size.mlir: identity, static shape, size, splat constants, outlined ub.poison kernel, dynamic-size rejection.

## Test plan

- pre-commit run --files lib/Conversion/HipToTosa/HipToTosa.cpp test/lit/Conversion/hip-to-tosa/test_shape_size.mlir
- python .../llvm-lit.py -v .../test/lit/Conversion/hip-to-tosa/test_shape_size.mlir

## Notes for reviewers

hip.gather and the other conversions deliberately stay in addIllegalOp, so a dynamic-shape form of those — and now of hip.size — still fails the pass. The same argument for dynamic legality applies to them, but that is pre-existing behaviour with its own diagnostics, and changing it is outside this change's scope. It is a reasonable follow-up.

The static-shape restriction carries over: TOSA has no shape-of, so tensor.dim / dynamic hip.size are left unconverted (or fail when the op is claimed). The difference from before is that a rock.kernel containing static Shape / Size / ConstantOfShape residues is now converted rather than leaving arith.constant / tensor.from_elements / tensor.splat / hip.size in the kernel.

There is no hip.identity. Adding one would undo the OnnxToHip SSA-forwarding that already matches tosa.identity after DCE.

## Checklist

- [ ] The change is focused, or links a design/series explaining its scope.


zz002:
Hi @pragyam32  @Pragya0302  
GitHub Actions cache space is limited — please don't open many PRs in parallel that each build their own LLVM rocMLIR  Triton caches. We need to leave room for the main branch caches, since those are what most PRs rely on to speed up CI.

https://github.com/ROCm/hip-ep/actions/caches?query=sort%3Asize-desc

image

PR #1066, opened 2026-09-18
fix(matmul_nbits): gate LDS barrier on archs that provide s_barrier

amd-mah:
## Summary
matmul_nbits_kernel.hip's HIPDNN_LDS_BARRIER emitted the raw s_barrier inline asm on every architecture except gfx12, using a denylist (!defined(__GFX12__)). Any architecture that also lacks the standalone s_barrier instruction then fails to device-link. 

s_barrier instruction got retired -- gfx12 replaced it with the s_barrier_signal/s_barrier_wait pair and later families lack it too -- so a denylist keeps breaking on each new one. This switches the gate to an allowlist of the families that provide s_barrier (gfx9/gfx10/gfx11); every other target, current or future, falls back to __syncthreads(), a strictly stronger full-workgroup fence, so the fallback is always correct and only reintroduces the redundant per-K-step L0 flush the raw asm skips.

Gate change in matmul_nbits_kernel.hip:

    -    !defined(__GFX12__)
    +    (defined(__GFX9__) || defined(__GFX10__) || defined(__GFX11__))

Also adds LdsBarrierArchGate, a GPU-free CTest registered under BUILD_HIP_UNIT_TESTS.

## Related issue or design
None.

## Notes for reviewers
gfx12 already used the __syncthreads() fallback before this change, so behavior on currently-supported architectures is unchanged; only families that previously failed to build are affected.

LdsBarrierArchGate extracts the actual macro from the kernel and, using only the HIP compiler, asserts each family emits the correct barrier (gfx9/gfx10/gfx11 use s_barrier; gfx12 uses the s_barrier_signal/s_barrier_wait fallback) and that a hypothetical future family falls back -- catching a regression back to a denylist. Verified it fails on the reverted denylist and passes here; gfx11/gfx12 device assembly is unchanged aside from the per-compile __hip_cuid symbol.

## Checklist


lib/Runtime/Kernels/test/lds_barrier_arch_gate/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+foreach(_cand IN ITEMS "${CMAKE_HIP_COMPILER}" "${_HIP_CLANGXX}")
+  if(_cand AND NOT _lds_barrier_hip_cc)
+    set(_lds_barrier_hip_cc "${_cand}")
+  endif()
+endforeach()
+if(NOT _lds_barrier_hip_cc)
+  find_program(_lds_barrier_hip_cc NAMES amdclang++ clang++)
+endif()
+
+if(Python3_Interpreter_FOUND AND _lds_barrier_hip_cc)
```
fhanuman, line 31: The `LdsBarrierArchGate` test doesn’t seem to run in CI. On Windows, compiler detection doesn't seem to consider HIPCC_EXECUTABLE, and build.py excludes the test from the CTest regex. Could you please double-check this?

fhanuman (approved):

lib/Runtime/Kernels/test/lds_barrier_arch_gate/CMakeLists.txt
amd-mah: Thanks @fhanuman - fixed both. The guard's CMake now also resolves the compiler via `HIPCC_EXECUTABLE`/`hip_HIPCC_EXECUTABLE` (so it registers on Windows), and `LdsBarrierArchGate` is added to `build.py`'s ctest `-R` regex so the GPU-free suite runs it. Verified it's now selected and passes with both `amdclang++` and `hipcc`. Please take another look when you get a chance.

fhanuman (approved):

PR #1037, opened 2026-09-16
feat(hipsr): emit HIP generate-interface after wrapping the graph ABI

zpye:
## Summary
- Record `@main_graph` input/output metadata as `hipdnn.*` module attributes before LLVM lowering (`hipsr-record-graph-io`).
- Wrap the expanded memref `@main_graph` in the runtime `(state, inputs) -> i32` ABI after `convert-to-llvm` (`hipsr-main-graph-abi`).
- Run HIP `generate-interface` at the end of `--hipsr-pipeline` so the artifact emits `inference_init` / `compute` / `cleanup` / `get_metadata_json`.

HIP generate-interface requires that wrapped signature and the I/O attributes. Generic `--convert-to-llvm` leaves an expanded memref calling convention, so the new passes sit on either side of it.

Substantial AI assistance was used to implement the passes, LIT coverage, and pipeline wiring.

PR #928, opened 2026-09-08
perf(nemotron): speed up QMoE prefill, linear-attention, and reshape host scalars

AMDmoore:
## Summary
- Speed up com.amd QMoE prefill with device-side expert grouping (GEMV default, optional ragged WMMA / matmul_nbits) and keep the fused decode path for `num_tokens==1`.
- Window the linear-attention chunked prefill (gated + gated_delta), copy recurrent state on the compute stream, and skip D2H sync when reshape extents are already host SSA.
- Extend kernel ABI (`hip_qmoe_amd_prefill_grouped_*`, `hip_matmul_nbits_grouped_wmma`, optional route scratch) and add LIT / numeric coverage for the new paths.

## Test plan
- [ ] LIT: `test/lit/Conversion/onnx-to-hip/test_reshape.mlir`
- [ ] Numeric: `test/numeric/tests/test_linear_attention.py`
- [ ] Nemotron 120B QMoE: TTFT/TPS vs main at prompt 128 and 2048 (PERF flags off)
- [ ] Nemotron PPL sanity (pruned and full-logits) vs the pre-opt baseline

## AI assistance
Ported and landed with Cursor. Changes were reviewed against `main` (3-way merge of `matmul_nbits_kernel.hip` / `hip_custom_kernels.h` so #903/#916 APIs are preserved) and formatted with pre-commit/lintrunner.

Made with [Cursor](https://cursor.com)

PR #1067, opened 2026-09-20
kernels: QMoE kernel optimization for baseline and MTP/EAGLE

penglu12-quark:
## Summary

`wrap_qmoe` only had a fused path at `num_tokens == 1`, so every burst wider than a single
token fell into the bucketed multi-pass path. This extends the fused path to serve narrow
bursts by iterating tokens, gated on `HIPDNN_EP_QMOE_FUSED_MAX_TOKENS` (default 8).

It is a dispatch change: the fused path computes the same per-token result, so output is
unchanged. Behaviour only differs for bursts of 2-8 tokens, which is the regime speculative
decoding (MTP / EAGLE-3) verifies in. `num_tokens == 1` and prefill both keep their existing
behaviour.

## Related issue or design

None filed. The measurements come from a speculative-decoding bring-up series on Strix Halo
(gfx1151) for MTP and EAGLE-3 draft heads; the relevant numbers are inlined under Performance.

## Why

The two paths trade weight traffic against dispatch overhead.

Bucketing reads each *distinct* expert once, which is strictly the less weight traffic. It pays
for that with a D2H copy of the per-expert counts plus a `hipStreamSynchronize` per MoE layer,
and then five kernel launches per active expert - on a 30-layer model verifying four positions,
30 pipeline stalls and roughly 2700 launches.

At the widths speculative decoding actually verifies, that trade inverts. Each expert serves one
or two tokens, so the per-expert GEMMs degenerate to GEMVs and the traffic bucketing saves is
small, while the stalls and launches are not. Iterating the fused path reads one expert set per
token - more traffic - but it is three launches per layer and never synchronizes.

The win is bounded rather than monotonic: once bursts are wide enough for expert reuse to repay
the stall, bucketing must win again. Hence a ceiling instead of an unconditional switch, with
the default stopping at the widest burst measured rather than extrapolating past it.

## What

- `lib/Runtime/real/qmoe.cpp`: the gate becomes `num_tokens <= hipdnn_ep_qmoe_fused_max_tokens()`
  and the body loops over tokens, offsetting `input`, `output`, `expert_indices` and
  `expert_weights` per token. The dp4a / non-dp4a choice is hoisted out of the loop.
- The `[k, inter]` and `[k, hidden]` slot buffers are reused across iterations rather than
  indexed by token. Each iteration is independent - one token's k experts, reduced into that
  token's output row - and iterations are serialized on the stream. They were already sized
  `max(num_tokens, k)` and every iteration uses `k`.
- `lib/Runtime/debug_log.h`: `hipdnn_ep_qmoe_fused_max_tokens()`, parsed once and clamped to
  [1, 64], default 8. Setting it to 1 restores the previous decode-only behaviour exactly, which
  is the A/B control used below.
- `test/numeric/tests/test_qmoe.py`: `SEQ_LENS` gains 2, 5, 8 and 9 to cover both sides of the
  dispatch boundary.

## Test plan

**Numerics.** All checks green on
[run 6546](https://github.com/ROCm/hip-ep/actions/runs/35559275386) (commit `6552aff`), including
`gpu_test / test`. `test_qmoe.py` now covers seq lengths 1, 2, 5, 8, 9 and 128: 2/5/8 take the
new fused loop, 9 and 128 stay bucketed, 1 is the pre-existing fused case.

**End to end.** Every model the CI exercises, from the same run, against the min-max spread of
the same benchmarks over the last 12 merged PRs (#1060, #1056, #1055, #1046, #1045, #1040,
#1037, #1035, #1033, #1028, #1026, #1023):

| Benchmark | This PR (run 6546) | Last 12 merged PRs |
|---|---:|---|
| OGA gpt-oss-20b (MoE 20B) - TPS | 78.2 | 78.0 - 79.1 |
| OGA gpt-oss-20b - TTFT ms | 177.7 | 139.8 - 196.5 |
| OGA Llama-3.1-8B - TPS | 46.3 | 46.1 - 46.4 |
| OGA Llama-3.1-8B - TTFT ms | 120.7 | 118.6 - 121.0 |
| OGA wheel smoke, Llama-3.1-8B - TPS | 45.9 | - |
| VLM Qwen3.8-27B - TPS | 13.8 | 13.8 - 13.8 |
| VLM Qwen3.8-27B - TTFT ms | 4949.9 | 4974.0 - 5180.4 |
| EP `full_model_seq128` - QPS | 8.16 | 8.15 - 8.19 |
| EP `GroupQueryAttention_seq128` - QPS | 4367.99 | 4223.3 - 4425.4 |
| EP `matmul_down_seq128` - QPS | 482.88 | 478.7 - 498.0 |
| L2 `QMoE_seq128` | 39.2997 | 39.2997 (all 12) |
| L2 `MatMulNBits_o_seq128` | 259.905 | 259.905 (all 12) |
| `hipgpu.dll` | 63.37 MB | 63.37 MB on main (+0.00) |

Every figure sits inside its baseline spread, except the VLM TTFT which lands just below the
baseline minimum (i.e. slightly faster).

One caveat on the two L2 gates: they agree to six significant figures across all 13 runs, but
that is expected rather than informative here, because both run at seq 128 and so stay on the
bucketed path this PR does not touch. Coverage of the changed path is `test_qmoe.py` at seq
2/5/8/9.

This branch is rebased onto `main` at 8c888c0 (#928, which reworked QMoE *prefill* via
`qmoe_amd.cpp`), so the table above reflects the two changes together. They do not overlap:
#928 touches the `qmoe_amd` operator, this PR touches `real/qmoe.cpp`.

## Performance

gemma-4-26B-A4B (128 experts, top-8, 30 MoE layers), verify time per burst:

| burst width | bucketed | fused | speedup |
|---:|---:|---:|---:|
| 2 | 34.0 ms | 23.1 ms | 1.47x |
| 4 | 50.2 ms | 32.1 ms | 1.56x |
| 8 | 81.2 ms | 58.3 ms | 1.39x |

End to end on the same target, MTP speculative decoding at k=1 goes from 0.76x to 1.07x against
non-speculative decode - i.e. from "speculation loses" to "speculation pays".

On gpt-oss-20b (32 experts, top-4, 24 MoE layers), verify is 1.13-1.17x faster at widths 2
through 4.

Hardware, workload, build configuration, and measurement method: AMD Strix Halo / Radeon 8060S
(gfx1151), Windows, release build of this branch, int4 block-32 targets driven through a
draft-verify loop. Each figure is the mean verify-step time over a 128-token generation. The A/B
is the same binary with `HIPDNN_EP_QMOE_FUSED_MAX_TOKENS` toggled between 1 (bucketed, the old
behaviour) and the default 8, so only dispatch differs between the two columns.

## Notes for reviewers

- **Scope, for anyone comparing perf dashboards.** The only behaviour that changes is a QMoE op
  called with 2-8 tokens. Dense models never reach `wrap_qmoe` at all; non-speculative decode
  presents exactly 1 token and is bit-identical; prefill is not chunked, so the op sees the full
  prompt and stays bucketed. A standard decode-plus-prefill benchmark on any model in the table
  above therefore runs the same code before and after this PR.
- The ceiling is deliberate and the default is the widest burst actually measured. If a future
  workload verifies wider bursts, the crossover should be re-measured before raising it rather
  than assumed to keep scaling.
- `HIPDNN_EP_QMOE_FUSED_MAX_TOKENS=1` is a complete revert of the behaviour at runtime, which
  makes this cheap to bisect or disable in the field without a rebuild.
- No doc file is updated, which I believe is correct rather than an omission: there is no
  central env-var reference in the tree, and the sibling gate `HIPDNN_EP_MATMUL_DP4A` is
  likewise documented in its `debug_log.h` comment. The table in `docs/quick_start_mi350.md` is
  specifically about knobs that go inert on MI350X, which does not apply here. Happy to add a
  reference table if reviewers would rather have one.

## Checklist

AI assistance: Cursor assisted with the implementation and with assembling the measurements
above; the commit carries a `Co-authored-by: Cursor` trailer. I reviewed the dispatch logic and
checked the per-token pointer arithmetic against the scratch-buffer layout (`expert_weights` is
allocated at `num_tokens * k * elem_size` and `expert_indices` at `num_tokens * k *
sizeof(int32_t)`, matching the strides used in the loop), and validated the result with the
numeric tests and the A/B above.


amd-mingw (changes requested):
Thanks @penglu12-quark , could you add e2e test link and result for all the models?

amd-yanjunz:

- PR: ROCm/hip-ep#1067
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/1067#issuecomment-5755147021)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1553/
- Main baseline build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1528/
- Package build run: https://github.com/ROCm/hip-ep/actions/runs/35484448348
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6ab0c3242225a21cf8e4c69e
- Compare with main baseline 8c888c0: http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fdatabase%3Ddevelopment%26collection%3Ddaily_hip_ep_genai%26id%3D6ab012a615d05d850a51da72&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=1553

penglu12-quark:
@amd-mingw thanks - added. The description now has a filled-in **Test plan** with the e2e
results for every model the CI exercises, linked to
[run 6513](https://github.com/ROCm/hip-ep/actions/runs/35484448348) (commit `d150a79`).

Rather than quoting this run's numbers on their own, I put each one next to the min-max spread
of the same benchmark over the last 12 merged PRs, so it is visible whether a delta is real or
run-to-run noise:

| Benchmark | This PR (run 6513) | Last 12 merged PRs |
|---|---:|---|
| OGA gpt-oss-20b (MoE 20B) - TPS | 78.8 | 78.0 - 79.1 |
| OGA Llama-3.1-8B - TPS | 46.7 | 46.1 - 46.4 |
| OGA wheel smoke, Llama-3.1-8B - TPS | 46.6 | - |
| VLM Qwen3.8-27B - TPS | 13.8 | 13.8 - 13.8 |
| EP `full_model_seq128` - QPS | 8.17 | 8.15 - 8.19 |
| EP `GroupQueryAttention_seq128` - QPS | 4429.86 | 4223.3 - 4425.4 |
| EP `matmul_down_seq128` - QPS | 483.26 | 478.7 - 498.0 |
| L2 `QMoE_seq128` | 39.2997 | 39.2997 (all 12) |
| L2 `MatMulNBits_o_seq128` | 259.905 | 259.905 (all 12) |
| `hipgpu.dll` | 63.37 MB | 63.36 MB on main (+0.00) |

Everything is inside its baseline spread except four figures that land just outside on the
favourable side (Llama TPS, GQA QPS, and two session times below the baseline minimum). Full
table including TTFT is in the PR description.

One thing worth being explicit about, since it affects how much these numbers can tell us: this
PR only changes behaviour for a QMoE op called with **2-8 tokens**. Dense models never reach
`wrap_qmoe`; non-speculative decode presents exactly 1 token and keeps the fused path it already
had; prefill is not chunked, so the op sees the full prompt and stays bucketed. A standard
decode-plus-prefill benchmark therefore runs identical code before and after this change, which
is why I would read the table above as a no-regression check rather than as evidence the
optimization works.

The evidence that it works is separate, and it is in the Performance section: on gemma-4-26B-A4B
verify time per burst drops 34.0 -> 23.1 ms at width 2, 50.2 -> 32.1 at 4 and 81.2 -> 58.3 at 8,
which takes MTP speculative decoding at k=1 from 0.76x to 1.07x end to end. That A/B is the same
binary with `HIPDNN_EP_QMOE_FUSED_MAX_TOKENS` toggled between 1 (old behaviour) and the default
8, so only dispatch differs.

If you would like the 2-8 token regime covered in CI as well as in `test_qmoe.py`, I can add a
narrow-burst case to the perf job - happy to do that in this PR if you think it belongs here.

amd-yanjunz:

- PR: ROCm/hip-ep#1067
- Source comment: [models=all](https://github.com/ROCm/hip-ep/pull/1067#issuecomment-5806011887)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: (GenAI job default)
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1687/
- Main baseline: pending package run selection

PR #1068, opened 2026-09-20
perf(allocator): reland large pinned-buffer pooling with the eviction leak fixed

BoarQing:
## Summary

Relands #961 with the one defect that caused its revert fixed. The pooling of
pinned host buffers above 16 MB returns, and with it the Gemma-4 decode win;
what changes versus the reverted commit is eight lines in `FreeImpl`.

Two commits, deliberately: the first is an exact revert of #1056, so reviewing
it is reviewing #961 again, and the second is the entire delta.

## Related issue or design

Relands #961, reverted by #1056.

## Why

#1056's report was accurate and reproducible: #961 grew pinned host memory by
roughly 200 MB per EPCert MMLU question and faulted the device with
`hipMemcpy* failed: unspecified launch failure (719)` around question 105. I
reproduced it first try on gfx1151.

The cause was not the pooling design and not its retention accounting, both of
which hold up under measurement. `FreeImpl` collects evicted buffers into
`to_release` and depends on the `ReleaseToDriver` at the end of the function to
hand them to `hipHostFree`. The retain path returned before reaching it, so every
eviction that actually succeeded in making room leaked the buffer it had just
evicted. `EvictLruLarge` has already erased that pointer from `ptr_to_size_`, so
the destructor could not reclaim it either — the pages stayed pinned for the life
of the process, untracked.

That is why review did not catch it. `large_retained_bytes_` was decremented, the
ceiling was respected, and every counter balanced. They were simply balancing
against a quantity that no longer described what was pinned. Instrumenting the
`hipHostFree` status that the shipped code discards is what makes it visible:
over 100 MMLU questions the pool makes 94 driver allocations and 84 evictions and
only **2** `hipHostFree` calls, those being the two that took the fall-through
path.

MMLU is the shape that exposes this and Gemma-4 is the shape that hides it. MMLU
prefills once per question, with a fresh `og.Generator` each time and an unpruned
`lm_head`, so the single large allocation is an fp16 logits buffer of 303,872
bytes per token. Across the dataset's 257–2382 token prompts that is 91 distinct
sizes in 94 requests, never more than one live, so nearly every free has to
evict. Gemma-4 26B decode reuses one capacity for hundreds of steps: 15,160
reuses against **2** evictions. The leak cost it about 96 MB, invisible beside the
10.5 GB the pool saves.

## What

`FreeImpl` becomes single-exit. The small-size-class branch and the large-retain
branch both fall through to the one `ReleaseToDriver` instead of returning early.
Done that way rather than by adding a release before the early return, so a later
branch added to that block cannot reintroduce the same bug.

No change to the pooling policy, the capacity function, the retention ceiling, or
the eviction order.

## Test plan

Everything below is gfx1151, one machine, one ORT, one `amdhip64`, one model per
workload, with **only `hipgpu.dll`** differing between arms. Both binaries are
built from the same commit of this branch's base.

### The reported failure, reproduced and then fixed

EPCert 2.10.1 MMLU, `DeepSeek-R1-Distill-Qwen-1.5B`, 300 questions:

| `hipgpu.dll` | result | accuracy | memory |
|---|---|---|---|
| main | 300/300 | 34.00% | commit flat from Q50, +9.3 GB |
| main + #961 as merged | **faults Q105–154, error 719** | n/a | **193 MB/question**, +19.6 GB when halted |
| main + #961 + this fix | 300/300 | 34.00% | commit flat from Q125, +10.4 GB |

Accuracy is identical to main, so the fix changes nothing observable except that
the memory comes back. The ~1 GB the fixed build sits above main is the pool
retaining buffers as designed, bounded by the 701 MB ceiling that run reported.

Note that the growth does **not** appear in process private bytes, which sat at
1894 MB in every arm. On this APU these pinned pages show up in system commit
charge and in the GPU shared-memory counter, 3.9 GB → 17.8 GB, which is the pool
that runs out and takes the device down.

### The win #961 exists for, still present

Gemma-4 26B-A4B, 8192-token prompt, 128 generated, `past_present_share_buffer:
false`. Arms interleaved and order-reversed, each with its own `TEMP` so the WMMA
autotune cache cannot make a DLL swap look like a cold-tune run:

| arm | TPS (round 1 / round 2) | peak RSS | commit delta | TTFT |
|---|---|---|---|---|
| main | 23.97 / 23.93 | 17,441 MB | +37.4 GB | 5,752 ms |
| #961 as merged | 40.93 / 40.82 | 7,033 MB | +28.0 GB | 5,750 ms |
| **#961 + this fix** | **41.01 / 41.04** | **6,902 MB** | **+27.8 GB** | 5,756 ms |

23.95 → 41.02 tok/s, **+71%**, with peak RSS down 60%. The fixed build is
marginally ahead of the leaky one rather than behind it, and uses slightly less
memory. Reversing arm order does not move anything, so this is the change and not
position or thermal drift. TTFT is identical across all three, as expected — the
churn is per decode step, so only decode is affected.

### Mechanism, measured rather than argued

Temporary instrumentation recording the pool's state and the discarded
`hipHostFree` status on every large operation (not in this diff):

| | Gemma-4 26B decode, 8K | EPCert MMLU, 100 questions |
|---|---|---|
| reuses | 15,160 | 6 |
| driver allocations | 182 | 94 |
| evictions — the leak path | **2** | **84** |
| `hipHostFree` calls before fix | 62 | 2 |
| `hipHostFree` calls after fix | 62 | 89 |

After the fix every eviction and drop is matched by a `hipHostFree` returning
`hipSuccess`, and the pool makes byte-for-byte identical decisions either side of
the change — 94 allocations, 84 evictions, 5 drops, 6 reuses on MMLU before and
after. Only the release differs. The Gemma-4 reuse/eviction figures also match
#961's own mechanism table (15,180 reuses against 182 driver calls at 16K).

### Offline reproduction

A host-only replay of the pool against stubbed `hipHostMalloc`/`hipHostFree`,
fed the captured request sequence, predicts 18,806 MB leaked against 18,823–19,609
MB measured on hardware across three runs, and 0 MB with the fix. Deterministic
and needs no GPU, so it is a two-second check rather than a 40-second EPCert run.

### Checks

- `pre-commit run --from-ref <base> --to-ref HEAD` — all hooks pass
- Release build, gfx1151, MSVC 14.50, ninja — clean

## Notes for reviewers

**What I could not measure.** #961's results table is at a 16384-token context;
mine is at 8192. A 16K Gemma-4 26B run needs about 44 GB of commit charge on top
of my box's ~30 GB resting charge and hits the 74.9 GB commit limit, on `main`
as much as on this branch. Physical memory was never short (24 GB free
throughout), so this is a commit ceiling on my machine and not the change. 8192
still exercises the target path fully — KV tensors are 32 MiB there, well past
the 16 MB cutoff, and all 60 are reallocated every step. The +71% here is
therefore smaller than #961's +164% at 16K, which is expected: a shorter context
has less transient to recover. #961's own 12B row showed +0.2% at 16K, so this
gain is strongly context- and model-dependent either way.

**A follow-up worth considering, not done here.** The pool is bounded correctly
but has no way to report that it is. This defect was a pure control-flow slip
that no counter could have caught, because the counters were the thing being
fooled. An earlier revision of #961 logged a one-line pool summary at teardown;
had it also reported driver `hipHostFree` calls against evictions, the mismatch
would have been visible in any run. I would rather add that as its own change
than grow this one.

**Ordering.** The companion PR carries this branch plus #599 so that CI builds
OGA from the fork and can exercise Gemma-4 end to end. It is for CI evidence
only and should not merge ahead of #599 itself.

## Checklist

## AI assistance

Reproduction, root-cause instrumentation, the offline replay harness, the A/B
measurement scripts, and this description were produced with Cursor (Claude Opus
5). The diagnosis was validated three independent ways before the fix was
written: on hardware via EPCert, via `hipHostFree` status instrumentation, and
offline via trace replay. I reviewed the change and own the claims above.


BoarQing:
Gemma Perf: 
http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F1477%2F&base_job=daily_hip_ep_genai&base=&base_commit=&curr_url=http%3A%2F%2Fxsjvitisaijenkins%3A8080%2Fjob%2Fdaily_hip_ep_genai%2F1493%2F&curr_job=daily_hip_ep_genai&curr=1493&curr_commit=
Gemma Acc:
http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1571/

amd-yanjunz:

- PR: ROCm/hip-ep#1068
- Source comment: [models=all node=xsjnucstrhalo09](https://github.com/ROCm/hip-ep/pull/1068#issuecomment-5754905547)
- MODELZOO: LLM_ROCm_dynamic
- TEST_MODELS: (all)
- FORBID_MODELS: (none)
- NODE: xsjnucstrhalo09
- CI_OPTIONS: (none)
- Jenkins build: http://xsjvitisaijenkins:8080/job/daily_hip_ep_genai/1549/
- Main baseline: not usable, the baseline uses a different NODE
- Package build run: https://github.com/ROCm/hip-ep/actions/runs/35556763797
- NTS report: http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6ab0c016d8be46f5a9e4c6d9
- Compare with latest daily baseline (fallback): http://xsjnts.xilinx.com/ROCm/compare?base_url=http%3A%2F%2Fxsjnts.xilinx.com%2FROCm%2Fhipdnn-detail%3Fids%3D6ab0217915d05d850a51da84%26collection%3DHIP-EP-Dynamic&base_job=&base=&curr_url=&curr_job=daily_hip_ep_genai&curr=1549

BoarQing:
Superseded by #1088, which keeps the throughput win and removes the cost this one charged other models.

The difference is what retention is budgeted against. This PR capped retained bytes against a high-water mark that only ever rises, so any workload could hoard in proportion to its largest-ever allocation whether or not that capacity would be asked for again. #1088 budgets by circulation instead: a large buffer is retained only once several are checked out at the same time (the per-step KV signature), and capacities that stop circulating age out on the alloc path.

Measured on the same hardware, Gemma-4 26B-A4B at 8K/16K/20K and Llama-3.1-8B under MMLU:

- Gemma-4 throughput and peak working set match this PR at every context (16K: 30.2 -> 32.0 tok/s; 20K: 28.0 -> 27.9; peak RSS within 0.04 GiB).
- Llama-3.1-8B under MMLU records 1176 large requests with all 1176 frees released straight to the driver, peak retained bytes exactly 0, and the pool never creating a bucket. Accuracy unchanged.

#1088 also carries the justification this PR was missing: why the graph and the OGA layer cannot bound the allocation, and the upstream model-builder root cause (no AMD entry in \is_gqa_supported()\) that makes the workaround necessary at all.

PR #1071, opened 2026-09-20
perf(qgemm): retune the K=32 tile from 64x64 to 128x16

donghanbj:
> Stacked on #1051 — that PR introduces the K=32 kernel this retunes. Base will
> need retargeting to `main` once #1051 lands.

## Why the square tile is wrong here

The K=32 staging kernel inherited the `[64 x 64]` output tile the other tiled kernels use. Logical traffic for one GEMM is

```
A: (N / TN) * M * K * 2      B: (M / TM) * N * K * 0.5
```

so weighting by element width argues for a **wide** N tile. Measurement says the opposite, because the two operands sit on different sides of the cache:

| operand | size at M=128 | fits L2 (2 MB)? | cost of a re-read |
|---|---|---|---|
| A panel | 0.5–2 MB | yes | ~free, served from L2 |
| B weights | 8 MB per projection | no | a miss every time |

So the tile wants `TM` as tall as `M` — one row of blocks, B fetched once — and `TN` as narrow as the block can go before it stops filling the machine.

## The sweep

Nine shapes, ranked on a whole-prefill replay (588 GEMMs, 916 MB of weights, so nothing stays resident):

| tile | block | acc VGPR | replay | vs baseline |
|---|---:|---:|---:|---:|
| 64×64 (baseline) | 256 | 32 | 89.6 ms | — |
| 64×128 | 256 | 64 | 98.3 | +9.7% |
| 64×128 | 512 | 32 | 96.5 | +7.7% |
| 64×256 | 512 | 64 | 116.8 | +30% |
| 128×128 | 512 | 64 | 103.9 | +16% |
| 128×64 | 512 | 32 | 85.3 | −4.8% |
| 128×32 | 256 | 32 | 85.2 | −4.9% |
| **128×16** | **128** | **32** | **81.0** | **−9.6%** |
| 128×8 | 64 | 32 | 104.9 | +17% |

Not a plateau: `TN=8` drops the block to two waves and regresses 17%, and every shape that grows the accumulator past `[4 x 4]` regresses on register pressure alone.

## Measured

Fixed 128-token prefill graph, no ktrace/ROCTx, decode autotune disabled, `-w 1 -r 5`, 3 processes per point:

| input | TTFT | prefill |
|---|---|---|
| 128 | 121.59 → **110.66 ms** (−8.99%) | +10.0% |
| 2K | 1.879 → **1.692 s** (−9.93%) | +10.9% |
| 8K | 7.720 → **7.007 s** (−9.24%) | +10.2% |

Decode is unaffected — this kernel is gated on `M == 128` and never dispatches there.

## Correctness

The standalone QGemm suite passes all 14 shapes. The three routed here also improve in isolation: the K=8192 projection goes 4.47 → 6.71 logical TOPS, K=2048/N=8192 goes 5.78 → 6.51.

The kernel now carries its own tile constants rather than borrowing `kTileM` / `kTileN` / `kTiledBlock` from the generic tiled path, and stages A and B at different rates since the tile is no longer square.


donghanbj:
Closing this one: a WMMA path for the same projections supersedes it.

`v_wmma_i32_16x16x16_iu8` retires a 16x16x16 int8 product per instruction and
measures ~54 int8-TOPS on gfx1151, against ~27 for `v_dot4_i32_iu8`. Once that
path takes the prefill projections, the shapes this retune targets -- M=128
with N>=8192 K=2048, or N=2048 K>=3072 -- are all admitted by it, and the K=32
kernel records **zero dispatches** across a full run (prefill plus 128 decode
steps) on gfx1151. The geometry tuned here no longer runs on the part it was
tuned for.

The retune is still correct and still helps on targets without WMMA, so this
can be revisited if gfx12 prefill becomes a priority. For now it would land an
optimization that is already dead on arrival on gfx1151, which is not worth the
review surface.

Measured for reference, model_benchmark.exe on gfx1151, medians of 3-4
processes:

| input | #1051 + this PR | #1051 + WMMA |
|---|---|---|
| 128 | 109.2 ms | 81.4 ms |
| 2K | 1.698 s | 1.262 s |
| 8K | 6.888 s | 5.189 s |

PR #1072, opened 2026-09-20
fix: decline the patch-embed conv-to-gemm rewrite for unit kernels

AMDmoore:
## Summary

`PatchEmbedConvToGemm` no longer rewrites unit (1x1) convolutions into
Reshape/Transpose/Gemm. The `outSpatialProd == 1` exemption that let them
through is removed, so the unit-kernel guard now applies unconditionally.

The visible effect is on quantized decode graphs: their 1x1 convolutions stay
on `hip.conv`, remain fusable into `hip.qconv`, and no longer materialize
dequantized f32 weights as pooled transients.

## Related issue or design

None. The ordering hazard this exposed is worth tracking separately -- see
"Notes for reviewers".

## Why

The guard declines unit kernels because a 1x1 convolution is already a
contraction over C, so the rewrite is pure added traffic with nothing to gain.
The exemption argued that when there is a single output position both
permutations degenerate to identities and are never emitted, making the
rewrite free. That is true, and it is also beside the point: free is not the
same as useful, and a rewrite with no gain should be declined whatever it
costs.

The distinction stopped being academic when #1046 moved QDQ fusion from the
ONNX layer to `hip-fusion-transform`. That inverted the ordering. Previously
fusion ran first and the rewrite found a fused quantized conv with nothing to
match. Now the rewrite runs first, turns `dequantize + conv + quantize` into
`reshape + gemm`, and the intervening reshape breaks the pattern that would
have fused it back into a quantized conv. The weights are left dequantized as
plain gemm operands.

Every quantized 1x1 convolution in a decode graph has `outSpatialProd == 1`
and so takes this path. The f32 weight copies are transients, which means
PoolAllocs sizes them into the pool. For a 40-layer model with hidden 5120 and
FFN 17920:

```
40 * (7680*5120 + 5120*5120 + 3*5120*17920) * 4 B = 54,525,952,000 B
```

which accounts for all but 81,920 B of the 54,526,033,920 B pool the compiler
was requesting. Allocation fails outright on any consumer-sized device.

## What

- `ProjectorOpsRewrites.cpp`: drop `outSpatialProd != 1 &&` from the
  unit-kernel guard and rewrite the comment to state why a single output
  position is not an exemption.
- `test_patch_embed_conv.mlir`: `unit_kernel_single_patch` asserted the old
  behaviour. Renamed to `guard_unit_kernel_single_patch`, alongside the other
  declined cases, and inverted to expect `hip.conv`.

Vision patch-embed shapes are unaffected, including the rank-5 case that
motivated the exemption: its kernel is 2x16x16, so the unit-kernel guard never
applied to it in the first place.

## Compiler IR

<details>
<summary>Before / after, quantized 1x1 conv in a decode graph</summary>

```mlir
// Before: the rewrite fires, the reshape strands the contraction, and the
// weights are dequantized into a plain gemm operand.
%wd = "onnx.DequantizeLinear"(%wq, %ws, %wzp) : (tensor<17920x5120xi4>, ...)
                                             -> tensor<17920x5120xf32>
%xr = "onnx.Reshape"(%x, %s0)  : ... -> tensor<1x5120xf32>
%g  = "onnx.Gemm"(%xr, %wd, %b) : ... -> tensor<1x17920xf32>
%y  = "onnx.Reshape"(%g, %s1)  : ... -> tensor<1x17920x1x1xf32>

// After: the rewrite is declined, the QDQ triple survives intact, and
// hip-fusion-transform folds it into a single quantized conv.
%y = hip.qconv %x, %wq, %ws, %wzp, %b
   : tensor<1x5120x1x1xf16>, tensor<17920x5120xi4>, ... -> tensor<1x17920x1x1xf16>
```

</details>

## Test plan

- `pre-commit run --files` on both changed files: all hooks pass.
- `ctest -R MorphizenMLIRLitTests`: 532 discovered, 9 unsupported, 0 failed.
- `test_patch_embed_conv.mlir` specifically, all 14 cases behaving as
  documented:

  | case | result |
  |---|---|
  | `qwen_patch_embed` (2x16x16) | `hip.gemm`, no transpose -- unchanged |
  | `gemma3_patch_embed` and variants (14x14) | transpose + `hip.gemm` -- unchanged |
  | `convnext_downsample` (2x2) | transpose + `hip.gemm` -- unchanged |
  | `guard_unit_kernel_single_patch` (1x1, one position) | `hip.conv` -- **changed** |
  | other `guard_*` (overlap/pad/ragged/group/dilate/auto_pad/1x1) | `hip.conv` -- unchanged |

- Compiled a 40-layer quantized model whose decode graph is built from these
  shapes. The decode subgraph's pool goes from 54,526,033,920 B to 256,000 B;
  the prefill subgraph is unchanged at 131,072,000 B. The ratio between them is
  now 1/512, matching their sequence lengths.
- Decode subgraph op histogram: the `hip.gemm` ops are replaced by `hip.qconv`.
- End-to-end 128-prompt / 128-token generation completes with exit 0 and
  produces byte-identical output text to a run from before the regression.

## Notes for reviewers

The exemption was not wrong when it was written -- it became wrong when #1046
inverted the ordering between an ONNX-level rewrite and a HIP-level fusion.
Nothing reported that inversion: the rewrite succeeded, the fusion silently
found no match, and the only symptom was a pool two orders of magnitude too
large at a much later stage. Any ONNX-level rewrite that inserts an op between
a quantize/dequantize pair and its consumer can do the same thing. That seems
worth a guard or a diagnostic rather than leaving it to the next person to
bisect, but it is a larger change than this fix and I would rather not fold it
in here.

## Checklist

Prepared with AI assistance (Cursor). The diagnosis, the IR and pool evidence
above, and the test update were reviewed by hand.


AMDmoore:
Superseded by #1074, which carries this commit unchanged as its first commit.

The numeric coverage added there depends on this fix: without it PatchEmbedConvToGemm rewrites a 1x1 convolution with a single output position into reshape + gemm, so every spatial_size == 1 case -- the GEMV path #1074 optimises -- never reaches the kernel under test. Keeping the two in one branch is what lets those tests actually exercise the change.

Closing rather than merging so the fix lands once. The branch behind this PR is left in place for reference.

PR #1074, opened 2026-09-20
perf(qconv): speed up both paths of the quantized 1x1 convolution

AMDmoore:
## Summary

Speeds up the fused quantized 1x1 convolution on both of its dispatch paths.
The GEMV path carries a decode step; the tiled path carries prefill, and between
them they account for the large majority of GPU time in a quantized decoder
pipeline.

- **GEMV / decode**: roughly **7x** per dispatch on the widest decode shape,
  worth about **5x** on decode throughput end to end.
- **Tiled / prefill**: roughly **5x** per dispatch across the four projection
  shapes of a 40-layer decoder, worth about **4.9x** on time to first token and
  **2.3x** on whole-generation wall clock.

The branch also carries a correctness fix the new coverage depends on, and the
dependency pins CI needs to build any of it.

## Why

Both paths started out losing to the same two things in different proportions:
weights fetched an element at a time with the unpacking branched at run time,
and a block re-reading the whole activation run once per output channel it
produced. Fixing those exposed a third, which turned out to be the larger one on
the tiled path.

**GEMV.** `qconv_gemv_kernel` was the single heaviest dispatch in a decode step.
The scalar loop fetched a weight and dereferenced it in the same iteration, so
every iteration ended in a wait for a single byte and a thread walked
`K / block` of those waits end to end -- nothing else was in flight to cover
them, and occupancy was already at the maximum. On top of that a block produced
one output channel, so it re-read the whole activation column per channel and
walked its reduction tree per channel.

A sweep of K at fixed `Cout` separates the two cost terms: kernel time barely
moves from K=512 to K=4096 and only the part above that scales, which puts the
per-block fixed cost (reduction plus epilogue) at roughly a third of the kernel
at the shapes that matter. The slope of the scaling part corresponds to an L1/L2
request rate several times the DRAM rate, four fifths of which is activations.

**Tiled, traffic.** With the GEMV path fixed, a fresh trace moved the hot spot
rather than removing it: prefill became the large majority of GPU time and is
almost entirely this kernel. The same activation-sharing argument applies, and
it is far starker here because the position tile makes the activation run long.
A block streams the whole of K for the positions it owns whatever number of
channels it produces, so the activations are charged `ceil(Cout / kTileCo)`
times against the weights' `ceil(P / kTileP)`. At `kTileCo = 8` that put the
activations at ~97% of everything the kernel moved on a wide FFN projection.

**Tiled, issue rate.** After the traffic fix the kernel was still four times off
its arithmetic roofline, and hardware counters said it was not waiting on
memory: the memory unit reported busy essentially always but stalled under one
percent, shared-memory bank conflicts were negligible, and sustained DRAM
bandwidth sat at a fifth of the part's roofline.

Disassembling settled it. The inner loop issued 101 instructions to perform 32
multiply-accumulates, and 32 of those instructions were sign extension: the
staged weights were int16, two to a dword, and the 24-bit multiply-add reads the
low 24 bits of a register, so every packed pair cost a bitfield extract and a
shift before it could be multiplied. Multiplying the whole instruction stream by
the wave and tile counts predicts the measured kernel time to within a few
percent, at two different clocks. The kernel is issue-limited, so the question
is only how few instructions a multiply-accumulate can cost.

## What

**`perf(qconv): stage GEMV weights by the word instead of the byte`** --
a step loads its weights and activations into registers and touches neither
until both have arrived, which leaves the loads outstanding together and
collapses the step's waits into one. Weights are read by the word rather than
the byte, which shortens the chain again by the packing factor and makes the
access contiguous per wave. Bit width and signedness become template parameters
so the nibble unpacking folds to constants. Requires the vector width to divide
`in_channels`; the caller checks that and falls back to the scalar body
otherwise.

**`perf(qconv): give each GEMV block several output channels`** -- a block now
covers `kGemvCoPerBlock` channels. The activation chunk a step stages is spent
against every one of them, the channels share one reduction tree so the barriers
are walked once rather than once per channel, and the weights -- the only
traffic that is genuinely per channel -- are unchanged. A tail block folds its
surplus channels onto its first one and discards the result, so no lane
addresses past the weight buffer.

Eight is where the two directions meet. Raising the count keeps dividing the
shared traffic but divides the grid by the same amount, and a narrow output
layer runs out of blocks to fill the machine before the saving pays for the
registers and shared memory the extra accumulators cost. Measured across
`Cout` of 512 / 4096 / 17920, sixteen wins only on the widest and loses on the
other two; eight is within roughly a tenth of the best on the widest and is the
best or tied elsewhere.

**`perf: cut the activation re-reads out of the tiled qconv path`** -- raises
`kTileCo` from 8 to 64, dividing the activation term by eight, and then takes
three smaller costs off the inner loop that the traffic fix exposed: the product
goes through `__mul24`, the staged weights are held transposed and padded so
that the channels one thread multiplies against a single activation arrive as
two 128-bit reads, and the weights are staged by the word with bit width and
signedness as template parameters.

The tile shape was swept rather than reasoned about: `kTileCo` of 32 and 128
both lose to 64, 128 badly, because the grid shrinks and occupancy follows;
`kTileK` of 64 pushes shared memory far enough to cost more occupancy than the
halved barrier count buys.

**`perf: widen the staged tiled-qconv weights to one per dword`** -- stores one
weight per dword and sign-extends on the staging write, where it happens once
per staged element instead of once per use. The extracts and shifts leave the
loop entirely, taking it from 3.16 to 1.94 instructions per multiply-accumulate.
The slab doubles in size, which is free because occupancy here is limited by
registers rather than shared memory.

**`perf: run the tiled qconv reduction on the integer dot product`** -- moves the
reduction onto `v_dot4_i32_iu8`, which folds four multiply-accumulates into one
instruction. Two things have to change to reach it, and both pay for themselves.

Its operands are packed bytes, so a 16-bit activation is split into high and low
halves that each get a dot and an accumulator. It reads both operands as
unsigned, which rules out subtracting a zero point first, so the product is
expanded into a plain dot plus a column sum of the activations, a column sum of
the weights, and a constant. The activation sum runs over the position a thread
already owns, and the weight sum is a block reduction over the staged slab, so
both cost a handful of instructions per tile.

Taking the zero point out of the loop also takes all unpacking out of the weight
staging: at 4 bits a masked word is already a valid dot operand, and a signed
weight is mapped onto the unsigned range with one exclusive-or whose bias
cancels between the weight and its zero point. A K-tile now issues about one
instruction per multiply-accumulate where it used to issue nearly three, and it
needs fewer registers and less shared memory, so occupancy rises as well.

The kernel keeps its two accumulators in int32 for the whole reduction rather
than widening per tile, which bounds K. Shapes past that bound, or whose weight
rows are not word-aligned, keep using the multiply-add kernel, which is still
the general fallback.

**`fix: decline the patch-embed conv-to-gemm rewrite for unit kernels`** -- this
was ROCm/hip-ep#1072, carried here because the new coverage needs it. Without it
`PatchEmbedConvToGemm` rewrites a 1x1 convolution with a single output position
into a reshape plus gemm, so every `spatial_size == 1` case -- exactly the GEMV
path this PR optimises -- never reaches the kernel under test. It also fixes a
pool blow-up: the intervening reshape breaks the QDQ fusion that would have
folded the pattern back into a quantized conv, the weights stay dequantized as
plain gemm operands, and for a 40-layer graph the transient pool grows from a
few hundred KiB to tens of GiB.

**`test(numeric): cover the fused quantized 1x1 convolution`** -- the kernel had
no numeric coverage at all, only conversion and fusion lit tests that stop at
the IR. See the test plan below.

**`ci: build dependencies against the ONNX Runtime 1.29 / AMDGPU OGA pins`** --
the dependency builds still resolved ONNX Runtime 1.27.0 and OGA v0.14.0 plus an
upstream PR patch, so a run here built a different stack than the GPU suites
measure against. Ported from ROCm/hip-ep#534.

## Test plan

- `test/numeric/tests/test_qconv.py`, 37 cases, all passing against ORT CPU with
  `--ep-dll`. Covers both dispatch paths and the boundary between them, odd
  `Cin` (which alternates the nibble parity of each weight row), ragged `Cin`
  that leaves a tail against 8/16/32-wide loads, ragged `Cout` that leaves a
  partial last channel group, signed and unsigned 4-bit, 8-bit both signs, and
  reductions wide enough to overflow int32. Most cases are bit-exact; the rest
  are within one output LSB.
- The wider channel tile left three gaps in the first 29, closed by the last
  eight: every prefill case had a position count that divided the position tile
  exactly, no case had more than one full output-channel tile, and the
  element-staging path was unreachable at 8 bits because the only 8-bit `Cin`
  divided by a word. Those same eight are what keep the multiply-add fallback
  covered now that the dot-product kernel takes the aligned shapes.
- Full LIT suite and `pre-commit run` pass.
- Generated ISA inspected for each rewrite to confirm the intended effect rather
  than inferring it from the timing: the GEMV loop's waits drop from ten to two
  and its weight loads widen from single bytes to vectors; the tiled loop's
  sign-extension pairs disappear and then its multiply-adds are replaced by
  dot-product instructions, with the instruction count per multiply-accumulate
  falling at each step.

Performance was measured on a gfx1151 part, Release build, GPU otherwise idle
and every run serial. Kernel time comes from a system-level trace of single-node
models at the real shapes; counter data comes from a per-dispatch hardware
capture of the same models. End-to-end numbers come from `model_benchmark` on a
quantized decoder pipeline, `-w 1 -r 5`, with all EP perf and debug tracing off;
each baseline was rebuilt and measured in the same session rather than quoted
from earlier runs. The per-token distribution is skewed enough that its mean is
not usable, so the comparison is on medians, whose spread was under 3% of the
value on both sides. Decode is unchanged by the tiled commits and prefill by the
GEMV commits, which makes each a control for the other.

## Notes for reviewers

- **Contains ROCm/hip-ep#1072**, which is now closed in favour of this branch.
  That fix is the first commit here, unchanged.
- **Duplicates ROCm/hip-ep#534** rather than depending on it, since #534 is
  still open and this branch needs the pins in its own CI now. The two should
  resolve to identical files.
- `test/numeric` is not wired into CI -- it needs a GPU and an explicit
  `--ep-dll` -- so the new coverage does not run on a PR check. It was run
  locally against every intermediate kernel, not only the final one.
- The `bias` argument of `hip_qconv` is dead: the kernel takes it and the
  epilogue would add it, but the fusion pattern only matches a four-operand
  `hip.conv`, so it is always null. Left alone here; flagging it because the new
  tests do not cover it and could look like they should.
- The dot-product kernel is gated on word-aligned weight rows and on K staying
  inside an int32 accumulator. Both hold for the shapes a decoder pipeline
  produces, and the multiply-add kernel remains correct and tested for the rest,
  but a reviewer should sanity-check the bound rather than take it on trust.
- What dominates the tiled path now is its staging: the compute loop is close to
  one instruction per multiply-accumulate, while a K-tile still spends roughly
  fifty 64-bit address-arithmetic pairs and several predicated blocks getting
  the slabs into shared memory. That is the next thing to attack, not part of
  this PR.

## AI assistance

Written with Cursor. The analysis, the kernel rewrites, the tests, and this
description were produced with the tool and reviewed by me; the performance
claims come from the measurements described above, not from the tool.


AMDmoore:
PSU_ORC-Model-3.4.0_kvInt8 performance on Sep 21:
http://xsjnts.xilinx.com/ROCm/hipdnn-detail?database=development&collection=daily_hip_ep_genai&id=6ab0e56e534ee95a11e4c6a7

AMDmoore:
Added: the wave-matrix prefill body now stages the activation a vector of positions at a time.

**Why.** The body issued more memory instructions than matrix ops. Each lane staged one activation element per request, so sixteen matrix ops were fed by sixteen two-byte activation loads plus two weight loads. The memory unit meters requests rather than bytes, and hardware counters put it at the edge of saturation (memory unit busy 96%, stalled 3%) -- the kernel was spending its time issuing staging requests, not doing arithmetic.

**What.** Specializing the body on every extent dividing its tile lets a lane take a run of four adjacent positions as one request (positions are the contiguous axis of the activation; the quad a lane assembles runs along k and is strided). One quad and one position group per lane covers the tile exactly once, so the staging repeat loop disappears too. The same specialization makes the range guards constant true.

Inner loop, gfx1151 ISA:

| | before | after |
|---|---|---|
| activation load | 16x \global_load_d16_u8\ | 4x \global_load_b64\ |
| weight load | 2x \global_load_b32\ | unchanged |
| LDS store | 8x \ds_store_b32\ | 4x \ds_store_2addr_b32\ |
| \_wmma_i32_16x16x16_iu8\ | 16 | unchanged |

**Effect.** The prefill convolution kernel is ~1.07x faster than the previous commit on a decoder-only model at a 128-token prompt, and the counters confirm the intended shift: memory unit busy 96% -> 84%, stalled 3% -> 16%. The limit moved from issuing requests to waiting on data return. LDS bank conflicts rise from 1.2% to 3.4% -- lanes are now four positions apart, and the fragment rows must stay 16-byte aligned, so the bank count halves -- but that is small enough not to offset the gain.

The general body keeps its per-element staging and its guards, so ragged shapes are unaffected. \	est_qconv.py\ passes 37/37, covering ragged, 4-bit, 8-bit, signed and unsigned.

AMDmoore:
Added one more commit to this branch: the wave-matrix prefill loop now issues a
tile's global loads one iteration ahead of the iteration that consumes them.

**Why.** The loop staged and consumed a tile in the same iteration — issue the
loads, wait on them in full, write LDS, then run the matrix ops. Nothing stood
between a load and its first use, so each iteration paid an exposed round trip.
The previous commit on this branch had already cut the staging request count by
widening the loads, which moved the limiter from request count to data return,
so what was left to win was the latency itself rather than more bytes per
request.

**What changed.** A tile is staged into registers immediately after the previous
tile reaches LDS, so its latency runs underneath that tile's matrix ops. One set
of staging registers serves both tiles because the registers are spent once a
tile is written to LDS, so the LDS footprint is unchanged and the occupancy
ceiling does not move. Both staging paths share the registers; the general path's
bounds checks moved into the staging step, and static assertions tie the two
register mappings together so a change to either tile geometry fails to compile.

**Evidence in the generated code.** In the inner loop the previous build issued
its loads and then walked the vector-memory counter down to zero before the
barrier; the new build rotates them past the barrier to the end of the iteration
and waits on a partial counter, with the tile's matrix ops covering the round
trip. Instruction counts are essentially unchanged (172 vs 173 in the loop body,
same six loads, same sixteen matrix ops) — only the schedule moved, which is the
point. Allocated VGPRs drop rather than rise, because issuing the loads at the
end of the iteration shortens the window in which the values must stay live.

**Measured, relative.** Prefill kernel time is about 7% lower. Two independent
baseline/candidate pairs were run back to back in one session with no overlap
between the two sample sets, and end to end the time to first token moves by
about the same amount. Decode is unaffected, as expected — it runs the GEMV path,
not this kernel.

**Correctness.** `test/numeric/tests/test_qconv.py` passes 37/37, which covers the
ragged shapes that exercise the general staging path. End to end, greedy
generation of 128 tokens from a 2k-token prompt produces byte-identical output
before and after.

amd-mingw (approved):

